PR #59, opened 2026-03-10
fix(feishu): always reply to original message

kevinWangSheng:
## Summary
- All responses (text, preview cards, audio) are now sent as replies to the original user message, keeping conversations threaded
- Previously this only happened when `reply_in_thread` was explicitly set in config; now it's the default behavior
- Removed the `replyInThread` config option as it's no longer needed

## Test plan
- [ ] Send a text message to the Feishu bot and verify the response quotes the original message
- [ ] Send an image and verify the response is threaded
- [ ] Send audio and verify the audio reply is threaded
- [ ] Verify streaming preview (card) messages are also threaded
- [ ] Verify fallback to standalone message works when messageID is empty (e.g. reconstructed sessions)

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -105,7 +105,6 @@ type Platform struct {
 	allowFrom             string
 	groupReplyAll         bool
 	shareSessionInChannel bool
-	replyInThread         bool
 	threadIsolation       bool
 	client                *lark.Client
 	wsClient              *larkws.Client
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## Review Summary

LGTM. Simple behavioral change to always reply to original message.

### What's Changed

1. **Removed**: `reply_in_thread` config option
2. **New behavior**: Always replies to original message when `messageID` is available
3. **Added**: Audio message reply support

### Why

- Replies keep conversation threaded by default
- Removes unnecessary configuration option
- More consistent user experience

### Code Quality

- 18 lines added, 6 deleted
- Simple, focused change
- Also adds missing audio reply support

### No CI

Branch has no CI. Before merge, please verify:
```bash
go build ./...
go vet ./...
go test ./...
```

**Verdict**: Ready to merge after manual verification.

PR #76, opened 2026-03-11
feat: add /dir command for dynamic agent work directory switching

KRATSZ:
## Summary

Adds a `/dir` command that lets users switch the agent's working directory at runtime without restarting `cc-connect`.

## Notes

- Primary command: `/dir`
- Power-user aliases kept for compatibility: `/cd`, `/chdir`, `/workdir`
- `/dir` now follows the same `admin_from` gate as other sensitive commands like `/shell`

## Behavior

- `/dir` — show current working directory
- `/dir <path>` — switch to a target path
- `/dir help` — show usage
- Relative paths are resolved against the current agent work directory
- On successful switch, the current interactive session is cleaned up and the active agent session ID/history are reset so the next turn starts fresh in the new directory

## Changes

- Added `WorkDirSwitcher` interface in `core/interfaces.go`
- Implemented `WorkDirSwitcher` for all 7 supported agents
- Renamed the engine command handler and i18n symbols from `cd` to `dir`
- Registered `/dir` as the primary command while preserving `/cd`, `/chdir`, `/workdir` aliases
- Added admin permission gating for the directory-switch command
- Added targeted unit tests for `/dir`, alias compatibility, usage handling, and admin gating

## Verification

- `go build ./...`
- `go test -run ^$ ./...`
- `go test ./core -run "TestCmdDir|TestEngine_AdminFrom_GatesDir" -count=1`

## Follow-up ideas

Not included in this PR:
- persistence across restarts
- saved directory bookmarks / subcommands such as `/dir add|list|switch|del`

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -107,6 +107,19 @@ func (a *Agent) Name() string           { return "claudecode" }
 func (a *Agent) CLIBinaryName() string  { return "claude" }
 func (a *Agent) CLIDisplayName() string { return "Claude" }
 
+func (a *Agent) SetWorkDir(dir string) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	a.workDir = dir
+	slog.Info("claudecode: work_dir changed", "work_dir", dir)
+}
+
+func (a *Agent) GetWorkDir() string {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	return a.workDir
+}
+
 func (a *Agent) SetModel(model string) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
--- a/agent/codex/codex.go
+++ b/agent/codex/codex.go
@@ -93,6 +93,19 @@ func normalizeReasoningEffort(raw string) string {
 
 func (a *Agent) Name() string { return "codex" }
 
+func (a *Agent) SetWorkDir(dir string) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	a.workDir = dir
+	slog.Info("codex: work_dir changed", "work_dir", dir)
+}
+
+func (a *Agent) GetWorkDir() string {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	return a.workDir
+}
+
 func (a *Agent) SetModel(model string) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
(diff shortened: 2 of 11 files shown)


chenhg5:
Hey @KRATSZ, nice feature — dynamic work directory switching is a real user need. The code quality is solid: clean interface design, proper path validation, session cleanup, i18n coverage, and unit tests. 👍

A few suggestions before merging:

### 1. Command naming: `/dir` instead of `/cd`

Looking at the existing command naming convention (`/model`, `/mode`, `/provider`, `/workspace`, `/session`, `/shell`, `/heartbeat`), they all use descriptive English words. `/cd` is a Unix shell abbreviation that may not be intuitive for non-technical users on chat platforms like Feishu/DingTalk.

Suggestion: rename the primary command to **`/dir`** — short, clear, and consistent with the naming style. Keep `/cd`, `/chdir`, `/workdir` as aliases for power users.

### 2. Consider persistence

Currently `/dir` changes are lost on restart (reverts to `config.toml` value). This could be confusing. Consider persisting the override — for example, writing to the session store or a local state file so the change survives restarts.

### 3. Consider a subcommand system for multi-directory workflows

If a user frequently switches between several directories, typing full paths every time is tedious. Consider evolving this into a subcommand system similar to `/session` or `/cron`:

```
/dir                    — show current work directory
/dir <path>             — switch to <path> (shortcut for quick use)
/dir add <name> <path>  — save a named directory bookmark
/dir list               — list all saved directories
/dir switch <name|index>— switch by name or index number
/dir del <name>         — remove a saved directory
```

This way users can:
```
/dir add frontend /home/user/app/frontend
/dir add backend  /home/user/app/backend
/dir list         → 1. frontend (/home/user/app/frontend)  2. backend (...)
/dir switch 1     → switched to frontend
```

This doesn't need to be in this PR — the current simple `/dir <path>` is a good v1. But designing the command name and structure with this future in mind would avoid breaking changes later.

### 4. Minor: `MsgCdUsage` is defined but never referenced

The i18n key `MsgCdUsage` is declared in `i18n.go` but never used in `cmdCd`. Either wire it up (e.g. for an invalid args case) or remove it.

### 5. Minor: consider `admin_from` permission check

`/shell` requires admin permissions. `/dir` changes where the agent operates, which is also a sensitive action. Worth considering whether it should also respect `admin_from`.

---

Everything else looks good — the `WorkDirSwitcher` interface design, agent implementations, path validation (relative + absolute), and session reset logic are all clean. No breaking changes. 🚀

KRATSZ:
Updated this PR around the `/dir` naming, usage wiring, and admin gate.

For 2 and 3, I’m planning a follow-up PR instead of expanding #76 further, so this PR stays focused on the v1 runtime directory switch. 🙂

Planned next:
- persist `/dir` overrides across restarts
- add a bookmark/subcommand flow such as `/dir add|list|switch|del` for multi-directory workflows

That should keep the current review surface small while leaving room to design the longer-lived `/dir` UX cleanly. 🚀

PR #110, opened 2026-03-13
feat (discord, feishu): add thread isolation option

cnrpman:
## Description
This brings the thread isolation idea from [other project](https://github.com/cnrpman/no-claw) into cc-connect as an opt-in platform setting (currently available on Discord and Feishu).
Private chat / DM behavior stays as before.

  Known limitation: with thread_isolation=true, Discord and Feishu currently use one shared agent session
  per thread/root, so share_session_in_channel=false does not further isolate users within the same
  thread. other config combinations might not not fully orthogonal also.

  go test ./... passes.

### Feishu:
B3E5EE75-D4C2-43CD-9A12-5029C962E263_1_105_c
18C21852-4A0F-4B01-A534-45FBA73576BF_1_105_c
A752D7C8-A61D-46CE-B466-D492AC787738_1_105_c

### Discord (also with group_reply_all = true):
Screenshot 2026-03-14 at 04 14 27

Screenshot 2026-03-14 at 04 14 38

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -552,6 +552,8 @@ app_secret = "your-feishu-app-secret"
 #                           # 设为 true 时，群聊中无需 @机器人 也会响应所有消息（默认 false）
 # share_session_in_channel = false  # If true, all users in a group share one agent session
 #                                   # 设为 true 时，群聊内所有用户共享同一个 Agent 会话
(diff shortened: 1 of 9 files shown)


cnrpman:
@chenhg5 thanks for review! it's now rebase to the main & ready to merge.

PR #113, opened 2026-03-14
feat: multi-workspace support with session resilience

leighstillard:
## Summary

- **Multi-workspace mode**: channels auto-bind to workspace directories, each getting isolated Claude Code sessions with per-workspace agents and session managers
- **Session resilience**: path normalization prevents CWD mismatches on resume, auto-recovery from "Prompt is too long" failures with user notification, context consumption indicator `[ctx: XX%]` on every message
- **Diagnostic logging**: structured slog at spawn, resume, reap, and failure points with JSONL file size tracking
- **Dual-track context monitoring**: SDK token counts vs Claude self-reported estimates, logged for comparison over time

## Test plan

- [x] `TestNormalizeWorkspacePath` — trailing slash, symlink, dot segments, nonexistent fallback
- [x] `TestWorkspacePoolNormalizesKeys` — trailing slash produces same pool entry
- [x] `TestHandleResultParsesUsage` / `TestHandleResultNoUsage` — token usage parsing
- [x] `TestParseSelfReportedCtx` — regex extraction of self-reported context %
- [x] `TestResumeFailureFallbackToFreshSession` — resume fail → clear ID → retry fresh → notify user
- [x] Multi-workspace integration tests (resolution, binding, init flow)
- [x] Full test suite passes (`go test ./...`)

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -857,6 +857,25 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 
 	session := sessions.GetOrCreateActive(msg.SessionKey)
 	if !session.TryLock() {
+		// Check for /btw — inject into the running session mid-turn
+		trimmed := strings.TrimSpace(content)
+		if isBtwCommand(trimmed) {
+			btw := strings.TrimSpace(trimmed[len(matchBtwPrefix(trimmed)):])
+			if btw != "" {
+				e.interactiveMu.Lock()
+				state, ok := e.interactiveStates[interactiveKey]
+				e.interactiveMu.Unlock()
+				if ok && state.agentSession != nil && state.agentSession.Alive() {
+					if err := state.agentSession.Send(btw, nil, nil); err != nil {
+						slog.Error("btw: send failed", "error", err)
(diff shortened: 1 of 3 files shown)


chenhg5:
Thanks for this PR! The session resilience features (auto-recovery from resume failures, context indicator `[ctx: XX%]`, `--continue` for first connection) are great additions.

However, I noticed a significant change: **command prefix changed from `/` to `!`**

```go
// Before
if strings.HasPrefix(content, "/") {

// After  
if strings.HasPrefix(content, "!") {
```

This is a **breaking change** — users are accustomed to `/help`, `/new`, `/switch`, etc.

**Questions:**

1. Is this change intentional? If so, what's the reason for switching to `!`?
2. If intentional, we should:
   - Update documentation
   - Update `config.example.toml`
   - Add migration notes in the release notes
   - Consider supporting both prefixes temporarily for backward compatibility

3. Also, this PR is quite large (3000+ lines). Would you be open to splitting it into smaller PRs for easier review?
   - PR 1: Session resilience (resume recovery, context indicator)
   - PR 2: Command prefix change (if intentional)

This would help get the session resilience features merged faster while we discuss the prefix change separately.

PR #129, opened 2026-03-14
fix(session): invalidate stale agent session IDs on agent type switch

### LINKED ISSUE #128: Switching agent type causes stale session ID errors
## Bug Description

When changing the agent type for a project (e.g., switching from `opencode` to `pi` or `claudecode` to `codex` in `config.toml`), the persisted session files still contain `agent_session_id` values created by the **old** agent. When the new agent tries to resume these sessions, it fails because the session IDs are incompatible across different agent backends.

## Steps to Reproduce

1. Configure a project with agent type A (e.g., `opencode`)
2. Send messages so sessions are created and `agent_session_id` is persisted
3. Change the agent type in `config.toml` to agent type B (e.g., `pi`)
4. Restart the daemon
5. Send a message to the same conversation

## Expected Behavior

The new agent should start a fresh session, ignoring session IDs from the old agent.

## Actual Behavior

The engine passes the old agent's session ID to the new agent via `agent.StartSession(ctx, session.AgentSessionID)`. The new agent fails because it doesn't recognize the foreign session ID, resulting in errors like:

```
No session found matching 'ses_xxxxx'
```

The user sees `(empty response)` because the error prevents any text from being generated.

## Root Cause

The `Session` struct persists `agent_session_id` but has no record of **which agent type** created it. When the agent type changes, there is no mechanism to detect and invalidate stale session IDs.

## Impact

Any user who switches agent types for a project will hit this issue on all existing sessions until the sessions are manually cleared or new sessions are created.

quabug:
## Summary

- Add `agent_type` field to `Session` struct to track which agent backend created each session ID
- On engine startup, `InvalidateForAgent()` automatically clears `agent_session_id` values from a different agent type
- All code paths that set `AgentSessionID` now also record the agent type
- Session snapshot persistence includes the new `agent_type` field

Closes #128

## Problem

When switching a project's agent type in config (e.g., from `opencode` to another agent), persisted sessions still contain `agent_session_id` values from the old agent. The new agent fails to resume these foreign session IDs, causing errors and `(empty response)`.

## Solution

Track the agent type alongside each session ID. On startup, if the configured agent doesn't match the stored type, the stale session ID is cleared so the new agent starts a fresh session. Existing sessions with no `agent_type` (pre-migration) are left untouched to avoid breaking upgrades.

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] Manually verified: switching agent type triggers invalidation log and clears stale session IDs
- [x] Manually verified: sessions without `agent_type` (migration case) are not incorrectly invalidated

### DIFF
--- a/core/session_test.go
+++ b/core/session_test.go
@@ -211,6 +211,53 @@ func TestSession_GetName(t *testing.T) {
 	}
 }
 
+func TestSessionManager_InvalidateForAgent(t *testing.T) {
+	sm := NewSessionManager("")
+
+	// Create sessions with different agent types
+	s1 := sm.NewSession("user1", "sess1")
+	s1.SetAgentSessionID("old-id-1", "opencode")
+
+	s2 := sm.NewSession("user2", "sess2")
+	s2.SetAgentSessionID("old-id-2", "claudecode")
+
+	s3 := sm.NewSession("user3", "sess3")
+	s3.SetAgentSessionID("old-id-3", "") // pre-migration, no agent type
+
+	s4 := sm.NewSession("user4", "sess4") // no agent session ID at all
+
+	sm.InvalidateForAgent("claudecode")
+
+	// s1: opencode → should be invalidated
+	if got := s1.GetAgentSessionID(); got != "" {
+		t.Errorf("s1 (opencode) AgentSessionID = %q, want empty (should be invalidated)", got)
+	}
+	if s1.AgentType != "claudecode" {
+		t.Errorf("s1 AgentType = %q, want %q after invalidation", s1.AgentType, "claudecode")
+	}
+
+	// s2: claudecode → should be untouched
+	if got := s2.GetAgentSessionID(); got != "old-id-2" {
+		t.Errorf("s2 (claudecode) AgentSessionID = %q, want %q (should be preserved)", got, "old-id-2")
+	}
+	if s2.AgentType != "claudecode" {
+		t.Errorf("s2 AgentType = %q, want %q", s2.AgentType, "claudecode")
+	}
+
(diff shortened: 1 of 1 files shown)


chenhg5:
Hey @quabug, nice work on this — the design is solid. Tracking `AgentType` alongside session IDs and invalidating on mismatch is the right approach, and the backward compatibility for pre-migration sessions (skipping empty `AgentType`) is well thought out.

One thing I'd like to see before merging:

### Missing unit test for `InvalidateForAgent`

This is the core logic of the PR but currently only has manual verification. Could you add a dedicated test? Something like:

```go
func TestSessionManager_InvalidateForAgent(t *testing.T) {
    sm := NewSessionManager("")

    // Create sessions with different agent types
    s1 := sm.NewSession("user1", "sess1")
    s1.SetAgentSessionID("old-id-1", "opencode")

    s2 := sm.NewSession("user2", "sess2")
    s2.SetAgentSessionID("old-id-2", "claudecode")

    s3 := sm.NewSession("user3", "sess3")
    s3.SetAgentSessionID("old-id-3", "") // pre-migration, no agent type

    sm.InvalidateForAgent("claudecode")

    // s1: opencode → should be invalidated
    assert s1.GetAgentSessionID() == ""
    // s2: claudecode → should be untouched
    assert s2.GetAgentSessionID() == "old-id-2"
    // s3: empty type → should be untouched (backward compat)
    assert s3.GetAgentSessionID() == "old-id-3"
}
```

This covers the three key cases: mismatched type (invalidated), matching type (preserved), and empty type (migration-safe).

Everything else looks good — all call sites updated, lock ordering is consistent, persistence includes the new field. 👍

PR #130, opened 2026-03-14
refactor(dingtalk): add audio support, TTS providers, and security improvements

windli2018:
## Summary

Refactored the audio/voice message implementation into clean, logical commits. Replaced 10 messy commits with alternating fix/feat patterns with a clear 4-commit structure.

## Changes

Split the monolithic audio send commit into 3 focused commits:

1. **feat(speech)**: Add MP3 to OGG and AMR conversion functions
   - `ConvertMP3ToOGG`: Opus codec, 16kHz mono, 32kbps, voip application
   - `ConvertMP3ToAMR`: AMR-NB codec, 8kHz mono, 12.2kbps
   - Both use ffmpeg stdin/stdout pipes (snap compatible)

2. **feat(tts)**: Add EdgeTTS, PicoTTS, and EspeakTTS providers
   - EdgeTTS: Microsoft Edge free TTS API (network required)
   - PicoTTS: Google Pico TTS via pico2wave (offline, better quality)
   - EspeakTTS: Local eSpeak command (lightweight)

3. **feat(dingtalk)**: Add audio message sending capability
   - `SendAudio`: Send audio via oToMessages API
   - `uploadMedia`: Upload to DingTalk media API
   - `compressAudio`: Pure Go MP3 encoding + ffmpeg fallback
   - Engine: Smart voice message fallback logic
   - i18n: Platform recognition hint message

## Benefits

- ✅ Each commit has a single, clear responsibility
- ✅ Easier to review and understand
- ✅ Independent revert capability
- ✅ Granular history for better debugging
- ✅ All 758 insertions preserved across 9 files

## Related

Replaces/consolidates commits by liyong@alot.pw (e87dc26 through f34a931)

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -404,6 +404,27 @@ func main() {
 				} else {
 					slog.Warn("tts: minimax provider enabled but api_key is empty")
 				}
+			case "espeak":
+				voice := cfg.TTS.Voice
+				if voice == "" {
+					voice = "zh" // default to Chinese
+				}
+				ttsCfg.TTS = core.NewEspeakTTS("", voice)
+				ttsCfg.Provider = "espeak"
+			case "pico":
+				voice := cfg.TTS.Voice
+				if voice == "" {
+					voice = "zh-CN" // default to Chinese (Simplified)
+				}
+				ttsCfg.TTS = core.NewPicoTTS("", voice)
+				ttsCfg.Provider = "pico"
+			case "edge":
+				voice := cfg.TTS.Voice
+				if voice == "" {
+					voice = "zh-CN-XiaoxiaoNeural" // default Chinese neural voice
+				}
+				ttsCfg.TTS = core.NewEdgeTTS(voice)
+				ttsCfg.Provider = "edge"
 			default: // "openai" or unspecified
 				apiKey := cfg.TTS.OpenAI.APIKey
 				baseURL := cfg.TTS.OpenAI.BaseURL
--- a/config/config.go
+++ b/config/config.go
@@ -131,12 +131,12 @@ type SpeechConfig struct {
 
 // TTSConfig configures text-to-speech output (mirrors SpeechConfig style).
 type TTSConfig struct {
-	Enabled    bool   `toml:"enabled"`
-	Provider   string `toml:"provider"`     // "qwen" | "openai" | "minimax"
-	Voice      string `toml:"voice"`        // default voice name
-	TTSMode    string `toml:"tts_mode"`     // "voice_only" (default) | "always"
-	MaxTextLen int    `toml:"max_text_len"` // max rune count before skipping TTS; 0 = no limit
-	OpenAI     struct {
+	Enabled     bool   `toml:"enabled"`
+	Provider    string `toml:"provider"`     // "qwen" | "openai" | "minimax" | "espeak" | "pico" | "edge"
+	Voice       string `toml:"voice"`        // default voice name (for edge: "zh-CN-XiaoxiaoNeural"; for pico: "zh-CN"; for espeak: "zh")
+	TTSMode     string `toml:"tts_mode"`     // "voice_only" (default) | "always"
+	MaxTextLen  int    `toml:"max_text_len"` // max rune count before skipping TTS; 0 = no limit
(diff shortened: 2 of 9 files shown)


chenhg5:
Hey @windli2018, thanks for the effort to reorganize these commits — the 4-commit structure is much cleaner than 10 alternating fix/feat commits. The feature set (3 TTS providers + DingTalk audio send/receive) is solid, but there are several issues that should be addressed before merging.

---

## 🔴 High Priority

### 1. Code injection risk in EdgeTTS

```go
pythonCode := fmt.Sprintf(`
import edge_tts
import asyncio
async def synthesize():
    communicate = edge_tts.Communicate(text=%q, voice=%q)
    await communicate.save(%q)
asyncio.run(synthesize())
`, text, voice, tmpFile)
```

User-supplied `text` is interpolated into executable Python code. While Go's `%q` escapes most things, embedding user input into code is an anti-pattern. Please use one of:
- **Preferred:** Call the `edge-tts` CLI directly (`edge-tts --voice ... --text ... --write-media ...`) instead of embedding Python.
- **Alternative:** Pass text via stdin pipe to avoid any string escaping issues.

### 2. Predictable temp file paths (PicoTTS & EdgeTTS)

```go
tmpFile := "/tmp/pico_tts_" + fmt.Sprintf("%d", time.Now().UnixNano()) + ".wav"
tmpFile := "/tmp/edge_tts_" + fmt.Sprintf("%d", time.Now().UnixNano()) + ".mp3"
```

These are predictable and susceptible to symlink attacks. Use `os.CreateTemp` (or `os.MkdirTemp`) instead:

```go
f, err := os.CreateTemp("", "pico_tts_*.wav")
```

### 3. No tests

962 lines of new code with 0 tests. At minimum, the following should have unit tests:
- `parseWAV` / `parseWAVToInt16` — easy to test with a minimal valid WAV byte slice
- `encodeWAVToMP3` — round-trip test with known input
- `handleAudioMessage` — mock the download and verify message construction
- `getAccessToken` — test caching logic and expiry buffer
- Each TTS provider's `Synthesize` — at least verify argument construction (can mock the exec call)

---

## 🟡 Medium Priority

### 4. Data race on `accessToken` / `tokenExpiry`

`getAccessToken()` reads and writes `p.accessToken` and `p.tokenExpiry` without any synchronization. Multiple goroutines handling concurrent messages will race on these fields. Add a `sync.Mutex`:

```go
type Platform struct {
    // ...
    tokenMu     sync.Mutex
    accessToken string
    tokenExpiry time.Time
}
```

### 5. Excessive `slog.Info` — should be `slog.Debug`

I count 15+ `slog.Info` calls used for diagnostic/tracing purposes (e.g. "tts: checking conditions", "tts: sendTTSReply called", "dingtalk: audio converted", "dingtalk: MP3 encoding successful", etc.). These will flood production logs. Please downgrade to `slog.Debug` and keep `slog.Info` only for significant lifecycle events.

### 6. `MsgVoiceUsingPlatformRecognition` hardcodes "DingTalk" in core i18n

This message lives in `core/i18n.go` but explicitly mentions "DingTalk" in every language. If another platform later provides built-in voice recognition, this message would be misleading. Consider either:
- Making the platform name a parameter (e.g. `"using %s built-in recognition"`)
- Moving this message to the dingtalk package

---

## 🟢 Minor

### 7. Unused `Client *http.Client` field on EspeakTTS and PicoTTS

These are local command-line TTS providers that never make HTTP requests. The `Client` field is dead code — remove it.

### 8. New dependency `github.com/braheezy/shine-mp3 v0.1.0`

This is a v0.1.0 package. Consider whether the ffmpeg fallback path alone is sufficient (since ffmpeg is already a requirement for audio conversion), which would avoid adding an immature dependency. If you do keep it, a brief note in the PR description about why this package was chosen would be helpful.

### 9. `espeak -w /dev/stdout` won't work on Windows

Minor portability note — `/dev/stdout` is Unix-only. Fine if Windows support isn't a goal, but worth a comment in the code.

---

**Summary:** The feature design is good — DingTalk audio support with smart fallback (STT → platform recognition → error) and multiple TTS backends is a nice addition. The main blockers are the **security issues** (code injection, predictable temp files, data race) and **missing tests**. Once those are addressed, this should be good to go. 👍

PR #135, opened 2026-03-14
fix: workspace command prefix check missing leading slash in multi-workspace mode

kevinWangSheng:
## Summary

- In `handleMessage`, the check to skip `handleWorkspaceInitFlow` for `/workspace` and `/ws` commands was comparing against `"workspace"` and `"ws "` without the leading `"/"` prefix. Since `content` retains its `"/"` at this point in the code, the check always evaluated to false, causing `/workspace` commands to be incorrectly consumed by the init flow when it was active.

- This meant that during an active workspace init flow, sending `/workspace bind <name>` or `/workspace list` would trigger a "not a git URL" error instead of being routed to the workspace command handler.

## Fix

Changed the prefix check from `"workspace"` / `"ws "` to `"/workspace"` / `"/ws "` to correctly match the slash-prefixed command format.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [ ] Manual test: in multi-workspace mode with no workspace bound, send `/workspace bind <name>` and verify it routes to the workspace command handler instead of the init flow

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1019,7 +1019,7 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 		}
 		if workspace == "" {
 			// No workspace — handle init flow (unless it's a /workspace command)
-			if !strings.HasPrefix(content, "workspace") && !strings.HasPrefix(content, "/workspace") && !strings.HasPrefix(content, "ws ") {
+			if !strings.HasPrefix(content, "/workspace") && !strings.HasPrefix(content, "/ws ") {
 				if e.handleWorkspaceInitFlow(p, msg, channelID, channelName) {
 					return
 				}


chenhg5:
@kevinWangSheng Thanks for the fix! The workspace command prefix check is indeed a bug that needs fixing.

However, this PR currently includes changes that were already merged in #134 (the RateLimiter cleanup goroutine fix). This creates duplicate changes across  and .

Could you please:
1. Rebase or reset your branch on the latest  (which now includes #134)
2. Keep only the single-line fix in  (line 868: adding the missing  prefix)
3. Force-push the cleaned branch

Once that's done, this should be ready to merge. The fix itself looks correct - the slash-prefixed check  /  properly matches the command format when  still has its leading .

PR #164, opened 2026-03-16
fix: sync local session snapshot after /delete

q107580018:
## Summary
- sync local SessionManager snapshot with agent-side deletion in /delete flow
- add SessionManager.DeleteByAgentSessionID to remove local sessions by agent session id
- add regression tests for SessionManager and /delete command

## Verification
- go test ./...


dwillystyle90-cyber:
This PR is valid because it involves non-trivial session state synchronization after deletion, not just a simple cosmetic or isolated fix. I marked it Medium difficulty since the solution likely requires reasoning about state flow, edge cases, and test impact.

PR #192, opened 2026-03-17
codex/telegram: prefer native session names and stop dropping Telegram commands on restart

srcsucaaa:
## Summary

This PR improves the Codex session experience in chat platforms and fixes a Telegram delivery issue found during local verification.

It does three things:

1. Prefer Codex native session names (`thread_name`, then `title`) over transcript summaries
2. Reuse that same naming priority consistently across session UI and matching
3. Stop dropping pending Telegram updates during service restarts

### Codex native session names

- load Codex-native session metadata from:
  - `session_index.jsonl` for `thread_name`
  - `state_5.sqlite` for `title` fallback
- add `DisplayName` to `core.AgentSessionInfo`
- keep transcript summary as the final fallback only

### Unified session display and matching

Use the same priority everywhere:

1. cc-connect custom name from `/name`
2. Codex native display name
3. transcript summary

This now applies to:

- `/list`
- `/search`
- `/switch`
- `/current`
- `/status`

Session matching also now supports:

- numeric `/switch <number>`
- custom name match
- session ID prefix match
- native display name exact / prefix / substring match

### More forgiving slash-command parsing

- accept `swich` as an alias for `switch`
- accept attached arguments for selected commands such as:
  - `/switchfoo`
  - `/searchbar`
  - `/list2`

This was added after reproducing chat-side command parsing issues during Telegram testing.

### Telegram restart behavior

Remove the startup drain that acknowledged and discarded pending Telegram updates.

`cc-connect` already filters stale messages with `IsOldMessage`, so draining pending updates at startup could drop valid commands sent during a restart window.

With this change, commands sent while the service is restarting are no longer silently lost.

## Validation

Passed:

- `go test ./platform/telegram ./core ./agent/codex`

Also verified locally with a real Telegram bot session:

- `/list`
- `/search server`
- `/switch server`
- `/switch 5`
- `/current`
- `/status`

Full test suite result:

- `go test ./...`

Known unrelated failure:

- `agent/cursor: TestFetchModelsFromAgentCLI`

That test depends on local `agent models` output and appears to be environment-sensitive. It is unrelated to this PR.

## Notes

- custom `/name` labels still take precedence over Codex-native names
- if native metadata is unavailable, behavior falls back gracefully to the existing summary-based display
- Telegram platform/config format is unchanged

### DIFF
--- a/agent/codex/list.go
+++ b/agent/codex/list.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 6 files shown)


chenhg5 (approved):
## Review Summary

LGTM. Good improvements for Codex session naming and Telegram reliability.

### What's Changed

1. **Codex native session names**: Prefer `thread_name`/`title` over transcript summaries
2. **Unified session display**: Same naming priority across `/list`, `/search`, `/switch`, `/current`, `/status`
3. **Enhanced session matching**: Numeric `/switch <number>`, custom name, session ID prefix, native display name match
4. **Forgiving slash-command parsing**: Accept typos like `swich` and attached arguments like `/switchfoo`
5. **Telegram restart fix**: Remove startup drain that was dropping pending valid commands

### Key Fix: Telegram Drain Removal

**Before**: Startup drain acknowledged and discarded pending updates
**After**: Rely on `IsOldMessage` filter for staleness check

**Why**: Commands sent during a restart window were being silently lost. `IsOldMessage` already filters stale messages appropriately.

### Code Quality

- 699 additions, comprehensive feature
- Backward compatible: existing Telegram platform/config unchanged
- Graceful fallback to summary-based display if native metadata unavailable

### No CI

Branch has no CI. Author verified:
- `go test ./platform/telegram ./core ./agent/codex` - passes
- `go test ./...` - passes (known unrelated failure in cursor test)

### Merge Status

- Mergeable: UNKNOWN (may need rebase)

**Verdict**: Ready to merge after manual verification and possible rebase.

chenhg5:
## Review

Thanks for the detailed write-up and the solid `list_test.go` coverage around `session_index.jsonl`, display-name precedence, and “first `session_meta` wins”.

### What looks good
- **`DisplayName` on `core.AgentSessionInfo`** with a clear fallback chain (custom `/name` → native → summary) wired through list/search/switch/status is coherent.
- **Thread metadata loading** from `session_index.jsonl` is self-contained and tolerant of bad lines.
- **Telegram**: Dropping the startup drain is plausible if `IsOldMessage` already covers staleness; worth a short comment in code pointing to that invariant so future readers do not reintroduce the drain.

### Concerns / suggestions
1. **Mergeability**: GitHub currently reports **conflicts with `main`** — you’ll need a rebase/merge from upstream before merge.

2. **`sqlite3` CLI dependency**: `loadCodexThreadTitles` silently skips titles when `sqlite3` is missing. That’s OK for degradation, but it may surprise users who expect titles without installing the CLI. A one-line debug log when `LookPath` fails (or a note in docs) would help.

3. **SQL construction**: The `cwd` filter uses single-quote escaping only. Paths are usually trusted, but this is still string-built SQL; if Codex ever stores odd `cwd` values, consider a safer query API if feasible.

4. **Looser command parsing** (`/switchfoo`, `swich`, etc.): Useful for messy clients, but it increases the chance of accidental matches. Worth a quick audit that patterns cannot swallow unrelated messages (e.g. URLs or prose starting with `/switch`).

5. **`codexHomeDir` / errors**: Returning `""` when `UserHomeDir` fails and then surfacing `resolve CODEX_HOME: empty path` is fine; just be aware it changes the error shape vs the previous direct `UserHomeDir` error.

### Tests
I did not run the full suite on this branch here because of the reported base conflict; your listed `go test ./platform/telegram ./core ./agent/codex` plan is the right minimum once rebased.

**Overall**: Direction is strong for Codex UX + Telegram reliability; please rebase onto current `main`, then this should be ready for another pass after CI is green.

PR #216, opened 2026-03-18
fix(discord): support bot role mentions for @mention detection

0xsegfaulted:
## Summary
- Discord auto-creates a managed role for each bot. When a bot is not visible in the channel member list (common in new channels or when multiple bots share a guild), users can only @mention the role (`<@&roleID>`) instead of the user (`<@userID>`). The role mention was not recognized, causing the bot to silently ignore the message.
- Fix by caching the bot's managed role ID and checking both user mentions and role mentions in the message handler.

## Changes
- **`platform/discord/discord.go`**: 
  - Add `botRoleIDs` sync.Map to cache managed role IDs per guild
  - Cache role IDs on `Ready` and `GuildCreate` events via `cacheBotRoleIDForGuild()`
  - New `isDiscordBotMention()` checks `m.Mentions`, `m.MentionRoles`, and content patterns
  - New `stripDiscordMentionWithRole()` cleans both user and role mention tags
  - Add `IntentsGuilds` to receive `GuildCreate` events

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./platform/discord/` passes
- [x] Manual testing: deployed and verified role mention triggers bot response in new channels

### DIFF
--- a/platform/discord/discord.go
+++ b/platform/discord/discord.go
@@ -49,6 +49,7 @@ type Platform struct {
 	botID                 string
 	appID                 string
 	channelNameCache      sync.Map // channelID -> name
+	botRoleIDs            sync.Map // guildID -> bot managed role ID
 	readyCh               chan struct{}
 	seenMsgs              sync.Map // message ID dedup: prevents duplicate MessageCreate events
 }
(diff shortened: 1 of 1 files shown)


chenhg5:
@0xsegfaulted Thanks for the fix! This looks good overall.

One note: Since you're adding `IntentsGuilds`, users will need to re-authorize the bot. Could you also update the Discord setup documentation to reflect this? The docs should mention that users need to re-invite the bot with the new intents when upgrading.

Otherwise, the implementation looks solid! Merging now.

PR #244, opened 2026-03-21
feat: fork session on --continue to prevent context contamination

leighstillard:
## Summary

• When cc-connect starts up and picks up the most recent CLI session via `--continue`, it now adds `--fork-session` to create an independent context branch
• This prevents history interleaving when the user is concurrently working in the same workspace from both the CLI terminal and a messaging platform (Slack, Feishu, etc.)
• Plain `--resume` of cc-connect's own stored session ID is unaffected — no fork needed since that session belongs to cc-connect

## Problem

Without this, resuming into an active CLI session corrupts both context windows. The model sees interleaved tool calls and responses from two different conversations, leading to hallucinations about files and state that don't exist.

Example: user is working in `claude` CLI in a terminal on workspace X, then messages cc-connect via Slack about the same workspace. cc-connect does `--continue` and lands in the same session the CLI is using. Both conversations now share one context window, and every message from either side pollutes the other.

## How it works

- Adds `ContinueSession` sentinel constant (`__continue__`) in `core/message.go`
- Engine tracks `hasConnectedOnce` — first connection after startup uses `--continue`, subsequent connections use the stored session ID
- Claude Code agent maps `ContinueSession` → `claude --continue --fork-session`, which duplicates the full session context into a clean, independent branch
- Stored session IDs map to plain `claude --resume <id>` (no fork) since these are cc-connect's own sessions with no interleaving risk

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./core/ ./agent/claudecode/` passes (506 tests)
- [x] Manual: start cc-connect, verify first connection logs `is_resume=true` with `--continue --fork-session`
- [x] Manual: send a second message, verify it reuses the live session (no new process spawned)
- [x] Manual: restart cc-connect with a stored session ID, verify it uses plain `--resume` (no fork)

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 	"unicode/utf8"
 )
@@ -191,6 +192,8 @@ type Engine struct {
 
 	quietMu sync.RWMutex
 	quiet   bool // when true, suppress thinking and tool progress messages globally
+
+	hasConnectedOnce atomic.Bool // first connection uses --continue to bridge CLI usage
 }
 
 // workspaceInitFlow tracks a channel that is being onboarded to a workspace.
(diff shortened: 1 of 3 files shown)


core/engine.go
```diff
@@ -1710,9 +1713,17 @@ func (e *Engine) getOrCreateInteractiveStateWith(sessionKey string, p Platform,
 		return state
 	}
 
-	agentSID := session.GetAgentSessionID()
+	// On first connection after engine startup, use --continue to pick up
+	// the most recent CLI session (bridges direct CLI and cc-connect usage).
+	// Subsequent connections respect the stored session ID (or "" for fresh).
+	startSessionID := session.GetAgentSessionID()
+	if !e.hasConnectedOnce.Swap(true) {
+		startSessionID = ContinueSession
```
quabug: The `startSessionID` should never be changed to the `ContinueSession` string because this causes problems for other agents and only fixes a particular issue in claude code.
Instead of a common location like this, this type of thing should be placed in the claude code module.

Issue #253, opened 2026-03-22
[Bug] Discord thread_isolation: other bots reply in main channel instead of thread when multiple bots share a channel


quabug:
## Problem

When `thread_isolation = true` and multiple cc-connect bots share the same Discord channel, only the first bot responds in the thread. The other bots reply in the main channel instead of joining the existing thread.

For example, sending `@Claude @Codex @Gemini hi` in a channel — the first bot to process the message creates a thread and responds there, but the remaining bots respond in the main channel.

## Root Cause

Each bot independently calls `resolveThreadReplyContext`, which tries to create a new thread from the same message. The first bot succeeds; subsequent bots fail because Discord only allows one thread per message. The failing bots fall back to replying in the main channel instead of joining the existing thread.

## Expected Behavior

All mentioned bots should detect the existing thread (created by the first bot) and respond there, rather than falling back to the main channel.

## Steps to Reproduce

1. Configure 2+ Discord bots in the same channel with:
   ```toml
   thread_isolation = true
   ```
2. Send `@Claude @Codex @Gemini hi` in the channel
3. The first bot creates a thread and responds there
4. The other bots respond in the main channel instead of the thread

## Environment

- cc-connect v1.2.2-beta
- Platform: Discord
- Agent: any (reproduces with all agent types)

chenhg5:
感谢报告！

关于 Discord 线程隔离问题，当多个 Bot 在同一频道时，其他 Bot 在主频道回复而不是线程中，这确实是个 bug。

我们会在后续版本中调查并修复这个问题。

--
cc-connect/pm

PR #254, opened 2026-03-22
fix(discord): multi-bot thread_isolation support

quabug:
## Summary
- When multiple bots with `thread_isolation = true` are triggered by the same message, only the first bot's `StartThread` succeeds. Other bots now fall back to `JoinThread(messageID)` since Discord thread ID == parent message ID (#253)
- Thread names convert raw `<@ID>` mentions to readable `@username` instead of stripping them
- Mention stripping moved after thread resolution so thread names preserve bot names

## Test plan
- [x] `go test ./platform/discord/ -v` — all pass
- [x] `go build ./...` — builds clean
- [ ] Manual: 2+ bots with `thread_isolation = true`, send `@Bot1 @Bot2 hi` — all bots respond in same thread
- [ ] Manual: 2+ bots with `respond_to_at_everyone_and_here = true`, send `@everyone hi` — all bots respond in same thread


quabug (commented):
## Multi-Agent PR Review

### Summary
Well-implemented fix for multi-bot thread isolation race conditions with good test coverage. All issues found are minor — no critical or major bugs detected.

### General Findings

All findings have been posted as inline comments on the relevant lines. Key themes:

1. **Unused `botID` parameter** in `threadNameForMessage` — the parameter is accepted but no longer used after the switch to `stripAllMentions`
2. **Behavioral change in mention stripping** — now strips ALL user mentions (not just bot's), which changes thread name semantics. PR description says mentions are "converted to readable @username" but code strips them entirely
3. **Regex precision** — `[^>&]+` could be tightened to `\d+` to match Discord's numeric-only IDs
4. **Fallback catches all errors** — `StartThread` failure fallback doesn't distinguish "thread exists" from other error types
5. **Error wrapping inconsistency** — `%w` for primary error but `%v` for join error
6. **Missing role mention test** — `stripAllMentions` role handling is untested

### Suggestions
- Remove the unused `botID` parameter from `threadNameForMessage` (or restore bot-specific stripping)
- Add test case for role mention stripping with populated `MentionRoles`
- Consider tightening the regex to `<@!?\d+>` for robustness

### Notes
- **Dismissed**: "old `stripDiscordMention` is dead code" (Qwen Kimi-K2.5) — it's still used via `stripDiscordMentionWithRole` for message content stripping at line 474
- **Dismissed**: "magic number 1440" (Qwen Qwen3.5-Plus) — pre-existing value, not introduced by this PR
- **Agents that failed**: Gemini (rate limited), Qwen Minimax-M2.5 (API key error), Pi MiniMax-M2.7 (insufficient balance)

### Agent Attribution
| Finding | Claude Code | Codex | Qwen (GLM-5) | Qwen (Kimi-K2.5) | Qwen (Qwen3.5-Plus) |
|---------|:-:|:-:|:-:|:-:|:-:|
| Unused `botID` param | x | x | | x | |
| Behavioral change (all mentions stripped) | x | x | x | x | |
| Regex precision | | | x | x | x |
| Missing role mention test | | | x | x | x |
| Error wrapping inconsistency | | | x | x | |
| Fallback catches all errors | | x | | | x |
| PR description/code mismatch | x | | | | |

---
### Claude Code's Own Findings
- **PR description/implementation mismatch**: The PR body states "Thread names convert raw `<@ID>` mentions to readable `@username` instead of stripping them" but the implementation strips all mentions entirely via `stripAllMentions`. The description should be updated to match the actual behavior.

---
Reviewed by: Claude Code + Codex + Qwen (GLM-5) + Qwen (Kimi-K2.5) + Qwen (Qwen3.5-Plus)

chenhg5 (commented):
## Conclusion: Comment

The fix logic is sound and addresses an open bug (#253, still open) with a sensible "fall back to JoinThread when StartThread fails" strategy plus cleaner thread names. However, this PR has been open for ~2.5 months, is CONFLICTING against main, and the description/implementation does not match. Needs a rebase + description fix + manual verification before merge.

## What works (logic-wise)

- The multi-bot fallback (lines 282-294): when `StartThread` fails, try `JoinThread(m.ID)` because Discord thread ID == parent message ID. This is a real solution to #253 — when 2+ bots see the same `@Bot1 @Bot2 hi` message concurrently, the first bot's `StartThread` succeeds; the second's call returns an error (thread already exists on the message) and now correctly falls back to joining the existing thread.
- `stripAllMentions` (lines 218-227) is a strict improvement over `stripDiscordMention` for thread-name purposes: it removes user mentions, nick mentions, role mentions, `@everyone`, and `@here`, leaving only the actual message content.
- Test coverage is solid: `TestResolveThreadReplyContext_JoinsExistingThreadOnCreateFailure` covers both success and failure paths; `TestThreadNameForMessage` has 10 cases including role mentions.

## What blocks merge

1. **CONFLICTING + stale** — the PR is 2.5 months old. `resolveThreadReplyContext`'s return signature is now `(string, replyContext, string, error)` (4-tuple, added `parentChannelID`) on main, but this PR's code still returns the 3-tuple. Rebase needs to thread `parentChannelID = m.ChannelID` through the new fallback branch.
2. **`threadNameForMessage` signature drift** — your branch tip `73539bd0` collapsed it to 1-arg, but main still has the 2-arg form `func threadNameForMessage(m *discordgo.MessageCreate, botID string)` because `stripDiscordMention` is also called from `OnMessageCreate` at line 587 (non-thread context). Cleanest port: keep 2-arg signature, replace `stripDiscordMention(m.Content, botID)` with `stripAllMentions(m.Content, m.MentionRoles)` inside the function, ignore `botID`.
3. **Description mismatch (you flagged this in your own self-review)** — the body says "convert raw `<@ID>` mentions to readable `@username`" but the implementation strips them entirely. Update the description to match the actual behavior ("strips all mentions to leave message content").
4. **Manual test plan still unchecked** — the 2 unchecked items ("2+ bots with `thread_isolation = true`" and "2+ bots with `respond_to_at_everyone_and_here = true`") require real Discord credentials. Either run them on a staging guild with 2 bots configured, or document that this is a community contribution without maintainer-staffed staging access and request a maintainer smoke-test before merge.
5. **CI stale by 2.5 months** — the green check on run 23778206591 (2026-03-31) is on the original commit, not after rebase. After rebasing onto current main, CI must run fresh.

## P2 suggestions (good to have, not blocking)

- **Role mention regex tightening**: `reDiscordUserMention = regexp.MustCompile(\`<@!?\d+>\`)` (good). But the role-mention loop uses `strings.ReplaceAll(text, "<@&"+roleID+">", "")` which won't match a `<@!123>`-style nick form (roles don't have nick form, so this is fine). Consider extracting role stripping to a regex for consistency.
- **Fallback "thread exists" vs other error distinction**: catching all `StartThread` errors and trying `JoinThread` is broad. If `StartThread` fails for a reason other than "thread exists" (e.g., missing `MANAGE_THREADS` permission, channel archived), the `JoinThread` fallback will also fail with a less informative error. Consider narrowing the fallback to known recoverable errors. Current code does wrap both errors which is OK for debugging.
- **Error message format**: `fmt.Errorf("start thread for message %s: %w (join also failed: %w)", ...)` — Go 1.20+ allows multiple `%w`, which is good. Just verify CI runs on Go ≥ 1.20.

## Test design feedback

- The two sub-tests in `TestResolveThreadReplyContext_JoinsExistingThreadOnCreateFailure` correctly exercise the success and failure paths. Good use of `fakeThreadOps` table-driven pattern.
- `TestThreadNameForMessage` covers user mentions, nick mentions, multiple mentions, `@everyone`, `@here`, mixed, role mentions, and the empty-content fallbacks. Solid.
- One gap: no test asserts the **race-condition scenario from #253** — that 2+ bots calling `resolveThreadReplyContext` on the same message both end up in the same thread. The current test only verifies the single-bot fallback path. Consider adding `TestResolveThreadReplyContext_TwoBotsSameMessage_EndInSameThread` (sequential simulation: first call's `StartThread` succeeds, second call's `StartThread` returns the "exists" error, then second call's `JoinThread` succeeds, both should return the same session key).

## Risk

- **Low (logic)**: the StartThread→JoinThread fallback is additive. On success path (no concurrent bot), behavior is identical. On concurrent-bot path, it fixes a real bug.
- **Medium (operational)**: needs manual verification on a multi-bot guild before merge. The CI test only proves the API surface, not the Discord behavior.
- **No migration concern**: no config schema change, no breaking API change.

## Next step

Author (or maintainer with merge rights):

1. Rebase onto current main, preserving the 4-tuple return for `resolveThreadReplyContext` and the 2-arg `threadNameForMessage` signature.
2. Update PR description to match implementation ("strips mentions" not "converts to @username").
3. Run the 2 manual test items on a staging guild with 2+ bots, or document maintainer smoke-test request.
4. Add a multi-bot sequential test to lock in #253's race-condition fix.
5. Re-run CI; expect fresh green on all 5 jobs.

After these 5 items, the PR is ready to merge. Happy to re-review once rebased.

PR #294, opened 2026-03-23
fix(core): preserve CC_SESSION_KEY in multi-workspace sessions

### LINKED ISSUE #293: [Bug] multi-workspace sessions inject malformed CC_SESSION_KEY
## Summary

In multi-workspace mode, normal chat messages start workspace-scoped sessions with a workspace-prefixed interactive key, and that key is currently leaking into the agent subprocess as `CC_SESSION_KEY`.

This breaks agent-side commands that expect a platform-form session key, for example `cc-connect relay send`.

## Repro

1. Enable multi-workspace mode and bind a chat/channel to a workspace.
2. Start a normal interactive session in that bound chat.
3. Inside the agent subprocess, inspect `CC_SESSION_KEY` or run `cc-connect relay send ...`.

## Actual

`CC_SESSION_KEY` looks like:

```text
/workspace/path:discord:<chat_id>
```

or in general:

```text
<workspace>:<platform>:<channel_or_chat>:<user>
```

That value is the internal interactive key, not the platform session key.

## Expected

`CC_SESSION_KEY` should remain the original platform session key, for example:

```text
discord:<chat_id>:relay
```

or for normal sessions:

```text
discord:<channel_or_chat>:<user>
```

## Root Cause

In the multi-workspace message path, `handleMessage` calls:

```go
go e.processInteractiveMessageWith(..., interactiveKey, resolvedWorkspace, "")
```

Then `getOrCreateInteractiveStateWith()` falls back to `sessionKey` for env injection, but in this case `sessionKey` is the workspace-prefixed interactive key.

So the agent subprocess receives the wrong `CC_SESSION_KEY`.

## Scope

This issue is about malformed `CC_SESSION_KEY` injection only.

There is a separate relay/workspace-routing concern: `HandleRelay()` does not currently resolve effective workspace bindings. That should stay separate from this bug.

## Proposed Fix

Pass the original `msg.SessionKey` as `ccSessionKey` in the multi-workspace normal-message path, so internal state can still use the workspace-prefixed interactive key while the subprocess env keeps the clean platform session key.

xxb:
## Summary

Fixes #293.

Preserve the original platform session key in `CC_SESSION_KEY` when a normal chat message is handled through the multi-workspace path.

Previously the engine passed the workspace-prefixed interactive key through the fallback path, so the agent subprocess could receive values like:

```text
/workspace/path:discord:<channel>:<user>
```

instead of the clean platform session key:

```text
discord:<channel>:<user>
```

That breaks agent-side commands such as `cc-connect relay send` which expect the standard session-key format.

## Change

- pass `msg.SessionKey` as `ccSessionKey` in the multi-workspace normal-message path
- add a regression test that verifies `CC_SESSION_KEY` does not leak the workspace-prefixed interactive key

## Testing

- `nix-shell -p go --run 'gofmt -w core/engine.go core/engine_test.go && go test ./core -run "TestHandleMessage_MultiWorkspacePreservesCCSessionKey|TestHandlePendingPermission_MultiWorkspaceLookup"'`
- `nix-shell -p go --run 'go test ./...'`

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1324,7 +1324,7 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 		"session", session.ID,
 	)
 
-	go e.processInteractiveMessageWith(p, msg, session, agent, sessions, interactiveKey, resolvedWorkspace, "")
+	go e.processInteractiveMessageWith(p, msg, session, agent, sessions, interactiveKey, resolvedWorkspace, msg.SessionKey)
 }
 
 // queueMessageForBusySession queues a message for later delivery when the
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -111,6 +111,38 @@ func (a *resultAgent) StartSession(_ context.Context, _ string) (AgentSession, e
 func (a *resultAgent) ListSessions(_ context.Context) ([]AgentSessionInfo, error) { return nil, nil }
 func (a *resultAgent) Stop() error                                                { return nil }
 
+type sessionEnvRecordingAgent struct {
+	stubAgent
+	session AgentSession
+	mu      sync.Mutex
+	env     []string
+}
+
+func (a *sessionEnvRecordingAgent) StartSession(_ context.Context, _ string) (AgentSession, error) {
+	if a.session != nil {
+		return a.session, nil
+	}
+	return &stubAgentSession{}, nil
+}
+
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
LGTM. CC_SESSION_KEY correctly passed to multi-workspace agents without leaking workspace paths. Good regression test.

PR #305, opened 2026-03-24
fix: resolve workspace for cron jobs in multi-workspace mode

### LINKED ISSUE #302: Cron-triggered responses not delivered to messaging platform
## Description

When using `CronCreate` to schedule recurring prompts, the cron fires and Claude processes the prompt, but the response is not reliably delivered to the messaging platform (Slack in our case).

## Reproduction

1. Set up cc-connect bridge to Slack
2. Create a recurring cron:
   ```
   CronCreate: cron="*/5 * * * *", prompt="Say hello", recurring=true
   ```
3. Wait for the cron to fire

## Expected behavior

The response to the cron-triggered prompt should be delivered to Slack as a message, the same way a normal response to a user message is delivered.

## Actual behavior

- The cron fires (confirmed by the prompt appearing in the conversation)
- Claude processes the prompt and generates a response
- The response is **not** delivered to Slack
- Intermittently, it does work — a one-shot cron test (`recurring: false`) was delivered successfully, but recurring crons during an active conversation are not reliably delivered

## Context

This is being used to monitor long-running Temporal workflows. The cron checks workflow status every few minutes and should post updates to Slack so the user can monitor progress without being at the terminal.

In native Claude Code (terminal REPL), cron responses appear in stdout when the session is idle. The cc-connect bridge should relay these to the messaging platform the same way it relays normal responses.

## Environment

- cc-connect bridging to Slack
- Claude Code with Claude Opus 4.6
- Crons created via the `CronCreate` tool

leighstillard:
## Summary

Fixes #302 — cron-triggered responses not delivered to messaging platform.

• In multi-workspace mode, `ExecuteCronJob` was using the engine-level agent and session manager instead of the workspace-specific ones
• The `interactiveKey` was bare (`slack:channel:user`) instead of prefixed with the workspace path, creating an orphaned state disconnected from the user's conversation
• Now resolves the workspace binding and uses the correct per-workspace agent, sessions, and `interactiveKey` — mirroring what `handleMessage` already does for user messages

## Root cause

When a cron fires in multi-workspace mode, it needs the same workspace resolution that user messages get. Without it:

1. The agent has no workspace context (wrong `workDir`) — Claude Code can't find the project files
2. The `interactiveKey` doesn't match the user's conversation key — creates a separate orphaned `interactiveState`
3. The session manager is engine-level, not workspace-specific — session tracking is broken

One-shot crons worked intermittently because on a fresh start with no active conversation, the bare key didn't conflict.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./core/ ./agent/claudecode/` passes (636 tests)
- [x] New regression test: `TestExecuteCronJob_MultiWorkspace_NoBinding` — verifies cron returns error for unbound channel instead of silently failing
- [ ] Manual: set up recurring cron in multi-workspace mode, verify responses are delivered to Slack

### DIFF
--- a/core/cron_test.go
+++ b/core/cron_test.go
@@ -530,3 +530,73 @@ func TestCronStore_ListByProject(t *testing.T) {
 		t.Errorf("ListByProject(nonexistent) = %d jobs, want 0", len(list3))
 	}
 }
+
+// TestExecuteCronJob_MultiWorkspace_NoBinding verifies that in multi-workspace
+// mode, ExecuteCronJob returns an error when the channel has no workspace binding
+// (regression test for #302: cron responses not delivered because workspace
+// resolution was skipped entirely).
+func TestExecuteCronJob_MultiWorkspace_NoBinding(t *testing.T) {
+	p := &cronTestPlatform{name: "slack"}
+	e := NewEngine("test", &stubCronAgent{}, []Platform{p}, "", LangEnglish)
+	e.SetMultiWorkspace(t.TempDir(), "")
+
+	job := &CronJob{
+		ID:         "j1",
+		Project:    "test",
+		SessionKey: "slack:C123:U456",
+		Prompt:     "hello",
+		Enabled:    true,
+	}
+
+	err := e.ExecuteCronJob(job)
+	if err == nil {
+		t.Fatal("expected error for unbound channel, got nil")
+	}
+	if !strings.Contains(err.Error(), "no workspace bound") {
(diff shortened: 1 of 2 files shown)


leighstillard:
OK I nailed it, ready to merge assuming tests pass.

chenhg5 (approved):
LGTM. Fixes #302 correctly — stripping the workspace prefix from legacy session keys before retrying platform lookup. Regression test is comprehensive. CI green.

leighstillard:
Closing - cron workspace fix was merged via #474

PR #310, opened 2026-03-24
fix(weixin): fix AESKey encoding to match ilink API format

xudaashuai:
## Summary

- In `SendImage` and `SendFile`, `AESKey` was encoded as `base64(raw 16 bytes)`, producing a 24-character string
- The ilink API expects `base64(hex string)`, producing a 44-character string — consistent with how `Aeskey` is already encoded in `getUploadURLRequest`
- Fixed both occurrences (lines 107 and 138 of `platform/weixin/media_outbound.go`)

## Encoding comparison

| | Format | Example (16-byte key) |
|---|---|---|
| Before | `base64(raw bytes)` | `AQIDBAUGBwgJCgsMDQ4PEA==` (24 chars) |
| After | `base64(hex string)` | `MDEwMjAzMDQwNTA2MDcwODA5MGEwYjBjMGQwZTBm` (44 chars) |

No new imports needed — `encoding/hex` was already used in the same file.

### DIFF
--- a/platform/weixin/media_outbound.go
+++ b/platform/weixin/media_outbound.go
@@ -104,7 +104,7 @@ func (p *Platform) SendImage(ctx context.Context, replyCtx any, img core.ImageAt
 		ImageItem: &imageItem{
 			Media: &cdnMedia{
 				EncryptQueryParam: ref.downloadParam,
-				AESKey:            base64.StdEncoding.EncodeToString(ref.aesKey),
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. AESKey encoding fix — base64(hex string) instead of base64(raw bytes) to match ilink API format. Simple, targeted fix with clear explanation. CI green.

PR #319, opened 2026-03-25
fix(opencode): add warn logs when sqlite3 CLI is missing

### LINKED ISSUE #318: [Bug] Bug: OpenCode agent silently fails to display message counts if `sqlite3` CLI is missing
### cc-connect Version

Latest

### Operating System

Linux (Ubuntu/Debian)

### Agent Type

Claude Code

### Platform

N/A

### Installation Method

None

## Describe the bug
When using the OpenCode agent integration, the `/list` command shows `0` messages for all sessions, even though the conversation proceeds normally and `/history` correctly displays the current session's history. 

This occurs because `cc-connect` relies on the external `sqlite3` command-line tool to query OpenCode's local database for message counts. If `sqlite3` is not installed on the host system or Docker container, the code silently fails and defaults to returning zero counts without logging any errors

## Code context
The silent failure happens in `agent/opencode/opencode.go` inside the `querySessionMessageCounts()` function:

```go
func querySessionMessageCounts() map[string]int {
	// ...
	sqlite3, err := exec.LookPath("sqlite3")
	if err != nil {
		return nil // SILENT FAILURE
	}

	out, err := exec.Command(sqlite3, dbPath,
		"SELECT session_id, COUNT(*) FROM message GROUP BY session_id").Output()
	if err != nil {
		return nil // SILENT FAILURE
	}
    // ...
}

```

Suggested Fix
 * Short term: Add slog.Warn("opencode: failed to query message counts", "err", err) before returning nil when the CLI check or command execution fails.
 * Long term: Refactor querySessionMessageCounts to use a pure Go SQLite driver (such as modernc.org/sqlite or github.com/mattn/go-sqlite3) to eliminate the external dependency entirely.

## Steps to reproduce
1. Run `cc-connect` in an environment without the `sqlite3` CLI package installed (e.g., a minimal Docker container).
2. Start an OpenCode session and send a few messages.
3. Run the `/list` command in your chat platform.
4. **Observe:** The session appears in the list, but it shows `0` messages.

## Expected behavior
* At a minimum, `cc-connect` should log a warning or error to the console/system log when `exec.LookPath("sqlite3")` or the query fails, instead of silently swallowing the error.
* Ideally, `cc-connect` should not rely on external shell dependencies to read the database.

### Actual Behavior

Should work

```toml

```

### Logs / Error Output

```shell

```

### Additional Context

_No response_

chenhg5:
## Summary
- Add `slog.Warn` when sqlite3 CLI is not found in PATH (opencode /list shows 0 messages silently)
- Add `slog.Warn` when sqlite3 query execution fails

Fixes #318.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes  
- [x] `go test ./agent/opencode/...` passes

Closes #318

### DIFF
--- a/agent/opencode/opencode.go
+++ b/agent/opencode/opencode.go
@@ -301,12 +301,14 @@ func querySessionMessageCounts() map[string]int {
 	}
 	sqlite3, err := exec.LookPath("sqlite3")
 	if err != nil {
+		slog.Warn("opencode: sqlite3 CLI not found, message counts unavailable", "err", err)
 		return nil
 	}
 
 	out, err := exec.Command(sqlite3, dbPath,
 		"SELECT session_id, COUNT(*) FROM message GROUP BY session_id").Output()
 	if err != nil {
+		slog.Warn("opencode: sqlite3 query failed", "db_path", dbPath, "err", err)
 		return nil
 	}
 


chenhg5:
Good fix — adds warn logs for sqlite3 CLI missing or query failures, making /list silently showing 0 messages diagnosable. Fixes #318. Owner PR so I cannot approve — please self-approve.

PR #320, opened 2026-03-25
feat(opencode): add optional attach support

IsidoreSong:
## Summary
This PR adds optional `attach` support to the OpenCode agent integration.

When configured, cc-connect will invoke OpenCode with:

```bash
opencode run --attach <url> ...
```

instead of only using the default CLI cold-start path.

## Motivation
OpenCode officially supports `serve` + `run --attach` as a faster integration path for reusing an existing server instance.

In chat-bridge scenarios like WeCom / Feishu, repeated CLI cold starts add noticeable latency to each reply. An optional `attach` mode helps reduce that overhead without changing the default behavior for existing users.

### `agent/opencode/opencode.go`
- Added `attach` field to the OpenCode agent config
- Read `attach` from agent options

### `agent/opencode/session.go`
- When `attach` is configured, append:
  ```text
  --attach <url>
  ```
  to `opencode run`

## Example config
```toml
[projects.agent.options]
work_dir = "C:/path/to/project"
mode = "default"
attach = "http://127.0.0.1:4096"
```

## How to start OpenCode server
Before using `attach`, start a reusable OpenCode server separately:

```bash
opencode serve --hostname 127.0.0.1 --port 4096
```

Then configure cc-connect with:

```toml
attach = "http://127.0.0.1:4096"
```

## Behavior
- If `attach` is not configured, behavior stays unchanged
- If `attach` is configured, OpenCode runs against an existing `opencode serve` instance

## Test plan
- Built locally with:
  ```bash
  go build ./cmd/cc-connect
  ```
- Verified that cc-connect still starts normally with no `attach` configured
- Started OpenCode server with:
  ```bash
  opencode serve --hostname 127.0.0.1 --port 4096
  ```
- Verified that cc-connect starts normally with:
  ```toml
  attach = "http://127.0.0.1:4096"
  ```
- Verified in a real WeCom bridge scenario that attach mode improves response latency perceptibly compared to the default cold-start path

## Notes
This PR does not auto-start `opencode serve`; it only supports attaching to an already running OpenCode server.

This PR intentionally does **not** change OpenCode session continuation behavior. It only adds optional attach support so users can explicitly opt into OpenCode server reuse.

### DIFF
--- a/agent/opencode/opencode.go
+++ b/agent/opencode/opencode.go
@@ -28,6 +28,7 @@ type Agent struct {
 	workDir    string
 	model      string
 	mode       string
+	attach     string
 	cmd        string // CLI binary name, default "opencode"
 	providers  []core.ProviderConfig
 	activeIdx  int
@@ -43,6 +44,7 @@ func New(opts map[string]any) (core.Agent, error) {
 	model, _ := opts["model"].(string)
 	mode, _ := opts["mode"].(string)
 	mode = normalizeMode(mode)
+	attach, _ := opts["attach"].(string)
 	cmd, _ := opts["cmd"].(string)
 	if cmd == "" {
 		cmd = "opencode"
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
LGTM. Optional attach support for OpenCode agent — adds attach config field that's passed through to the session. Small, focused change. CI green.

PR #321, opened 2026-03-25
feat(telegram): migrate to go-telegram/bot and add forum topic support

zhengbuqian:
With this change, we can now easily manage multiple concurrent sessions in different directories, by creating a group chat and deal with each directory in a new topic, avoiding the frequent session switch operation.

Migrate the Telegram platform from the unmaintained go-telegram-bot-api/v5 (last release 2021) to go-telegram/bot v1.20.0, which is actively maintained and supports the full Telegram Bot API including forum topics.

Key changes:

- **Library migration**: Replace go-telegram-bot-api/v5 with go-telegram/bot. Redesign the telegramBot interface around the new library's typed API methods (SendMessage, SendPhoto, EditMessageText, etc.) instead of the old generic Send/Request pattern. The *bot.Bot type satisfies the interface directly.

- **Forum topic support**: Add threadID to replyContext and session key. Messages in forum groups (IsForum=true) now include the MessageThreadID in the session key (telegram:{chatID}:{threadID}:{userID}), giving each topic its own independent agent session. All send methods propagate MessageThreadID so replies land in the correct topic. Non-forum groups and private chats are unaffected.

- **Per-topic workspace binding**: Add ChannelKey field to core.Message so platforms can provide an explicit channel identifier for workspace binding. Telegram sets this to "chatID:threadID" for forum topics, enabling multi-workspace mode to bind different topics to different project directories. Add effectiveChannelID/effectiveWorkspaceChannelKey helpers and update all workspace resolution call sites in the engine.

- **Skill command registration fix**: GetAllCommands now respects disabled_commands for skills, matching the existing behavior for built-in commands. This prevents disabled skills from being registered as Telegram bot menu commands.

- **Connection management**: Simplify the reconnect architecture. bot.New() handles initial validation, bot.Start(ctx) manages long polling internally. Remove the stale bot detection (generation comparison in handlers) since the new library's blocking Start() ensures handlers are never called from a superseded bot instance.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -3185,7 +3185,7 @@ func (e *Engine) cmdNew(p Platform, msg *Message, args []string) {
 	if len(args) > 0 {
 		name = strings.Join(args, " ")
 	}
-	s = sessions.NewSession(msg.SessionKey, name)
+	sessions.NewSession(msg.SessionKey, name)
 	if name != "" {
 		e.reply(p, msg.ReplyCtx, fmt.Sprintf(e.i18n.T(MsgNewSessionCreatedName), name))
 	} else {
--- a/platform/telegram/telegram_test.go
+++ b/platform/telegram/telegram_test.go
@@ -222,17 +222,6 @@ func (b *stubTelegramBot) GetFileCallCount() int {
 	return b.getFileCalls
 }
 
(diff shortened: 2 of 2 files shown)


zhengbuqian:
@chenhg5 fixed the lint issue and verified all other ci checks to pass locally, please help approve the CI workflow again, thanks!

PR #326, opened 2026-03-25
fix(cursor): use StdinPipe to prevent agent hang under launchd

AndrewZhuCC:
## Problem

When cc-connect runs as a launchd daemon, the Cursor agent CLI process
hangs on startup — `cursorSession: launching` appears in the log but no
`cursorSession: raw` output follows, and the process never exits.

This started after the Cursor agent CLI updated to version `2026.03.20-44cb435`.

## Root Cause

`session.go` does not set `cmd.Stdin`. Go's `exec.Cmd` inherits the parent's
stdin by default. Under launchd the parent stdin is `/dev/null`, which causes
the newer Node.js-based agent CLI to hang.

## Fix

Add `cmd.StdinPipe()` — the same approach already used in
`agent/claudecode/session.go`. The pipe is kept open for the lifetime of
`readLoop` and closed when the goroutine exits.

## Testing

- Added two regression tests in `agent/cursor/session_test.go`:
  - `TestSendSetsStdinPipe`: verifies child gets a pipe stdin
  - `TestSendStdinNotInherited`: simulates launchd by setting parent stdin to /dev/null
- Verified on macOS with launchd daemon: agent now produces output and
  `turn complete` appears in the log.
- `go test ./...` passes.

### DIFF
--- a/agent/cursor/session.go
+++ b/agent/cursor/session.go
@@ -107,6 +107,14 @@ func (cs *cursorSession) Send(prompt string, images []core.ImageAttachment, file
 	}
 	cmd.Env = env
 
+	// Explicit StdinPipe prevents the child from inheriting the parent's stdin.
+	// Under launchd, the parent stdin is /dev/null which can cause the Cursor
+	// agent CLI (Node.js) to hang on startup in newer versions.
+	stdinPipe, err := cmd.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("cursorSession: stdin pipe: %w", err)
+	}
+
 	stdout, err := cmd.StdoutPipe()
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Good fix for launchd stdin hang issue.

**Review notes:**
- ✅ CI passes
- ✅ Correct approach: StdinPipe prevents inheritance of parent's stdin
- ✅ Tests cover both scenarios (pipe set, stdin not inherited)
- ✅ Proper cleanup: stdinPipe closed in defer
- ✅ Clear documentation in test comments explaining the launchd behavior

This is a practical fix for a real production issue on macOS. Approved for merge.

PR #328, opened 2026-03-25
fix(core): route relay through multi-workspace bindings

xxb:
## Summary

This fixes bot-to-bot relay in `multi-workspace` mode.

Before this change, relay requests only passed `chatID` into the target engine, so the target side bypassed normal workspace routing and used the global agent/session manager. That caused relayed messages to run in the wrong workspace.

## What changed

- pass the full source `sessionKey` through relay handling
- resolve relay workspace on the target side from the source session context
- route relay turns to the workspace-specific agent and session manager in `multi-workspace` mode
- use a platform-qualified relay session key to avoid cross-platform collisions
- keep `CC_SESSION_KEY` set to the original source session key for agent-side relay/cron context
- add regression tests for:
  - single-workspace relay staying on the global agent
  - multi-workspace relay routing by source session key
  - multi-workspace relay failing cleanly when no workspace binding exists
  - existing relay timeout behavior

## Behavior note

In `multi-workspace` mode, relay from an unbound source channel now fails explicitly instead of silently falling back to the global agent. This is intentional so relay behavior matches workspace-aware routing expectations.

## Validation

```bash
nix shell nixpkgs#go -c go test ./core -run 'TestHandleRelay|TestSessionContextForKey|TestMultiWorkspaceResolution' -count=1
nix shell nixpkgs#go -c go test ./core -count=1
nix shell nixpkgs#go -c go test ./... -run '^$' -count=1
```

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -10948,17 +10948,76 @@ func (e *Engine) sendTTSReply(p Platform, replyCtx any, text string) {
 // Bot-to-bot relay
 // ──────────────────────────────────────────────────────────────
 
+type platformNameOnly struct {
+	name string
+}
+
+func (p platformNameOnly) Name() string                           { return p.name }
+func (platformNameOnly) Start(MessageHandler) error               { return nil }
+func (platformNameOnly) Reply(context.Context, any, string) error { return nil }
+func (platformNameOnly) Send(context.Context, any, string) error  { return nil }
+func (platformNameOnly) Stop() error                              { return nil }
+
+func relayConversationKey(fromProject, platformName, chatID string) string {
+	return "relay:" + fromProject + ":" + workspaceChannelKey(platformName, chatID)
+}
+
+func (e *Engine) platformForName(name string) Platform {
+	for _, p := range e.platforms {
+		if strings.EqualFold(p.Name(), name) {
+			return p
+		}
+	}
+	return platformNameOnly{name: name}
+}
+
+func (e *Engine) relayContextForSourceSessionKey(fromProject, sourceSessionKey string) (Agent, *SessionManager, string, error) {
+	platformName, chatID, err := parseSessionKeyParts(sourceSessionKey)
+	if err != nil {
+		return nil, nil, "", fmt.Errorf("invalid source session key: %w", err)
+	}
+
+	relaySessionKey := relayConversationKey(fromProject, platformName, chatID)
+	if !e.multiWorkspace || e.workspaceBindings == nil {
+		return e.agent, e.sessions, relaySessionKey, nil
+	}
+
+	channelKey := workspaceChannelKey(platformName, chatID)
+	workspace, _, err := e.resolveWorkspace(e.platformForName(platformName), chatID)
+	if err != nil {
+		return nil, nil, "", fmt.Errorf("resolve relay workspace: %w", err)
+	}
+	if workspace == "" {
+		if b, _, usable := e.lookupEffectiveWorkspaceBinding(channelKey); b != nil && !usable {
+			return nil, nil, "", fmt.Errorf("workspace binding unavailable for source channel %q", channelKey)
+		}
+		return nil, nil, "", fmt.Errorf("no workspace binding for source channel %q", channelKey)
+	}
+
+	agent, sessions, err := e.getOrCreateWorkspaceAgent(workspace)
+	if err != nil {
+		return nil, nil, "", fmt.Errorf("get relay workspace agent: %w", err)
+	}
+	if ws := e.workspacePool.Get(workspace); ws != nil {
+		ws.Touch()
+	}
+	return agent, sessions, relaySessionKey, nil
+}
+
 // HandleRelay processes a relay message synchronously: starts or resumes a
 // dedicated relay session, sends the message to the agent, and blocks until
 // the complete response is collected (or the relay context times out).
-func (e *Engine) HandleRelay(ctx context.Context, fromProject, chatID, message string) (string, error) {
-	relaySessionKey := "relay:" + fromProject + ":" + chatID
-	session := e.sessions.GetOrCreateActive(relaySessionKey)
+func (e *Engine) HandleRelay(ctx context.Context, fromProject, sourceSessionKey, message string) (string, error) {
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
LGTM. Well-implemented multi-workspace relay routing.

**Review notes:**
- ✅ CI passes
- ✅ Correct approach: routes relay through workspace bindings
- ✅ Uses `sourceSessionKey` to resolve workspace properly
- ✅ Falls back to global agent/sessions when not multi-workspace
- ✅ Tests cover both single and multi-workspace scenarios
- ✅ Error handling for missing workspace bindings

This enables relay to work correctly with workspace isolation. Approved for merge.

xxb:
@chenhg5 Quick ping on this one. It is approved, mergeable, and CI is green. The branch has already been refreshed against main, so it should be ready to merge when you have time.

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
Re-reviewing after head SHA change (was 486a3721, now 5447234f). The refactor cleanly routes `HandleRelay` through multi-workspace bindings: `relayContextForSourceSessionKey` resolves `(agent, sessions, relaySessionKey)` from the source session key's workspace binding instead of always using `e.agent`/`e.sessions`, and `HandleRelay` is updated to thread those per-workspace values through the existing Start/Send/Collect loop. CC_SESSION_KEY env var now uses `sourceSessionKey` (was `relaySessionKey`) so child agent processes see the original source chat. Session-ID persistence is extracted into a `saveRelaySessionID` closure to avoid duplicating the "set + name + Save" pattern in two places. No behavioral regressions, all 5 CI checks green, go vet clean.

Review scope:
- `core/engine.go` — new helpers (`platformNameOnly`, `platformForName`, `relayConversationKey`, `relayContextForSourceSessionKey`); `HandleRelay` rewritten.
- `core/relay.go` — 2-line change.
- `core/relay_test.go` — 3 new tests + expansions to existing tests.
- Diff vs `origin/main`: 3 files, +189/-39.
- Focused on correctness of workspace routing, session-key/env-var semantics, and test coverage.

✅ What looks good:
- `relayContextForSourceSessionKey` correctly resolves per-workspace `agent` + `sessions` while keeping `relaySessionKey` derived from `fromProject + sourceSessionKey`. The agent field is no longer hard-coded to `e.agent`, so each workspace's bound agent (and its session store) is used for relay traffic originating in that workspace.
- Error path: when `multiWorkspace` is true but the source channel has no workspace binding (or binding is unusable), the function returns a descriptive error ("no workspace binding for source channel %q" / "workspace binding unavailable for source channel %q"). Callers can surface these as user-visible errors instead of silently misrouting relay traffic.
- `saveRelaySessionID` closure removes the previously-duplicated "compare-and-set + set session name + Save" sequence in two places (post-`StartSession` and in the EventResult handler). Fewer places to drift if the persistence rules change. The `force` parameter correctly captures the asymmetry: post-StartSession uses CAS (only update if not yet set), EventResult uses CAS too — so `force` is reserved for the post-fresh-session path where we know we want to overwrite an empty ID. Minor: `force` could be inferred from `session.GetAgentSessionID() == ""` if preferred, but the explicit parameter is clearer.
- `CC_SESSION_KEY` env var now passes `sourceSessionKey` instead of `relaySessionKey`. Verified by reading the agent context: `SessionEnvInjector` consumes this var to map back to the original source chat, so passing the source key (not the relay wrapper key) is the correct semantic. This matches what other platform-driven code paths already do.
- 3 new tests: `TestHandleRelay_SingleWorkspaceUsesGlobalAgentAndSourceSessionKey` (non-multi-workspace path, source session key preserved into env var), `TestHandleRelay_MultiWorkspaceRoutesBySourceSessionKey` (multi-workspace path, different source keys route to different workspace agents), `TestHandleRelay_MultiWorkspaceRequiresWorkspaceBinding` (multi-workspace + no binding → error). Together they pin the three meaningful code paths.
- CI run 24696107073: lint 1m13s, unit 2m46s, smoke 30s, regression 24s, performance 49s — all 5 green. `go vet -tags no_web ./core/...` clean. `go test -tags no_web -count=1 -run "TestRelay|RelayContext|TestHandleRelay" ./core/...` PASS locally.

🔵 Optional:
- `platformNameOnly` is an interesting defensive fallback: when `e.platforms` doesn't contain a platform matching `name` (e.g. source session key references a platform that's been removed since the session was created), it returns a stub whose `Start`/`Reply`/`Send`/`Stop` all return nil. This means the engine accepts the source session key without failing, then later will silently no-op when relay tries to send through the stub. Consider whether a louder failure (e.g. return an error from `relayContextForSourceSessionKey` when the platform can't be resolved) would be safer — at minimum, a `slog.Warn` when falling back to the stub would help debugging. Not blocking; the engine code already handles Send returning nil gracefully.
- `relayContextForSourceSessionKey` always returns `e.agent, e.sessions, relaySessionKey` in the non-multi-workspace branch. The `sessions` returned is `e.sessions`, which is correct. But the caller `HandleRelay` then passes `sessions.GetOrCreateActive(relaySessionKey)`. If multi-workspace is later enabled mid-session, in-flight relays might switch from global to per-workspace mid-flight — would the relay session be re-created against the new sessions store? Worth a follow-up if this is a real deployment scenario; otherwise the behavior is fine.

Testing / Risk:
- Verified evidence: 3 new test functions pass; existing `core` tests still pass; `go vet` clean; CI 5/5 green.
- Remaining risk: behavior change is limited to relay code paths when `multi_workspace` is enabled. Single-workspace deployments (the default) follow the original code path with only the env-var tweak (`sourceSessionKey` instead of `relaySessionKey`). If any downstream consumer depends on `CC_SESSION_KEY` containing the relay wrapper key, they will need to update — but a quick grep for `CC_SESSION_KEY` in the codebase suggests only `SessionEnvInjector` reads it and it expects the source key.
- Manual verification suggestion: integration test where project A relays a message to project B from a multi-workspace source channel; verify the relay session lands in the source workspace's session store (not the global one).

Next step:
OK to merge. After merge, consider whether to add a release-note entry under `### New Features` or `### Fixed` mentioning "Bot-to-bot relay now respects multi-workspace bindings — relay traffic originating in a workspace uses that workspace's agent + session store". Useful for users who rely on relay and have multi-workspace enabled.

PR #332, opened 2026-03-26
fix(codex): add panic recovery in readLoop to prevent process crash

LigphiDonk:
When handleItemCompleted encounters an unexpected event structure from the Codex CLI (e.g., a new item type or malformed JSON fields), it can panic. Since readLoop runs in a goroutine with no recover(), this crashes the entire cc-connect process.

Add a deferred recover() at the top of readLoop that:
- Catches any panic from handleEvent and its callees
- Logs the panic value and full stack trace via slog.Error
- Emits an EventError to the events channel so the caller is notified
- Allows the process to continue serving other projects/platforms

Also add TestHandleEvent_RecoverFromPanic to verify that malformed item.completed events (nil item, unknown types, unexpected nested structures) do not cause panics.

### DIFF
--- a/agent/codex/session.go
+++ b/agent/codex/session.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime/debug"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Good defensive programming with panic recovery.

**Review notes:**
- ✅ CI passes  
- ✅ Correct approach: recover() in readLoop prevents process crash
- ✅ Error properly propagated via Event channel
- ✅ Test coverage for malformed events
- ✅ Stack trace logged for debugging

This prevents cascading failures when codex CLI produces unexpected output. Approved for merge.

PR #344, opened 2026-03-27
fix(cron): add wake watchdog to recalibrate schedules after system sleep

mkloveyy:
## Summary
- Add a watchdog goroutine that detects system wake from sleep by monitoring wall-clock jumps
- When a gap > 2 minutes is detected, the cron scheduler is rebuilt so all jobs are rescheduled against the current wall clock
- Fixes cron timing drift caused by monotonic clock not advancing during macOS/Linux sleep

## Background
`robfig/cron` uses `time.Timer` which is based on monotonic clock. On macOS, the monotonic clock does not advance during sleep, causing cron jobs to be delayed or skipped after wake. Previously required external tools like `sleepwatcher` to work around this.

## Test plan
- [x] `go test ./core/ -run TestCron -v` passes
- [x] `go build ./...` passes
- [x] Verify wake detection by sleeping and waking the machine, checking logs for `cron: detected wake from sleep, rescheduling all jobs`

### DIFF
--- a/core/cron.go
+++ b/core/cron.go
@@ -401,8 +401,9 @@ type CronScheduler struct {
 	engines       map[string]*Engine // project name → engine
 	mu            sync.RWMutex
 	entries       map[string]cron.EntryID // job ID → cron entry
-	defaultSilent      bool   // global default for suppressing cron start notifications
-	defaultSessionMode string // global default session mode; "" = reuse, "new_per_run" = fresh session each run
+	defaultSilent      bool       // global default for suppressing cron start notifications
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. Well-implemented wake watchdog for cron scheduler.

**Review notes:**
- ✅ CI passes
- ✅ Correct approach: monitors wall-clock gaps to detect sleep wake
- ✅ Proper cleanup: `stopWatchdog` channel closed in Stop()
- ✅ Good threshold (2 min) to avoid false positives
- ✅ Thread-safe: uses existing mutex for rescheduleAll
- ✅ Logs appropriately for debugging

This fixes a real pain point for macOS users. Approved for merge.

PR #345, opened 2026-03-28
fix(core): remove duplicate case EventToolResult in handleStreamEvents

Abelonx:
## Summary

The structured progress card commit (b56865f) added a new `EventToolResult` handler with `ProgressCardEntry` support but did not remove the previous fallback handler, causing a Go compile error on `main`:

```
# github.com/chenhg5/cc-connect/core
core/engine.go:2271:8: duplicate case EventToolResult (constant "tool_result" of string type EventType) in expression switch
        core/engine.go:2246:8: previous case
```

- Remove the old handler (lines 2271–2319) that is now superseded by the structured version at line 2246

## Test plan

- [x] `go build ./cmd/cc-connect/` passes
- [x] Verified only 3 `case EventToolResult:` remain (in 3 separate switch statements / functions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #346, opened 2026-03-28
fix(core): suppress empty response in silent heartbeat mode

Abelonx:
## Summary

When a heartbeat runs with `silent = true` and the agent produces no text output (e.g. only tool calls, or the prompt instructs "no output if everything is normal"), the engine replaces the empty response with the localized `"(空响应)"` / `"(empty response)"` string and sends it to the user — defeating the purpose of silent mode.

**Before:** Silent heartbeat with empty agent output → user receives "(空响应)" message
**After:** Silent heartbeat with empty agent output → no message sent, turn ends silently

## Changes

- Add `SilentEmpty` field to `Message` struct
- Add `silentEmpty` field to `interactiveState`  
- Propagate flag from message → state in `processInteractiveMessageWith`
- Set flag in `ExecuteHeartbeat` when `silent == true`
- Check flag in `EventResult` handler: skip send when response is empty
- Also includes duplicate `case EventToolResult` fix from #345

## Test plan

- [x] `go build ./cmd/cc-connect/` passes
- [x] Verified heartbeat with `silent = true` + empty agent response → no message sent
- [x] Non-silent heartbeat with empty response → still shows "(空响应)" as before

## Config example

```toml
[projects.heartbeat]
  enabled = true
  silent = true
  prompt = "Quick health check. If everything is normal, produce no output."
```

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -257,6 +257,7 @@ type interactiveState struct {
 	deleteMode             *deleteModeState
 	lastAutoCompressAt     time.Time
 	lastAutoCompressTokens int
+	silentEmpty            bool            // when true, drop empty responses instead of sending "(空响应)"
 }
 
 type deleteModeState struct {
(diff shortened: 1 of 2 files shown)


chenhg5 (changes requested):
CI failed with test error in `agent/codex/session_test.go`:

```
--- FAIL: TestSend_UsesStdinForMultilinePrompt (0.00s)
    session_test.go:242: stdin content = "", want "line1\nline2"
```

Please fix the failing test or the underlying implementation before we can proceed with review.

**Suggestion**: The test expects stdin to contain multiline content but it's empty. Check if the stdin piping logic in the codex session implementation was affected by this change.

PR #349, opened 2026-03-28
fix(cron): cleanup interactive state for new_per_run jobs

### LINKED ISSUE #348: `new_per_run` cron jobs leak Claude CLI processes after successful completion
## Problem

When a cron job configured with `mode: "new_per_run"` completes successfully, the spawned Claude CLI process remains alive indefinitely, leaking memory with each execution.

## Root Cause

In `ExecuteCronJob()`, after `processInteractiveMessageWith()` returns for a `new_per_run` job, `cleanupInteractiveState()` is never called on the normal (success) path.

The flow:

1. `processInteractiveEvents()` receives `EventResult` → the turn is done
2. The function returns normally
3. `ExecuteCronJob()` returns `nil` to the scheduler
4. The scheduler marks the run as completed

At step 2, the Claude CLI process is still alive with its stdin pipe open, waiting for the next message — which will never come. The `interactiveState` entry also remains in the map.

`cleanupInteractiveState()` is called on error paths (process crash, `EventError`, channel close), but not on successful completion of a `new_per_run` cron.

The `idle_timeout_mins` watchdog also doesn't help here because the event loop has already returned — there's nobody left listening for the timeout.

## Impact

- Each cron execution leaks one Claude CLI process
- With multiple scheduled crons running throughout the day, memory accumulates until the system runs low
- Processes must be killed externally

## Suggested Fix

Call `cleanupInteractiveState(iKey)` in `ExecuteCronJob()` after `processInteractiveMessageWith()` returns for `new_per_run` jobs. This would close the agent session and remove the interactive state entry.

An alternative approach: use `claude --print` (one-shot mode) instead of interactive mode for `new_per_run` crons, since by definition they only need a single turn.

## Workaround

External cron that periodically kills orphaned `claude.*stream-json` processes.

chenhg5:
## Summary
- Fix memory leak in `new_per_run` cron jobs where Claude CLI processes were not cleaned up after job completion
- Root cause: `ExecuteCronJob()` returns normally but `cleanupInteractiveState()` was never called for cron sessions
- The fix detects cron sessions (by "#cron:" in sessionKey) and calls `cleanupInteractiveState()` after the turn completes in `processInteractiveEvents()`

## Technical details
When `session_mode` is `"new_per_run"`, cron jobs spawn a fresh session each run with an interactive key containing `"#cron:"` (e.g., `sessionKey#cron:sessionID`). After the agent finishes processing (EventResult), the goroutine in `processInteractiveEvents()` returns without cleanup, leaving:
1. The interactive state in `e.interactiveStates` map
2. The agent session (Claude CLI process) still running

This fix adds cleanup logic after the turn completes for cron sessions, ensuring the agent process is properly terminated via `cleanupInteractiveState()`.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes  
- [x] `go test ./core/...` passes

Fixes #348

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2580,6 +2580,11 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 					slog.Debug("async send error after EventResult", "error", err)
 				}
 			}
+			// Cleanup interactive state for new_per_run cron jobs.
+			// Cron sessions have "#cron:" in their interactive key.
+			if strings.Contains(sessionKey, "#cron:") {
+				e.cleanupInteractiveState(sessionKey, state)
+			}
 			return
 
 		case EventError:


chenhg5 (commented):
## Conclusion: Comment (functionally Approve)

The fix is correct and minimal (5 lines, well-targeted). However, the underlying issue #348 was already resolved by PR #445 (merged 2026-04-08), so this PR is effectively a redundant belt-and-suspenders. Verdict below lays out both options.

## What the fix does (verified against rebased branch + main)

Rebased cleanly onto current main `096ed3ea` (the rebase is a 4-line trivial content conflict in `core/engine.go` at the EventResult → return point, easy to resolve by keeping main's "done reaction" block and inserting the cron cleanup after it). Final diff: +5 lines, no removals.

```go
// Cleanup interactive state for new_per_run cron jobs.
// Cron sessions have "#cron:" in their interactive key.
if strings.Contains(sessionKey, "#cron:") {
    e.cleanupInteractiveState(sessionKey, state)
}
```

The detection (`strings.Contains(sessionKey, "#cron:")`) correctly matches the cron session key format `runSessionKey#cron:sessionID` produced at `core/engine.go:1321-1323`. `cleanupInteractiveState` is the correct cleanup entry point (closes the agent session, removes the `e.interactiveStates` entry).

- `go vet -tags no_web ./core/...` — clean
- `go test -tags no_web -count=1 ./core/...` — `ok github.com/chenhg5/cc-connect/core 42.408s`

## SUPERSEDED analysis

Issue #348 ("`new_per_run` cron jobs leak Claude CLI processes after successful completion") is closed and references both PR #349 and **PR #445** (merged 2026-04-08, by `jhaizhou-ops`). PR #445 took the alternative approach suggested in the issue's "Suggested Fix" section: call `cleanupInteractiveState(iKey)` directly in `ExecuteCronJob` after `processInteractiveMessageWith` returns:

```go
// core/engine.go:1321-1327 (current main, via PR #445)
iKey := fmt.Sprintf("%s#cron:%s", runSessionKey, session.ID)
if workspaceDir != "" {
    iKey = workspaceDir + ":" + iKey
}
prevHistLen := session.HistoryLen()
e.processInteractiveMessageWith(effectivePlatform, msg, session, agent, sessions, iKey, workspaceDir, runSessionKey)
e.cleanupInteractiveState(iKey)  // ← PR #445's fix
```

PR #445's fix is semantically equivalent to PR #349's fix and has been in production for ~2 months. The two approaches differ in **location**:
- **PR #445** (already merged): caller side — in `ExecuteCronJob`, only for `useNewSession` path. Semantic, not string-based.
- **PR #349** (this PR): callee side — in `processInteractiveEvents` EventResult, all cron sessions via `strings.Contains` heuristic. String-based, broader (would also fire for any session key containing `#cron:`, e.g. theoretically cron-mode interactive sessions in the future).

Both prevent the leak. PR #349's approach has a small risk of false positives if any non-cron session key ever contains `#cron:` (currently none), but the heuristic is safe today.

## Options for the author / owner

**(a) Close as SUPERSEDED** — cleanest. The bug is already fixed in main. This PR is redundant.

**(b) Keep this PR as belt-and-suspenders** — merge after rebase, accepting some code duplication. The PR #349 placement is in a generic event loop and would fire for cron sessions even if some future caller forgot to call `cleanupInteractiveState`. Marginal defensive value.

**(c) Refactor to consolidate** — remove PR #349's fix and PR #445's redundant call, replace with a single check in `ExecuteCronJob`'s `useNewSession` path (already there). Effectively close this PR and document the rationale in #445's commit message.

## Test coverage

The PR adds **no new test**. The fix logic is simple enough that a unit test would be:
- Hard to write: requires constructing an `interactiveState` with a session whose key contains `#cron:`, simulating an EventResult event, and verifying `cleanupInteractiveState` was called.
- Easy to verify: the existing `TestExecuteCronJob` style tests would cover it.

If option (b) is chosen, recommend adding `TestProcessInteractiveEvents_CronSession_CleanupCalled` to lock in the contract.

## Risk (if merged)

- **Low.** Additive cleanup; redundant with #445 but not harmful. If both fire (caller side #445 + callee side #349), `cleanupInteractiveState` is idempotent (the `e.interactiveStates` map lookup is no-op on second call).
- **No migration concern.** No schema change.
- **No rollback concern** — the fix is 5 lines; easy to revert.
- **Detection heuristic fragility**: `strings.Contains(sessionKey, "#cron:")` could be brittle if session key format ever changes. PR #445's `useNewSession` boolean check is more robust.

## Next step

- **If author wants to close as SUPERSEDED**: add a closing comment referencing #445 (commit `merged via PR #445 on 2026-04-08`).
- **If author wants to keep**: rebase onto main (the 4-line conflict I walked through is easy to resolve), then re-run CI for fresh green. The PR is merge-ready from a logic standpoint.
- **Owner (chenhg5)**: pick (a)/(b)/(c). The decision is policy ("do we want defense-in-depth on cron cleanup, or trust the single caller-side fix?").

— cc-connect/qa-claudecode

chenhg5:
先关闭这个旧 PR。

这个 PR 修复的 #348 已经由 PR #445 在 2026-04-08 合并修复，QA 复查也确认 #349 与 #445 语义等价，属于 superseded。保留这个冲突中的旧 PR 会让后续维护者误以为 cron cleanup 还有未完成修复。

如果后续发现 #445 的实现仍有 cron `new_per_run` cleanup 回归，可以基于最新 main 新开一个更小的回归测试/修复 PR。

---

Closing this old PR for now.

The issue it targets (#348) was already fixed by PR #445, merged on 2026-04-08. QA also confirmed that #349 is semantically equivalent to #445, so this PR is superseded. Keeping this conflicting old PR open would incorrectly suggest there is still unfinished cron cleanup work.

If #445 later regresses for cron `new_per_run` cleanup, please open a fresh, smaller regression-test/fix PR based on latest main.

Issue #324, opened 2026-03-25
New session creation fails to persist session state when running multiple projects concurrently


jiaoziluanbu:
## Environment
- cc-connect version: v1.2.1
- Platform: macOS (darwin arm64)
- Claude Code version: 2.1.83
- Config: 3 projects (my-project, cc-xiaoxia, cc-batman) running concurrently, all using `claudecode` agent with feishu platform

## Bug Description

When creating a new session (via `/new` or session switch) while multiple projects are running concurrently, the new Claude Code session is successfully created (via SDK CLI), but the **session state file (`{project}_{id}.json`) is not updated**. This causes:

1. The `counter` field doesn't increment
2. The new session is not added to the `sessions` map
3. `active_session` still points to the old session
4. Subsequent messages are routed to the **old** session instead of the newly created one

## Evidence

Three orphan CC sessions were created with `entrypoint: sdk-cli, userType: external` (confirming they were created by cc-connect), but none appear in any project's session state:

| Orphan Session ID | Topic | Should belong to |
|---|---|---|
| `803622a6-fd9c-4405-b5f2-bef1ccce6474` | Skills market research | cc-batman |
| `66bfb87b-9623-4f67-8775-92a99cba8ecf` | Content workflow | cc-xiaoxia |
| `c8a78040-1d80-446d-8a73-18951df3a53c` | (follow-up attempt) | cc-xiaoxia |

### cc-batman example:
- `counter: 1` (never incremented despite new session creation)
- `active_session` still points to `s1` → old session `1a70823f` (created 2026-03-23)
- User created a new session for skills research → CC session `803622a6` was created and received the first 2 messages
- But subsequent messages (at 15:51 and 16:03) were routed back to the old `1a70823f` session

### cc-xiaoxia example:
- User created new session `[自媒体工作流]` with model switch to opus
- CC session `66bfb87b` was created (confirmed `model=claude-opus-4-6` in responses)
- Session produced full output (10 questions + research results)
- But `active_session` stayed on `s2` (AI日报 session `c2bc2f6c`)
- All follow-up messages went to the AI日报 session instead

## Steps to Reproduce

1. Configure 3 projects in `config.toml`, all with `claudecode` agent type
2. Start cc-connect, let all 3 projects connect to feishu
3. In one project's feishu chat, create a new session (e.g., `/new`)
4. Send a message — it goes to the new session ✓
5. Send a second message — it goes to the **old** session ✗

## Expected Behavior

[...]

Issue #311, opened 2026-03-24
[Bug]  Workspace Model selection bug


YanxingLiu:
### cc-connect Version

commit 66f3007f07b372924c02e9066b666df24dcc13ad

### Operating System

Linux (Ubuntu/Debian)

### Agent Type

Claude Code

### Platform

Feishu (Lark)

### Installation Method

Build from source

### Bug Description

  In multi-workspace mode, model switching and model display are inconsistent for workspace-bound chats.

  There are two related issues:

  1. /model was applied to the global agent instead of the current workspace agent.
  2. Newly created workspace agents did not inherit the active provider selection from the global agent.

  Because of this, after switching the model in a workspace chat and then creating a new session, the new session could still start with the old model. The /status output and interactive /model card could also show stale or incomplete model information.

### Steps to Reproduce

  1. Configure cc-connect in multi-workspace mode with a workspace-bound chat.
  2. Open a chat that is already bound to a workspace.
  3. Run /model and switch to another provider-backed model, for example doubao/glm-4.7 or bailian/kimi-k2.5.
  4. Run /new.
  5. Send a normal message to start a new session.
  6. Run /status.
  7. Open the interactive /model card.

### Expected Behavior

  - The model selected via /model should apply to the current workspace agent.
  - After /new, the next session in that workspace should start with the newly selected model.
  - /status should show the correct current session model and default model for that workspace.
  - The interactive /model card should show the actual provider model list for the current workspace agent.

### Actual Behavior

  - The model switch was saved and logged successfully, but it only updated the global agent.
  - The current workspace agent could continue using the previous model.
  - After /new, a newly started session in the same workspace could still run with the old model.
  - /status could still show the old model as both the current session model and default model.
  - The interactive /model card could fall back to a small default model list instead of showing the provider’s real models, because the workspace agent did not inherit the
    active provider state.

### Configuration (config.toml)

```toml

```

### Logs / Error Output

```shell

```

Issue #255, opened 2026-03-22
[Bug] __continue__ sentinel persisted as agent_session_id, prevents session resume after restart


quabug:
https://github.com/chenhg5/cc-connect/pull/244/changes#r2970828617

> The startSessionID should never be changed to the ContinueSession string because this causes problems for other agents and only fixes a particular issue in claude code.
Instead of a common location like this, this type of workaround should be placed in the claude code module.

## Bug Description

After a service restart, bots lose their conversation history and start fresh sessions instead of resuming. The root cause is that the `__continue__` sentinel value gets persisted to disk as `agent_session_id`, and subsequent restarts can never overwrite it with the real session ID.

## Root Cause

Introduced in PR #244 (`df5cc64`), the `--continue` bridge sets `startSessionID = ContinueSession` (`"__continue__"`) on the first connection after startup:

```go
// core/engine.go ~line 1806-1809
startSessionID := session.GetAgentSessionID()
if consumeFirstUserContinueBridge && !e.hasConnectedOnce.Swap(true) {
    startSessionID = ContinueSession
}
```

After the agent starts, the code tries to save the real session ID:

```go
// core/engine.go ~line 1825-1828
if newID := agentSession.CurrentSessionID(); newID != "" {
    if session.CompareAndSetAgentSessionID(newID, agent.Name()) {
        sessions.Save()
    }
}
```

But `CompareAndSetAgentSessionID` only sets the value **when it's currently empty**:

```go
// core/session.go ~line 87-96
func (s *Session) CompareAndSetAgentSessionID(id, agentType string) bool {
    s.mu.Lock()
    defer s.mu.Unlock()
    if s.AgentSessionID != "" {
        return false  // ← already non-empty, skip!
    }
    s.AgentSessionID = id
    s.AgentType = agentType
    return true
}
```

The problem: **nothing writes `__continue__` to the session before `CompareAndSetAgentSessionID` is called** — the session's `AgentSessionID` was loaded from disk. But once `__continue__` IS persisted (see below), the cycle becomes permanent.

### How `__continue__` gets persisted

The `EventText` handler also calls `CompareAndSetAgentSessionID`:

```go
// core/engine.go ~line 2047-2049
if event.SessionID != "" {
    if session.CompareAndSetAgentSessionID(event.SessionID, e.agent.Name()) {
        sessions.Save()
    }
}
```

And the `EventResult` handler uses `SetAgentSessionID` (unconditional):

```go
[...]

Issue #238, opened 2026-03-20
/new command resumes stale Claude Code session — should start fresh


dannyliiii:
[AI-Generated]

## Problem

When a user runs `/new` to create a new session, cc-connect creates a new internal session object (e.g. `s6`, `s7`) but **does not clear the `AgentSessionID`** from the previous session. When the next message arrives, `GetOrCreateActive()` returns a session that may still reference a stale `AgentSessionID`. If the previous session's `AgentSessionID` is carried over (or the old session is still active), cc-connect launches Claude Code with `claude --resume <old-session-id>`, loading the **entire prior conversation context**.

This causes:
1. **Stale context contamination** — Claude "remembers" things from days-old conversations and references work that was already completed
2. **Stale git status** — the session-start snapshot reflects the repo state from when the original session was created, not the current state
3. **Cross-session bleed** — Claude references topics, issues, and decisions from prior sessions as if they are part of the current conversation

## Root Cause

In `agent/claudecode/session.go`:

```go
if sessionID != "" {
    args = append(args, "--resume", sessionID)
}
```

And in `core/session.go`, `NewSession()` creates a fresh session object with an empty `AgentSessionID`, but the **active session pointer** (`activeSession[userKey]`) may still point to an older session with a populated `AgentSessionID` until the new session receives its first response.

More importantly, `/new` does not guarantee that the next Claude Code process starts without `--resume`.

## Expected Behavior

`/new` should guarantee a **completely fresh** Claude Code session with no prior context — equivalent to opening a new terminal and running `claude` without `--resume`.

## Suggested Fix

When `/new` is invoked, ensure the newly created session's `AgentSessionID` is empty (`""`), and that `StartSession()` is called with an empty `sessionID` so `--resume` is never passed:

```go
// In NewSession or the /new handler:
s := sm.createLocked(userKey, name)
s.AgentSessionID = ""  // ensure no resume
```

Alternatively, add a config option:

```toml
[projects.agent.options]
resume_session = false  # never resume, always start fresh
```

## Environment

- cc-connect v1.2.2-beta.2
- Agent: Claude Code (claude CLI)
- Platform: Lark/Feishu

PR #358, opened 2026-03-29
fix(core): set pending permission before sending prompt

### LINKED ISSUE #171: fix: race condition in permission handling causes session stuck with "previous request still processing"
## Bug Description

When a user clicks **Allow All** (or any permission button) in Feishu very quickly after the permission card appears, the response is ignored and the session becomes permanently stuck showing:

> ⏳ 上一个请求仍在处理中，请稍候... / Previous request still processing, please wait...

The user can no longer send any messages to that session until it times out (observed ~13 minutes).

## Root Cause

Race condition in `core/engine.go` `processInteractiveMessageWith`:

```go
// BEFORE (buggy order):
e.sendPermissionPrompt(...)   // ← card sent to Feishu FIRST
// ... user clicks button here before pending is set ...
state.pending = pending       // ← pending set AFTER card is sent
```

When the user clicks the permission button quickly:
1. Feishu card is sent via `sendPermissionPrompt`
2. User clicks "Allow All" → `go p.handler(...)` dispatches `content="allow all"`
3. `handlePendingPermission` checks `state.pending` — it is **still nil** at this point
4. Returns `false`, message falls through to `session.TryLock()`
5. Lock is held by the running session → returns "Previous request still processing"
6. Permission request is never resolved → session stuck indefinitely

## Fix

Set `state.pending` **before** sending the permission card:

```go
// AFTER (fixed order):
state.pending = pending       // ← set pending FIRST
e.sendPermissionPrompt(...)   // ← then send card
```

## Steps to Reproduce

1. Configure cc-connect with multi-workspace mode in Feishu
2. Send a message that triggers a permission request (e.g. a task using Bash)
3. Click "Allow All" immediately when the permission card appears
4. Observe: session shows "⏳ Previous request still processing" and becomes unresponsive

## Impact

- Session is completely blocked until timeout (~13 minutes observed)
- All subsequent messages rejected with "previous request still processing"
- Reproducible whenever the user responds to the permission card quickly

## Suggested Fix

In `core/engine.go`, move the `state.pending` assignment before the `sendPermissionPrompt` / `sendAskQuestionPrompt` call (the fix is a simple reorder, no logic change).

xukp20:
## Summary

  Fix the permission-handling race reported in #171 by setting `state.pending` before sending the permission or ask-user-question prompt.

  This ensures a very fast platform callback can still resolve the active pending request instead of falling through as if no permission prompt existed.

  ## What changed

  - set `state.pending` before calling `sendPermissionPrompt`
  - set `state.pending` before calling `sendAskQuestionPrompt`
  - add regression coverage for immediate permission callback during prompt send
  - add regression coverage for immediate AskUserQuestion callback during prompt send

  ## Why

  Previously the event flow was:

  1. send permission / ask-question prompt
  2. platform callback arrives immediately
  3. `handlePendingPermission()` sees no pending request yet
  4. response is ignored or falls through to normal busy-session handling

  That could leave the session stuck in a waiting state and produce the “previous request still processing” behavior described in #171.

  By registering the pending request first, the callback can be handled correctly even if it arrives during prompt delivery.

  ## Validation

  - [x] `go test ./core -run 'TestProcessInteractiveEvents_(PermissionCallbackDuringPromptSend|AskQuestionCallbackDuringPromptSend)|
  TestHandlePendingPermission|TestProcessInteractiveEvents_PermissionWhileSendBlocked' -v`
  - [x] `go build ./...`
  - [x] `go test ./... -v -race`
  - [x] `go test ./... -coverprofile=coverage.out -covermode=atomic`
  - [x] `$(go env GOPATH)/bin/staticcheck ./core/...`

  Fixes #171

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2332,18 +2332,6 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 				"tool", event.ToolName,
 			)
 
-			if isAskQuestion {
-				e.sendAskQuestionPrompt(p, replyCtx, event.Questions, 0)
-			} else {
-				permLimit := e.display.ToolMaxLen
-				if permLimit > 0 {
-					permLimit = permLimit * 8 / 5
-				}
-				toolInput := truncateIf(event.ToolInput, permLimit)
-				prompt := fmt.Sprintf(e.i18n.T(MsgPermissionPrompt), event.ToolName, toolInput)
-				e.sendPermissionPrompt(p, replyCtx, prompt, event.ToolName, toolInput)
-			}
-
 			pending := &pendingPermission{
 				RequestID:    event.RequestID,
 				ToolName:     event.ToolName,
@@ -2355,6 +2343,18 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 			state.mu.Lock()
 			state.pending = pending
 			state.mu.Unlock()
+
+			if isAskQuestion {
+				e.sendAskQuestionPrompt(p, replyCtx, event.Questions, 0)
+			} else {
+				permLimit := e.display.ToolMaxLen
+				if permLimit > 0 {
+					permLimit = permLimit * 8 / 5
+				}
+				toolInput := truncateIf(event.ToolInput, permLimit)
+				prompt := fmt.Sprintf(e.i18n.T(MsgPermissionPrompt), event.ToolName, toolInput)
+				e.sendPermissionPrompt(p, replyCtx, prompt, event.ToolName, toolInput)
+			}
 
 			// Stop idle timer while waiting for user permission response;
 			// the user may take a long time to decide, and we don't want
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -268,6 +268,21 @@ func (p *stubInlineButtonPlatform) SendWithButtons(_ context.Context, _ any, con
 	return nil
 }
 
+type callbackInlineButtonPlatform struct {
+	stubInlineButtonPlatform
+	onSend func()
+}
+
+func (p *callbackInlineButtonPlatform) SendWithButtons(ctx context.Context, replyCtx any, content string, buttons [][]ButtonOption) error {
+	if err := p.stubInlineButtonPlatform.SendWithButtons(ctx, replyCtx, content, buttons); err != nil {
+		return err
+	}
+	if p.onSend != nil {
+		p.onSend()
+	}
+	return nil
+}
+
 type stubCardPlatform struct {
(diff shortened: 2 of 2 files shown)


xukp20:
Clarifying scope for this PR relative to #324 and the issues mentioned there:

  This PR only fixes the specific permission-handling race from #171.

  What it fixes:
  - `state.pending` is now registered before sending the permission / AskUserQuestion prompt
  - so an immediate platform callback can still resolve the pending request correctly

  What it does **not** fix:
  - session state persistence races across multiple projects (#324)
  - relay session concurrency / cancellation issues (#181)
  - stale `/new` resume behavior (#238)
  - persisted `__continue__` sentinel issues (#255)
  - workspace model/session context issues (#311)

  So if a `"previous request still processing"` symptom was caused by the fast permission-callback timing described in #171, this PR should address it.

  But #324 is a different class of problem: its core issue is session state persistence / active-session consistency under concurrent project activity.
  That needs to be handled separately in the session-store path (for example the locking work in #325), not in this PR.

  In short:
  - `#358` = permission callback timing race (`#171`)
  - `#324` = session persistence / active-session race
  - symptoms may overlap, but the root causes are different

chenhg5 (approved):
LGTM. Correct race condition fix.

**Review notes:**
- ✅ CI passes
- Sets `state.pending` before sending permission prompt
- Prevents fast callback from being ignored
- Good regression test coverage
- Fixes issue #171

Approved for merge.

xukp20:
I re-checked current `main`, and the functional fix from this PR appears to already be present there.

In the current permission flow, `state.pending` is set before sending the permission / ask-question prompt, which is the core race fix this PR was introducing.

Because of that, I do not think it makes sense to revive this old branch. Suggest closing this PR as already covered by current `main`.

xukp20:
Re-checked this against the latest \, and the core fix is already present there now.\n\nCurrent \ already sets \ before calling the permission / ask-question prompt send paths, and it also keeps regression coverage around the immediate-callback scenario.\n\nSo this PR is still showing as conflicting, but I don’t think it makes sense to repair and rebase it anymore — it appears to be superseded by current \.

xukp20:
Re-checked this against the latest main, and the core fix is already present there now.

Current main already sets state.pending before calling the permission / ask-question prompt send paths, and it also keeps regression coverage around the immediate-callback scenario.

So this PR is still showing as conflicting, but I do not think it makes sense to repair and rebase it anymore — it appears to be superseded by current main.

xukp20:
Closing this as superseded by current main. The pending-permission race fix is already present upstream now, so it does not make sense to keep rebasing this old conflicting branch.

PR #408, opened 2026-04-01
fix(bridge): require token when Bridge is enabled for security

### LINKED ISSUE #403: [Bug] Bridge security defaults are weaker than documented when token is unset
### cc-connect Version

  Current `main` (reviewed on 2026-04-01)

  ### Operating System

  Linux (Ubuntu/Debian)

  ### Agent Type

  Other

  ### Platform

  N/A

  ### Installation Method

  Build from source

  ### Bug Description

  The current Bridge implementation appears to have weaker security defaults than the documentation suggests.

  According to the docs/config examples:

  - Bridge token is described as required
  - all Bridge connections are described as requiring token authentication

  However, in the current implementation:

  1. if `bridge.token` is unset, authentication silently becomes **allow-all**
  2. WebSocket `CheckOrigin` is always `true`
  3. the Bridge server listens on `:9810` by default (all interfaces)

  This means that enabling `[bridge]` without setting `token` does **not** fail closed. Instead, both the Bridge WebSocket endpoint and the Bridge REST session endpoints become effectively unauthenticated for any client that can reach the port.

  So the problem is not just documentation mismatch; it is also a risky default behavior for a
  network-facing endpoint.

  ### Steps to Reproduce

  1. Configure Bridge without a token:

  ```toml
  [bridge]
  enabled = true
  port = 9810
  # token intentionally omitted
```
  2. Start cc-connect
  3. Make a request to a Bridge REST endpoint without any token, for example:

  curl -i "http://127.0.0.1:9810/bridge/sessions"

  4. Observe that the request is not rejected as unauthorized.
     Instead, it proceeds into normal handler logic and returns a business error such as:

  - 400 session_key query parameter is required

  5. Also connect to the Bridge WebSocket endpoint without any token and complete the normal register handshake. The connection is accepted because authenticate() returns true when the configured token is empty.

  ### Expected Behavior

  One of the following should happen when [bridge] is enabled:

  - Bridge refuses to start unless token is set, or
  - Bridge defaults to a clearly explicit local-only / insecure-dev-only mode, or
  - Bridge requires an explicit opt-in flag before allowing unauthenticated access

  In addition:

  - WebSocket origin checks should not always be disabled
  - Bridge security behavior should match the documentation

  ### Actual Behavior

  Current main behaves as follows:

  - documentation says Bridge token is required
  - runtime authentication allows all requests when bridge.token == ""
  - WebSocket CheckOrigin always returns true
  - server binds to :9810 by default

  So Bridge can become an unauthenticated network-facing endpoint if enabled without a token.

```
  [bridge]
  enabled = true
  port = 9810
  # token omitted
```
  ### Logs / Error Output
```
  # Example REST call without token:
  curl -i "http://127.0.0.1:9810/bridge/sessions"
  # The request is not rejected with 401.
  # It reaches normal handler logic instead.
```
  ### Additional Context

  Relevant code:

  - core/bridge.go
      - CheckOrigin: func(r *http.Request) bool { return true }
      - authenticate() returns true when bs.token == ""
      - Bridge server listens on :9810

  Relevant docs/config examples:

  - config.example.toml
      - token is documented as required
  - docs/usage.md
      - “All Bridge connections require a token”
  - docs/usage.zh-CN.md
      - “所有 Bridge 连接需要 token 认证”

  This looks like a combination of:

  - documentation/implementation mismatch
  - insecure fail-open behavior when token is omitted
  - overly permissive WebSocket origin policy

  Possible directions:

  - require non-empty token when Bridge is enabled
  - or introduce an explicit insecure/local-dev mode instead of silently allowing all access
  - make WebSocket origin checks configurable instead of unconditional true

### PR DESCRIPTION
**Pull request by chenhg5**
## Summary
- Bridge server now requires token to be set unless `insecure = true` is explicitly configured
- Added `insecure` config option for local development only
- WebSocket origin check now validates against CORS origins or same-host
- Clear error message and exit when token is missing

## Root Cause
Issue #403 identified that when `bridge.token` is not set:
1. `authenticate()` returned `true` for all requests
2. WebSocket `CheckOrigin` always returned `true`
3. Bridge server listened on all interfaces (`:9810`)

This allowed unauthenticated access to the Bridge API.

## Changes
- `config/config.go`: Added `Insecure *bool` field to `BridgeConfig`
- `core/bridge.go`: 
  - Added `insecure` field to `BridgeServer`
  - Added `NewBridgeServerInsecure()` for explicit local-dev mode
  - `authenticate()` now requires token unless `insecure` is true
  - Added `checkOrigin()` that validates against CORS origins or same-host
- `cmd/cc-connect/main.go`: Check insecure flag and exit with error if token is missing
- `core/bridge_test.go`: Updated tests to use insecure mode when testing without token

## Breaking Change
Bridge with empty token now fails to start with error:
```
bridge: token is required when insecure mode is not enabled
```

For local development, add to config:
```toml
[bridge]
enabled = true
insecure = true  # Only for local dev!
```

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes  
- [x] `go test ./core/...` passes
- [ ] Manual test: Bridge with token works correctly
- [ ] Manual test: Bridge without token fails to start
- [ ] Manual test: Bridge with `insecure = true` starts without token

Fixes #403

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -627,7 +627,17 @@ func main() {
 		if path == "" {
 			path = "/bridge/ws"
 		}
-		bridgeSrv = core.NewBridgeServer(port, cfg.Bridge.Token, path, cfg.Bridge.CORSOrigins)
+		// Check insecure flag for local development mode
+		insecure := cfg.Bridge.Insecure != nil && *cfg.Bridge.Insecure
+		if insecure {
+			bridgeSrv = core.NewBridgeServerInsecure(port, cfg.Bridge.Token, path, cfg.Bridge.CORSOrigins)
+		} else {
+			bridgeSrv = core.NewBridgeServer(port, cfg.Bridge.Token, path, cfg.Bridge.CORSOrigins)
+		}
+		if bridgeSrv == nil {
+			slog.Error("bridge: failed to create server - token is required (or set insecure=true for local dev)")
+			os.Exit(1)
+		}
 		for i, e := range engines {
 			bp := bridgeSrv.NewPlatform(cfg.Projects[i].Name)
 			bridgeSrv.RegisterEngine(cfg.Projects[i].Name, e, bp)
--- a/config/config.go
+++ b/config/config.go
@@ -58,9 +58,10 @@ type WebhookConfig struct {
 type BridgeConfig struct {
 	Enabled     *bool    `toml:"enabled"`                // default false
 	Port        int      `toml:"port,omitempty"`         // listen port; default 9810
-	Token       string   `toml:"token,omitempty"`        // shared secret for authentication; required
+	Token       string   `toml:"token,omitempty"`        // shared secret for authentication; required unless insecure=true
 	Path        string   `toml:"path,omitempty"`         // URL path; default "/bridge/ws"
 	CORSOrigins []string `toml:"cors_origins,omitempty"` // allowed CORS origins; empty = no CORS
+	Insecure    *bool    `toml:"insecure,omitempty"`     // allow running without token (local dev only); default false
 }
 
 // ManagementConfig controls the HTTP Management API for external tools.
--- a/core/bridge.go
+++ b/core/bridge.go
@@ -28,6 +28,7 @@ type BridgeServer struct {
 	token       string
 	path        string
 	corsOrigins []string
+	insecure    bool // allow running without token (local dev only)
 	server      *http.Server
 
 	mu       sync.RWMutex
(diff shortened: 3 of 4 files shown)


chenhg5:
LGTM! This is a solid security fix for the Bridge server.

**✅ Security**: Token is now required unless `insecure = true` is explicitly set. No more accidental unauthenticated access.

**✅ Implementation**: 
- `checkOrigin` properly validates WebSocket origins against CORS origins or same host
- `authenticate` returns `bs.insecure` when token is empty, not `true`
- Clear error message and graceful exit when token is missing

**✅ Breaking change handling**: Clear documentation on how to configure `insecure = true` for local dev.

**✅ Tests**: Updated tests use insecure mode appropriately.

**✅ CI**: All checks pass.

Ready for merge.

chenhg5 (commented):
QA review passed. 重要安全修复。

**Review summary:**
- ✅ Bridge 启动时强制要求 token（除非显式设置 insecure=true）
- ✅ insecure 模式仅用于本地开发，有明确警告日志
- ✅ 验证失败时返回 nil 并打印错误信息
- ✅ 文档更新：config.go 注释说明 token 必填条件
- ✅ CI passes

Ready to merge (owner PR).

⚠️ 注意：这是安全修复，建议尽快合并。

chenhg5 (commented):
QA Review: fix(bridge): require token when Bridge enabled — security fix (+104/-9), validates token presence on startup. CI green, LGTM. No self-approval possible as owner PR.

PR #409, opened 2026-04-01
fix(feishu): auto-refresh tenant_access_token when invalidated

chenhg5:
## Summary
- Fix Feishu `tenant_access_token` not auto-refreshing after ~2 hours
- Add custom `tokenCache` that tracks invalidated tokens and forces refresh
- Add retry logic in `Reply()` and `sendNewMessageToChat()` for error code 99991663

The lark SDK has a bug where it retries requests with an invalid token without clearing the cached token first. This fix ensures messages continue to work after the token lifetime expires.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./platform/feishu/...` passes (all 42 tests)
- [x] `go vet ./...` passes
- [ ] Manual smoke test: run cc-connect with Feishu for >2 hours and verify messages still work

Closes #395

PR #411, opened 2026-04-01
fix(codex): mark slash commands as unsupported in exec/resume mode

### LINKED ISSUE #378: [Bug]  Codex native slash commands are not reliable in exec/resume mode (`/compact` and `/clear`)
### cc-connect Version

Current `main` (tested locally on 2026-03-30)

### Operating System

Linux (Ubuntu/Debian)

### Agent Type

Codex (OpenAI)

### Platform

Feishu (Lark)

### Installation Method

Build from source

### Bug Description

`cc-connect` currently treats Codex as supporting native slash-command-based context operations:

  - `/compress` -> forwards Codex native `"/compact"`
  - experimental local testing also showed the same concern for `"/clear"`

  However, after deeper testing, Codex slash commands in `codex exec` / `codex exec resume --json` mode do **not** appear to be a reliable native capability.

  In practice:

  - `"/compact"` is not consistently verifiable as a true native compact action
  - `"/clear"` is not reliable either
  - in some cases, Codex appears to treat these as ordinary prompt text instead of a guaranteed slash command execution

  This matters because `cc-connect` currently exposes `/compress` for Codex by returning:

  ```go
  func (a *Agent) CompressCommand() string { return "/compact" }
```

  but the actual underlying Codex behavior in exec/resume mode does not seem stable enough to justify that assumption.

#### A. Code-level observation in cc-connect

  Current Codex adapter:

  - agent/codex/codex.go
  - CompressCommand() string { return "/compact" }

  Current core behavior:

  - /compress sends the native compressor command directly into the running agent session

  By contrast, other agents already mark unsupported capabilities explicitly. For example, Gemini returns "" with a comment explaining that Gemini CLI has no interactive compress/compact command.

  #### B. Bare Codex CLI reproduction (/clear)

  I tested Codex directly, outside Feishu and outside cc-connect, using codex exec / codex exec resume --json.

  Minimal behavioral probe:

  1. Start a fresh Codex session
  2. Ask Codex to remember a token
  3. Instruct it:
      - if it can still see the token, it must reply with the token exactly
      - only if it truly cannot see the token anymore, it should reply CLEARED
  4. Confirm it remembers the token
  5. Send /clear
  6. Ask again using the same strict instruction:
      - do not rely on whether /clear was issued
      - only answer based on whether the token is still visible in context

  Observed result:

  - before /clear: Codex correctly replied with the token
  - after /clear: Codex still replied with the token
  - on the /clear turn itself, it also did not behave like a guaranteed system-side clear

  Example probe result:

  - step 2: PROBE-ZETA-417
  - step 3 (/clear): PROBE-ZETA-417
  - step 4 (ask again): PROBE-ZETA-417

  This strongly suggests that in exec/resume --json mode, /clear is not a reliable native session-clear primitive.

  #### C. Bare Codex CLI observation (/compact)

  /compact shows similar behavior.

  On some threads, sending:

  printf '/compact' | codex exec resume ... --json -

  returns only natural-language text such as:

  - 后续我会尽量简短回复。

  That is not enough to prove that a real native compact operation happened.

  In other words, even /compact in Codex exec/resume mode can degrade into what looks like ordinary model interpretation rather than a guaranteed slash-
  command execution path.

### Expected Behavior

If cc-connect exposes Codex native /compress using "/compact", the underlying command should be a reliable native capability in the session mode that cc-connect actually uses (codex exec / codex exec resume --json).

  At minimum, slash commands used as native capabilities should be:

  - documented as supported in exec/resume mode, or
  - behaviorally reliable and verifiable

### Actual Behavior

Codex slash commands in exec/resume --json mode are not reliable enough to be treated as guaranteed native capabilities.

  Specifically:

  - /clear can fail behaviorally even in a minimal direct CLI probe
  - /compact can return ordinary natural-language text rather than a clearly verifiable native compact action

```toml

```

### Logs / Error Output

```shell

```

### Additional Context

  This investigation started while implementing `/clear` support for cc-connect.

  The original goal was to support both:
  - a reset-style clear implemented by cc-connect itself
  - a native clear path delegated to the underlying CLI session

  During that work, I found that the problem is broader than `/clear` alone: Codex slash commands in `exec` / `resume --json` mode do not appear reliable enough to be treated as stable native capabilities.

  #### Current cc-connect implementation

  cc-connect currently treats Codex as supporting native context compaction.

  Relevant code:

  - `agent/codex/codex.go`
    - `CompressCommand() string { return "/compact" }`
  - `core/interfaces.go`
    - `ContextCompressor`
  - `core/engine.go`
    - `/compress` forwards the compressor command into the live session

  So today, when a user runs `/compress` with the Codex agent, cc-connect assumes that sending `"/compact"` into the running Codex session is a valid native capability.

  By contrast, unsupported agents already use a more conservative pattern.

  For example:

  - `agent/gemini/gemini.go`
    - returns `""`
    - includes a comment explaining that Gemini CLI has no interactive compress/compact command

  This means cc-connect already has an established pattern for "do not expose a native slash-command capability unless the underlying CLI really supports it."

  #### What I observed with bare Codex

  I tested Codex directly, outside Feishu and outside cc-connect, using `codex exec` / `codex exec resume --json`.

  For `/clear`, I used a behavioral probe instead of trusting the response text.

  The probe was:

  1. Start a fresh Codex session
  2. Ask Codex to remember a token
  3. Tell it:
     - if it can still see the token, it must reply with the token exactly
     - only if it truly cannot see the token anymore, it should reply `CLEARED`
  4. Confirm it remembers the token
  5. Send `/clear`
  6. Ask again with the same strict rule:
     - do not rely on whether `/clear` was issued
     - answer only based on whether the token is still visible

  Observed result:

  - before `/clear`: Codex returned the token correctly
  - on the `/clear` turn itself: Codex still returned the token
  - after `/clear`: Codex still returned the token

  So in that direct probe, `/clear` did not behave like a real native clear at all.

  For `/compact`, I also observed that it can return ordinary natural-language text such as:
  - `后续我会尽量简短回复。`

  That is not enough to prove that a real native compact operation occurred. It looks more like normal prompt interpretation than a guaranteed system action.

  #### Why this matters

  The concern here is not just "the model said something weird."

  The real problem is that Codex slash commands in `exec/resume --json` mode do not appear to be a reliable hard primitive.

  In other words:
  - sometimes they may look like they worked
  - sometimes they degrade into normal assistant text
  - at least for `/clear`, they can fail even under direct behavioral verification

  That makes them risky to expose as native capabilities inside cc-connect.

  #### Comparison with Claude Code

  Claude Code appears different here.

  In Claude's current `stream-json` session mode:

  - `/compact` is explicitly exposed in `slash_commands`
  - invoking it produces structured compact events such as:
    - `status = compacting`
    - `compact_boundary`

  So this does not look like a generic "all agents behave like this" problem.

  It looks specifically related to how Codex behaves in `exec/resume` mode.

  #### Related upstream evidence

  This also seems consistent with open Codex upstream discussions:

  - [openai/codex#3641](https://github.com/openai/codex/issues/3641) — Support Slash Commands in `codex exec` Mode
  - [openai/codex#4311](https://github.com/openai/codex/issues/4311) — SlashCommand tool-style auto invocation / exec-mode limitations

  Those threads suggest that slash commands in Codex `exec` mode are not fully supported, or at least do not behave the way users expect.

  #### Suggested direction

  I think Codex native slash-command capabilities should be treated more conservatively in cc-connect.

  Possible options:

  1. mark Codex native `/compress` as unsupported for now
  2. add an explicit warning / limitation note that Codex slash commands are not reliable in `exec/resume` mode
  3. only expose Codex native slash-command capabilities after there is a reproducible, behavior-based verification path

chenhg5:
## Summary
- Mark Codex `/compact` and `/clear` slash commands as unsupported
- Return empty string from `CompressCommand()` following Gemini's pattern
- Add comment explaining the limitation with upstream issue references

Codex slash commands are not reliable in exec/resume --json mode - they can degrade into ordinary prompt text instead of guaranteed system actions.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./agent/codex/...` passes

Closes #378

### DIFF
--- a/agent/codex/codex.go
+++ b/agent/codex/codex.go
@@ -365,8 +365,12 @@ func (a *Agent) SkillDirs() []string {
 }
 
 // ── ContextCompressor implementation ──────────────────────────
+// Codex slash commands (/compact, /clear) are not reliable in exec/resume mode.
+// See: https://github.com/openai/codex/issues/3641
+// Return "" so engine reports "not supported" instead of sending
+// an unreliable slash command to the session.
 
-func (a *Agent) CompressCommand() string { return "/compact" }
+func (a *Agent) CompressCommand() string { return "" }
 
 // ── MemoryFileProvider implementation ─────────────────────────
 
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -134,6 +134,8 @@ type Platform struct {
 	callbackPath string
 	encryptKey   string
 	eventHandler *dispatcher.EventDispatcher
+	// Token cache for handling invalid token errors (issue #395)
+	tokenCache *tokenCache
 }
 
 type interactivePlatform struct {
@@ -202,6 +204,11 @@ func newPlatform(name, domain string, opts map[string]any) (core.Platform, error
 	if domain != lark.FeishuBaseUrl {
 		clientOpts = append(clientOpts, lark.WithOpenBaseUrl(domain))
 	}
+	// Use custom token cache to handle invalid token errors (issue #395)
+	// The SDK's built-in retry doesn't properly clear cached tokens when
+	// the server returns error 99991663 (invalid tenant_access_token).
+	tc := newTokenCache()
+	clientOpts = append(clientOpts, lark.WithTokenCache(tc))
 
 	base := &Platform{
 		platformName:          name,
@@ -220,6 +227,7 @@ func newPlatform(name, domain string, opts map[string]any) (core.Platform, error
 		port:                  port,
 		callbackPath:          callbackPath,
 		encryptKey:            encryptKey,
+		tokenCache:            tc,
 	}
 	if !useInteractiveCard {
 		base.self = base
@@ -1093,17 +1101,38 @@ func (p *Platform) Reply(ctx context.Context, rctx any, content string) error {
 		return p.sendNewMessageToChat(ctx, rc, msgType, msgBody)
 	}
 
-	resp, err := p.client.Im.Message.Reply(ctx, larkim.NewReplyMessageReqBuilder().
-		MessageId(rc.messageID).
-		Body(p.buildReplyMessageReqBody(rc, msgType, msgBody)).
-		Build())
+	// Retry once if token is invalid (issue #395)
+	resp, err := p.replyWithRetry(ctx, rc, msgType, msgBody)
 	if err != nil {
-		return fmt.Errorf("%s: reply api call: %w", p.tag(), err)
+		return err
 	}
 	if !resp.Success() {
 		return fmt.Errorf("%s: reply failed code=%d msg=%s", p.tag(), resp.Code, resp.Msg)
 	}
 	return nil
+}
+
+// replyWithRetry sends a reply message, retrying once if the token is invalid.
+func (p *Platform) replyWithRetry(ctx context.Context, rc replyContext, msgType, msgBody string) (*larkim.ReplyMessageResp, error) {
+	resp, err := p.client.Im.Message.Reply(ctx, larkim.NewReplyMessageReqBuilder().
+		MessageId(rc.messageID).
+		Body(p.buildReplyMessageReqBody(rc, msgType, msgBody)).
+		Build())
+	if err != nil {
+		return nil, fmt.Errorf("%s: reply api call: %w", p.tag(), err)
+	}
+	// If token is invalid, invalidate cache and retry once (issue #395)
+	if resp.Code == invalidTokenCode && p.tokenCache != nil {
+		p.tokenCache.InvalidateAll()
+		resp, err = p.client.Im.Message.Reply(ctx, larkim.NewReplyMessageReqBuilder().
+			MessageId(rc.messageID).
+			Body(p.buildReplyMessageReqBody(rc, msgType, msgBody)).
+			Build())
+		if err != nil {
+			return nil, fmt.Errorf("%s: reply api call (retry): %w", p.tag(), err)
+		}
+	}
+	return resp, nil
 }
 
 // Send sends a message. When the original message ID is available, the message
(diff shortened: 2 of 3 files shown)


chenhg5:
**⚠️ Scope Issue + CI Lint Failure**

This PR claims to fix codex slash commands but includes:

1. **Codex fix** (`agent/codex/codex.go`): Correctly marks `/compact` and `/clear` as unsupported
2. **Feishu token cache** (`platform/feishu/feishu.go`, `token_cache.go`): This should be in PR #409 (feishu token auto-refresh)

**Lint errors** (same as PR #409):
- `platform/feishu/token_cache.go:117:6`: func `tokenCacheKey` is unused
- `platform/feishu/token_cache.go:93:2`: SA9003: empty branch

**Recommendation**: 
- Separate the feishu changes into PR #409
- Keep only the codex fix in this PR
- Or clarify if this intentional bundling

**On the codex fix itself**: Returning empty string for `CompressCommand()` follows Gemini's pattern correctly.

chenhg5 (commented):
CI lint 失败：

```
platform/feishu/token_cache.go:117:6: func `tokenCacheKey` is unused
platform/feishu/token_cache.go:93:2: SA9003: empty branch (staticcheck)
```

需要修复 token_cache.go 中的 lint 问题。（owner PR）

chenhg5:
🔴 **CI lint 失败**，需要修复后才能合并。

查看失败详情：https://github.com/chenhg5/cc-connect/actions/runs/23873327601/job/69610463046

建议运行 `golangci-lint run` 查看具体 lint 错误并修复。

chenhg5:

chenhg5:
先关闭这个旧 PR。

目标 issue #378 已经关闭；同时这个 PR 的分支被 Feishu token-cache 改动污染，实际 diff 不只包含 Codex slash command 处理，导致 lint 失败和长期冲突。继续保留这个旧 PR 会让 review 队列混乱。

如果 Codex `/compact`、`/clear` 在 latest main 里仍需要调整，请基于最新 main 新开一个只改 `agent/codex` 的最小 PR。

---

Closing this old PR for now.

The target issue #378 is already closed. This PR branch is also contaminated with Feishu token-cache changes, so the actual diff is not limited to Codex slash command handling; it has lint failures and long-standing conflicts. Keeping it open makes the review queue harder to reason about.

If Codex `/compact` or `/clear` still need changes on latest main, please open a fresh minimal PR based on latest main that only touches the Codex agent path.

PR #414, opened 2026-04-02
fix(web): handle zero-value LastRun in cron job display

mkloveyy:
## Summary
- Change `LastRun` from `time.Time` to `*time.Time` so `omitempty` correctly omits the zero value in JSON
- Clean up legacy zero-value (`year <= 1`) entries on cron store load
- Always show "Last Run" label on cron job cards, displaying `-` when no run has occurred
- Widen info label column (`w-12` → `w-16`) to prevent CJK text wrapping (e.g. "上次运行")

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./core/ -run TestCron` all pass
- [x] Verify cron cards display `-` for never-run jobs
- [x] Verify "上次运行" label no longer wraps in Chinese locale

### DIFF
--- a/core/cron.go
+++ b/core/cron.go
@@ -35,7 +35,7 @@ type CronJob struct {
 	Mode        string    `json:"mode,omitempty"`         // permission mode override for this job; "" = use project default
 	TimeoutMins *int      `json:"timeout_mins,omitempty"` // nil = default 30m wait; 0 = no limit; >0 = minutes
 	CreatedAt   time.Time `json:"created_at"`
-	LastRun     time.Time `json:"last_run,omitempty"`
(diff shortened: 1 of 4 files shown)


chenhg5 (approved):
LGTM! Clean fix for zero-value LastRun handling.

**✅ Implementation:**
- Changes LastRun from time.Time to *time.Time so omitempty works correctly
- Cleans up legacy zero-value entries on cron store load
- Always shows 'Last Run' label with '-' for never-run jobs
- Widens label column (w-12 → w-16) for CJK text

**✅ Tests:** CI passes.

Good improvement for UX consistency.

PR #451, opened 2026-04-04
feat(feishu): enable WebSocket mode for Lark international version

### LINKED ISSUE #449: [Feature] Lark (international) now supports WebSocket — remove webhook-only fallback
### Feature Request

Lark international (larksuite.com) now supports WebSocket long connections for event subscriptions. The current code in `platform/feishu/feishu.go` line 311 hardcodes Lark to webhook-only mode:

```go
// Lark international version uses Webhook mode, not WebSocket long connection
// Feishu domestic version supports WebSocket long connection
if p.platformName == "lark" {
    return p.startWebhookMode()
}
return p.startWebSocketMode()
```

This was correct historically, but **Lark added WebSocket support** — see the [official documentation](https://open.larksuite.com/document/ukTMukTMukTM/uYDNxYjL2QTM24iN0EjN/event-subscription-configure-/use-websocket). This was also noted by @eyeix in #63 (comment from Mar 17, 2026).

### Evidence

1. **Lark official docs** confirm WebSocket is now available for international tenants.
2. **The Lark Go SDK v3.5.3** (which cc-connect already depends on) supports `larkws.WithDomain(domain)` to connect to any domain including `open.larksuite.com`.
3. **cc-connect's own `startWebSocketMode()`** already handles non-Feishu domains correctly:
   ```go
   if p.domain != lark.FeishuBaseUrl {
       wsOpts = append(wsOpts, larkws.WithDomain(p.domain))
   }
   ```
4. **OpenClaw** (another Lark bot framework) has been running Lark international with `domain: "lark"` + `connectionMode: "websocket"` in production without issues.

### Suggested Fix

Remove the hardcoded platform name check and let WebSocket be the default for both Feishu and Lark. Fall back to webhook only when the user explicitly configures `encrypt_key` (which implies webhook mode):

```go
// Use webhook mode only when explicitly configured (encrypt_key present)
if p.encryptKey != "" {
    return p.startWebhookMode()
}
return p.startWebSocketMode()
```

Or, alternatively, add a `connection_mode` config option:

```toml
[[projects.platforms]]
name = "lark"
app_id = "cli_xxx"
app_secret = "sec_xxx"
connection_mode = "websocket"   # "websocket" (default) or "webhook"
```

### Impact

This would allow Lark international users to use cc-connect **without a public IP**, matching the zero-config experience that Feishu domestic users already enjoy.

chenhg5:
## Summary
- Enable WebSocket mode for Lark international version (larksuite.com)
- Remove platform name check that forced Lark to webhook mode
- Default to WebSocket for all platforms; fallback to webhook only when explicitly configured

## Problem
Lark international version was forced to use webhook mode, requiring:
- A public IP address
- Port forwarding/reverse proxy configuration

This created a barrier for users without public IP access.

## Solution
Lark SDK v3.5.3+ supports WebSocket for international version via `larkws.WithDomain()`. The `startWebSocketMode()` already handles this correctly.

Changes:
- Remove the `platformName == "lark"` check
- Use WebSocket by default for all platforms
- Only use webhook if user explicitly configures `encrypt_key`, `port`, and `callback_path`

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [ ] Manual test with Lark international account using WebSocket mode

Fixes #449

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -306,9 +306,11 @@ func (p *Platform) Start(handler core.MessageHandler) error {
 			return p.onBotMenu(event)
 		})
 
-	// Lark international version uses Webhook mode, not WebSocket long connection
-	// Feishu domestic version supports WebSocket long connection
-	if p.platformName == "lark" {
+	// Lark international version now supports WebSocket long connection (since SDK v3.5.3).
+	// Use WebSocket by default for all platforms; only fallback to webhook if explicitly configured.
+	// Users can force webhook mode by setting encrypt_key + port + callback_path together.
+	if p.encryptKey != "" && p.port != "" && p.callbackPath != "" {
+		slog.Info(p.tag()+": using webhook mode (encrypt_key configured)", "port", p.port)
 		return p.startWebhookMode()
 	}
 


chenhg5:

chenhg5 (commented):
## Conclusion: Comment (functionally Approve)

The fix's **intent** is correct and valuable (allow Lark international users to use WebSocket without a public IP), and the issue's exact "Suggested Fix" has **already been merged into main** as `shouldUseWebhookMode()`. The PR's specific 3-field check (`encryptKey + port + callbackPath`) is over-conservative — main's 1-field check (`encryptKey`) matches the issue body's suggested fix and the SDK semantics (`encryptKey` is webhook-specific). Rebase is non-trivial because main has also refactored shared group fan-out in the same Start function. I walked through the rebase resolution below; after rebase, ready-to-merge.

## What the PR does (verified against diff + main)

| File | Change | Lines |
|---|---|---|
| `platform/feishu/feishu.go` | Removes `p.platformName == "lark"` check (3 lines) + adds `encryptKey + port + callbackPath` 3-field check (4 lines) | +5 / -3 |

## Code-side verification (all PASS)

| Claim | Code evidence | Verdict |
|---|---|---|
| Lark SDK v3.5.3+ supports WebSocket for international via `larkws.WithDomain()` | `larkws.WithDomain(p.domain)` already in `startWebSocketMode` | ✓ Accurate |
| `startWebSocketMode()` already handles non-Feishu domains | `feishu.go:494-496`: `if p.domain != lark.FeishuBaseUrl { wsOpts = append(wsOpts, larkws.WithDomain(p.domain)) }` | ✓ Accurate |
| The fix allows Lark international users to avoid public IP | Removes forced webhook for Lark | ✓ Accurate |
| Fixes #449 | The fix targets the issue's exact code location | ✓ Accurate |

## What main has already done (issue body's suggested fix)

The issue #449 body contains an explicit "Suggested Fix":

```go
// Use webhook mode only when explicitly configured (encrypt_key present)
if p.encryptKey != "" {
    return p.startWebhookMode()
}
return p.startWebSocketMode()
```

Main has implemented this **exactly** via the `shouldUseWebhookMode()` helper:

```go
// platform/feishu/feishu.go:458-464 (current main)
if p.useInteractiveCard {
    slog.Info(p.platformName + ": interactive card mode enabled, ...")
}

if p.shouldUseWebhookMode() {
    return p.startWebhookMode()
}
return p.startWebSocketMode()
}

// platform/feishu/feishu.go:465-467 (helper)
func (p *Platform) shouldUseWebhookMode() bool {
    return strings.TrimSpace(p.encryptKey) != ""
}
```

So the issue's intent is **already in main**. The PR adds nothing new at the semantic level.

## Why the PR's 3-field check is over-conservative

The PR checks `encryptKey + port + callbackPath` for webhook mode. Main's 1-field check (`encryptKey`) is more correct because:
- `encryptKey` is the **encryption key for webhook event payloads** — its presence is the canonical signal that the user wants webhook mode (WebSocket doesn't use encryption in Lark).
- `port` and `callbackPath` are HTTP server config — they're required for webhook to function, but checking them as a "user wants webhook" signal is brittle (a user could set them by accident without intending webhook mode).
- The SDK itself uses `encryptKey` as the webhook mode indicator (see `dispatcher.NewEventDispatcher("", p.encryptKey)` at `feishu.go:394`).

If a user sets only `encryptKey` (forgetting port/callback_path), main's behavior falls into webhook mode and `startWebhookMode` would fail at runtime (server tries to listen on `:` with empty port). The PR's 3-field check would correctly fall through to WebSocket in this edge case. **But this is a user config error, not a library bug** — a user who set `encrypt_key` clearly wants webhook and should also set `port` and `callback_path`. The PR's 3-field check adds defensive code at the cost of clarity.

**Verdict on this point**: main's 1-field check is correct. The PR's 3-field check is over-defensive. Recommend the author drop the 3-field check in favor of main's behavior (use `shouldUseWebhookMode()` after rebase).

## Rebase analysis (significant conflict)

I rebased onto current main `096ed3ea`. The conflict is more substantial than a 1-line change because main has refactored the surrounding `Start()` function extensively:

1. **Shared group fan-out**: `OnP2CardActionTrigger` and `OnP2BotMenuV6` now iterate over `p.sharedGroup.allPlatforms()` for multi-platform webhook sharing (added by a separate refactor).
2. **Webhook/websocket helper**: `shouldUseWebhookMode()` extracted as a method.
3. **Interactive card log**: added as a separate `if p.useInteractiveCard` block before the webhook/websocket decision.

After resolution, the rebased commit is effectively a no-op (just a 3-line comment addition; the actual logic matches main):

```diff
+ // Use WebSocket by default for all platforms (Feishu + Lark international).
+ // Fall back to webhook only when encrypt_key is explicitly set (which implies webhook mode).
+ // This matches the issue body's suggested fix for #449.
  if p.shouldUseWebhookMode() {
      return p.startWebhookMode()
  }
```

`go vet -tags no_web ./platform/feishu/...` — clean. `go test -tags no_web -count=1 ./platform/feishu/...` — `ok github.com/chenhg5/cc-connect/platform/feishu 25.341s`.

## What looks good

- **The fix's intent is correct**: Lark international users should not be forced to webhook mode when they have no public IP.
- **Backed by official Lark docs** (linked in issue #449) and production evidence (OpenClaw has been running Lark international with WebSocket).
- **Strictly better than the original code**: original forced webhook for all `lark` platform; main's `shouldUseWebhookMode` correctly defaults to WebSocket.
- **The 3-field check** is over-defensive but technically correct (it would prevent the user-config-error scenario described above). After rebase, this difference is moot because main's helper is used.

## Should improve (P2/P3, not blocking)

1. **P2 — Use main's `shouldUseWebhookMode()` after rebase**: don't keep the 3-field check. The SDK semantics + issue body both support the 1-field check. Less code, clearer intent.
2. **P2 — Comment about SDK version**: the PR's comment says "since SDK v3.5.3". This is a useful historical note but should be a doc comment near the helper, not in the body. Recommend moving to a `// shouldUseWebhookMode` doc comment.
3. **P3 — Test coverage**: no test added for the mode selection. The current `startWebhookMode` / `startWebSocketMode` behavior isn't covered by unit tests either. Recommend adding `TestShouldUseWebhookMode_*` for: `encryptKey=""` → false; `encryptKey="abc"` → true; `encryptKey="   "` → false (whitespace handling).
4. **P3 — Manual test item** in test plan ("use Lark international account with WebSocket mode") is the only unchecked item. End-to-end integration; defer to post-merge smoke.

## Risk

- **Very low.** After rebase to main's logic, the only diff is a 3-line comment. No behavior change. No schema change. Easy to revert.
- **If PR's 3-field check is preserved instead of main's 1-field**: small behavior change — a user setting `encryptKey` but forgetting `port`/`callback_path` would get WebSocket instead of webhook. This is arguably better (defensive), but contradicts the issue body's intent.
- **No migration concern.**
- **No rollback concern.**

## Next step

- **Author (chenhg5)**:
  1. Rebase onto current main (conflict requires resolving `OnP2CardActionTrigger`/`OnP2BotMenuV6` shared-group fan-out + replacing PR's 3-field check with main's `shouldUseWebhookMode()`). Walked through above.
  2. Drop the PR's 3-field check in favor of main's `shouldUseWebhookMode()`.
  3. Re-run CI for fresh green (2-month-old green is stale).
  4. Consider adding tests for `shouldUseWebhookMode`.
- **Owner**: this PR's intent is already in main. If author doesn't rebase within a cycle or two, recommend **close as SUPERSEDED** (the fix is already shipped via the refactor that introduced `shouldUseWebhookMode`). Closing is cleaner than carrying a redundant PR.
- **PM**: 4th SUPERSEDED pattern this cycle (this + #349 + #254 + #581). Many "rebase + re-verify" tasks accumulating. Consider a dev-claudecode sprint to triage the entire qa backlog.

— cc-connect/qa-claudecode

PR #475, opened 2026-04-06
feat(slack): implement FileSender for direct file uploads

ht290:
Adds SendFile to the Slack platform, mirroring SendImage. Both use Slack's UploadFileV2Context, which already handles arbitrary file types. The engine routes attachments from `cc-connect send --file` (invoked by agents via the system-prompt-documented CLI) to this method via the core.FileSender interface assertion.

https://claude.ai/code/session_01WyqXEst7Z8tiMbzeh4rTfX

### DIFF
--- a/platform/slack/slack.go
+++ b/platform/slack/slack.go
@@ -414,6 +414,34 @@ func (p *Platform) SendImage(ctx context.Context, rctx any, img core.ImageAttach
 
 var _ core.ImageSender = (*Platform)(nil)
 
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM! Clean implementation of FileSender for Slack.

**What I like:**
- Mirrors the existing `SendImage` pattern using `UploadFileV2Context`
- Implements `core.FileSender` interface assertion
- Proper error handling with context propagation

Thanks for the contribution!

PR #477, opened 2026-04-07
fix(feishu): avoid model card callback timeouts

q107580018:
## Summary
- route Feishu `/model` card actions through async command dispatch instead of synchronous card navigation
- return a quick toast response so OpenCode model switching no longer blocks the Feishu callback window
- add a regression test covering async `/model switch` card behavior

## Test Plan
- [x] `go test ./platform/feishu -run TestInteractivePlatform_ModelCardActionDispatchesCommandAsync -count=1`
- [x] `go test ./platform/feishu -count=1`
- [x] `go test ./core -run 'TestHandleCardNav|TestRenderModelCard|TestResolveModelAlias|TestParseModelSwitchArgs' -count=1`
- [ ] `go test ./...` currently fails on unrelated existing test `TestCmdShell_MultiWorkspaceIgnoresMissingSharedBinding`

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -106,19 +106,19 @@ type replyContext struct {
 }
 
 type Platform struct {
-	platformName          string
-	domain                string
-	appID                 string
-	appSecret             string
-	progressStyle         string
-	useInteractiveCard    bool
-	self                  core.Platform
-	reactionEmoji         string
-	allowFrom             string
+	platformName               string
+	domain                     string
+	appID                      string
+	appSecret                  string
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM! Good fix for Feishu model card callback timeout.

**What I like:**
- Routes `/model` card actions through async command dispatch instead of synchronous card navigation
- Returns quick toast response to avoid blocking Feishu callback window
- Good regression test
- Clean pattern: `act:/model` prefix matching + goroutine dispatch + toast response

Thanks for the contribution!

PR #483, opened 2026-04-07
fix(claudecode): handle mid-turn compaction events in readLoop

### LINKED ISSUE #481: bug: Claude Code context compaction mid-turn causes task interruption
## Problem

When Claude Code performs automatic context compaction (condensing conversation history to fit within context limits) **during a turn** (mid-task), cc-connect may misinterpret compaction-related events as turn completion, causing the task to be interrupted prematurely.

## Root Cause

In `agent/claudecode/session.go`, the `readLoop` event switch (L177-191) has **no default case** — any unrecognized event type from Claude Code is silently dropped:

```go
switch eventType {
case "system":    cs.handleSystem(raw)
case "assistant": cs.handleAssistant(raw)
case "user":      cs.handleUser(raw)
case "result":    cs.handleResult(raw)
case "control_request": cs.handleControlRequest(raw)
case "control_cancel_request": // handled
// ← no default: unknown events silently ignored
}
```

If Claude Code emits a compaction-related event followed by an `EventResult`, the engine's `processInteractiveEvents` (in `core/engine.go`) treats that `EventResult` as turn completion and returns, dropping any subsequent tool calls or assistant responses.

## Observed Behavior

- User sends a complex task to cc-connect
- Claude Code begins processing (tool calls, file reads, etc.)
- Claude Code hits context limit and triggers automatic compaction
- cc-connect loses track of the in-progress task
- User receives an incomplete or truncated response

## Expected Behavior

cc-connect should either:
1. Recognize compaction events and continue waiting for the actual task completion
2. Or at minimum, log unknown event types so the issue is diagnosable

## Notes

- cc-connect's own **auto-compress feature** (`[auto_compress]` config) does NOT have this problem, because it triggers `/compact` between turns (after `EventResult`), uses a dedicated `processCompressEvents` handler, and holds the session lock throughout
- The issue only occurs when **Claude Code itself** initiates compaction mid-turn
- Adding a `default` case with debug logging in `readLoop` would be a good first step to understand the exact event sequence during compaction

## Environment

- cc-connect: v1.2.2-beta.5
- Agent: Claude Code (claudecode)
- Observed on: macOS (darwin-arm64)

chenhg5:
## Summary
- Add default case in event switch to log unknown event types for diagnostic purposes
- Handle compaction result events by checking `subtype` field (`compact` or `compaction`)
- Set `Done=false` for compaction result events so the event loop continues processing

## Problem
When Claude Code performs automatic context compaction mid-turn, cc-connect could misinterpret compaction-related events as turn completion, causing the task to be interrupted prematurely. Unknown event types were silently dropped, making diagnosis difficult.

## Solution
1. **Diagnostic logging**: Unknown event types are now logged with their full content, helping identify new event types Claude Code may emit
2. **Compaction detection**: Result events with `subtype: "compact"` or `subtype: "compaction"` are treated as intermediate events (Done=false), not turn completion

## Test plan
- [x] `go vet ./agent/claudecode/...` passes
- [x] `go test ./agent/claudecode/...` passes
- [x] Unit tests for Done=true on normal result
- [x] Unit tests for Done=false on compaction subtypes

Closes #481

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -188,6 +188,11 @@ func (cs *claudeSession) readLoop(stdout io.ReadCloser, stderrBuf *bytes.Buffer)
 		case "control_cancel_request":
 			requestID, _ := raw["request_id"].(string)
 			slog.Debug("claudeSession: permission cancelled", "request_id", requestID)
+		default:
+			// Log unknown event types for diagnostic purposes.
+			// Claude Code may emit new event types (e.g., during context compaction)
+			// that we don't yet handle. Logging helps identify and fix issues.
+			slog.Debug("claudeSession: unknown event type", "type", eventType, "event", raw)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


chenhg5 (commented):
LGTM. Correctly handles compaction events by checking subtype field, preventing premature turn interruption. Good diagnostic logging for unknown events.

chenhg5:

PR #491, opened 2026-04-08
fix(claudecode): send SIGTERM instead of SIGKILL for graceful session close

### LINKED ISSUE #490: bug: SessionEnd hook never fires for interactive sessions due to SIGKILL
# bug: SessionEnd hook never fires for interactive sessions due to SIGKILL

## Problem

Claude Code's `SessionEnd` hook (and effectively any hook that runs at process termination) **never fires** when cc-connect closes an interactive session — whether via `/new`, `/switch`, `/clear`, `/delete`, idle timeout, daemon restart, or process crash recovery.

This means any post-processing logic configured in `.claude/settings.json` under `hooks.SessionEnd` or `hooks.PreCompact` is silently skipped for all cc-connect interactive sessions.

In contrast, the same hooks **work correctly** for:
- `claude --print` direct invocations (e.g. one-shot scripts, Paperclip-style workflows)
- `claude` interactive sessions running natively in a terminal (without cc-connect)

So the issue is specific to cc-connect's interactive bridge.

## Root Cause

In `agent/claudecode/session.go` (verified in both `v1.2.1` and `v1.2.2-beta.5`), the Claude process is started with:

```go
// session.go:92 (v1.2.1) / :92 (beta.5)
cmd := exec.CommandContext(sessionCtx, "claude", args...)
cmd.Dir = workDir
// ❌ no cmd.Cancel
// ❌ no cmd.WaitDelay
```

Per Go's `os/exec` documentation:

> CommandContext sets the command's Cancel function to invoke the Kill method on its Process. [...] By default, CommandContext sets Cancel to call the Kill method on the command's Process.

`Process.Kill()` on Unix translates to **SIGKILL**, which is uncatchable. The `claude` process dies instantly without running any cleanup, including its registered hooks.

The `Close()` method (line 467 in 1.2.1, line 565 in beta.5) then makes things worse by also explicitly calling `Process.Kill()` as a fallback after an 8-second timeout — both branches use SIGKILL:

```go
func (cs *claudeSession) Close() error {
    cs.cancel()  // SIGKILL via default cmd.Cancel
    select {
    case <-cs.done:
        return nil
    case <-time.After(8 * time.Second):
        slog.Warn("claudeSession: graceful close timed out, killing process")
        if cs.cmd != nil && cs.cmd.Process != nil {
            _ = cs.cmd.Process.Kill()  // SIGKILL again
        }
```

There is no SIGTERM, no `stdin.Close()` (which would trigger Claude's natural EOF exit), no graceful path of any kind for the Claude Code agent.

## Reproduction

1. Configure `.claude/settings.json` in any project with a `SessionEnd` hook that writes to a log file:
   ```json
   {
     "hooks": {
       "SessionEnd": [
         {
           "matcher": "",
           "hooks": [{"type": "command", "command": "echo SessionEnd >> /tmp/hook.log"}]
         }
       ]
     }
   }
   ```
2. Open a Discord session bridged through cc-connect, send a message, wait for response
3. Send `/new` (or `/delete`, `/clear`, or just wait for idle timeout)
4. Check `/tmp/hook.log` — **empty**

For comparison, run `claude --print "hello"` directly in the same project — `/tmp/hook.log` will contain `SessionEnd`.

## Comparison with Codex agent

The Codex agent in cc-connect has dedicated `proc_unix.go` / `proc_windows.go` files implementing nuanced termination handling. The Claude Code agent has no equivalent — it's a clear gap.

## Suggested Fix

Set `cmd.Cancel` and `cmd.WaitDelay` on the `exec.Cmd` after creation, before `cmd.Start()`. This requires only ~3 lines:

```go
cmd := exec.CommandContext(sessionCtx, "claude", args...)
cmd.Dir = workDir

// Send SIGTERM first to allow Claude Code to run cleanup hooks (SessionEnd, PreCompact).
// After WaitDelay, escalate to the default SIGKILL behavior.
cmd.Cancel = func() error {
    return cmd.Process.Signal(syscall.SIGTERM)
}
cmd.WaitDelay = 8 * time.Second
```

With this change:
1. `cs.cancel()` triggers `cmd.Cancel`, which sends SIGTERM
2. Claude Code receives SIGTERM, runs its registered hooks, exits cleanly
3. If Claude Code doesn't exit within `WaitDelay` (8 seconds), Go's runtime escalates to `Process.Kill()` (SIGKILL) automatically
4. The existing `Close()` 8-second timeout fallback is preserved as a safety net

The behavior is fully backward-compatible: misbehaving processes still get killed, just with a clean shutdown opportunity first.

## Related issues

- **#457**: `UserPromptSubmit` hooks also don't fire in stream-json mode (different root cause but related theme: cc-connect has gaps in Claude Code hook support)
- **#481**: cc-connect drops unknown events from Claude Code in `readLoop`, potentially related to hook events being lost

## Environment

- cc-connect: tested on `v1.2.1` and `v1.2.2-beta.5`
- Agent: Claude Code (claudecode)
- OS: Ubuntu 24.04 LTS (vmi2660635), kernel 6.8

## Use case

We use `SessionEnd` hooks to capture conversation summaries into a long-term memory system (similar to https://github.com/coleam00/claude-memory-compiler). The hooks work perfectly for direct `claude --print` invocations (e.g. in cron jobs) and for native interactive sessions on a workstation, but are completely silent for any session bridged through cc-connect — which is the majority of our daily traffic.

Happy to send a PR with the fix if the maintainers agree on the approach.

chenhg5:
## Summary
Set `cmd.Cancel` and `cmd.WaitDelay` to allow Claude Code to run cleanup hooks (SessionEnd, PreCompact) before the process is terminated.

Before this fix, `exec.CommandContext`'s default Cancel behavior sent SIGKILL, which is uncatchable and prevented Claude Code from running any registered hooks on session close (via `/new`, `/switch`, `/clear`, `/delete`, idle timeout, daemon restart, etc.).

## Changes
- `cmd.Cancel` sends SIGTERM instead of SIGKILL
- `cmd.WaitDelay = 8 * time.Second` for graceful shutdown window
- Go runtime automatically escalates to SIGKILL if process doesn't exit within WaitDelay
- Backward compatible: misbehaving processes still get killed after 8 seconds

## Test plan
- [x] `go vet ./agent/claudecode/...` passes
- [x] `go test ./agent/claudecode/...` passes (all 26 tests)
- [x] Manual verification: SIGTERM sent on session close

## How it works
1. `cs.cancel()` triggers `cmd.Cancel`, which sends SIGTERM
2. Claude Code receives SIGTERM, runs its registered hooks, exits cleanly
3. If Claude Code doesn't exit within `WaitDelay` (8 seconds), Go's runtime escalates to `Process.Kill()` (SIGKILL) automatically
4. The existing `Close()` 8-second timeout fallback is preserved as a safety net

Closes #490

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/chenhg5/cc-connect/core"
@@ -98,6 +99,13 @@ func newClaudeSession(ctx context.Context, workDir, model, sessionID, mode strin
 		env = core.MergeEnv(env, extraEnv)
 	}
 	cmd.Env = env
+
+	// Send SIGTERM first to allow Claude Code to run cleanup hooks (SessionEnd, PreCompact).
+	// After WaitDelay, escalate to the default SIGKILL behavior.
+	cmd.Cancel = func() error {
+		return cmd.Process.Signal(syscall.SIGTERM)
+	}
+	cmd.WaitDelay = 8 * time.Second
 
 	stdin, err := cmd.StdinPipe()
 	if err != nil {


chenhg5:

chenhg5 (commented):
QA review (comment — same account as PR author, so using `--comment` instead of `--request-changes`):

> **Note:** My GitHub CLI is authenticated as `@chenhg5`, the PR author. GitHub blocks `--request-changes` on a PR you opened yourself, so I'm filing this as a comment with an explicit "needs author action" call-out. The role-doc conclusion for these blockers would normally be `Request changes`.

Conclusion: Comment (functionally request-changes, see note above)

Overall assessment:
- The original issue #490 (SessionEnd hook never fires) is **mostly solved** by commits on main that landed after this PR was opened. The 8 lines this PR adds are useful for a narrow remaining gap (context-cancel path) but the rest is redundant. The branch is also 2 months stale and **CONFLICTING with main** — it cannot merge as-is.

Review scope:
- Reviewed `agent/claudecode/session.go` (the only changed file).
- Compared with the current main implementation (`eb868664`) of `claudeSession.Close()` and `prepareCmdForKill` in `proc_unix.go`.
- Re-read issue #490 to understand the original bug.

What looks good:
- The intent is correct: SIGTERM gives Claude Code a chance to run cleanup hooks (SessionEnd, PreCompact, claude-mem session summary, etc.) before SIGKILL.
- `WaitDelay = 8s` is a reasonable default for Claude Code's cleanup.
- No regressions in the 8 lines themselves; the diff is minimal and easy to review.

Needs author action (P1):
1. **The branch is CONFLICTING with main.** `gh pr view 491` reports `mergeable: CONFLICTING`. The insertion point moved after the run-as-user refactor. Author must rebase before any merge attempt.
2. **Scope is partially redundant** with the main implementation. The current `claudeSession.Close()` already does a 3-phase shutdown (verified in main `agent/claudecode/session.go:822-863`):
   - Phase 1: stdin EOF — Claude Code runs Stop hooks, exits cleanly.
   - Phase 2: SIGTERM to the entire process group via `signalProcessGroup` (`agent/claudecode/proc_unix.go:31`).
   - Phase 3: SIGKILL to the entire process group after 5s timeout.
   This means the original #490 bug is fixed for the **normal Close() path** by main (since `eadc52a0 fix: graceful session close — stdin EOF before signals (#509)` landed).

Should improve (P2):
1. **No tests** for the SIGTERM-on-context-cancel path. A unit test that uses a fake `claude` binary trapping SIGTERM and reporting via stdout would catch the regression and prove the fix.
2. **PR description doesn't acknowledge the related `Close()` evolution.** After rebase, the PR should explain what's still missing: namely, the context-cancel path (when the parent process receives SIGTERM, the test framework cancels the context, or the daemon exits without calling `Close()`).
3. **The remaining gap is bigger than this PR addresses.** On context cancellation, `cmd.Cancel` (defaulting to `cmd.Process.Kill()`) SIGKILLs the **direct child only**, leaving MCP grandchildren as orphans (the 100% CPU bug from #490's discussion). The fix should use `signalProcessGroup(cmd, syscall.SIGTERM)` from `agent/claudecode/proc_unix.go:31`, not `cmd.Process.Signal(SIGTERM)`. Without the group-wide signal, this PR fixes half the problem.
4. **8s WaitDelay is hard-coded.** The current main has a configurable `gracefulStopTimeout` field on `claudeSession`. If the PR is kept, `WaitDelay` should match that config (or document why 8s is the right default).

Alternative path to consider:
- Given that 90% of the original issue is solved by main, the most efficient path is for the author to either:
  - (a) Update this PR to use `signalProcessGroup(cmd, SIGTERM)` for proper group coverage + add a unit test + rebase onto current main, **or**
  - (b) Close this PR as superseded and open a new (smaller) one for the context-cancel-path gap only.

Testing / Risk:
- Verified in this review:
  - `go vet ./agent/claudecode/...` — clean on the PR branch.
  - `gh pr checks 491` — all green on the stale branch (but on stale code, so not meaningful).
  - Diff is +8 lines, scoped to the cmd construction block.
- Not covered:
  - No test on this PR for the SIGTERM path.
  - No live test of `cmd.Cancel` firing (would require killing the parent context — hard to reproduce in CI).
- Regression risk if merged as-is: low. The conflict is the actual blocker.

Next step:
- **Author (@chenhg5):** please decide between (a) rebase + use `signalProcessGroup` + add a test, or (b) close as superseded. I lean toward (b) since the original #490 is solved by main; the remaining context-cancel gap is small and deserves a fresh, focused PR.
- **PM / release-codex:** if author chooses (b), no follow-up needed.
- **QA:** happy to re-review after rebase.

chenhg5:
Hi, closing this older conflicting PR as superseded. Current main already has staged graceful Claude Code shutdown using stdin close, SIGTERM, then SIGKILL fallback.\n\n你好，这个较早且已冲突的 PR 先按已被替代关闭。当前 main 已经有分阶段的 Claude Code 优雅停止逻辑：stdin close、SIGTERM，然后 SIGKILL 兜底。

If this still contains missing behavior on current main, please open a fresh PR rebased on current main and link back here.

如果当前 main 仍然缺少这里的能力，请基于当前 main 重新提交一个 PR，并链接回这个 PR。

chenhg5:
Clarification / 补充说明：closing as superseded because current main already has staged graceful Claude Code shutdown using stdin close, SIGTERM, then SIGKILL fallback. If this PR still covers a missing edge case, please reopen as a fresh PR rebased on current main.

补充说明：按已被替代关闭，因为当前 main 已有分阶段 Claude Code 优雅停止逻辑：stdin close、SIGTERM，然后 SIGKILL 兜底。如果仍有遗漏边界，请基于当前 main 重新提交 PR。

Issue #492, opened 2026-04-08
[BUG] Forced Context Drift via 15-Minute Session Termination and --continue Loop


leighstillard:
### cc-connect Version

v1.2.2-beta.4-54-g227c011-dirty

### Operating System

Linux (Ubuntu/Debian)

### Agent Type

Claude Code

### Platform

Slack

### Installation Method

Build from source

### Bug Description

The current session management lifecycle in cc-connect creates a "forced drift" pattern. Because sessions are automatically terminated after 15 minutes of inactivity/timeout, cc-connect as default uses the --continue flag to resume their work within the Slack interface and I presume all others.

While --continue is intended to maintain state, the combination of the Slack-mediated interface and the frequent session restarts creates a feedback loop that rapidly pollutes the model's context window with fragmented, noisy, and potentially erroneous historical data, leading to architectural erosion and "hallucination anchoring." This is further compounded by the reduced effort and inference quality being experienced in Claude subscription users currently. 

### Steps to Reproduce

1.  **Initialize a Complex Task:** Start a session via the Slack interface with a high-level architectural requirement (e.g., *"Refactor the existing API layer to use a Repository pattern and ensure all database calls are abstracted."*).
2.  **Execute Incremental Changes:** Perform several turns of coding. During these turns, intentionally or accidentally introduce a minor deviation—such as a typo in a variable name, a failed terminal command, or a slightly non-compliant code snippet.
3.  **Trigger Session Timeout:** Allow the session to remain idle for the 15-minute threshold to trigger a session termination.
4.  **Resume via `--continue`:** Use the `--continue` command to restart the session and pick up where the previous one left off.
5.  **Iterate the Cycle:** Repeat steps 2 through 4 several times (e.g., 3–4 session restarts).
6.  **Observe Model Output:** Ask the model to perform a new task that requires adherence to the original architectural requirement established in Step 1.

### Expected Behavior

*   **Architectural Persistence:** The agent should maintain a high "signal-to-noise" ratio. The original system instructions and architectural goals should remain the primary drivers of the model's logic, regardless of how many times the session has been restarted.
*   **Contextual Integrity:** The model should be able to distinguish between the "current task state" and "historical noise" (such as previous error logs or failed command outputs) contained within the continued context.
*   **Clean Continuity:** The `--continue` mechanism should ideally provide a seamless transition of *intent*, not just a raw dump of *transcript history*.

### Actual Behavior

[...]

PR #494, opened 2026-04-08
feat(core): default reset_on_idle_mins to 30 to prevent context drift

leighstillard:
## Summary

Defaults `reset_on_idle_mins` to **30 minutes** when unset, with a one-time startup log nudging users toward the docs. Builds directly on @xukp20's #400 — no new mechanism, just a default-behavior change.

## Why

The workspace pool evicts idle session entries after ~15 minutes. The next message after eviction rehydrates the session via Claude Code's `--continue`, which re-ingests the entire prior transcript. Across many cycles this accumulates stale chat history (failed commands, debugging noise, abandoned tangents) and the model's attention drifts from the original architectural intent — the "context drift loop" reported in #492.

#400 already shipped the right mechanism for this: idle-driven rotation to a fresh active session, with the prior session preserved for `/list` / `/switch`. The only gap was that it was opt-in and defaulted to disabled, so most deployments still hit the drift loop. As @xukp20 noted on #492:

> #492 may still remain valid as a product/default-behavior issue, because `reset_on_idle_mins` is currently opt-in rather than the default.

This PR closes that gap.

## What changed

- `cmd/cc-connect/main.go`: introduce `defaultResetOnIdleMins = 30` and a small `resolveResetOnIdle()` helper that both call sites now share. When the field is unset, the default is applied and a one-time `slog.Info` is emitted per project at startup pointing users at the docs and the opt-out.
- `cmd/cc-connect/main_test.go`: unit test for `resolveResetOnIdle` covering nil / explicit 0 / explicit positive.
- `README.md`, `docs/usage.md`, `config.example.toml`: document the new default and the opt-out (`reset_on_idle_mins = 0`).

## Why 30 minutes

Twice the workspace pool TTL (`15 * time.Minute` in `core/engine.go`):

- long enough that a coffee break or a meeting doesn't trigger rotation mid-task
- short enough that a "come back after lunch" *does* rotate, which is exactly the drift trigger #492 describes

## Behavior change & opt-out

Existing deployments that rely on always-resume will see new behavior. They can restore the prior behavior with one line:

```toml
[[projects]]
reset_on_idle_mins = 0
```

The startup log calls this out per project so it isn't silent.

## Test plan

- [x] `go build ./...`
- [x] `go test ./cmd/cc-connect/ ./core/ ./config/`
- [x] Manual: verify a project with no `reset_on_idle_mins` logs the nudge at startup and rotates after 30 min idle
- [x] Manual: verify `reset_on_idle_mins = 0` suppresses both the rotation and the nudge
- [x] Manual: verify `reset_on_idle_mins = 5` honors the explicit value

Refs: #492
Credit: @xukp20 for #400, which this builds on


chenhg5 (approved):
LGTM. Clean, minimal change with good rationale — defaults reset_on_idle_mins to 30 min to prevent context drift from accumulating stale transcripts. The resolveResetOnIdle helper is simple and correct, the test coverage is complete, and the one-time startup log is a nice touch to help users discover the opt-out. Documentation is thorough without being verbose.

chenhg5 (approved):
LGTM — CI green, conflict-free, implementation reviewed. Approved for merge.

PR #499, opened 2026-04-08
fix(core): fix flaky shell test on macOS due to symlink path resolution

happyTonakai:
## Summary
  - Fix `TestCmdShell_MultiWorkspaceIgnoresMissingSharedBinding` failing on macOS
  - On macOS, `/var/folders/` is a symlink to `/private/var/folders/`. The test was comparing shell output (which
  uses the original path) against `normalizeWorkspacePath` result (which resolves the symlink), causing a
  mismatch.
  - The fix checks both the original and resolved paths in the assertion.

  ## Test plan
  - [x] `go test ./core/ -run TestCmdShell_MultiWorkspaceIgnoresMissingSharedBinding` passes on macOS
  - [x] `go test ./...` all pass


chenhg5 (approved):
LGTM. Tiny, correct fix for macOS symlink path resolution in a flaky test. The assertion now checks both original and resolved paths to handle /var/folders -> /private/var/folders symlink. Minimal change, focused fix.

PR #517, opened 2026-04-09
fix: prevent /new session race by detaching state before async close

ding-dan:
## Problem

A race condition in `/new`, `/switch`, and `/dir` commands causes session confusion. When the user sends `/new` while an agent session is active, `cleanupInteractiveState()` blocks for up to 130 seconds waiting for the old agent process to exit. During this blocking window:

1. The old event loop exits (stopCh closed) → the session lock is released
2. A new message arrives → `TryLock` succeeds on the **old** session (not yet cleared)
3. `getOrCreateInteractiveStateWith` reads the stale `AgentSessionID` → starts `claude --resume <old-id>`
4. The user's message is sent to the **old** conversation context
5. `cmdNew` finally unblocks → clears session data → creates new session → orphans the process from step 3

## Root Cause

`cleanupInteractiveState()` performs two logically independent operations in sequence:
- **Fast**: remove interactive state from routing map + stop event loop
- **Slow**: close the agent process (stdin EOF → wait for exit → SIGTERM → SIGKILL)

Session data (`AgentSessionID`, history) was only cleared **after** both operations completed, leaving a window where stale data could be read by concurrent message handlers.

## Fix

Introduced `detachInteractiveState()` — a non-blocking function that handles only the fast path (map removal + stop signal) and returns the `AgentSession` for later closing.

Command handlers (`cmdNew`, `cmdSwitch`, `dirApply`) now follow a 3-phase pattern:

1. **Detach** — `detachInteractiveState()` removes the state from the routing map and stops the event loop (non-blocking)
2. **Clear session data** — reset `AgentSessionID`, clear history, create new session record — while the state is already detached but the process may still be exiting
3. **Async close** — `closeAgentSessionAsync()` terminates the old process in a background goroutine

This ensures any message arriving during the close window sees a cleared `AgentSessionID` and starts a fresh agent session instead of resuming the old one.

## Additional Changes

- Fixed log prefix in `closeAgentSessionWithTimeout` (`cleanupInteractiveState:` → `closeAgentSessionWithTimeout:`)
- Added nil guards before `closeAgentSessionAsync` calls, consistent with `cleanupInteractiveState` style
- Added `slog.Debug` after detach in `cmdNew` for troubleshooting

## Tests

- `TestDetachInteractiveState_ReturnsAgentSession` — verifies map removal and return value
- `TestDetachInteractiveState_SkipsWhenStateReplaced` — verifies CAS (compare-and-swap) behavior
- `TestCmdNew_ClearsSessionDataBeforeAsyncClose` — regression test using a `slowCloseSession` stub that blocks in `Close()`, verifying that `AgentSessionID` is cleared and the async close goroutine is started before `cmdNew` returns

All existing tests pass (`go test ./...`).

Refs: #509, #510

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2204,12 +2204,25 @@ func (e *Engine) getOrCreateInteractiveStateWith(sessionKey string, p Platform, 
 // a stale goroutine (still running after /new created a fresh Session object and
 // a new turn started on it) from accidentally destroying the replacement state.
 func (e *Engine) cleanupInteractiveState(sessionKey string, expected ...*interactiveState) {
+	agentSession := e.detachInteractiveState(sessionKey, expected...)
+	if agentSession != nil {
+		e.closeAgentSessionWithTimeout(sessionKey, agentSession)
+	}
+}
+
+// detachInteractiveState removes the interactive state from the routing map,
+// stops its event loop, and returns the agent session for later closing.
+// It does NOT close the agent session — the caller is responsible for calling
+// closeAgentSessionWithTimeout or closeAgentSessionAsync on the returned value.
+// When an expected state is provided, detach is skipped if the map entry has
+// been replaced by a different state.
+func (e *Engine) detachInteractiveState(sessionKey string, expected ...*interactiveState) AgentSession {
 	e.interactiveMu.Lock()
 	state, ok := e.interactiveStates[sessionKey]
 	if len(expected) > 0 && expected[0] != nil && state != expected[0] {
 		// Another turn has already replaced the state — skip cleanup.
 		e.interactiveMu.Unlock()
-		return
+		return nil
 	}
 	delete(e.interactiveStates, sessionKey)
 	e.interactiveMu.Unlock()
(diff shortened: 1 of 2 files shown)


chenhg5 (changes requested):
请将这个 PR 拆分为两个独立 PR：

**1. Session race 修复**（core/engine.go、core/engine_test.go、agent/claudecode/session.go）
这部分是好的 — detachInteractiveState 模式正确解决了 /new /switch /dir 在 session 忙时的 race condition。请单独提交这个修复。

**2. 需要移除的文件**（不应包含在这个 PR 里）：
- `cc-connect-private` — 二进制文件，不应该提交到代码库
- `cmd/cc-connect/inject.go`（带 `//go:build private` tag）— private feature，未在 PR description 中说明
- `core/inject.go`、`core/inject_stub.go`、`core/api.go`、`core/management.go` — inject 功能，PR description 完全没提
- `cmd/cc-connect/main.go` — 需要检查是否只有 inject 相关的改动

请拆分后只保留 session race 修复部分，重新提交 review。

chenhg5:

PR #520, opened 2026-04-09
feat(bridge): preserve progress rendering hints in reply contexts

xxb:
## Summary
- record bridge adapter metadata on registration and derive progress-style hints from it
- carry those hints into reconstructed and preview reply contexts so bridge replies keep the right progress rendering mode
- make the compact progress writer prefer reply-context hints before falling back to platform defaults
- add regression coverage for bridge message routing, reconstructed reply contexts, and progress payload handling

## Why
Bridge adapters decide whether progress should render as legacy text, compact previews, or cards. That decision is adapter-specific, but reconstructed reply contexts and preview handles previously lost the adapter hint, which could make bridge flows fall back to the wrong rendering mode.

## Testing
- `nix shell nixpkgs#go -c go test ./...`


chenhg5 (approved):
LGTM. Bridge adapters now record metadata on registration, and ReconstructReplyCtx returns progress style hints (legacy/compact/card) derived from that metadata. This prevents reconstructed reply contexts from falling back to the wrong rendering mode in bridge flows. Clean approach with private struct fields.

PR #523, opened 2026-04-09
feat(core): add busy_action inject mode

### LINKED ISSUE #397: feat: Add `busy_action = "inject"` config to auto-inject messages when session is busy
### Current Behavior

When a session is actively processing a task (bot is "typing"), any incoming message is queued as a new task:

> "Previous request still processing. Use `/btw <message>` to add context to the current turn."

### Desired Behavior

Add a per-project config option `busy_action` that controls what happens when a message arrives while the session is busy:

```toml
[[projects]]
  name = "myproject"
  
  [projects.agent]
    busy_action = "inject"   # "queue" (default) | "inject" (auto-btw)
```

- `"queue"` (default): current behavior — message is queued as next task
- `"inject"`: message is automatically injected into the current turn via btw, same as if the user had typed `/btw <message>`

### Use Case

On mobile, when I see the bot is actively working on a long task and want to add quick context ("are you stuck?", "stop and summarize", "focus on X instead"), I currently have to remember to prefix my message with `/btw`.

With `busy_action = "inject"`, messages sent while the bot is typing would automatically become mid-task context injections — which is almost always what I want during an active session.

### Implementation Notes

The btw injection path already exists and works correctly. This would only require routing incoming messages through that path when the session is busy, conditioned on the project config value.

### Environment

- cc-connect: v1.2.2-beta.5
- Platform: Telegram
- Agent: Claude Code

yzlu0917:
## Summary
  - add project-level `busy_action = "inject"` support
  - auto-forward busy pure-text messages through the existing mid-turn `/btw`
  send path
  - keep images, files, and voice messages on the existing queueing path
  - expose `busy_action` in config details and document it in
  `config.example.toml`
  - add regression tests for config validation, config detail export, and busy-
  session inject behavior

  ## Why
  cc-connect already supports `/btw` for sending a follow-up message into a
  running interactive turn, but the default busy-session behavior still queues
  normal text messages. This adds an opt-in project mode for users who prefer
  busy text messages to behave like `/btw` automatically.

  ## Behavior
  - default: `busy_action = "queue"`
  - opt-in: `busy_action = "inject"`
  - inject mode only applies to plain text while a session is busy
  - attachments and voice still queue to avoid changing side-channel delivery
  behavior

  ## Validation
  - ran `go test ./config`
  - ran `go test ./core`
  - ran `git diff --check`

  Closes #397

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -215,6 +215,7 @@ func main() {
 		}
 		engine.SetShowContextIndicator(showCtx)
 		engine.SetAttachmentSendEnabled(cfg.AttachmentSend != "off")
+		engine.SetBusyAction(proj.BusyAction)
 		engine.SetBaseWorkDir(workDir)
 		engine.SetProjectStateStore(projectState)
 
@@ -396,6 +397,7 @@ func main() {
 		if proj.InjectSender != nil {
 			engine.SetInjectSender(*proj.InjectSender)
 		}
+		engine.SetBusyAction(proj.BusyAction)
 
 		// Wire speech-to-text if enabled
 		if cfg.Speech.Enabled {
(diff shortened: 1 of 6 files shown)


chenhg5 (approved):
LGTM. busy_action inject mode is well-designed — opt-in project-level config, default "queue" preserves existing behavior, inject mode only applies to plain text (attachments/voice still queue), clean integration with existing /btw send path, good test coverage for both inject and queue behaviors.

chenhg5:

PR #525, opened 2026-04-09
fix(feishu): preserve session continuity across nested threads and deduplicate root context

leoliang1997:
## Summary                                                                                    
                                                                                               
  Fixes two related issues with `thread_isolation` mode in the Feishu platform.                 
                                                                                               
  ### Bug 1: Session breaks when bot reply creates nested thread                                
                                             
  **Problem:** When a user @mentions the bot inside a Feishu thread, the bot calls              
  `Im.Message.Reply(userMessageID, ReplyInThread=true)`. Feishu interprets this as creating a  
  **new nested thread** on the user's message, rather than replying within the existing thread. 
  Subsequent messages in that nested thread have a different `root_id` (the user's message ID  
  instead of the original thread root), which produces a different session key — causing
  cc-connect to start a **new agent session** and lose conversation continuity.                

  **Example from real logs:**                                                                   
  Message 1 (user @bots in thread A):
    message_id = ...3c1b                                                                        
    root_id    = ...0913  ← thread A root    
    → session key: feishu:chat:root:...0913
    → session: s92                                                                              
   
  Message 2 (user continues in nested thread created by bot):                                   
    message_id = ...a37                      
    root_id    = ...3c1b  ← changed to message 1's ID!                                          
    → session key: feishu:chat:root:...3c1b  ← different!                                       
    → session: s93  ← new session, conversation context lost                                    
                                                                                                
  **Fix:** Add a `msgSessionMap` (`sync.Map`) on the Platform struct that records `messageID →  
  sessionKey` for every processed message. In `makeSessionKey()`, when computing the session key
   from `root_id`, first check if that root ID exists in the map. If so, reuse the mapped      
  session key instead of generating a new one. This lets nested threads inherit their parent    
  message's session.                                                                           
                                         
  ### Bug 2: Thread root message content repeated on every message                             

  **Problem:** When a user posts directly in a Feishu thread (not quote-replying a specific     
  sub-message), Feishu sets `parent_id = root_id`. The existing `fetchQuotedMessage(parentID)`
  was called on **every message**, prepending the root message content each time. This caused:  
  1. Wasted tokens sending redundant context to the agent
  2. Confusing UX when the AI echoes user input (e.g., during permission prompts) — the root    
  message content is visible to the user in every response                                      
                                                                                                
  **Fix:** Distinguish between "thread context" (`parent_id == root_id`) and "explicit          
  quote-reply" (`parent_id != root_id`):     
  - **Thread context:** Use `rootContextSent` (`sync.Map`) with `LoadOrStore` to inject root    
  message content only on the **first message** of a session. Uses a distinct `[Thread root     
  message from ...]` prefix and truncates at 2000 characters.
  - **Explicit quote-reply:** Preserve existing `fetchQuotedMessage` behavior unchanged (always 
  fetch).                                    
                                         
  ## Changes                                                                                   

  - `platform/feishu/feishu.go`:
    - Add `msgSessionMap sync.Map` and `rootContextSent sync.Map` to Platform struct
    - `makeSessionKey()`: look up `rootID` in `msgSessionMap` to inherit parent session         
    - `onMessage()`: record `messageID → sessionKey` mapping; extract and pass `rootID` to      
  `dispatchMessage()`                                                                           
    - `dispatchMessage()`: branch on `parentID == rootID` for dedup vs always-fetch             
    - New `fetchThreadRootMessage()`: wraps `fetchQuotedMessage` with prefix replacement and    
  truncation                                                                                    
  - `platform/feishu/platform_test.go`:                                                         
    - `TestThreadIsolation_NestedThreadInheritSessionKey`: verifies nested thread inherits      
  parent session key                                                                            
    - `TestThreadRootContext_FirstMessageMarksSession`: verifies first message marks           
  rootContextSent                                                                               
    - `TestThreadRootContext_SecondMessageSkipsRootContext`: verifies subsequent messages skip
  root context                                                                                  
    - `TestFetchThreadRootMessage_Truncation`: verifies 2000 char truncation with suffix       
                                                                                                
  ## Test plan                                                                                 
                                                                                                
  - [x] `go build ./cmd/cc-connect` passes                                                      
  - [x] `go test ./platform/feishu/` — all tests pass (including 4 new tests)
  - [x] Manual test: @bot in Feishu thread → bot replies → continue in same thread → verify same
   session                                                                                      
  - [x] Manual test: verify root message content appears only on first message of thread session
  - [x] Manual test: explicit quote-reply of a sub-message still fetches quoted content every   
  time

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -136,6 +136,8 @@ type Platform struct {
 	botOpenID        string
 	userNameCache    sync.Map // open_id -> display name
 	chatNameCache    sync.Map // chat_id -> chat name
+	msgSessionMap    sync.Map // messageID -> sessionKey; tracks which session a message belongs to
+	rootContextSent  sync.Map // sessionKey -> bool; tracks which sessions already received thread root context
 	// Webhook mode fields (for Lark international version)
 	server       *http.Server
 	port         string
@@ -730,8 +732,15 @@ func (p *Platform) onMessage(ctx context.Context, event *larkim.P2MessageReceive
 	}
 	mentions := msg.Mentions
 	parentID := stringValue(msg.ParentId)
+	rootID := stringValue(msg.RootId)
 
 	sessionKey := p.makeSessionKey(msg, chatID, userID)
+	// Record which session this message belongs to, so that if a bot reply
+	// creates a nested thread on this message, subsequent messages in that
+	// thread can be routed back to the same session.
+	if p.threadIsolation {
+		p.msgSessionMap.Store(messageID, sessionKey)
+	}
 	rctx := replyContext{messageID: messageID, chatID: chatID, sessionKey: sessionKey}
 	slog.Debug(p.tag()+": routed inbound message",
 		"message_id", messageID,
@@ -743,7 +752,7 @@ func (p *Platform) onMessage(ctx context.Context, event *larkim.P2MessageReceive
 	// blocked by IO-heavy operations (image/audio download, handler HTTP calls).
 	// The dedup and old-message checks above remain synchronous to guarantee
 	// correctness before spawning the goroutine.
-	go p.dispatchMessage(ctx, msgType, content, mentions, messageID, sessionKey, userID, userName, chatName, rctx, parentID)
+	go p.dispatchMessage(ctx, msgType, content, mentions, messageID, sessionKey, userID, userName, chatName, rctx, parentID, rootID)
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Bug fix 清晰，两部分都设计合理：
1. `msgSessionMap` (messageID → sessionKey) 让嵌套线程的消息能查到原 session key，维持会话连续性
2. `rootContextSent` (sessionKey → bool) 的 LoadOrStore 确保根消息内容每个 session 只注入一次
全局唯一 messageID 保证了跨聊天不会冲突，测试覆盖了核心路径。

PR #526, opened 2026-04-09
fix(restart.sh): safe deploy with correct backup order and atomic install

leighstillard:
## Summary

Introduces a `restart.sh` deploy script for cc-connect that owns the build/backup/install/restart flow end-to-end, replacing the ad-hoc \"copy a new binary into place and hope\" pattern. Motivated by an incident where a bad manual build left both the live binary and the rollback backup corrupted at the same time.

## What it does

- **Script owns the build.** `go build -o \$STAGED ./cmd/cc-connect` runs in the foreground so compile errors surface to the caller immediately; `\$BINARY` is never touched if the build fails.
- **Post-build verification** via `file` + exec-bit check, to catch the failure mode where `go build` without a package arg produces an ar-archive instead of an ELF executable.
- **Backup captures the OLD binary**, not the new one — the previous script shape backed up after the new binary was in place, which defeated the rollback entirely.
- **Install uses `install -m 0755`**, not `cp`. `cp` over an existing file silently preserves the destination's mode, which can strip the execute bit if the file was ever touched to 0664.
- **`systemctl --user reset-failed`** before every `start`, so a parked `StartLimitBurst` state can't block recovery.
- **`systemd-run --collect`** on both the simple restart and the deploy, so transient units auto-clean after exit and don't block the next invocation.
- **Rollback path** copies the backup over the staged binary on health-check failure, then re-runs the health check.
- **Ownership contract** documented in the header: \$BINARY and \$BACKUP must be writable by the invoking user; the script cannot fix chown issues itself.

`.gitignore` additions for the deploy artifacts (`cc-connect.new`, `cc-connect.known-good`, `web/dist/*`) so they stay out of git.

## Test plan

- [x] `./restart.sh` (simple bounce) exercises the systemd-run detachment and the `--collect` flag; service comes back up without manual `reset-failed`
- [x] `./restart.sh --deploy` builds, backs up, installs atomically, and restarts the service; verified against the run_as_user + Slack delivery work on a separate branch
- [x] Failure injection: a deliberately broken build is caught by the pre-install `file` verification and does not touch \$BINARY
- [x] Rollback path: manually corrupted the new binary; rollback restored the known-good and the service came back up


chenhg5 (changes requested):
This PR has merge conflicts that need to be resolved before merging. Please rebase onto the latest main branch and push the updates.

**PR**: https://github.com/chenhg5/cc-connect/pull/526

chenhg5 (approved):
LGTM. Confirmed MERGEABLE — conflicts resolved. The restart.sh script is well-documented with clear rationale for each design choice. CI all green.

**PR**: https://github.com/chenhg5/cc-connect/pull/526

leighstillard:
Merged locally into feature/thread-affinity-router — restart.sh now included in #533

PR #527, opened 2026-04-09
fix(slack): delivery correctness — threading, MCP blocklist, context injection

leighstillard:
## Summary

Four interlocking fixes to how cc-connect delivers messages on the Slack platform. Individually they look like small edges; together they close the failure mode where a cc-connect session would (a) post as the human user's claude.ai account instead of the bot, (b) drop intermediate events into the main channel instead of a thread, and (c) have the in-progress PlatformContext feature blocked behind a broken CI.

## Commits

1. **`feat(slack): inject channel/thread context block into agent prompt`**
   Adds a structured \`## Slack Message Context\` block (channel_id, channel_name, user_id, user_name, message_ts, thread_ts) to \`core.Message.PlatformContext\`. Threaded through \`buildSenderPrompt\` and populated by the Slack platform at all three message sites (real messages, app_mentions, slash commands). Lets the agent address the right channel/thread when making Slack-side decisions.

2. **`fix(test): update buildSenderPrompt call sites for PlatformContext param`**
   Unblocks CI: the PlatformContext commit added a fifth parameter to \`buildSenderPrompt\` and updated every production call site, but missed the four test call sites in \`core/engine_test.go\`. Pass an empty \`platformContext\` in each test — the existing assertions about the sender header remain valid because the new prefix path is a no-op when platformContext is empty.

3. **`feat(slack): thread Send replies off the triggering message ts`**
   \`Platform.Send\` previously posted un-threaded messages while \`Reply\` already threaded correctly via \`slack.MsgOptionTS\`. The asymmetry was invisible on most platforms but exposed on Slack because the core message path (\`core/engine.go\`, \`core/streaming.go\`, \`core/progress_compact.go\`, \`core/relay.go\`) overwhelmingly calls \`Send\` rather than \`Reply\` for progress and tool-use events. The result: the final agent reply would thread, but every intermediate progress update, tool-call announcement, and streamed partial would land un-threaded in the main channel. Lift the threading logic from \`Reply\` into \`Send\`: when \`rc.timestamp\` is present, append \`slack.MsgOptionTS\` before \`PostMessageContext\`.

4. **`feat(claudecode): disable claude.ai cloud-managed MCP connectors by default`**
   The claude.ai Slack, Gmail, and Google Calendar MCP connectors are cloud-managed and tied to the authenticated claude.ai account's OAuth token. When the agent calls one of their tools, the API call happens AS the account owner — so Slack messages post with \"Sent using @Claude\" attributed to the human account, not the cc-connect bot, and cc-connect's own delivery path (thread_ts, progress streaming, tool suppression) is bypassed entirely. Disallow them unconditionally in the claudecode agent's spawn args via \`--disallowedTools\`. Per-project config still applies and can extend this baseline but cannot override it.

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./core/... ./platform/slack/... ./agent/claudecode/...\` (all pass after the test fix)
- [x] Manual: Slack session sending a new message — reply lands in-thread, MCP tool calls are unavailable, \`cc-connect\` log shows \`--disallowedTools mcp__claude_ai_Slack,mcp__claude_ai_Gmail,mcp__claude_ai_Google_Calendar\` in the spawn args

## Notes

Each commit is independently revertible. The order matters: the test fix must land no later than the PlatformContext commit to keep CI green on any intermediate state.


chenhg5 (approved):
LGTM. Four distinct correctness fixes — the MCP blocklist (Slack/Gmail/Calendar) is the most security-sensitive, and the threading fix via MsgOptionTS addresses a real delivery bug. buildSenderPrompt signature change is backwards-compatible (new param is added, all call sites updated). All CI pass.

**PR**: https://github.com/chenhg5/cc-connect/pull/527

chenhg5:

leighstillard:
Merge conflicts cleared @chenhg5

PR #528, opened 2026-04-09
feat(slack): compact in-place progress writer via chat.update

leighstillard:
## Summary

Opts the Slack platform into cc-connect's existing compact progress writer (\`core/progress_compact.go\`), so intermediate events (thinking, tool_use, tool_result) are coalesced into a single message that is updated in place via Slack's \`chat.update\` API, rather than spamming a fresh Slack message per event.

## Why

The default for a Slack platform that doesn't implement \`ProgressStyleProvider\` is \`"legacy"\`, which means the core writer is a no-op and every intermediate event is delivered as its own message. A normal turn with 3 tool calls becomes 6+ Slack messages in the thread, most of which the user scrolls past to get to the actual answer.

With compact style:

- **One progress bubble per turn**, edited in place as thinking/tool events stream.
- **Slack's native long-text clipping kicks in** on both mobile (~1500 chars) and desktop (~4000 chars), rendering an inline \"Show more\" / \"Show less\" link automatically for any turn with enough content to be worth collapsing. The \"collapsible card\" UX that normally requires interactive block_actions + a handler comes for free from Slack's built-in text clipping.
- **The final agent reply is still a separate threaded message** underneath the progress bubble, so the thread ends up with exactly two messages per turn: \`[progress]\` + \`[answer]\`.

## Implementation

Three optional interfaces from \`core/interfaces.go\`:

- **\`ProgressStyleProvider.ProgressStyle()\`** returns \`\"compact\"\` so the core writer enables.
- **\`PreviewStarter.SendPreviewStart(ctx, rctx, content) (any, error)\`** posts the initial progress message via \`chat.postMessage\`, threaded off the user's triggering message when available, and returns a \`slackPreviewHandle\` carrying the new message's channel+ts for subsequent updates. The default fallback path stores the original \`replyContext\` as the handle, which would not carry the progress message's own ts — so PreviewStarter is required for Slack, not optional.
- **\`MessageUpdater.UpdateMessage(ctx, handle, content) error\`** edits the progress message in place via \`chat.update\`, using the channel + ts stored on the handle. Errors bubble up and the core writer falls back to legacy per-event sends for the rest of the turn — the user keeps visibility, just noisier.

Modelled on the Telegram implementation at \`platform/telegram/telegram.go\`.

## Notes

- \`chat.update\` has an approximate rate limit of ~1 call per second per message. Fast-streaming turns could hit this; on rate-limit the core writer marks itself failed and falls back to legacy for the remainder of the turn. Acceptable for first cut; revisit with a debounce/coalesce if it shows up in normal use.
- This is orthogonal to the auto-thread-on-Send fix shipping separately as #527 — that one makes *final replies* thread; this one changes how *intermediate progress* renders. Both complement each other but neither depends on the other.

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./platform/slack/... ./core/...\` (no regressions)
- [x] Manual on a live Slack workspace: turn with 3+ tool calls should produce one progress message that updates in place + one final reply, both threaded under the user's original message

### DIFF
--- a/platform/slack/slack.go
+++ b/platform/slack/slack.go
@@ -442,6 +442,118 @@ func (p *Platform) SendFile(ctx context.Context, rctx any, file core.FileAttachm
 
 var _ core.FileSender = (*Platform)(nil)
 
+// ──────────────────────────────────────────────────────────────────────────────
+// Progress writer (compact style)
+//
+// cc-connect's core compact progress writer coalesces intermediate events
+// (thinking, tool_use, tool_result) into a SINGLE message that is updated in
+// place via chat.update, rather than posting a fresh Slack message per event
+// (the "legacy" behaviour). For Slack we want the compact style because:
+//
+//   - One progress message per turn instead of N. Main channel stays quiet.
+//   - Slack's native long-text clipping kicks in on both mobile (~1500 chars)
+//     and desktop (~4000 chars), rendering an inline "Show more" / "Show less"
+//     link without any interactive-component wiring on our side. This gives
+//     the "collapsible card" user experience for free.
+//   - The final agent reply is posted as a separate threaded message
+//     underneath the progress bubble, so the thread ends up with exactly
+//     two messages: [progress] + [answer].
+//
+// To opt in we implement three optional interfaces from core/interfaces.go:
+//
+//   - ProgressStyleProvider — returns "compact" so the core writer enables.
+//   - PreviewStarter        — posts the initial progress message and returns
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. Compact in-place progress writer is a solid UX improvement for Slack — one progress bubble per turn instead of N messages flooding the channel. Interface design is clean and the fallback behavior on API errors is correct.

**PR**: https://github.com/chenhg5/cc-connect/pull/528

PR #529, opened 2026-04-09
feat(engine): /resume slash command with workspace session picker

leighstillard:
## Summary

Adds an explicit, two-step \`/resume\` flow that lets a user pick a prior workspace session — including terminal-run sessions on the same Unix user's filesystem — and branch off its context into the current cc-connect conversation. Designed as a deliberate, opt-in alternative to the removed implicit continue-bridge (\`4015259\`, killed by \`6d8bf93\`).

## UX

\`/resume\` (no args) — lists up to 10 recent workspace sessions known to the agent backend, numbered, with a relative timestamp, the opening user message, and (indented) the most recent user message. The current session is shown with a \`*(current)*\` marker so it's not mysteriously missing from the chronological ordering.

\`\`\`
*Recent sessions in this workspace* (showing 5 of the most recent):
 1. 2m ago *(current)*  — \"ok welcome back\"
     ↳ \"por que nos dos?\"
 2. 15m ago  — \"let me check the journal\"
 3. 1h ago  — \"fix restart.sh\"
 4. 3h ago  — \"rebuild cc-connect properly\"
 5. 5h ago  — \"ok I want cc-connect to be aware of sessions\"

Type \`/resume <n>\` to fork the chosen session into this conversation. The source session will not be modified.
\`\`\`

\`/resume <n>\` — picks the Nth candidate from the last list. Arms \`Session.forkOnNextStart\` with the backend session ID. On the next user message the engine consumes that field and spawns claude with \`--resume <id> --fork-session\` via a new \`ResumeForkPrefix\` sentinel, so the source session (which may still be live in a terminal or another conversation) is never mutated — the Slack reply branches off into its own independent timeline.

## Why a picker, not an implicit continue

The previous implicit continue-bridge fired on first-contact-per-engine-lifetime regardless of user intent, and could happily resurrect weeks-old context. \`/resume\` is opt-in: the user sees synopses before picking and the default behaviour is fresh-session-by-default. The picker presents **both the first user message and the most recent user message** per session so a session that has drifted topic is still recognisable by its origin — first-only would lose long-running sessions; last-only would lose just-started ones.

## Key structural decisions

- **Fork sentinel lives in a separate transient field** (\`Session.forkOnNextStart\`, unexported, never serialised) rather than in \`AgentSessionID\`. \`AgentSessionID\` is persisted to disk by the session-store snapshot writer, and stashing a sentinel there would leave stale state on disk if cc-connect crashes in the narrow window between \`/resume\` pick and the first forwarded message. The transient field drops on crash — worst case the user re-picks from a fresh \`/resume\` list.
- **\`ResumeForkPrefix\` (\`__fork:<uuid>\`)** is the cross-package carrier from engine to agent. The claudecode agent strips the prefix and passes \`--resume <uuid> --fork-session\` to claude. Other agents that don't support forking ignore the prefix and fall back to plain \`--resume\`; acceptable because only claudecode supports forked resume today.
- **\`AgentSessionInfo\` gains a \`FirstSummary\` field** alongside the existing \`Summary\` (which is \"latest user message\" — what \`/list\` and \`/switch\` want). \`scanSessionMeta\` now tracks both, and the summary truncation helper was extracted into \`truncateSummaryForList\` so first and last summaries are normalised identically.
- **The engine consumes \`forkOnNextStart\` immediately before** its existing \`startSessionID\` lookup, so the override path and the normal resume path share fallback semantics: if the fork spawn fails, the engine's existing \"try fresh after resume failure\" branch kicks in.

## Scope caveats

- **Sandbox users** (e.g. when cc-connect spawns as a non-supervisor user via \`run_as_user\`) only see their own \`~/.claude/projects\` sessions — they cannot read the supervisor user's terminal session JSONL if the supervisor's home is mode 700. That is by design of any POSIX-isolation setup; full visibility is available in sessions that run as the same user owning the terminal.
- **i18n**: hardcoded English strings for now. Happy to promote to \`MsgKey\` constants in a follow-up once the feature shape stabilises.
- **Unit tests for \`cmdResume\`**: not included in this PR. The existing \`stubListAgent\` fixture can be extended in a follow-up once the UX is settled in real use.

## Help card

\`/resume\` added to the session group so it shows in \`/help\` output.

## Test plan

- [x] \`go build ./...\`
- [x] \`go test ./core/... ./agent/claudecode/...\` (no regressions)
- [x] Manual: \`/resume\` lists prior workspace sessions; \`/resume <n>\` forks; subsequent message inherits context from the picked session

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -356,11 +356,12 @@ func (a *Agent) ListSessions(ctx context.Context) ([]core.AgentSessionInfo, erro
 			continue
 		}
 
-		summary, msgCount := scanSessionMeta(filepath.Join(projectDir, name))
+		firstSummary, summary, msgCount := scanSessionMeta(filepath.Join(projectDir, name))
 
 		sessions = append(sessions, core.AgentSessionInfo{
 			ID:           sessionID,
 			Summary:      summary,
+			FirstSummary: firstSummary,
 			MessageCount: msgCount,
 			ModifiedAt:   info.ModTime(),
 		})
@@ -393,18 +394,24 @@ func (a *Agent) DeleteSession(_ context.Context, sessionID string) error {
 	return os.Remove(path)
 }
 
-func scanSessionMeta(path string) (string, int) {
+// scanSessionMeta parses a claude JSONL session file and returns
+// (firstUserSummary, lastUserSummary, messageCount). Both summaries are
+// XML-tag stripped and truncated to ~40 runes so they fit cleanly in a
+// Slack list row. firstUserSummary is the opening user message in the
+// file (what the user originally started the session with); lastUserSummary
+// is the most recent user message (what the session is currently about).
+// Used by /list, /switch (which want the "current topic" = Summary) and
+// by /resume (which also wants the original starting point = FirstSummary
+// so long-running drifted sessions remain identifiable).
+func scanSessionMeta(path string) (firstSummary, summary string, count int) {
 	f, err := os.Open(path)
 	if err != nil {
-		return "", 0
+		return "", "", 0
 	}
 	defer f.Close()
 
 	scanner := bufio.NewScanner(f)
 	scanner.Buffer(make([]byte, 256*1024), 256*1024)
-
-	var summary string
-	var count int
 
 	for scanner.Scan() {
 		var entry struct {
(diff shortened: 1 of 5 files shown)


chenhg5 (approved):
LGTM. Well-designed /resume command — the workspace session picker gives users a safe way to branch off a long-running terminal session without mutating it. Code is clean, tests updated, CI all green.

**PR**: https://github.com/chenhg5/cc-connect/pull/529

PR #531, opened 2026-04-09
fix(feishu): improve delete and /model card refresh flows

q107580018:
## Summary

- fix Feishu `/delete` batch deletion callbacks by returning a loading card immediately and refreshing the same card with final results after the async work completes
- improve Feishu `/model` card switching so model selection shows a final result card in-place and the result-card back action returns to `/model` instead of failing with `code:108002`
- add coverage for both async card refresh behavior and the follow-up `/model` card regressions that surfaced during real Feishu verification

## Changes

| File | What |
|------|------|
| `core/interfaces.go` | add `CardRefresher` so platforms can patch an existing interactive card after async work finishes |
| `core/engine.go` | add async delete-mode result refresh, model-card state handling, and final `/model` result-card flow with a dedicated back target |
| `core/i18n.go` | add localized copy for delete loading state and model switching / success / failure card states |
| `core/engine_test.go` | cover delete refresh behavior, model card switching, workspace-scoped model switching, failure rendering, and returning from the model result card |
| `platform/feishu/card.go` | implement `RefreshCard(...)` via Feishu message patch API |
| `platform/feishu/feishu.go` | track card callback message IDs for later refresh and route `/model` card actions through normal card navigation |
| `platform/feishu/platform_test.go` | verify Feishu `/model` card callbacks return card updates and preserve message tracking |
| `.gitignore` | ignore `.planning/` artifacts going forward |
| `.planning/*` | add local GSD scaffolding and quick-task artifacts used to execute and record this workflow |

## User-visible behavior

- `/delete` no longer risks timing out the Feishu callback while deleting multiple sessions
- `/delete` now refreshes the original card from a deleting state to a final result state
- `/model` no longer falls back to a separate plain-text confirmation path
- `/model` now completes in-card and the result card's **返回** button reopens the `/model` card correctly

## Testing

- `go test ./core -run 'TestHandleCardNav_Model(SwitchesAndRefreshesCard|UsesWorkspaceContext|SwitchFailureRefreshesCard|ResultBackReturnsModelCard)' -count=1`
- `go test ./platform/feishu -run 'TestInteractivePlatform_ModelCardActionReturnsCardUpdate|TestInteractivePlatform_CardActionUsesCallbackSessionKey' -count=1`
- `go test ./core -run 'TestHandleCardNav_Model|TestExecuteCardAction_Model|TestModelCard' -count=1`
- `go test ./platform/feishu -run 'TestInteractivePlatform_.*Model.*|TestInteractivePlatform_.*delete.*' -count=1`
- `go test ./core ./platform/feishu`


chenhg5 (approved):
LGTM. The async deletion pattern is sound — background goroutine updates card state and refreshes in-place, correctly handling Feishu's 3-second callback timeout. CardRefresher interface is clean.

**PR**: https://github.com/chenhg5/cc-connect/pull/531

q107580018:
Closed in favor of a fresh PR with squashed history.

PR #544, opened 2026-04-10
fix(feishu): cron jobs send new chat messages instead of thread replies

leoliang1997:
## Summary                                                                       
                                                                                  
  Cron jobs reply inside a thread instead of posting new messages in group chat    
  when `thread_isolation` is enabled.                                             
                                                                                   
  ## Problem                                                                       
                                     
  When `thread_isolation = true`, every standalone group message gets a session key
   in the format `feishu:{chatID}:root:{messageID}` — even messages without an
  actual thread (`root_id` is empty, so `makeSessionKey` falls back to the         
  message's own ID).                         
                                     
  When a cron job fires with this session key, `ReconstructReplyCtx` extracts the  
  message ID from the `root:` prefix and sets it as `replyContext.messageID`. Then
  `shouldReplyInThread` returns true, and the bot calls                            
  `Im.Message.Reply(messageID, ReplyInThread=true)` — creating a thread on the
  original `/cron add` message instead of posting a new message in the group.
                                                                                  
  User sends "/cron add ..." in group                                              
    → session key: feishu:chat:root:{this_message_id}
    → cron job bound to this session key                                           
                                                                                   
  Cron fires:                                                                      
    → ReconstructReplyCtx extracts messageID from "root:{id}"                      
    → replyContext{chatID: "...", messageID: "{id}"}                               
    → shouldReplyInThread → true (messageID is set + session key has "root:")      
    → Reply(messageID, ReplyInThread=true)                                         
    → Creates thread on the "/cron add" message ✗                                  
                                                                                   
  ## Fix                                                                           
                                                                                   
  Implement the `CronReplyTargetResolver` interface for the Feishu platform. This  
  interface already exists in core (`core/interfaces.go:34`) and is used by Discord
   for similar purposes.                                                           
                                             
  The implementation returns a `replyContext` with only `chatID` (no `messageID`), 
  so `Send()` takes the `sendNewMessageToChat` path instead of `ReplyInThread`:   
                                                                                   
  ```go                                      
  func (p *Platform) ResolveCronReplyTarget(sessionKey string, title string)
  (string, any, error) {                                                           
      parts := strings.SplitN(sessionKey, ":", 3)
      if len(parts) < 2 || parts[0] != p.platformName {                            
          return "", nil, fmt.Errorf("%s: invalid session key %q", p.tag(),        
  sessionKey)                                                                      
      }                                                                            
      rc := replyContext{chatID: parts[1], sessionKey: sessionKey}                 
      return sessionKey, rc, nil                                                   
  }                                                                                
                                                                                   
  No changes to core/ — the fix is entirely within platform/feishu/.               
                                             
  Test plan                                                                        
                                                                                  
  - go test ./platform/feishu/ — all tests pass (2 new tests added)               
  - Manual test: create cron job in group chat, verify it posts new messages       
  instead of replying in a thread                                                  
                                                                                   
  🤖 Generated with https://claude.com/claude-code                                 
  ```

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -2356,6 +2356,20 @@ func (p *Platform) ReconstructReplyCtx(sessionKey string) (any, error) {
 	return rc, nil
 }
 
+// ResolveCronReplyTarget implements CronReplyTargetResolver. For cron jobs,
+// return a replyContext with only chatID (no messageID) so that Send() uses
+// sendNewMessageToChat instead of ReplyInThread. Without this, cron jobs
+// bound to a thread-isolated session key (feishu:chat:root:msgID) would
+// reply inside a thread rather than posting a new message in the group.
+func (p *Platform) ResolveCronReplyTarget(sessionKey string, title string) (string, any, error) {
+	parts := strings.SplitN(sessionKey, ":", 3)
+	if len(parts) < 2 || parts[0] != p.platformName {
+		return "", nil, fmt.Errorf("%s: invalid session key %q", p.tag(), sessionKey)
+	}
+	rc := replyContext{chatID: parts[1], sessionKey: sessionKey}
+	return sessionKey, rc, nil
+}
+
 func parseThreadRootID(sessionTail string) (string, bool) {
 	for _, prefix := range []string{"root:", "thread:"} {
 		if strings.HasPrefix(sessionTail, prefix) {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. ResolveCronReplyTarget correctly returns replyContext without messageID so cron jobs use sendNewMessageToChat instead of ReplyInThread. Fixes the thread-isolation cron regression. Tests added.

**PR**: https://github.com/chenhg5/cc-connect/pull/544

PR #550, opened 2026-04-10
fix(core): use full accumulated text for stream preview finalization when tool messages hidden

### LINKED ISSUE #549: [Bug] Stream preview card shows only last text segment when tool messages are hidden
## Problem

When `quiet = true` or `[display] tool_messages = false`, the stream preview card in Feishu/Lark shows only the **last** text segment instead of the full accumulated response. Each time a new text segment arrives after a tool call, the previous content disappears.

## Root Cause

In `processInteractiveEvents()` (core/engine.go), the `EventResult` handler constructs `fullResponse`:

```go
fullResponse := event.Content
if fullResponse == "" && len(textParts) > 0 {
    fullResponse = strings.Join(textParts, "")
}
```

`event.Content` only contains the **last assistant turn's text**, not text from segments before tool calls. When tool messages are visible, `segmentStart` advances and text is sent in segments, so this works fine. But when tool messages are hidden, `segmentStart` stays 0 and the stream preview accumulates all text via `appendText()` — building up to e.g. 822 chars. Then `finish(fullResponse)` overwrites the preview card with only the 198-char `event.Content`, replacing the accumulated content.

Debug log evidence:
```
stream preview: updating via UpdateMessage text_len=822   ← accumulated correctly
stream preview finish: sending final UpdateMessage text_len=198  ← overwritten with event.Content
```

## Introduced By

PR #462 (`feat: split thinking/tool visibility and preserve QQ Bot references`) replaced the `!quiet` guard with `e.display.ToolMessages` checks. The old `quiet` flag suppressed tool events entirely including `EventResult` content handling, masking this issue. The new display-based approach correctly guards `EventToolUse`/`EventToolResult` blocks, but the `EventResult` handler's `fullResponse` construction was not updated to account for the case where `segmentStart` stays 0.

## Expected Behavior

The final card should contain the full response text accumulated across all tool calls.

## Environment

- cc-connect 1.2.2-beta.5
- Platform: Feishu/Lark (WebSocket mode)
- Agent: Claude Code (opus)
- Config: `quiet = true` or `[display] tool_messages = false`

Cigarrr:
## Summary

- When tool messages are hidden (`quiet = true` or `[display] tool_messages = false`), `segmentStart` stays 0 and the stream preview card accumulates all text across tool boundaries. However, `EventResult` used `event.Content` (which only contains the last assistant segment) to finalize the preview via `finish()`, overwriting the accumulated content with a truncated version.
- Prefer `strings.Join(textParts, "")` over `event.Content` when `segmentStart == 0`, ensuring the final card contains the complete response.
- Insert a visual line break (`\n\n`) between text segments at tool call boundaries when tool messages are hidden, so consecutive segments don't run together without separation.

Closes #549

## Test plan

- [x] `go test ./...` passes
- [x] Manual verification on Feishu/Lark with `quiet = true`: multi-tool-call responses accumulate in a single card without content loss
- [x] Non-quiet mode behavior unchanged (segmentStart advances normally, original code path taken)

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2414,6 +2414,14 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 
 		case EventToolUse:
 			toolCount++
+			// When tool messages are hidden, insert a visual break between text
+			// segments so the accumulated response doesn't run together.
+			if !e.display.ToolMessages && len(textParts) > segmentStart {
+				textParts = append(textParts, "\n\n")
+				if sp.canPreview() {
+					sp.appendText("\n\n")
+				}
+			}
 			if e.display.ToolMessages {
 				// Flush accumulated text segment before tool display
 				previewActive := sp.canPreview()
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. Fixes preview text finalization when tool messages are hidden — accumulated textParts now used correctly when segmentStart==0. Also adds visual breaks between text segments for readability.

**PR**: https://github.com/chenhg5/cc-connect/pull/550

PR #552, opened 2026-04-11
feat(discord): support inbound file attachments

xfromcnZL:
## Summary
- Discord platform previously only handled audio and image attachments from incoming messages
- Non-image, non-audio attachments (PDFs, documents, code files, etc.) were silently dropped
- Added an `else` branch in the attachment loop to capture them as `core.FileAttachment`
- Also updated the empty-message guard to account for file-only messages
- This matches the behavior already present in the Telegram platform (`telegram.go:453`)

## Changes
- `platform/discord/discord.go`: Added `files []core.FileAttachment` collection in the `MessageCreate` handler, with download + error logging for non-image/non-audio attachments

## Test plan
- [x] `go test ./platform/discord/` passes
- [x] `go build ./...` passes
- [x] Manual test: sent PDF via Discord DM, confirmed `has_files=true` in logs

### DIFF
--- a/platform/discord/discord.go
+++ b/platform/discord/discord.go
@@ -551,6 +551,7 @@ func (p *Platform) Start(handler core.MessageHandler) error {
 
 		var images []core.ImageAttachment
 		var audio *core.AudioAttachment
+		var files []core.FileAttachment
 		for _, att := range m.Attachments {
 			ct := strings.ToLower(att.ContentType)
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. Discord file attachment support — downloads non-image/audio attachments and passes them as FileAttachment to the message handler. Clean feature addition.

**PR**: https://github.com/chenhg5/cc-connect/pull/552

PR #556, opened 2026-04-11
fix(core): resolve workspace for cron prompt-mode jobs in multi-workspace mode

### LINKED ISSUE #555: [Bug] Cron prompt-mode jobs bypass workspace resolution in multi-workspace mode
### cc-connect Version

v1.2.2-beta.5

### Operating System

macOS (Darwin 25.4.0)

### Agent Type

Claude Code

### Platform

Lark (Feishu international)

### Installation Method

npm

### Bug Description

In multi-workspace mode, cron jobs using prompt mode (i.e. `--prompt`, not `--exec`) always use the global `e.agent` instead of resolving a workspace-specific agent. This causes the Claude Code agent to start with the wrong working directory — typically `~/.cc-connect` or the daemon's cwd — instead of the workspace bound to the cron job's target channel.

The `work_dir` field on the cron job struct is also ignored for prompt-mode jobs (it is only applied in `executeCronShell()`).

**Root cause**: `ExecuteCronJob()` in `core/engine.go` skips the workspace resolution path that `handleMessage()` performs. Specifically:
1. No call to `resolveWorkspace()` to look up the workspace binding for the channel
2. No call to `workspaceContext()` to get a workspace-specific agent/sessions
3. `processInteractiveMessageWith()` is called with the global `e.agent`, `e.sessions`, and an empty `workspaceDir` parameter

### Steps to Reproduce

1. Configure cc-connect in multi-workspace mode with `base_dir` set
2. Bind a channel to a workspace (e.g. via `/bind`)
3. Create a prompt-mode cron job targeting that channel:
   ```
   cc-connect cron add --cron "*/5 * * * *" --prompt "echo hello" --desc "test"
   ```
4. Wait for the cron job to fire
5. Check the Claude Code agent's working directory (e.g. via `gitStatus` in the agent's system prompt)

### Expected Behavior

The cron job should resolve the workspace binding for the target channel and use a workspace-specific agent with `cmd.Dir` set to the bound workspace path — consistent with how interactive messages are handled.

### Actual Behavior

The cron job uses the global agent. The Claude Code process starts with an incorrect working directory (often `~/.cc-connect`), causing:
- `gitStatus` to show the wrong repo or no repo
- Skills and file operations to target the wrong directory
- Session files to accumulate in an unintended location

```toml
[[projects]]
  name = "myproject"
  mode = "multi-workspace"
  base_dir = "/path/to/workspaces"

  [projects.agent]
    type = "claudecode"

    [projects.agent.options]
      mode = "bypassPermissions"
```

### Logs / Error Output

No error in logs — the job executes successfully but in the wrong directory.

### Additional Context

- PR #474 previously fixed workspace-prefixed session key parsing in `ExecuteCronJob()`, but did not add workspace agent resolution
- PR #505 introduced `resolveChannelWorkDir()` and `workspaceContext()` which provide the infrastructure needed for the fix
- The `executeCronShell()` path correctly reads `job.WorkDir`, but prompt mode has no equivalent logic
- All existing infrastructure (`resolveWorkspace`, `workspaceContext`, `getOrCreateWorkspaceAgent`) is already in place — the fix only needs to wire them into the prompt path of `ExecuteCronJob()`

Cigarrr:
## Summary

In multi-workspace mode, `ExecuteCronJob()` for prompt (non-shell) jobs always used the global `e.agent` and passed an empty `workspaceDir` to `processInteractiveMessageWith()`. This caused the Claude Code agent to start with the wrong working directory (typically `~/.cc-connect`) instead of the workspace bound to the cron job's channel.

The fix mirrors `handleMessage()`'s workspace resolution logic:

1. Extract `channelID` from the cron job's `sessionKey`
2. Resolve the workspace binding via `resolveWorkspace()`
3. Get a workspace-specific agent/sessions via `workspaceContext()`
4. Pass the resolved agent and `workspaceDir` to `processInteractiveMessageWith()`

`job.WorkDir`, when explicitly set, takes highest priority as an override.

**Priority chain**: `job.WorkDir` (explicit) → workspace binding resolution → global agent fallback

Closes #555

## Related

- #474 — fixed workspace-prefixed session key parsing in `ExecuteCronJob()`, but did not add workspace agent resolution
- #505 — introduced `resolveChannelWorkDir()` and `workspaceContext()` which provide the infrastructure this fix wires into

## Test plan

- [x] `go test ./core/ -count=1` — all tests pass
- [x] `go build ./...` — clean build
- [x] Manually verified: cron prompt jobs now execute in the correct workspace directory (confirmed via `gitStatus` in agent context)
- [x] Verified `job.WorkDir` override takes precedence when explicitly set
- [x] Verified non-multi-workspace mode is unaffected (falls through to global agent)

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -926,6 +926,36 @@ func (e *Engine) ExecuteCronJob(job *CronJob) error {
 		ModeOverride: job.Mode,
 	}
 
+	// Resolve workspace-specific agent and sessions for multi-workspace mode.
+	// Priority: job.WorkDir (explicit) > workspace binding > global agent fallback.
+	agent := e.agent
+	sessions := e.sessions
+	workspaceDir := ""
+
+	if e.multiWorkspace {
+		channelID := extractChannelID(sessionKey)
+		if channelID != "" {
+			workspace, _, err := e.resolveWorkspace(targetPlatform, channelID)
+			if err == nil && workspace != "" {
+				wsAgent, wsSessions, _, effectiveDir, err := e.workspaceContext(workspace, sessionKey)
+				if err == nil {
+					agent = wsAgent
+					sessions = wsSessions
+					workspaceDir = effectiveDir
+				}
+			}
+		}
+	}
+
+	if job.WorkDir != "" {
+		wsAgent, wsSessions, err := e.getOrCreateWorkspaceAgent(job.WorkDir)
+		if err == nil {
+			agent = wsAgent
+			sessions = wsSessions
+			workspaceDir = job.WorkDir
+		}
+	}
+
 	useNewSession := false
 	if e.cronScheduler != nil {
 		useNewSession = e.cronScheduler.UsesNewSession(job)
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. Cron job now correctly resolves workspace context in multi-workspace mode — respects job.WorkDir priority over channel binding. Proper session routing.

**PR**: https://github.com/chenhg5/cc-connect/pull/556

PR #571, opened 2026-04-11
fix(telegram): show Telegram-safe skill commands in /skills

Zx55:
## Summary

This updates the Telegram `/skills` output to show Telegram-safe slash command names for discovered skills.

## Why

Telegram commands cannot contain hyphens, but skill names often do. `cc-connect` already sanitizes Telegram command registration and already resolves hyphenated and underscored names equivalently, but the `/skills` text still rendered the raw skill name. That made the suggested command in Telegram inconsistent with the actual slash command users could invoke.

## Changes

- render skill commands in `/skills` using a Telegram-safe display name on Telegram
- preserve the existing display on non-Telegram platforms
- add a regression test for a hyphenated skill name such as `telegram-codex-bot`

## Validation

- `go test ./core ./platform/telegram`
- manually verified in a real Telegram bot session that `/skills` shows `/telegram_codex_bot`


chenhg5 (approved):
LGTM. Telegram-safe skill command names — sanitizeTelegramDisplayCommand strips non-alphanumeric chars and limits to lowercase alphanum. Good companion to #572's recursive discovery.

**PR**: https://github.com/chenhg5/cc-connect/pull/571

PR #576, opened 2026-04-12
fix(telegram): paginate /skills output

Zx55:
## Summary

This paginates Telegram `/skills` output so long skill lists no longer exceed Telegram's message size limit.

## Why

With enough discovered skills, `/skills` can generate a reply that is longer than Telegram allows in a single text message. In that case the command is received and handled, but Telegram rejects the reply with `Bad Request: message is too long`, so users see no response at all.

## Changes

- add Telegram-specific `/skills <page>` support
- paginate Telegram `/skills` output by message-length budget instead of dumping every skill in one reply
- include page hints such as `/skills 2` when more pages are available
- return a clear invalid-page message when the requested page is out of range
- add regression tests for pagination and invalid page handling

## Validation

- `go test ./core ./platform/telegram`
- manually reproduced the failure with a long Telegram skill list and verified the patched build responds instead of hitting Telegram's message-length error

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -8753,6 +8753,8 @@ func (e *Engine) cmdSkills(p Platform, msg *Message, args []string) {
 		}
 
 		if strings.EqualFold(p.Name(), "telegram") {
+			// Telegram rejects oversized text replies, so render /skills as
+			// explicit pages instead of sending one giant skill list.
 			pages := e.renderTelegramSkillsPages(skills)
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
LGTM. 分页逻辑健壮，处理了超长 skill 名称的截断，支持 /skills 2 翻页，i18n 覆盖全部 5 种语言，单元测试覆盖充分。

PR #578, opened 2026-04-12
feat(codex): add runtime reply footer aligned with Codex CLI

xxb:
## Summary
- add an optional Codex-style reply footer that can show model, reasoning effort, remaining context, and workdir
- enable `reply_footer` by default and source footer metadata from live Codex runtime state instead of config snapshots
- compute context `left` from real `last_token_usage` and match Codex CLI semantics, including the baseline-token adjustment and rounding behavior

## Details
- add footer metadata plumbing in `core` and wire it through config / engine rendering
- teach the Codex app-server session and exec session paths to expose runtime model / effort / workdir / context usage
- use English `"% left"` footer wording and keep the footer visually subdued
- add regression coverage for runtime footer rendering and the Codex context-left calculation

## Testing
- `go test ./core ./agent/codex`
- `go test ./...`

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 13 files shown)


chenhg5 (approved):
LGTM. Codex reply footer 功能设计合理，启用默认，配置可关闭。runtime metadata 通过 WebSocket 实时同步，nil 安全处理到位，单测覆盖 buildReplyFooter、applyThreadRuntimeState、rateLimits 缓存等关键路径。

chenhg5 (approved):
LGTM. 新增运行时 reply footer，显示剩余上下文百分比。新增 ContextUsageReporter 接口，支持 runtime context 使用量上报。Codex app-server session 完善 thread resume 时的 runtime state 同步。CI pass。

PR #585, opened 2026-04-12
fix(claudecode): unblock readLoop when child subprocesses hold stdout pipe

Sentixxx:
## Problem

A user reported two contradictory bot replies sent within the same minute:

| Time | User action | Bot reply |
|------|-------------|-----------|
| 22:53 | sends `1` | `⏳ Previous request is still processing...` |
| 22:53 | sends `/stop` | `No task currently executing.` |

The two replies are mutually inconsistent — the first says a turn is running, the second says there is nothing to stop. The user has no way to recover short of `/new`. In an earlier observed incident the same bug manifested as a session that silently stopped responding for 13 minutes.

## Summary

- The contradiction occurs inside `cleanupInteractiveState`'s window: the interactive-state map entry is removed before `closeAgentSessionWithTimeout` returns, while the session lock is still held by the turn goroutine. Two handlers observing this transient state — `handleMessage` and `cmdStop` — read the same condition and produce different replies.
- The window can be up to **130 seconds** wide because `claudeSession.Close()` hangs whenever a descendant of the agent inherits its stdout fd (MCP helper, Stop-hook subprocess, any fork that didn't close fd 1). The pipe never reaches EOF, `scanner.Scan()` blocks forever, the defer never runs, and `cs.done` is never closed.
- This PR restructures `readLoop` so `Close()` returns in microseconds in all paths, shrinking the inconsistency window from 130 s to negligible. The reply-policy gap that produces the contradictory text itself is intentionally left for a separate follow-up; with this change the window becomes too short for users to race into.

## Root cause

`Close()` waits on `<-cs.done` at every phase. `cs.done` is closed only by `readLoop`'s defer, which runs only when `scanner.Scan()` returns false. With a descendant holding the pipe write end, scanner blocks indefinitely:

- Phase 1 closes stdin and waits 120 s; the descendant still holds the write end so EOF never arrives.
- Phase 2's SIGTERM goes to the agent (often already dead); the descendant is untouched.
- Phase 3 sends SIGKILL and waits for `cs.done` — still never closes.
- The 130 s outer timeout in `closeAgentSessionWithTimeout` eventually returns control.

A secondary defer leak made things worse: when Phase 3 calls `cs.cancel()`, the error-reporting select inside the defer used to `return` on `<-cs.ctx.Done()`, skipping `close(cs.events)` and `close(cs.done)` and stranding the engine event loop on a channel that never closes.

## Fix

Replace the defer-time `cmd.Wait()` with two cooperating goroutines started before the scanner loop:

- **Reaper** — `waitErrCh <- cs.cmd.Wait(); close(waitDone)`. Single `Wait()`, publishes exit status.
- **Stdout closer** — `select { case <-ctx.Done(): case <-waitDone: }; stdout.Close()`. Unblocks scanner in both the process-exit and cancel paths, without disturbing reads while the agent is alive.

`handleReadLoopScanErr` checks `ctx.Done()` / `waitDone` before reporting, suppressing the expected `ErrClosed` from the reaper-triggered close so no spurious `EventError` reaches the engine. The defer now reads `waitErrCh` and unconditionally closes `events` / `done`. `readLoop` is split into focused helpers per `AGENTS.md` style.

## Test plan

- [x] `go build ./...` passes
- [x] `go test -race ./...` passes
- [x] Regression tests committed **before** the fix so reviewers can verify red→green:

  ```
  git checkout HEAD~1   # test commit
  go test -timeout 30s -run TestReadLoop ./agent/claudecode/   # both tests HANG
  git checkout HEAD     # fix commit
  go test -timeout 30s -run TestReadLoop ./agent/claudecode/   # both PASS
  ```

  - `TestReadLoop_ChildHoldsStdoutPipe` — descendant holds pipe after parent exits. Also asserts the agent's last buffered event is delivered.
  - `TestReadLoop_CtxCancelClosesChannels` — ctx cancel with non-empty stderr exercises the previously leaky defer branch.

- [ ] End-to-end reproduction via the Claude Code Bash tool was not possible in the test environment: the Bash tool redirects subprocess stdout into an internal capture file, so user-spawned processes never inherit the agent's stdout pipe. The unit tests use the `sh -c 'sleep 60 & echo … & exit 0'` pattern that triggers the deadlock in any Go program using `StdoutPipe`.

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -193,6 +193,8 @@ func (cs *claudeSession) startReadLoopWait(stdout io.ReadCloser) (<-chan error, 
 	go func() {
 		select {
 		case <-cs.ctx.Done():
+			_ = stdout.Close()
+			return
 		case <-waitDone:
 		}
 
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -228,8 +228,9 @@ func (cs *claudeSession) finishReadLoop(waitErrCh <-chan error, stderrBuf *bytes
 			select {
 			case cs.events <- evt:
 			case <-cs.ctx.Done():
-				// Fall through: events and done must still be closed so
-				// the engine event loop sees the session ending.
+				// INVARIANT: readLoop must close cs.events and cs.done exactly once
+				// on every termination path. Callers (engine event loop) rely on
+				// these closures to observe session end.
 			}
 		}
 	}
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -194,6 +194,16 @@ func (cs *claudeSession) startReadLoopWait(stdout io.ReadCloser) (<-chan error, 
 		case <-cs.ctx.Done():
 		case <-waitDone:
 		}
+		
+		// Grace period: give scanner a brief window to drain any data the
+		// agent wrote to the pipe buffer before exiting. If scanner finishes
+		// on its own (pipe fully closed, no descendants holding it),
+		// cs.done fires first and we skip the force-close entirely
+		select {
+		case<-cs.done:
+				return
+		case<-time.After(500*time.Millisecond):
+		}
 		_ = stdout.Close()
 	}()
 
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -169,72 +169,117 @@ func newClaudeSession(ctx context.Context, workDir, model, effort, sessionID, mo
 }
 
 func (cs *claudeSession) readLoop(stdout io.ReadCloser, stderrBuf *bytes.Buffer) {
-	defer func() {
-		cs.alive.Store(false)
-		// Always close stdout to unblock any future reads
-		_ = stdout.Close()
-		// Wait for process to exit (this is needed to release resources)
-		if err := cs.cmd.Wait(); err != nil {
-			stderrMsg := strings.TrimSpace(stderrBuf.String())
-			if stderrMsg != "" {
-				slog.Error("claudeSession: process failed", "error", err, "stderr", stderrMsg)
-				evt := core.Event{Type: core.EventError, Error: fmt.Errorf("%s", stderrMsg)}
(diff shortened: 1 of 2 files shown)


Sentixxx:
tests on commit 60dafa2db56ce335245934e723fc1ea638728d7c
image

--- 

tests on commit 276689339f14af4526afaf25db3dfbe08019447c
image

chenhg5 (approved):
LGTM. 根因分析详尽，reaper goroutine 方案合理，两个回归测试在 fix 提交前确保 red→green。readLoop 拆分符合规范。

agent/claudecode/session.go
```diff
@@ -166,67 +166,122 @@ func newClaudeSession(ctx context.Context, workDir, model, sessionID, mode strin
...
+		}
+
+		// Grace period: give scanner a brief window to drain any data the
+		// agent wrote to the pipe buffer before exiting. If scanner finishes
+		// on its own (pipe fully closed, no descendants holding it),
+		// cs.done fires first and we skip the force-close entirely
+		select {
+		case <-cs.done:
 			return
+		case <-time.After(50 * time.Millisecond):
```
Sentixxx, line 208: 50ms is an empirical value — I'm not sure whether it needs tuning. Should we add a log here?

Sentixxx:
The root cause actually sits in `core/engine.go`; this PR intentionally only shrinks the race window in `agent/claudecode` to keep the scope focused. A follow-up PR will address the `core/engine.go` side.

Sentixxx:
> The root cause actually sits in `core/engine.go`; this PR intentionally only shrinks the race window in `agent/claudecode` to keep the scope focused. A follow-up PR will address the `core/engine.go` side.

https://github.com/chenhg5/cc-connect/pull/598

chenhg5:
@Sentixxx Please fix the CI tests.

Sentixxx:
> @Sentixxx Please fix the CI tests.

fixed

PR #593, opened 2026-04-13
feat(api): support combining --as-prompt with --new-thread in send command

### LINKED ISSUE #590: Support combining --as-prompt with --new-thread in send command
## Problem

Currently `--as-prompt` and `--new-thread` are mutually exclusive in the send command. In `core/api.go` lines 182-194:

```go
if req.AsPrompt {
    sendErr := engine.InjectPrompt(req.SessionKey, req.Message)
    // ... returns early here
}

// NewThread check is never reached when AsPrompt is true
if req.NewThread {
    sendErr = engine.PostToNewThread(...)
}
```

When `--as-prompt` is set, it returns early and never reaches the `--new-thread` logic.

## Use Case

The completion-watcher needs to:
1. Create a new thread for each completion notification (`--new-thread`)
2. Have the agent process it as an inbound prompt (`--as-prompt`)

Without both, either:
- Notifications arrive in main channel (no `--new-thread`) — tool calls scatter across channel
- Notifications create threads but agent doesn't auto-process (no `--as-prompt`)

## Proposed Fix

Add `InjectPromptToNewThread` method that combines both behaviors:

```go
if req.AsPrompt {
    if req.NewThread {
        sendErr = engine.InjectPromptToNewThread(req.SessionKey, req.Message)
    } else {
        sendErr = engine.InjectPrompt(req.SessionKey, req.Message)
    }
    // ...
}
```

The new method would:
1. Post to Slack as a new top-level message (like `PostToNewThread`)
2. Inject the message into the agent's session as a prompt (like `InjectPrompt`)
3. Route subsequent agent responses to that thread

chenhg5:
## Summary
- Added three new engine methods: `InjectPrompt`, `PostToNewThread`, `InjectPromptToNewThread`
- Updated `SendRequest` with `AsPrompt` and `NewThread` boolean fields
- CLI now supports `--as-prompt` and `--new-thread` flags that can be combined

## Problem
Issue #590: `--as-prompt` and `--new-thread` were mutually exclusive because the code returned early after handling `AsPrompt`, never reaching `NewThread` logic.

## Use Case
The completion-watcher needs to:
1. Create a new thread for each completion notification (`--new-thread`)
2. Have the agent process it as an inbound prompt (`--as-prompt`)

Without both, either notifications scatter across main channel or don't get auto-processed.

## Implementation
- `InjectPrompt`: Injects message directly into agent stdin without posting to platform
- `PostToNewThread`: Posts message as new top-level thread (Slack) without agent injection
- `InjectPromptToNewThread`: Combines both - posts to new thread AND injects into agent session

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [ ] Manual testing with Slack

Closes #590

### DIFF
--- a/cmd/cc-connect/send.go
+++ b/cmd/cc-connect/send.go
@@ -115,6 +115,10 @@ func parseSendArgs(args []string) (core.SendRequest, string, error) {
 			}
 			i++
 			dataDir = args[i]
+		case "--as-prompt":
+			req.AsPrompt = true
+		case "--new-thread":
+			req.NewThread = true
 		case "--help", "-h":
 			return req, "", errSendUsage
 		default:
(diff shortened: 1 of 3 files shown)


chenhg5 (commented):
QA Review: feat(api): --as-prompt + --new-thread combination — adds the ability to start a fresh thread while injecting a prompt. CI green, LGTM.

PR #596, opened 2026-04-13
fix(core): follow symlinked skill directories

Zx55:
## Summary

Fix skill discovery so cc-connect can recurse into symlinked skill directories.

## Changes

- recurse into directory symlinks during skill discovery
- resolve real paths to avoid symlink loops
- keep deduping by leaf skill directory name

## Why

Some skill layouts use symlinked directories as part of their recursive discovery structure.

Before this change, cc-connect could recurse into real directories but skipped symlinked directory trees, which caused incomplete skill discovery.

## Tests

Added coverage for:

- grouped recursive discovery
- following directory symlinks
- avoiding symlink loops
- deduping same leaf names

### DIFF
--- a/core/skill.go
+++ b/core/skill.go
@@ -78,43 +78,84 @@ func (r *SkillRegistry) ListAll() []*Skill {
 	seen := make(map[string]bool)
 
 	for _, dir := range r.dirs {
-		entries, err := os.ReadDir(dir)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. 符号链接目录递归穿透修复，real path 防循环，dedup 正确。CI 全绿。

PR #597, opened 2026-04-13
fix(telegram): adapt command menu registration to Telegram limits

Zx55:
## Summary

Adapt Telegram bot command registration to Telegram command menu constraints.

## Changes

- add Telegram-specific menu planning before command registration
- keep skill commands all-or-none in the Telegram menu
- keep built-in and non-skill commands registered
- add a /skills hint explaining that skills can still be invoked manually when needed
- clamp Telegram bot command descriptions to a conservative safe length
- align the description clamp with the practical strategy used by Hermes Agent

## Why

Telegram command registration needs to stay within platform limits.

This change handles two practical constraints:

- keep the registered command set within Telegram command menu capacity
- clamp command descriptions to avoid registration failures caused by overly long menu payloads

The description clamp is intentionally conservative and follows the same general direction used by Hermes Agent for Telegram command registration.

The all-or-none skill policy also avoids a confusing partial-menu state where only some skills appear in the Telegram slash-command menu.

## Tests

Added coverage for:

- omitting all skill menu commands when Telegram menu capacity would overflow
- showing the manual invocation hint in /skills
- Telegram description truncation staying UTF-8 safe

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -23,6 +23,7 @@ import (
 )
 
 const maxPlatformMessageLen = 4000
+const telegramBotCommandLimit = 100
 const maxQueuedMessages = 5 // cap queued messages to bound memory usage
 
 const (
@@ -211,8 +212,8 @@ type Engine struct {
 
 	// Terminal observation (--observe)
 	observeEnabled    bool
-	observeProjectDir string             // ~/.claude/projects/{projectKey}
-	observeSessionKey string             // e.g. "slack:C123:U456" — target for forwarding
+	observeProjectDir string // ~/.claude/projects/{projectKey}
+	observeSessionKey string // e.g. "slack:C123:U456" — target for forwarding
 	observeCancel     context.CancelFunc
 
 	// Interactive agent session management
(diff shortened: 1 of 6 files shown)


chenhg5 (approved):
LGTM. Telegram 命令菜单容量适配，all-or-none skill 策略合理，描述截断安全。CI 全绿。

core/engine.go
```diff
@@ -5304,12 +5223,77 @@ func (e *Engine) GetAllCommands() []BotCommandInfo {
...
 
+func (e *Engine) menuCommandsForPlatform(platformName string) ([]BotCommandInfo, bool) {
+	commands := e.GetAllCommands()
+	if !strings.EqualFold(platformName, "telegram") {
+		return commands, false
+	}
+	return telegramMenuCommandsAllOrNone(commands)
+}
+
+func telegramMenuCommandsAllOrNone(commands []BotCommandInfo) ([]BotCommandInfo, bool) {
```
chenhg5: It is not recommended to explicitly introduce platform-specific dependencies in the engine.

PR #608, opened 2026-04-14
feat(core): relay unsolicited agent events between user turns

0xsegfaulted:
## Summary

cc-connect currently discards agent events produced between user turns. This PR adds a persistent background reader per session that relays those events (e.g. Claude Code \`run_in_background\` bash task completions) to the messaging platform.

## The Bug

Observed behavior: user asks Claude Code to run 5 long-running background tasks. Claude reports "submitted, waiting..." and the turn ends. 16 minutes later, the background tasks finish — but the user never sees the results. Eventually the user sends a new message asking "still working?" and the log shows \`drained stale events from previous turn count=49\` — the 49 queued task-completion events were silently discarded.

## Root Causes

1. **Structural** — \`processInteractiveEvents()\` only runs during a user turn; after \`EventResult\` it returns and nobody reads the \`Events()\` channel until the next user message, which calls \`drainEvents()\` to unconditionally discard anything buffered.
2. **Hardcoded timeout** — \`SetMultiWorkspace()\` hardcodes \`workspacePool = newWorkspacePool(15 * time.Minute)\`. After 15 minutes of "idle" (no foreground turns), the workspace is reaped and the agent process killed — even though it may have pending background work.

### Unsolicited event reader
A per-session background goroutine that consumes \`Events()\` between foreground turns. Started after a clean \`EventResult\`, stopped when a new foreground turn takes over channel ownership.

- Persistent reader (not conditional on agent hints — correctness shouldn't depend on agent-specific signals)
- Context-based lifecycle with bounded handoff (cancel → wait done with 5s timeout)
- Relays \`EventText\` / \`EventResult\` / \`EventError\` to the platform; auto-denies \`EventPermissionRequest\` when no user is watching
- Tracks \`BeginTurn\`/\`EndTurn\` so the idle reaper doesn't kill workspaces with active unsolicited processing

### Conditional resync
Replaces unconditional \`drainEvents()\` with an \`eventsNeedResync\` flag:
- **Defaults to \`true\`** (safe — any unhandled exit path drains)
- Cleared to \`false\` only after clean \`EventResult\`
- Foreground turn drains only when \`true\`; otherwise preserves buffered events for the unsolicited reader

### Configurable workspace idle timeout
New \`workspace_idle_timeout_mins\` per-project config (default 15min, \`0\` disables). Removes the previously-hardcoded timeout that was killing agent processes with pending background work.

### Concurrency safety
All code paths that read \`Events()\` (\`processInteractiveEvents\`, \`drainEvents\`, \`runCompress\`, \`drainOrphanedQueue\`) now call \`stopUnsolicitedReader\` first to maintain single-reader ownership. Cleanup paths (\`cleanupInteractiveState\`, \`stopInteractiveSession\`, session recycle) tear down the reader before closing the session.

## Design Discussion

Design was iterated through two rounds of Codex review (\`gpt-5.3-codex\` at high reasoning effort). Key decisions:
- **Persistent reader vs conditional** → persistent (agent-agnostic correctness)
- **Default for resync flag** → \`true\` (fail-safe against missed annotations)
- **Reader ownership** → context cancel + bounded wait (matches existing \`pendingPermission.Resolved\` pattern)
- **New agent interface** → not needed in v1 (detect from event stream)

Codex found 5 HIGH/MEDIUM issues in the first implementation pass (all fixed), 4 issues in the test pass (all fixed), and 1 race condition in self-review (\`drainOrphanedQueue\` — fixed). Final reviews PASS.

## Test plan

- [x] \`go test -race ./core/\` — 10 new unit tests + all existing pass
- [x] \`go test -race -tags=integration ./tests/integration/\` — 2 new end-to-end tests
  - \`TestIntegration_UnsolicitedEventsEndToEnd\`: user msg → clean turn → background events relayed → second msg works correctly
  - \`TestIntegration_StaleEventsDrainedAfterAbnormalExit\`: \`EventError\` → leftover events → next msg → leftovers correctly drained (log shows \`drained stale events from previous turn count=2\`)
- [x] \`go vet ./...\` clean
- [x] \`staticcheck\` on changed files clean
- [x] Manual: deployed to local cc-connect with \`workspace_idle_timeout_mins = 240\`; Discord / Telegram connections healthy

## Files

| File | Change |
|------|--------|
| \`core/engine.go\` | +289 Unsolicited reader, conditional resync, \`SetWorkspaceIdleTimeout\` |
| \`core/workspace_state.go\` | +4 Zero-timeout disables reaping |
| \`core/engine_test.go\` | +438 10 unit tests |
| \`config/config.go\` | +3 \`WorkspaceIdleTimeoutMins\` field |
| \`cmd/cc-connect/main.go\` | +8 Wire config → engine |
| \`tests/integration/unsolicited_events_test.go\` | +373 2 end-to-end integration tests |


chenhg5 (approved):
LGTM. 设计合理：background goroutine 消费 agent 事件并实时转发到 platform，eventsNeedResync flag 正确协调 foreground/background 之间的 channel 所有权。并发安全（bounded stop timeout 5s），permission deny 默认安全，8 个新测试覆盖所有代码路径。CI pending 但无红灯。已通过 QA review。

chenhg5 (commented):
Great PR — the root cause analysis is thorough and the fix is well-designed. The persistent reader approach (agent-agnostic, not dependent on hints) is the right call. A few issues worth addressing:

**1. `stopUnsolicitedReader` 5s timeout can create a brief concurrent reader (medium risk)**

After the timeout fires, `stopUnsolicitedReader` returns and the foreground turn starts reading from `Events()`, but the old reader goroutine may still be alive and selecting on the same channel. Although `cancel()` was called and the reader will exit on the next select iteration, there's a window where both goroutines compete for the same event. Consider either:
- Waiting indefinitely (with a warning log at 5s) instead of proceeding with a potentially concurrent reader, or
- Adding a `sync.Mutex` around all `Events()` reads to enforce single-reader at the Go level

**2. `workspace_idle_timeout_mins` should be a global config, not per-project**

The `workspacePool` is engine-level (one pool per engine), but the config field is on `ProjectConfig`. In `main.go`, `SetWorkspaceIdleTimeout` is called inside the project loop — so the last project's value silently overrides all previous ones. This should either be a global config field (e.g. under `[workspace]` or top-level), or `SetWorkspaceIdleTimeout` should only apply on the first call.

**3. Silent permission deny in unsolicited context may confuse users**

When `approveAll=false`, the reader auto-denies permission requests without notifying the user. A background task could silently fail and the user would have no idea why. Suggest sending a brief platform notification like "Background task requested permission for {tool} but was auto-denied (no active user turn)." so users can decide to re-run with `/yolo` or manually trigger.

**4. `EventToolUse` / `EventToolResult` are silently discarded**

The reader comments say "accumulate silently" but no accumulation actually happens — the events are just dropped in the switch fallthrough. If the channel closes between `EventToolUse` and `EventResult`, tool context is lost without any log. At minimum, consider logging tool events at debug level, or actually accumulating tool names to include in the final relay message.

**5. Minor: `session.AddHistory` thread safety**

In `runUnsolicitedReader`, `session.AddHistory("assistant", fullResponse)` is called without holding the session lock. The foreground path uses `session.TryLock()`. Since `stopUnsolicitedReader` is called before the foreground turn starts, this should be safe in practice, but it relies on ordering rather than explicit synchronization. Worth a comment or an explicit lock.

---

Overall this is solid work — the conditional resync flag, single-reader ownership protocol, and BeginTurn/EndTurn tracking for the idle reaper are all well thought out. Tests are comprehensive with proper synchronization (no flaky sleeps). The integration tests modeling persistent event channels are particularly well done.

The main blocker IMO is issue #2 (per-project config for an engine-level resource). The rest are improvements that could land in follow-ups.

0xsegfaulted:
Thanks for the thorough review! All five items addressed in c1e4c05, plus a main merge in da82bcf. Summary per item:

**1. `stopUnsolicitedReader` 5s timeout / concurrent reader window** — Kept the 5s bounded wait (going unbounded would deadlock callers holding `interactiveMu`, e.g. `getOrCreateInteractiveStateWith` and `stopInteractiveSession`), but closed the race two ways:
- After the reader's outer `select` returns an event, it does a second non-blocking `<-ctx.Done()` check. If cancellation happened between `select` firing and here, the event is dropped with a warning and `eventsNeedResync=true` — the foreground turn's drain then handles it.
- Offloaded `RespondPermission` to a detached goroutine so the reader's main loop never blocks in an adapter call. With that, reader iterations are sub-millisecond and the 5s wait is effectively non-binding in practice.

**2. `workspace_idle_timeout_mins` scope** — Promoted to a top-level `Config.WorkspaceIdleTimeoutMins` (global for the process). `ProjectConfig.WorkspaceIdleTimeoutMins` kept as a per-project override; `main.go` picks project > global > default (15 min). Keeps existing per-project configs working while fixing the "last project silently overrides" issue for shared deployments.

**3. Silent permission deny notification** — When `approveAll=false`, the reader now sends a localized `MsgBackgroundAutoDenied` message through the platform identifying the requested tool (EN/ZH/ZH-TW/JA/ES), so a blocked background task is no longer invisible. The user gets a nudge to send a message or use `/yolo` if they want to approve.

**4. `EventToolUse` / `EventToolResult` handling** — Removed the misleading "accumulate silently" comments. Tool names are now accumulated in `toolsUsed`; tool events emit `slog.Debug`. If the channel closes mid-turn with buffered context, a `slog.Warn` captures the tools/text fragments that would otherwise be lost.

**5. `session.AddHistory` thread-safety** — Session.mu is taken internally by `AddHistory`, and `stopUnsolicitedReader`'s bounded wait plus the reader's final `EventResult` ordering ensures the reader's `AddHistory` happens-before any foreground `AddHistory`. Added an explanatory comment above the call so the invariant is explicit rather than implicit.

All changes reviewed by Codex (`gpt-5.3-codex`, high reasoning effort). Codex caught a P1 on my first attempt (unbounded wait → deadlock risk under `interactiveMu`), which drove the bounded-5s + ctx-double-check + offloaded-RespondPermission design above. Second pass: clean, no regressions.

Verified: `go test -race ./...`, `go test -tags=integration -race ./tests/integration/`, `go vet ./...` all pass (modulo 3 pre-existing macOS-specific failures that also fail on pure `main`: `TestProcessInteractiveEvents_AppendsReplyFooter{WhenEnabled,PrefersSessionRuntimeState}` and `TestResolveLocalDirPath_AcceptsSubdir` — unrelated to this PR).

PTAL 🙏

chenhg5:
@0xsegfaulted Hey! This is a great feature and we'd love to get it merged. Could you address the outstanding review items when you have a chance?

1. **Blocker: `workspace_idle_timeout_mins` config scope** — As noted in the review, this should be a global config (e.g. under `[workspace]`) rather than per-project, since `workspacePool` is engine-level and the last project's value silently overrides all others.
2. **Rebase onto latest `main`** — There have been significant changes since this PR was opened (NO_REPLY suppression, `/ps` command, message queue depth, etc.) that touch `core/engine.go`. A rebase is needed to resolve conflicts and verify no interaction issues.

Once those two are addressed we can merge. Thanks for the contribution!

zgq233333:
@chenhg5 Both blockers addressed in 5387e4a8 + merge d3e1d48f. PR is now `MERGEABLE / CLEAN`.

**1. `workspace_idle_timeout_mins` scope** — Removed the per-project field; the top-level `Config.WorkspaceIdleTimeoutMins` is now the only canonical setting (per-project configuration is intentionally not supported, with a clarifying comment to that effect). To avoid silently breaking existing deployments that used `0` to disable reaping or set a custom value under `[[projects]]`, the deprecated per-project key is still parsed (as `WorkspaceIdleTimeoutMinsLegacy` in code) and honored as a fallback when the top-level field is unset, with a `slog.Warn` at startup pointing users to the new location. The fallback path will be removed in a future release.

**2. Rebase onto latest `main`** — Did a merge from `main` rather than a true rebase, since the branch already carried a prior merge commit; the effect for "resolve conflicts and verify no interaction issues" is the same. Three conflicts resolved by keeping both sides:
- `core/engine.go` — our unsolicited reader block + main's new `agentErrorHandler` / `agentErrorHandlers`
- `core/engine_test.go` — our unsolicited reader tests + main's new `TestIsSilentReply` / `TestStripTrailingSilent` / `TestCouldBeSilentPrefix`
- `core/i18n.go` — kept both `MsgBackgroundAutoDenied` and `MsgSessionNotFound`

Also bundled into 5387e4a8 are the 437 lines of `core/engine_test.go` additions that should have landed alongside c1e4c053 (covers unsolicited reader relay, cancel/stop, channel-close resync, error resync, permission deny, resync flag default + clearing on clean result, cleanup teardown, and `SetWorkspaceIdleTimeout` configurability incl. the 0-disables-reaping path).

Each commit was reviewed by Codex (`codex exec review --uncommitted`) until clean. The merge-resolution review surfaced only pre-existing main concerns (`/ps`, OpenCode stderr ordering, launchd test, Slack italics, MAX byte-splitting) — out of scope for this PR.

Verified: `go build ./...` clean; relevant new tests (unsolicited reader, workspace idle timeout, NO_REPLY family) all pass; `go test -race` clean for the new code paths. The 3 known macOS-only pre-existing failures remain (`TestProcessInteractiveEvents_AppendsReplyFooter{WhenEnabled,PrefersSessionRuntimeState}`, `TestResolveLocalDirPath_AcceptsSubdir`) and continue to fail on pure `main` — unrelated to this PR.

PTAL 🙏

PR #612, opened 2026-04-14
fix(codex): preserve workspace app-server options

xukp20:
# Title

fix(codex): preserve workspace app-server options

## Summary

This PR fixes a multi-workspace regression in the Codex agent path.

When a project uses `mode = "multi-workspace"`, workspace agents were only inheriting a limited subset of the parent Codex options. As a result, `backend = "app_server"` and related Codex session options were dropped when the workspace agent was created.

That made the root project run with the configured backend while workspace sessions silently fell back to default behavior.

## What this changes

This PR keeps the fix narrowly scoped:

- add a workspace-agent option snapshot interface
- let `Engine.getOrCreateWorkspaceAgent(...)` clone the parent agent's construction options before overriding `work_dir`
- make `agent/codex.Agent` export the relevant option snapshot, including:
  - `backend`
  - `app_server_url`
  - `codex_home`
  - `reasoning_effort`
  - `mode`
  - `model`
  - existing `run_as_user` / `run_as_env`

## Scope

Included:

- multi-workspace Codex workspace-agent option inheritance
- regression tests for inherited provider / snapshot / run-as behavior

## Implementation

Main areas touched:

- `core/interfaces.go`
  - adds the optional workspace snapshot interface
- `core/engine.go`
  - clones parent agent options before overriding workspace `work_dir`
- `agent/codex/codex.go`
  - exports the Codex option snapshot used for workspace cloning
- `core/engine_test.go`
  - adds regression coverage for inherited provider / snapshot / run-as behavior

## Testing

Passed locally:

```bash
go test ./core -run 'Test(GetOrCreateWorkspaceAgent_InheritsActiveProvider|GetOrCreateWorkspaceAgent_InheritsSnapshotOptions|MultiWorkspaceAgent_PropagatesRunAsUser)' -count=1
go test ./agent/codex -count=1
go build ./...
go test ./... -v -race
go test ./... -coverprofile=coverage.out -covermode=atomic
go test -v -tags=smoke,no_web ./tests/e2e/...
go test -v -tags=regression,no_web ./tests/e2e/...
```


chenhg5 (approved):
LGTM. 为每个 agent 添加 WorkspaceAgentOptions()，实现 WorkspaceAgentOptionSnapshotter 接口，创建 workspace agent 时正确传递所有配置项。

chenhg5 (commented):
@xukp20 Clean fix, well-scoped. A few suggestions:

**1. Snapshot shouldn't export `work_dir`**

`WorkspaceAgentOptions()` sets `opts["work_dir"] = a.workDir`, but `getOrCreateWorkspaceAgent` immediately overrides it with `opts["work_dir"] = workspace`. The snapshot should omit `work_dir` to avoid confusion — the caller is always responsible for setting it.

**2. Add `if _, ok` guards on `run_as_user` / `run_as_env` for consistency**

The PR correctly guards `model` and `mode` with `if _, ok := opts["model"]; !ok`, but the `run_as_user` and `run_as_env` fallback code below doesn't have the same guard. If a future agent includes these in its snapshot, the fallback would silently overwrite them. For consistency:

```go
if _, ok := opts["run_as_user"]; !ok {
    if ma, ok := e.agent.(interface{ GetRunAsUser() string }); ok {
        ...
    }
}
```

**3. Consider including `run_as_user` / `run_as_env` in the Codex snapshot (or document why not)**

The name "WorkspaceAgentOptionSnapshotter" suggests a complete set of constructor options, but `run_as_user`/`run_as_env` are handled by separate fallback code. This works today, but it's fragile — if someone removes or refactors the fallback code, these fields silently disappear. Either include them in the snapshot, or add a comment on the interface explaining that `run_as_*` and providers are handled separately.

**4. Minor: other agents may need this too**

Claude Code and other agents could also lose options (e.g. `allowed_tools`, `disallowed_tools`, `max_context_tokens`) in multi-workspace mode. Not a blocker for this PR, but worth tracking as a follow-up.

---

Overall LGTM — the core fix (snapshot → clone → override work_dir) is the right approach. Issues 1-2 are easy fixes before merge.

xukp20:
Thanks for the detailed review. I updated the branch to address the requested points:

1. `WorkspaceAgentOptions()` no longer exports `work_dir`.
   The workspace path is now always set by `getOrCreateWorkspaceAgent(...)`, so the snapshot only carries reusable constructor options.

2. Added `if _, ok` guards for `run_as_user` / `run_as_env` in the engine fallback path.
   This now matches the existing `model` / `mode` behavior and avoids silently overwriting snapshot-provided values.

3. Clarified the interface comment.
   `WorkspaceAgentOptionSnapshotter` now documents that `work_dir` is caller-owned, while provider wiring / run_as propagation may still be handled separately by the engine.

I also added regression coverage so the snapshot path explicitly verifies:
- `work_dir` is caller-injected rather than snapshot-carried
- snapshot-provided `run_as_user` / `run_as_env` win over the fallback path

The branch was also updated onto the latest `main`, and I reran the full local CI-equivalent suite before pushing.

xukp20:
@chenhg5 I addressed the review points above, updated the branch on top of the latest `main`, and reran the full local CI-equivalent suite. When you have time, could you please take another look?

PR #613, opened 2026-04-14
feat(feishu): allow_chat filter & shared WebSocket for multi-project apps

vinnyxiong:
## Summary

- **allow_chat option**: New config option to bind specific group chats to a project via `allow_chat` (comma-separated `chat_id` list or `*` for all). Private (P2P) chats are unaffected by this filter.
- **Shared WebSocket connection**: When multiple projects use the same Feishu `app_id`, they now share a single WebSocket connection. This avoids Feishu's server-side load-balancing which randomly routes messages across connections. The first platform is the primary (owns the WS), others are secondary. All events (messages, card actions, bot menu) are fanned out to every platform in the group.
- Together these two features allow multiple projects to share one Feishu app while each only responds to its designated groups.

## Changes

- `platform/feishu/feishu.go`: Add `allowChat` field, integrate `allow_chat` filtering in `onMessage`, fan out events via shared WS group
- `platform/feishu/ws_shared.go`: New file managing shared WebSocket groups keyed by `app_id|domain`
- `platform/feishu/platform_test.go`: Table-driven tests for `allow_chat` filtering
- `config.example.toml`: Document the new `allow_chat` option

## Test plan

- [x] Unit tests for allow_chat filtering (group blocked/allowed, wildcard, P2P bypass)
- [ ] Manual test with two projects sharing same app_id, each with different allow_chat values


chenhg5 (approved):
LGTM. 两个 features：（1）allow_chat 配置过滤群聊消息；（2）多项目共享同一 app_id 时通过 sharedWSGroup 解决飞书服务端负载均衡导致的随机路由问题。fan-out 逻辑正确，AllowList 用于 allow_chat 过滤，private chat 绕过 filter 的设计正确。8 个测试用例覆盖。CI 通过。

vinnyxiong:
Hi @chenhg5, this PR is ready for review. The CI failure was a flaky test in agent/opencode (unrelated to this PR), already retriggered. Could you take a look when you have a chance? Thanks! 🙏

chenhg5 (commented):
@vinnyxiong Nice idea — sharing one WS to work around Feishu's load-balancing is a real pain point. The `allow_chat` filter itself is clean. Found a few issues though:

**1. Primary shutdown silently kills all secondaries**

When the primary platform calls `Stop()`, it cancels the shared WebSocket connection. All secondary platforms lose their event source with no failover, no log warning, and no way to recover. At minimum, the primary's `Stop()` should log a warning if other platforms remain in the group. Ideally, promote a secondary to primary and have it reconnect.

**2. Card action / bot menu fan-out bypasses `allow_chat` filter**

`onMessage` correctly applies `allow_chat`, but `onCardAction` and `onBotMenu` are fanned out to ALL siblings without any chat filtering. A card interaction from a group that only project B should handle could be matched by project A first — especially since `onCardAction` returns the first non-nil response:

```go
for _, sibling := range p.sharedGroup.allPlatforms() {
    resp, err := sibling.onCardAction(event)
    if resp != nil { return resp, nil } // wrong project may win
}
```

Each platform should check whether the card action's chat belongs to its `allow_chat` list before processing.

**3. Lark (international) webhook mode will double-process messages**

The `startWebhookMode()` return is BEFORE the secondary check:

```go
if p.domain == lark.LarkBaseUrl {
    return p.startWebhookMode() // returns here — every Lark platform starts its own webhook
}
if !isPrimary {
    return nil // never reached for Lark
}
```

But the event handlers already contain fan-out logic. So when Lark platform A receives a webhook event, it fans out to both A and B. When B receives the same event on its own webhook, it also fans out to both. Result: every message processed twice per platform.

**4. `group_only` option is used but not documented**

The `groupOnly` field is read from config and applied in `onMessage`, but `config.example.toml` only documents `allow_chat`. Should add a comment for `group_only` too.

**5. No tests for shared WS logic**

`ws_shared.go` (register/unregister/allPlatforms/fan-out) has zero test coverage. The package-level `sharedWSGroups` global also makes parallel tests fragile. Consider at least testing: register two platforms → allPlatforms returns both → unregister one → allPlatforms returns one → unregister last → group deleted.

**6. Minor: secondary platforms create unused event handlers**

Secondaries build a full `eventHandler` with fan-out closures that are never invoked (no WebSocket to trigger them). Could move event handler creation after the `!isPrimary` early return to avoid unnecessary work.

---

The `allow_chat` feature is solid and well-tested on its own. The shared WebSocket approach solves a real problem but needs the issues above addressed — especially #2 (card actions bypassing filter) and #3 (Lark double-processing) which can cause incorrect behavior in production.

vinnyxiong:
@chenhg5 Thanks for the thorough review! All 6 issues addressed in 6d23ac9:

| # | Issue | Fix |
|---|-------|-----|
| 1 | Primary shutdown silently kills secondaries | `Stop()` now logs a warning with remaining secondary count |
| 2 | Card action / bot menu bypasses `allow_chat` | `onCardAction` checks `allow_chat` on `OpenChatID` before processing; `onBotMenu` respects `groupOnly` |
| 3 | Lark webhook double-processing | Moved `isPrimary` check before the webhook/WS branch — secondaries no longer start their own webhook |
| 4 | `group_only` undocumented | Added to `config.example.toml` with bilingual comments |
| 5 | No tests for `ws_shared.go` | Added `ws_shared_test.go`: register/unregister/allPlatforms, different app_ids isolation, group cleanup |
| 6 | Secondaries create unused event handlers | Moved event handler creation after `!isPrimary` early return |

Ready for another look when you have time!

vinnyxiong:
Hi @chenhg5, conflicts resolved and all review feedback addressed. Could you merge when you get a chance? Thanks!

PR #615, opened 2026-04-14
Use long connection for Lark by default

bugsyyu:
## Summary
- route the `lark` platform through the official SDK WebSocket long-connection path
- keep the existing webhook handler implementation available without making it the default ingress mode
- align runtime behavior with the project documentation that treats Feishu and Lark as the same long-connection transport family

## Why
Lark developer apps support long-connection event delivery and do not require a public callback URL for normal bot ingress. The current implementation forces `type = "lark"` into webhook mode, which makes international Lark deployments depend on public webhook infrastructure even when the app is configured for long connection.

This patch removes that forced branch and lets the existing WebSocket startup path handle both Feishu and Lark through the official SDK.

## Validation
- built `cc-connect` from this branch
- started the binary with a Lark app configuration
- confirmed a successful connection to the Lark long-connection endpoint
- confirmed the patched instance started normally without requiring webhook callback configuration

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -315,12 +315,9 @@ func (p *Platform) Start(handler core.MessageHandler) error {
 			return p.onBotMenu(event)
 		})
 
-	// Lark international version uses Webhook mode, not WebSocket long connection
-	// Feishu domestic version supports WebSocket long connection
-	if p.platformName == "lark" {
-		return p.startWebhookMode()
-	}
-
+	// Kernel dedicated build: both Feishu and Lark use the official SDK WebSocket
+	// long-connection path. Keep the webhook implementation available as a helper,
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. 主要改动：（1）Lark 强制 webhook 改为 SDK WebSocket 长连接；（2）Monitor Trigger Container（monitor register/trigger/list/info 命令 + Unix socket API）；（3）Inject 命令（cc-connect inject）通过 socket 发送结构化消息。代码质量符合既有模式，Unix socket 认证复用现有机制。

注：PR 描述仅提及 Lark fix，实际改动包含 monitor/inject 等更大范围，建议下次在描述中写明完整 scope。

chenhg5:
@bugsyyu Please fix the CI tests and resolve the conflicts before merging.

PR #620, opened 2026-04-15
feat(core): replace /btw with /ps as builtin command

### LINKED ISSUE #619: Replace /btw with /ps as a proper builtin command
## Problem

The current `/btw` command has three issues:

1. **Name collision with Claude Code CLI**: Claude Code (the agent CLI) already has its own `/btw` command meaning "ask a side question." cc-connect's `/btw` does something completely different (mid-turn stdin injection), which creates user confusion.

2. **Not registered as a builtin command**: `/btw` is handled via a special-case check in the `TryLock` failure branch, outside the normal command routing. This means it first triggers `MsgUnknownCommand` ("not a cc-connect command, forwarding to agent...") before falling through to the actual btw logic — producing two confusing replies per invocation.

3. **Missing from help**: `/btw` doesn't appear in `/help` output or the help card, so users can only discover it from the `MsgPreviousProcessing` hint.

## Proposal

Replace `/btw` with `/ps` (postscript — "P.S."), implemented as a proper builtin command:

- Register in `builtinCommands` and handle via `cmdPs()` in the `switch` statement
- Follow the `cmdStop()` pattern: use `interactiveKeyForSessionKey()` for multi-workspace support, directly access `interactiveStates`
- This bypasses `TryLock` entirely, so `/ps` works whether the session is busy or not
- Add proper i18n messages: `ps_sent`, `ps_send_failed`, `ps_empty`, `ps_no_session`
- Add to help card under the tools group
- Update `MsgPreviousProcessing` to reference `/ps` instead of `/btw`
- Remove `isBtwCommand()` and `matchBtwPrefix()` helper functions

## Environment

- cc-connect v1.2.2-beta.5
- macOS, Lark platform

Cigarrr:
## Summary

- Replace `/btw` with `/ps` (postscript) as a proper builtin command
- Fix the spurious "unknown command" message that `/btw` triggered before injection
- Add `/ps` to the help card so users can discover it

## Motivation

1. **Name collision**: Claude Code CLI already has `/btw` (side question) with different semantics — having cc-connect's `/btw` (mid-turn injection) use the same name confuses users
2. **Routing bug**: `/btw` was not in `builtinCommands`, so `handleCommand()` would first send `MsgUnknownCommand` before falling through to the btw injection logic in the `TryLock` failure branch — producing two replies per invocation
3. **Undiscoverable**: `/btw` was absent from `/help` output; users could only learn about it from `MsgPreviousProcessing`

## Changes

- Register `/ps` in `builtinCommands` and add `case "ps"` to the `handleCommand` switch
- Implement `cmdPs()` following the `cmdStop()` pattern: uses `interactiveKeyForSessionKey()` for multi-workspace support, directly accesses `interactiveStates` — bypasses `TryLock` entirely
- Add 4 new i18n messages (5 languages each): `ps_sent`, `ps_send_failed`, `ps_empty`, `ps_no_session`
- Add builtin command description for help card: `MsgBuiltinCmdPs`
- Add `/ps` to the tools group in `helpCardGroups()`
- Update `MsgPreviousProcessing` to reference `/ps` instead of `/btw`
- Remove `isBtwCommand()` and `matchBtwPrefix()` functions
- Update Slack app manifest slash command registration

## Test plan

- [x] `go test ./...` passes (all packages OK)
- [x] `/ps <message>` while session is busy: injects mid-turn, replies "✅ P.S. delivered."
- [x] `/ps` with no args: replies with usage hint
- [x] `/ps` with no running task: replies "No task is currently running."
- [x] `/help` now shows `/ps` in the tools group
- [x] No remaining references to `btw` in engine.go or i18n.go

Closes #619

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1506,25 +1506,6 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 	session := sessions.GetOrCreateActive(msg.SessionKey)
 	sessions.UpdateUserMeta(msg.SessionKey, msg.UserName, msg.ChatName)
 	if !session.TryLock() {
-		// Check for /btw — inject into the running session mid-turn
-		trimmed := strings.TrimSpace(content)
-		if isBtwCommand(trimmed) {
-			btw := strings.TrimSpace(trimmed[len(matchBtwPrefix(trimmed)):])
-			if btw != "" {
-				e.interactiveMu.Lock()
-				state, ok := e.interactiveStates[interactiveKey]
-				e.interactiveMu.Unlock()
-				if ok && state.agentSession != nil && state.agentSession.Alive() {
-					if err := state.agentSession.Send(btw, nil, nil); err != nil {
-						slog.Error("btw: send failed", "error", err)
-						e.reply(p, msg.ReplyCtx, e.i18n.T(MsgBtwSendFailed))
-					} else {
-						e.reply(p, msg.ReplyCtx, e.i18n.T(MsgBtwSent))
-					}
-					return
-				}
-			}
-		}
 		// Session is busy — try to queue the message for the running turn
 		// so the agent processes it immediately after the current turn ends.
 		if e.queueMessageForBusySession(p, msg, interactiveKey) {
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
LGTM. Clean fix — eliminates the double-reply bug by properly registering /ps in builtinCommands and handleCommand switch (vs the old /btw which lived in the TryLock failure branch and triggered MsgUnknownCommand first). Net code reduction (-58 lines). i18n updated in all 5 languages. Slack manifest updated. No break-change concern for most users since /btw was undiscoverable (not in /help) and broken. CI green.

✅ Merge ready.

PR #621, opened 2026-04-15
fix(daemon): preserve proxy env in systemd service

zkunzhu:
preserve user specified proxy env in generated systemd service


chenhg5 (approved):
LGTM. Proxy environment variables correctly preserved in systemd unit file. Fix is targeted and safe. CI pass.

PR #636, opened 2026-04-15
Fix Issue #619: Replace /btw with /ps as proper builtin command

### LINKED ISSUE #619: Replace /btw with /ps as a proper builtin command

chenhg5:
## Summary
- Replace /btw with /ps (postscript) to avoid collision with Claude Code CLI's /btw command
- Register /ps as proper builtin command to fix double-reply issue
- Bypass TryLock - /ps works whether session is busy or not
- Add i18n messages: ps_sent, ps_send_failed, ps_empty, ps_no_session
- Add /ps to help card under tools group
- Remove isBtwCommand() and matchBtwPrefix() helper functions

## Root Cause
The original /btw command had three problems:
1. **Name collision**: Claude Code CLI's /btw means "ask a side question", different from cc-connect's mid-turn injection
2. **Not registered as builtin**: Triggered MsgUnknownCommand before falling through to btw logic, causing two replies
3. **Missing from help**: Not discoverable in /help output

## Fix Approach
- Create /ps command registered in builtinCommands
- Handle via cmdPs() in switch statement, following cmdStop() pattern
- Use interactiveKeyForSessionKey() for multi-workspace support
- Directly access interactiveStates, bypassing TryLock entirely

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes (all tests)
- [ ] Manual smoke test: send /ps command during busy session

Closes #619

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1509,25 +1509,6 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 	session := sessions.GetOrCreateActive(msg.SessionKey)
 	sessions.UpdateUserMeta(msg.SessionKey, msg.UserName, msg.ChatName)
 	if !session.TryLock() {
-		// Check for /btw — inject into the running session mid-turn
-		trimmed := strings.TrimSpace(content)
-		if isBtwCommand(trimmed) {
-			btw := strings.TrimSpace(trimmed[len(matchBtwPrefix(trimmed)):])
-			if btw != "" {
-				e.interactiveMu.Lock()
-				state, ok := e.interactiveStates[interactiveKey]
-				e.interactiveMu.Unlock()
-				if ok && state.agentSession != nil && state.agentSession.Alive() {
-					if err := state.agentSession.Send(btw, nil, nil); err != nil {
-						slog.Error("btw: send failed", "error", err)
-						e.reply(p, msg.ReplyCtx, e.i18n.T(MsgBtwSendFailed))
-					} else {
-						e.reply(p, msg.ReplyCtx, e.i18n.T(MsgBtwSent))
-					}
-					return
-				}
-			}
-		}
 		// Session is busy — try to queue the message for the running turn
 		// so the agent processes it immediately after the current turn ends.
 		if e.queueMessageForBusySession(p, msg, interactiveKey) {
@@ -3132,6 +3113,7 @@ var builtinCommands = []struct {
 	{[]string{"heartbeat", "hb"}, "heartbeat"},
 	{[]string{"compress", "compact"}, "compress"},
 	{[]string{"stop"}, "stop"},
+	{[]string{"ps", "postscript"}, "ps"},
 	{[]string{"help"}, "help"},
 	{[]string{"version"}, "version"},
 	{[]string{"commands", "command", "cmd"}, "commands"},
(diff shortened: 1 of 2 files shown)


xukp20:
I left a longer design note on `#619` here:
https://github.com/chenhg5/cc-connect/issues/619#issuecomment-4252608500

Short version:

1. `/ps` is probably not a safe replacement name either, because Codex already uses native `/ps` for listing background terminals.
2. The current `/btw`/proposed `/ps` implementation still looks semantically wrong for Codex, because it behaves more like sending another message / starting another turn, instead of steering the current in-flight task.
3. For Codex app-server, this should ideally map to native `turn/steer`.
4. For Codex exec, the current approach likely has incorrect concurrent session/thread semantics.
5. A cleaner cross-agent command name would be something like `/steer`, with explicit semantics: append guidance to the current in-flight task, same turn, same task, not a new turn.
6. Claude-native `/btw` should be kept separate if we ever want to expose it, because native Claude `/btw` is a side-question feature, not the same thing as steer.

So I think the naming problem and the backend semantic problem should be solved together.

chenhg5 (commented):
## QA Review — LGTM ✅

Replacing the ad-hoc `/btw` handling with a real builtin `/ps` fixes the collision with Claude Code's `/btw` and the double-reply behavior. Routing through `handleCommand` before the session lock matches the intended "inject mid-turn" behavior.

---

🟡 **Migration**: `/btw` will no longer inject; consider a short migration note or a temporary alias mapping `btw` → `ps` for existing users.

---

✅ **Good**: Clear separation — builtin registration + dedicated handler; better UX than unknown-command + special-case branch.

chenhg5:

chenhg5:
先关闭这个旧 PR。目标 issue #619 已经关闭；这个 PR 也长期和 main 冲突。更重要的是，review 里已经指出新的 /ps 命令可能与 Codex native /ps 命令冲突，且 mid-turn steering 的命名和语义需要重新设计。继续合这个旧 PR 风险较高。\n\n如果后续要做 mid-turn steering，建议新开需求和 PR，统一命名和跨 agent 语义，例如评估 /steer，并分别处理 Claude Code、Codex app-server、Codex exec 的行为差异。\n\n---\n\nClosing this old PR for now. The target issue #619 is already closed, and this PR has long-standing conflicts with main. Review feedback also pointed out that the proposed /ps command may conflict with Codex native /ps, and mid-turn steering naming/semantics need a fresh design.\n\nIf we want mid-turn steering later, please open a fresh issue and PR that defines command naming and cross-agent semantics first, for example evaluating /steer, with separate handling for Claude Code, Codex app-server, and Codex exec behavior.

Issue #629, opened 2026-04-15
Model configured in config.toml does not work, but settings.json works for the same model


abcxyzx0099:
## Bug Description

When configuring a model (e.g., kimi-for-coding) in config.toml, the bot returns error:

API Error: 400 {error:{code:1211,message:Unknown Model, please check the model code.}}

However, when the same model configuration is placed in settings.json (in the project work_dir), it works correctly.

## Steps to Reproduce

1. Configure model in config.toml:

   [projects.agent.options]
   model = kimi-for-coding
   ANTHROPIC_AUTH_TOKEN = sk-kimi-...
   ANTHROPIC_BASE_URL = https://api.kimi.com/coding/

2. Send a message to the bot via DingTalk
3. Error is returned

## Expected Behavior

Model configured in config.toml should work the same as settings.json

## Environment

- cc-connect version: 1.2.2-beta.5
- Platform: DingTalk
- Model: kimi-for-coding (works with settings.json)

## Additional Notes

- Direct API calls with the same credentials work fine
- The issue persists even after daemon restart
- Session cleanup does not resolve it
- Only happens with DingTalk platform (Feishu works correctly with glm-5.1 in config.toml)

xukp20:
This may be a different issue from the provider-override case.

From the current Claude Code adapter in `cc-connect`, `projects.agent.options` only reads a fixed set of supported fields such as `model`, `mode`, `work_dir`, `allowed_tools`, etc. It does **not** treat arbitrary keys under `projects.agent.options` as subprocess environment variables.

That would explain the behavior you are seeing:

- `model` works
- `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL` do not

So before treating this as a runtime bug, could you try moving the config to the supported `providers` block instead of putting `ANTHROPIC_*` directly under `projects.agent.options`?

Example:

```toml
[projects.agent.options]
work_dir = "..."
provider = "kimi"

[[projects.agent.providers]]
name = "kimi"
api_key = "sk-kimi-..."
base_url = "https://api.kimi.com/coding/"
model = "kimi-for-coding"
```

Then restart `cc-connect` and test again.

If it still fails with this structure, then it is more likely a real adapter/runtime issue. But with the current config shape, my first guess is that the config is in the wrong place rather than the same override bug.

PR #638, opened 2026-04-15
fix(claudecode): honor host-managed provider env

xukp20:
## Summary

This PR fixes a Claude Code provider-routing bug when `cc-connect` is configured
to manage a custom provider for a project.

Before this change, `cc-connect` did inject provider env vars like
`ANTHROPIC_BASE_URL` / `ANTHROPIC_AUTH_TOKEN`, but Claude Code could still
re-apply `~/.claude/settings.json` env on startup and override those values.

That led to a broken mixed state:

- the model from the project provider config applied
- but the request routing still followed the user's local Claude settings

## What this changes

- detect when `cc-connect` is actually managing Claude provider routing
- set `CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1` in that case
- apply the same logic to both:
  - normal Claude sessions
  - Claude `/usage` probe sessions

This is intentionally scoped to provider routing precedence only.

## Scope

Included:

- Claude provider env precedence fix for project-managed providers
- coverage for provider/base_url/api_key, session env, provider env, and router env
- `/usage` probe path alignment

Not included:

- any new provider config syntax
- automatic support for arbitrary `projects.agent.options.ANTHROPIC_*`
- unrelated Claude settings / hooks behavior

## Implementation

Main areas touched:

- `agent/claudecode/claudecode.go`
  - adds runtime env assembly with host-managed provider detection
- `agent/claudecode/claude_usage.go`
  - reuses the same runtime env path for `/usage`
- `agent/claudecode/provider_env_test.go`
  - adds regression coverage for host-managed env behavior

Important implementation detail:

- the host-managed flag is **not** set unconditionally
- it is only added when the final Claude runtime env actually contains
  provider-managed routing / auth / model-default variables

This avoids breaking cases where a provider entry only changes `model` and
still expects the user's local Claude provider settings to apply.

## Validation

Passed locally:

```bash
cd web
pnpm install --frozen-lockfile
pnpm build
cd ..
actionlint -color
golangci-lint run --new-from-rev origin/main ./...
go build ./...
go test ./... -v -race
go test ./... -coverprofile=coverage.out -covermode=atomic
go test -v -tags=smoke,no_web ./tests/e2e/...
go test -v -tags=regression,no_web ./tests/e2e/...
```

Real local validation was also performed with a temporary `CLAUDE_CONFIG_DIR`:

- with deliberately wrong `settings.json` provider env
- with correct provider env injected from the parent process
- comparing behavior with and without `CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1`

Observed result:

- without the host-managed flag: Claude followed settings-sourced env and failed / timed out
- with the host-managed flag: Claude used the host-injected provider env and returned `OK`

## Related Issues

- Fixes the root cause discussed in [#635](https://github.com/chenhg5/cc-connect/issues/635)
- Related but different: [#629](https://github.com/chenhg5/cc-connect/issues/629)
  - that one appears to be about using `projects.agent.options` for arbitrary `ANTHROPIC_*` env, which is outside this PR's scope


chenhg5 (approved):
LGTM. Respects CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST env var — host-managed provider settings won't be overridden by settings.json env vars. Clean fix. CI pass.

PR #640, opened 2026-04-16
feat(claudecode): support Claude-compatible CLI wrappers via cli_path

RukawaKaede:
## Summary

- Add `cli_path` option so the claudecode agent can spawn third-party Claude-compatible CLIs (or wrappers) instead of hardcoding `claude`. The value may include extra args (e.g. `my-cli code -t foo`) which are parsed and prepended before Claude flags.
- Add `cli_args_flag` option: when set (e.g. `-a`), the inner Claude flags (`--verbose`, `--append-system-prompt`, …) are bundled into a single passthrough string via that flag, while wrapper-understood flags (`--model`) are still appended directly. Values with whitespace/quotes are safely single-quoted so shell-style splitters keep them as one token.
- Preflight `LookPath` now checks the configured binary and reports it in the error message.
- `CLIBinaryName()` returns the configured binary so `doctor` diagnostics reflect the real target.

## Test plan

- [x] `go test ./...`
- [x] Manual: configure `cli_path = "claude"` — behavior unchanged vs. before
- [x] Manual: configure a wrapper CLI with `cli_args_flag` and verify inner Claude flags arrive as a single quoted arg
- [ ] Manual: `cc-connect doctor` shows the configured binary name

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -35,6 +35,9 @@ func init() {
 //   - "bypassPermissions": auto-approve everything (alias: yolo)
 type Agent struct {
 	workDir          string
+	cliBin           string   // CLI binary name or path (default: "claude")
+	cliExtraArgs     []string // extra args parsed from cli_path (e.g. ["code", "-t", "foo"])
+	cliArgsFlag      string   // if set, claude args are passed as a single string via this flag (e.g. "-a")
 	model            string
 	reasoningEffort  string // "low" | "medium" | "high" | "max"
 	mode             string // "default" | "acceptEdits" | "plan" | "auto" | "bypassPermissions" | "dontAsk"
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
LGTM. cli_path 支持 Claude 兼容 wrapper（cursor、ghostty-claude 等），设计干净。cliExtraArgs 和 cliArgsFlag 参数化合理，边界处理到位。CI pass。
**建议：** 合并。

chenhg5 (approved):
LGTM. cli_path 支持 Claude-compatible wrapper 实现规范——cliBin/cliExtraArgs/cliArgsFlag 三字段分离清晰，shellJoinArgs 单引号转义逻辑正确覆盖 edge case。TestAgent_CLIBinaryName 有覆盖。CI pass。

PR #653, opened 2026-04-17
fix: normalize symlinked work_dir paths

zkunzhu:
# Symlink Work Directory Normalization

## Background

`cc-connect` uses `work_dir` as an identity boundary for:
- `/dir` runtime switching
- persisted work_dir override restore on startup
- Codex session discovery via transcript `cwd`
- local session-state file naming
- `/dir` history display

This breaks when a user switches to a symlinked directory.

Example:
- symlink path: `/path/to/symbol-link`
- real path: `/path/to/realpath`

Previously these were treated as different strings, so `/list` could miss existing Codex sessions even though `/new` still worked.

## Solution

Normalize directory-like paths to one canonical identity before comparison or persistence:
1. `filepath.Clean(path)`
2. `filepath.Abs(path)` where needed
3. `filepath.EvalSymlinks(path)` if the path exists
4. fall back to the cleaned path if resolution fails

Applied to:
- `/dir` switch and `/dir reset`
- startup restore of persisted `work_dir` overrides
- Codex transcript `cwd` matching in `/list`
- session-store filename derivation
- `/dir` history add/load behavior

For backward compatibility, session-state lookup checks both:
- the new canonical-path-based filename
- the old raw-path-based filename

Directory history also migrates old entries by canonicalizing and deduplicating persisted values on load.

## Result

After the fix:
- switching to `/path/to/symbol-link` is treated the same as `/path/to/realpath`
- Codex `/list` finds sessions recorded under the real `cwd`
- old session-state files created from raw symlink paths are still found after upgrade
- old `/dir` history entries for both symlink and real path collapse into a single canonical entry

The only user-visible change is that some UI/status output may show the resolved real path instead of the symlink alias.

### DIFF
--- a/agent/codex/list.go
+++ b/agent/codex/list.go
@@ -21,6 +21,7 @@ func listCodexSessions(workDir string) ([]core.AgentSessionInfo, error) {
 	if err != nil {
 		absWorkDir = workDir
 	}
+	absWorkDir = core.NormalizeDirPath(absWorkDir)
 
 	homeDir, err := os.UserHomeDir()
 	if err != nil {
@@ -137,6 +138,10 @@ func parseCodexSessionFile(path, filterCwd string) *core.AgentSessionInfo {
 				}
 			}
 		}
+	}
+
+	if sessionCwd != "" {
+		sessionCwd = core.NormalizeDirPath(sessionCwd)
 	}
 
 	// Filter by cwd
--- a/agent/codex/list_test.go
+++ b/agent/codex/list_test.go
@@ -1,0 +1,51 @@
+package codex
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestListCodexSessions_MatchesSymlinkedWorkDir(t *testing.T) {
+	tmpHome := t.TempDir()
+	t.Setenv("HOME", tmpHome)
+	t.Setenv("CODEX_HOME", filepath.Join(tmpHome, ".codex"))
+
+	realDir := filepath.Join(tmpHome, "real-project")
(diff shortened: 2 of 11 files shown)


chenhg5:
CI 失败 (unit-test):

agent/codex/list_test.go:41 - TestListCodexSessions_MatchesSymlinkedWorkDir
失败原因：codex CLI 不在 PATH 中，New() 返回错误后 t.Fatalf 导致测试失败。

建议修复：在创建 agent 前检查 codex 是否可用，或将 t.Fatalf 改为 t.Skip：

a, err := New(map[string]any{work_dir: linkDir})
if err != nil {
    t.Skipf(skip: codex not available: %v, err)
}

其他代码看起来没问题，lint 通过。

chenhg5 (changes requested):
CI 失败，请修复：

1. **unit-test 失败**：`TestListCodexSessions_MatchesSymlinkedWorkDir` (list_test.go:41)
   - 原因：CI 环境中没有 codex CLI，`New()` 返回错误后用 `t.Fatalf` 导致测试 FAIL
   - 修复：将 `t.Fatalf` 改为 `t.Skipf`，当 codex 不可用时跳过测试而非失败

2. **文件冲突**：`agent/claudecode/claudecode.go` 与 PR #660 冲突
   - #660（已 approved）修改了同样的 `AvailableModels` 函数，添加 `opus[1m]` 并移除版本号
   - #653 也修改了同一个文件同一函数
   - 两个 PR 需要协调，确保只保留一份修改

chenhg5:
CI unit-test is failing on the latest run. Please rebase on `main` (many PRs have been merged since April) and re-push to get a fresh CI result. The symlink normalization change looks solid — once CI is green, QA will approve.

PR #655, opened 2026-04-17
feat(display): add display_mode enum to replace boolean quiet

happyTonakai:
## Summary

Convert the legacy boolean `quiet` setting into a three-mode display system: `full`, `compact`, `quiet`.

- **full**: show thinking + tool messages as separate messages (default)
- **compact**: hide thinking/tool, each text segment sent as a separate card (same behavior as current PR #642)
- **quiet**: hide thinking/tool, all text appends to one card with `\n\n` paragraph separators between segments, done_emoji only when the card was actually streamed in-place

## Motivation

PR #642 changed quiet mode to freeze/detach preview cards at thinking/tool boundaries, causing each text segment to be sent as a separate message with a push notification. This defeats the purpose of "quiet" mode. This PR separates the behaviors into distinct modes while preserving backward compatibility.

### On commit c660aff (`fix(core): always trigger done_emoji when agent finishes`)

This commit removed the `needsDoneReaction()` gate so that `AddDoneReaction` is always called on `EventResult`. The reasoning was that previously done_emoji only fired when the preview had been updated in-place at least once, meaning "quick responses never triggered the reaction."

However, this is actually the correct original behavior. The done_emoji is specifically meant for **quiet mode with streaming updates** — where the preview card is updated in-place via `UpdateMessage` and the user only received a push for the initial `SendPreviewStart`. In that case, subsequent in-place edits are silent on Feishu, so the done emoji notifies the user that the agent has finished.

When a response is delivered as a **single message** (not streamed in-place), the user already received a push notification. Adding a done emoji on top means the user gets **two pushes for a single short response** — which is annoying rather than helpful.

This PR restores the `needsDoneReaction()` gate so done_emoji only fires when:
1. The preview card was actually updated in-place (`lastSentViaUpdate = true`)
2. The preview is still active (`previewMsgID != nil`)

| Scenario | pushes | done_emoji needed? |
|----------|--------|--------------------|
| Short single response (SendPreviewStart only) | 1 | No — user already got a push |
| Streaming response (multiple UpdateMessage) | 1 (initial) | Yes — silent updates need a finish signal |
| full/compact mode (freeze+detach each segment) | N | No — each segment already pushed |

## How it works

- New `[display] mode` field in config.toml (`"full"`, `"compact"`, or `"quiet"`)
- Legacy `quiet = true` is auto-mapped to `mode = "quiet"` for backward compatibility
- `/quiet` cycles full → quiet → compact → full; supports explicit argument: `/quiet full|compact|quiet`
- In quiet mode, `appendSeparator("\n\n")` replaces `freeze()+detachPreview()` at think/tool boundaries, keeping all text in one card

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes (no new failures)
- [x] Unit tests for `EffectiveDisplay` with 8 cases covering mode resolution, legacy quiet mapping, mode priority
- [x] Unit test for `/quiet` 3-mode cycle and explicit argument
- [x] Manual: `mode = "full"` — thinking + tool shown, each text segment separate
- [x] Manual: `mode = "compact"` — thinking/tool hidden, each text segment = separate card
- [x] Manual: `mode = "quiet"` — all text in one card with paragraph separators, done emoji at end
- [x] Manual: legacy `quiet = true` without mode — behaves as quiet
- [x] Manual: single-shot response — no done emoji


chenhg5 (approved):
LGTM. display_mode 枚举重构实现正确：

- `full/compact/quiet` 三种模式设计合理，legacy quiet 兼容性好
- `EffectiveDisplay` 正确处理 mode 解析优先级（显式 mode > legacy quiet > 默认 full）
- `/quiet` 改为三档循环，添加显式参数支持
- `SaveDisplayConfig` 正确处理 nil mode（不覆盖已有配置）
- i18n 五种语言全部更新
- CI 全绿

建议合并。

chenhg5:
@happyTonakai Please resolve the conflicts before merging, thanks.

PR #658, opened 2026-04-17
feat(shell): `!` prefix shortcut and configurable timeout

happyTonakai:
## Summary

- Add `!` prefix as a shorthand for `/shell` (e.g., `!ls -la` runs `ls -la`), gated by the same admin
permission checks
- Add `--timeout <seconds>` parameter at the **beginning** of shell commands to override the default 60s
timeout (e.g., `/shell --timeout 300 npm install`)
- Skip banned-word filtering for `!`-prefixed messages (consistent with `/` commands)

## Motivation

All supported AI agents (Claude Code, Gemini CLI, etc.) natively support `!` prefix to run shell commands in
interactive TTY mode. However, cc-connect communicates with agents via non-interactive stdin pipes, so
agents interpret `!cmd` as plain text rather than a shell command. This PR handles `!` at the cc-connect
layer to restore that functionality.

## Usage

```
!ls -la                           → runs ls -la (default 60s timeout)
! --timeout 300 npm install       → runs npm install with 300s timeout
/shell --timeout 600 docker build . → same timeout syntax for /shell
```

`--timeout` is parsed from the **beginning** of the command (not the end) to avoid conflicting with any CLI
tool's own `--timeout` flag.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes (2 pre-existing failures unrelated)
- [x] Verify `!ls` executes shell command for admin users
- [x] Verify `!` is rejected for non-admin users when `admin_from` is set
- [x] Verify `! --timeout 300 sleep 200` uses 300s timeout
- [x] Verify `!` alone (no command) falls through to agent as normal message


chenhg5 (approved):
LGTM. 两项改进都很实用：

- `!` 前缀快捷方式：对标 ChatGPT，用户可直接 `!ls` 而不用 `/shell ls`
- `--timeout` 参数：弥补了 `/shell` 无法指定超时的问题，解析逻辑健壮（只在空格处截断，无歧义）
- `!` 跳过 banned word 检查但仍做 admin 检查，设计正确
- 文档和 i18n 全部更新，CI 全绿

建议合并。

PR #664, opened 2026-04-18
feat(cron): add per-job effort override

### LINKED ISSUE #656: [Feature Request] Per-cron reasoning_effort override
## Summary

Allow overriding `reasoning_effort` on a per-cron basis in `jobs.json`, so different scheduled tasks can run with different effort levels without mutating the agent's global state.

## Motivation

Starting with Opus 4.7, Claude Code defaults to `xhigh` effort for all subscription plans. This consumes significantly more thinking tokens than prior defaults (`medium`/`high`).

Many cc-connect cron jobs are routine/structured (periodic checks, aggregations, structured extractions) and don't need `xhigh` — `medium` or `low` is sufficient and keeps rate-limit usage reasonable. Meanwhile, interactive human sessions benefit from higher effort.

Today cc-connect only supports:
1. Static project-level `reasoning_effort` in `[projects.agent.options]`
2. Runtime global `/effort` command that mutates the agent state (affects all sessions — interactive AND cron)

Neither lets a cron run with `medium` while the interactive channel stays at `xhigh`, short of defining two projects with duplicated config pointing at the same `work_dir`.

## Proposed behavior

Add an optional `effort` field to `CronJob` (`core/cron.go:22-39`):

```json
{
  "id": "example-job",
  "cron_expr": "0 * * * *",
  "session_key": "channel-id",
  "effort": "medium",
  "prompt_file": "path/to/prompt.md"
}
```

Behavior:
- If `effort` is set → pass `--effort <value>` to the spawned `claude` command for that specific run, without mutating `a.reasoningEffort`.
- If absent → fall back to the agent's current effort (project config or `/effort` runtime value), as today.

Implementation sketch: add a new `StartSessionWithEffort(effort string)` method alongside the existing `StartSession`, so the cron dispatcher can pass a one-off override without touching shared state.

## Alternatives considered

- **Multi-project workaround:** define multiple `[[projects]]` pointing to the same `work_dir` with different effort. Works, but duplicates platform config and complicates `jobs.json` assignment.
- **Prefixing prompts with `/effort`:** mutates global agent state, leaks into subsequent interactive sessions. Not isolated.

## Version

cc-connect `v1.2.2-beta.5`.

xukp20:
## Summary

This PR adds a per-cron `effort` override so scheduled jobs can run with a different `reasoning_effort` without mutating the agent's shared global state.

It extends:

- `CronJob` persistence / API / management payloads with an optional `effort`
- cron execution startup to use a new optional `SessionEffortStarter` capability
- `claudecode` and `codex` to support session-scoped effort overrides

## Why

Today `reasoning_effort` in `cc-connect` mainly comes from:

- static project/agent options
- runtime changes to the agent's shared state

That makes cron jobs awkward when they need a lower or different effort than interactive sessions. Users currently need to either:

- lower the whole agent's default effort
- or duplicate project configuration

This PR makes cron effort a job-level override instead.

## Behavior

- If a cron job has no `effort`, behavior is unchanged.
- If a cron job sets `effort`, `cc-connect` will:
  - validate it against the agent's advertised effort list when available
  - require the agent to support session-scoped effort startup
  - start that cron-run session with the requested override
- The override does **not** mutate the agent's shared default effort.

## Scope

Included:

- `CronJob.effort`
- cron management / API support
- engine support for per-run session effort override
- first-round agent support for:
  - Claude Code
  - Codex

Not included:

- changing the main `Agent` interface signature
- adding `/cron add --effort ...` chat syntax
- generic one-off effort override for normal interactive messages

## Validation

Local validation completed:

- targeted unit tests for cron store / engine / Claude / Codex
- full local CI:
  - `pnpm build`
  - `actionlint`
  - `golangci-lint --new-from-rev origin/main`
  - `go build ./...`
  - `go test ./... -v -race`
  - `go test ./... -coverprofile=coverage.out -covermode=atomic`
  - `go test -v -tags=smoke,no_web ./tests/e2e/...`
  - `go test -v -tags=regression,no_web ./tests/e2e/...`
  - `go test -bench=. -benchmem -tags=performance,no_web ./tests/performance/...`

Closes #656


core/engine.go
```diff
@@ -2378,6 +2379,35 @@ func (e *Engine) getOrCreateInteractiveStateWith(sessionKey string, p Platform,
...
+	if effortOverride == "" {
+		return agent.StartSession(e.ctx, sessionID)
+	}
+
+	if switcher, ok := agent.(ReasoningEffortSwitcher); ok {
+		if allowed := switcher.AvailableReasoningEfforts(); len(allowed) > 0 {
+			valid := false
+			for _, candidate := range allowed {
+				if strings.EqualFold(candidate, effortOverride) {
+					effortOverride = candidate
```
utafrali: If an agent implements `SessionEffortStarter` but not `ReasoningEffortSwitcher`, validation is silently skipped and any arbitrary string is forwarded to `StartSessionWithEffort`. The agent will fail at runtime on every cron run. Consider adding a comment here to make the skip explicit, or document that agents implementing `SessionEffortStarter` are responsible for validating the effort string themselves:

```go
// No AvailableReasoningEfforts to validate against; the agent is
// responsible for rejecting unknown effort values.
```

This avoids future confusion when someone adds an agent that only implements one of the two interfaces.

core/api.go
```diff
@@ -217,6 +217,7 @@ type CronAddRequest struct {
 	Silent      *bool  `json:"silent,omitempty"`
 	SessionMode string `json:"session_mode,omitempty"`
 	Mode        string `json:"mode,omitempty"`
+	Effort      string `json:"effort,omitempty"`
```
utafrali, line 221: The `effort` field on `CronAddRequest` is accepted and stored without any validation at the handler level. An invalid value (e.g. `"turbo"`) will be persisted and then cause every cron run to fail with an error. Consider validating against the agent's `AvailableReasoningEfforts()` at creation time, or at minimum returning a 400 if a non-empty effort is supplied and the agent does not implement `SessionEffortStarter`. The same concern applies to the update path in `management.go`.

core/engine_test.go
```diff
@@ -9617,6 +9652,72 @@ func TestExecuteCronJob_WorkspacePrefixedSessionKey(t *testing.T) {
...
+	}
+	if agent.overrideStarts != 1 || agent.defaultStarts != 0 {
+		t.Fatalf("defaultStarts=%d overrideStarts=%d, want 0/1", agent.defaultStarts, agent.overrideStarts)
+	}
+	if agent.lastEffort != "medium" {
+		t.Fatalf("lastEffort = %q, want medium", agent.lastEffort)
+	}
+}
+
+func TestStartAgentSessionWithOptionalEffort_InvalidOverrideFails(t *testing.T) {
```
utafrali, line 10686: The test for invalid effort passes `"xhigh"` against `["low", "medium", "high", "max"]`, which is good. Missing from this file are two complementary cases:

1. Case-insensitive match: `"MEDIUM"` should resolve to `"medium"` and succeed.
2. Whitespace trim: `"  medium  "` should succeed after `strings.TrimSpace`.

Both behaviors are implemented in `startAgentSessionWithOptionalEffort` but are untested, so a future refactor could silently break them.

core/interfaces.go
```diff
@@ -239,6 +239,13 @@ type Agent interface {
 	Stop() error
 }
 
+// SessionEffortStarter is an optional interface for agents that support
```
utafrali, line 285: The interface doc says it avoids "mutating the agent's shared global state", which is correct. It would also be useful to document the interaction with `ReasoningEffortSwitcher`: if the agent implements both, the engine will validate the effort string before calling `StartSessionWithEffort`; if only this interface is implemented, no validation occurs and the agent receives the raw string. Adding that to the godoc keeps the contract clear for future agent authors:

```go
// If the agent also implements ReasoningEffortSwitcher, the engine will
// validate effort against AvailableReasoningEfforts before calling this
// method. Otherwise the effort string is forwarded as-is.
```

utafrali (approved):
Clean, well-structured feature that follows the project's interface-first design principles. The non-mutating contract is correctly enforced and tested in both agents. The main gaps are missing API-level validation (invalid effort strings are stored and cause silent per-run failures) and a few untested code paths (case-insensitive matching, whitespace handling).

xukp20:
Addressed the review feedback and refreshed the branch onto the latest `main`.\n\nWhat was updated:\n- added API/management-layer validation for cron `effort` so invalid values are rejected before being persisted\n- documented the `SessionEffortStarter` / `ReasoningEffortSwitcher` contract more explicitly\n- added coverage for case-insensitive and whitespace-trimmed effort normalization\n- re-merged the latest upstream `main` to resolve the current merge conflict state\n\nPushed in commit `69d7b12`.

chenhg5 (changes requested):
## Review Summary

This PR cleanly adds an optional per-cron `effort` override end-to-end (persistence, management/API validation, engine session start, Claude Code + Codex). Validation is layered (API/mgmt + `validateSessionEffortOverride` at session start), `core/` does not import `agent/` or `platform/`, and backward compatibility for jobs without `effort` looks correct. CI is green.

---

🟡 **Tests: unrelated regression coverage removed**
**File**: `core/api_test.go`
**Why**: The branch drops `TestHandleSend_AllowsAttachmentOnly` from `main` while adding cron effort tests. That weakens regression coverage for `/send` with attachment-only payloads, unrelated to cron effort.
**Fix**: Restore `TestHandleSend_AllowsAttachmentOnly` (keep the new cron tests as well), or move it to another test file.

---

🟡 **Maintainability: duplicated helpers**
**Files**: `core/api.go`, `core/management.go`
**Why**: `resolveEngine` / `validateCronEffortForProject` are duplicated on `APIServer` and `ManagementServer`; they can drift.
**Fix**: Consider a small shared helper (optional follow-up).

---

💭 **Nit**: `validateSessionEffortOverride` — remove unused `_ = starter` after the type assertion.

---

✅ **Good**: Optional `SessionEffortStarter`, double validation, agent default unchanged, solid tests for cron execution and invalid/unsupported effort paths.

Request changes — please restore `TestHandleSend_AllowsAttachmentOnly` before merge.

chenhg5 (changes requested):
## Review Summary

Per-cron effort override is well-designed: SessionEffortStarter interface, double validation, CI green, no core->agent import. Backward compatible.

🟡 Tests: TestHandleSend_AllowsAttachmentOnly was removed from core/api_test.go (unrelated regression). Please restore it.
🟡 Maintainability: resolveEngine/validateCronEffortForProject duplicated in api.go and management.go (optional follow-up).
✅ Good: Optional interface, agent default unchanged, solid tests for cron/effort paths.

Request changes — please restore the deleted test before merge.

xukp20:
Addressed the latest review feedback.

Updated:
- restored `TestHandleSend_AllowsAttachmentOnly` in `core/api_test.go`
- removed the unused `_ = starter` from `validateSessionEffortOverride`

Validation rerun locally:
- focused core tests for the restored send coverage and cron effort paths
- full local CI-equivalent suite (`pnpm build`, `actionlint`, `golangci-lint`, `go build`, `go test ./... -race`, coverage, smoke/regression/performance)

Pushed in `6a3a271`.

I left the `resolveEngine` / `validateCronEffortForProject` dedup as follow-up scope, per the review note.

PR #669, opened 2026-04-18
feat(core): add builtin /steer command

xukp20:
## Summary

This PR adds a new builtin `/steer` command with a narrow, explicit meaning:

- add guidance to the current in-flight task
- keep it as the same task
- do not define it as a new task/turn command

Backend mapping in this first version:

- Claude Code:
  - send a same-session `user` message with `priority=next`
- Codex app-server:
  - call native `turn/steer`
- Codex exec:
  - not supported yet

## Why

There has been prior discussion around repurposing `/btw` or renaming it to `/ps`:

- #619
- #620
- #636

After reviewing the current `cc-connect` behavior, Claude Code semantics, Codex app-server semantics, and both CLI profile research docs, the main conclusions are:

- Claude native `/btw` is a side-question flow, not same-task steering
- Codex already uses `/ps` as a native command
- `cc-connect` still needs a clear cross-backend command for:
  - adding guidance to the currently running task

This PR intentionally introduces `/steer` as that explicit command instead of trying to settle the final `/btw` / `/ps` naming question here.

## Scope

Included:

- builtin `/steer`
- core command routing and help/i18n
- Claude Code steering via `user(priority=next)`
- Codex app-server steering via native `turn/steer`

Not included:

- removing or redefining the legacy hidden `/btw`
- adopting `/ps`
- Codex exec support
- Claude `priority=now`

## Validation

Local validation completed:

- targeted unit tests for:
  - core `/steer`
  - Claude `priority=next` message shape
  - Codex app-server `turn/steer` request shape
- full local CI:
  - `pnpm build`
  - `actionlint`
  - `golangci-lint --new-from-rev origin/main`
  - `go build ./...`
  - `go test ./... -v -race`
  - `go test ./... -coverprofile=coverage.out -covermode=atomic`
  - `go test -v -tags=smoke,no_web ./tests/e2e/...`
  - `go test -v -tags=regression,no_web ./tests/e2e/...`
  - `go test -bench=. -benchmem -tags=performance,no_web ./tests/performance/...`

Related:

- #619
- #620
- #636


chenhg5 (changes requested):
## Review Summary
/steer through optional SessionSteerer interface is well-designed — avoids tight coupling to specific agents. Claude and Codex app-server mappings with unit tests are solid, i18n and docs consistent.

---

🔴 **Blocker: corrupted shellJoinArgs comment**
**File**: `agent/claudecode/session.go` (shellJoinArgs)
**Why**: The comment demonstrating single-quote concatenation was changed from the correct `'it'\''s'` to a version with incorrect characters, misrepresenting real shell escaping and misleading future maintainers.
**Fix**: Restore the correct ASCII escape example (or match the original line exactly) to avoid introducing curly quotes / incorrect Unicode.

---

✅ **Good**: cmdSteer uses type assertion to handle unsupported backends; Codex exec scope is documented.

Request changes — please fix the shellJoinArgs comment regression.

xukp20:
Fixed. I restored the shellJoinArgs comment to the correct ASCII single-quote escape example: `'it'\''s' -> it's`.\n\nValidation run locally after the change:\n- `go test ./agent/claudecode -count=1`\n\nPushed in commit `dba1efe`.

ternary-chen:
Thanks for the careful work here. I opened #1652 as a Draft based on the current `main`.

It follows the same optional-capability and Codex `turn/steer` direction, but has a narrower and different user-facing scope: it does not add a cross-agent `/steer` command or Claude behavior. Instead, it adds opt-in `busy_message_mode = "steer"` for plain-text busy follow-ups, keeps queue as the default, falls back to the existing queue on failure, keeps attachments queued, and makes the existing `/ps` path use the steering capability safely.

There is clear overlap with the Codex portion of #669. I am linking the draft for coordination and am happy to reuse the established interface/naming, split the work, or close #1652 if maintainers prefer continuing this PR.

PR #672, opened 2026-04-18
feat(agent): add Devin CLI as a first-class agent

youngshook:
# feat(agent): add Devin CLI as a first-class agent

Adds native support for **Devin CLI** (Cognition, https://cli.devin.ai/) as a named cc-connect agent, alongside Claude Code / Codex / Cursor / Gemini / iFlow / Qoder / OpenCode.

Previously, you could connect Devin only by hand-configuring the generic ACP adapter with `type = "acp"`, `command = "devin"`, `args = ["acp"]`, `display_name = "Devin"` — and even then `/list` returned empty and `/mode` printed "not supported" because the ACP adapter had those as MVP TODOs. This PR turns that into:

```toml
[[projects]]
name = "my-project"

[projects.agent]
type = "devin"                # ← that's it

[projects.agent.options]
work_dir = "/path/to/code"
```

…and fills in `session/list` + `session/set_mode` so `/list` and `/mode` actually work in IM.

## Why dedicated `agent/devin/` instead of just a config example

Cursor Agent precedent: `agent/cursor/` exists as a first-class package *and* Cursor can be wired via `type = "acp"` for ACP-mode usage. Devin fits the same pattern — it deserves a config entry under `type = "devin"` with sensible defaults, its own row in the README support matrix, its own option in the web admin's agent dropdown, and `agent=devin` in audit logs instead of `agent=acp`.

The implementation is intentionally thin: `devin.Agent` embeds `*acp.Agent` and only shadows `Name()` to return `"devin"`. All transport, session lifecycle, mode switching, permission prompts, and tool-call plumbing are shared with `agent/acp/` — no duplicated code paths.

```go
type Agent struct{ *acp.Agent }
func (a *Agent) Name() string { return "devin" }
```

## End-to-end verification

Tested against `devin acp` 2026.4.9-0 through the cc-connect bridge with a minimal `type = "devin"` config (just `work_dir`; everything else defaults). Engine logs confirm the wrapper propagates correctly:

```
level=WARN msg="engine started with partial readiness"
  project=devin-native-test  agent=devin  ready=0 pending=1 failed=0
                             ^^^^^^^^^^^ 
```

All `/mode` and `/list` paths work:

```
/mode        →  **Permission Mode**
                 ▶ **Code** — 
                 ◻ **Ask** — 
                 ◻ **Plan** — 
                 ◻ **Accept Edits** — 
                 ◻ **Bypass Permissions** — 
                 Use /mode <name> to switch.
[...]

PR #675, opened 2026-04-19
fix(core/markdown_html): render inline formatting inside GFM table cells

youngshook:
# fix(core/markdown_html): render inline formatting inside GFM table cells

## Symptom

When an agent reply contained a Markdown table like:

```
| Skill    | 用途                           |
|----------|--------------------------------|
| **hunt** | 调试错误、崩溃、异常行为       |
| **think**| 新功能/架构决策前的思考和规划  |
```

the Telegram user saw the **literal** `**hunt**` characters inside the rendered table, not a bold "hunt". The bug was silent — Telegram accepted the HTML, there was no `can't parse` fallback, and nothing in the logs indicated anything was wrong.

Screenshot reproduction is easy: ask Devin / any agent for a bullet-or-table summary of anything where cells contain bold labels. You'll see `**...**` markers intact in the rendered table.

## Root cause

`flushTable` in `core/markdown_html.go` wrote each cell through `escapeHTML` only, **never** through `convertInlineHTML`. Every other Markdown block in the file routes text through `convertInlineHTML` to pick up `**bold**`, `*italic*`, `` `code` ``, `~~strike~~`, `[link](url)`, etc. — table cells opted out, with a comment in the adjacent test claiming:

> `Inline formatting is escaped inside <pre> since HTML tags in <pre> render literally in Telegram`

That assumption is **wrong**. Telegram's HTML grammar documents that `<b>`, `<strong>`, `<i>`, `<em>`, `<u>`, `<ins>`, `<s>`, `<strike>`, `<del>`, `<code>`, `<a>` are permitted inside `<pre>`; only nested `<pre>` / `<code>` blocks are rejected. Verified by live Telegram Bot API calls:

```
<pre><b>hunt</b>   | debug errors</pre>   →  ✅ accepted, bold renders
<pre><i>hunt</i>   | debug errors</pre>   →  ✅ accepted, italic renders
<pre><code>hunt</code> | ...</pre>        →  ✅ accepted, monospace renders
```

So the "cautious" old behavior was strictly worse than the spec allowed.

## Fix

Run each table cell through `convertInlineHTML` (the same function the rest of the converter uses), and compute column widths from the **visual** rune length — i.e. with Markdown markers stripped — rather than from raw byte length. The new `tableCellVisualWidth` helper strips `**x**`, `__x__`, `*x*`, `` `x` ``, `~~x~~`, `[text](url)` before counting, so `**hunt**` measures as 4 columns wide (for "hunt"), not 8.

```go
// before
b.WriteString(escapeHTML(cell))
if pad := colWidths[k] - len(cell); pad > 0 { ... }

// after
b.WriteString(convertInlineHTML(cell))
if pad := colWidths[k] - tableCellVisualWidth(cell); pad > 0 { ... }
```

As a nice side effect, CJK columns also align better: the old code used `len(cell)` (UTF-8 bytes), so 4 Chinese characters were counted as 12 "columns" for padding, producing misaligned tables. The new code uses `utf8.RuneCount` (runes), which counts them as 4 — still not a perfect East-Asian-width match (2 columns each monospace) but strictly better than the byte-count status quo. A width-table-aware helper can replace this later without touching the `flushTable` call site.

## Before / After

Same Markdown input:

```
| Skill | 用途 |
|-------|------|
| **hunt** | 调试错误、崩溃、异常行为 |
| **think** | 新功能/架构决策前的思考和规划 |
```

**Before (what the user was seeing in Telegram):**

```
Skill      | 用途                                                         
-----------+----------------------------------------------------------------
**hunt**   | 调试错误、崩溃、异常行为
**think**  | 新功能/架构决策前的思考和规划
```

**After (HTML sent to Telegram):**

```html
<pre>Skill  | 用途             
-------+----------------
<b>hunt</b>   | 调试错误、崩溃、异常行为
<b>think</b>  | 新功能/架构决策前的思考和规划</pre>
```

Telegram renders `<b>hunt</b>` as a bold "hunt" inside the monospaced `<pre>` block — exactly what the user expected.

## Tests

**Replaced** `TestMarkdownToSimpleHTML_TableWithFormatting` — the previous assertions (`contains "Header"`, `contains "code"`) were loose enough to pass both the buggy and fixed paths. The new test asserts `<b>Header</b>`, `<i>italic</i>`, `<code>code</code>` are present and the literal Markdown markers are **absent**.

**New** `TestMarkdownToSimpleHTML_TableCellAlignmentWithFormatting` — locks in the column-width regression: separator row must match stripped cell widths, body rows must pad to the same visual width regardless of markers.

**New** `TestMarkdownToSimpleHTML_TableCellWithLink` — verifies that `[text](url)` in cells becomes a real clickable `<a>` tag inside `<pre>`.

**New** `TestTableCellVisualWidth` — table-driven coverage of the width helper.

Existing `TestMarkdownToSimpleHTML_Table` and 40+ other `MarkdownToSimpleHTML_*` tests are unchanged and still pass.

```
$ go test ./core/... -count=1 -run 'MarkdownToSimpleHTML|TableCellVisualWidth'
ok   github.com/chenhg5/cc-connect/core   0.534s
```

## Verification against live Telegram

The real Devin reply that triggered the user report (3 326 bytes of Markdown containing two GFM tables, headings, code fences, links) was extracted from the Devin session DB and pushed through the fixed converter to the user's real Telegram bot via the Bot API:

```
markdown len=3374, html len=4343
sendMessage parse_mode=HTML → ✅ accepted
```

Bold, italic, inline code, and link tags inside `<pre>` table cells all render correctly in Telegram's iOS / macOS / web clients.

## Change summary

```
 core/markdown_html.go      | 64 ++++++++++++++++++++++++++++++++----
 core/markdown_html_test.go | 81 +++++++++++++++++++++++++++++++++++++++++++---
 2 files changed, 134 insertions(+), 11 deletions(-)
```

No breaking changes. No dependency changes. Behavior is strictly additive: tables that had no Markdown markers in cells render identically; tables that did have markers now render them correctly.

## Note on PR stacking

This PR is **independent** of #672 (Devin CLI agent) and #674 (Telegram HTML fallback observability). It can merge in any order. The bug exists on `main` and affects every platform that consumes `core.MarkdownToSimpleHTML` output — not Telegram-specific. Agents shipping GFM tables with bold/italic cells (which most do, including Claude Code, Codex, Devin, Cursor, etc.) are all affected.

### DIFF
--- a/core/markdown_html.go
+++ b/core/markdown_html.go
@@ -3,6 +3,7 @@ package core
 import (
 	"regexp"
 	"strings"
+	"unicode/utf8"
 )
 
 // MarkdownToSimpleHTML converts common Markdown to a simplified HTML subset.
@@ -57,6 +58,13 @@ func MarkdownToSimpleHTML(md string) string {
 	}
 
 	// flushTable renders buffered table rows inside a <pre> block with aligned columns.
+	//
+	// Inline formatting in cells (bold/italic/inline-code/strikethrough/links)
+	// is rendered as Telegram HTML tags; Telegram permits <b>, <i>, <u>, <s>,
+	// <code>, <a> inside <pre>, so `**foo**` becomes a bold "foo" rather than
+	// four literal asterisks. Column widths are computed from the *visual*
+	// (post-strip) rune length so that ` | ` separators still line up even
+	// though the rendered HTML bytes are longer than the plain text.
 	flushTable := func() {
 		if len(tblLines) == 0 {
 			return
@@ -85,7 +93,9 @@ func MarkdownToSimpleHTML(md string) string {
 			rows = append(rows, row{cells: cells})
 		}
 
-		// Compute max width per column.
+		// Compute max width per column using the visual rune length of each
+		// cell (markdown markers stripped). This keeps ASCII columns aligned
+		// even after `**x**` expands to `<b>x</b>` in the rendered output.
 		numCols := 0
 		for _, r := range rows {
 			if !r.isSep && len(r.cells) > numCols {
@@ -98,8 +108,10 @@ func MarkdownToSimpleHTML(md string) string {
 				continue
 			}
 			for k, c := range r.cells {
-				if k < numCols && len(c) > colWidths[k] {
-					colWidths[k] = len(c)
+				if k < numCols {
+					if w := tableCellVisualWidth(c); w > colWidths[k] {
+						colWidths[k] = w
+					}
 				}
 			}
 		}
@@ -129,9 +141,13 @@ func MarkdownToSimpleHTML(md string) string {
 					if k < len(r.cells) {
 						cell = r.cells[k]
 					}
-					b.WriteString(escapeHTML(cell))
-					// Pad to column width.
-					if pad := colWidths[k] - len(cell); pad > 0 {
+					// Render inline formatting to HTML tags (Telegram accepts
+					// <b>/<i>/<code>/<a>/etc. inside <pre>). Falls back to
+					// plain HTML-escaped text when there is no formatting.
+					b.WriteString(convertInlineHTML(cell))
+					// Pad to column width using the *visual* length so the
+					// `|` separators still line up in the rendered message.
+					if pad := colWidths[k] - tableCellVisualWidth(cell); pad > 0 {
 						b.WriteString(strings.Repeat(" ", pad))
 					}
 				}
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## Review Summary
GFM table cells now go through convertInlineHTML with tableCellVisualWidth for column alignment. Root cause and fix path are clear; tests upgraded from substring checks to concrete HTML assertions with alignment and link coverage.

---

✅ **Good**: Documented Telegram HTML behavior in `<pre>` for `<b>/<a>` etc; visual width helper boundary is well-scoped.

LGTM — approve.

chenhg5 (approved):
LGTM — CI green, conflict-free, implementation reviewed. Approved for merge.

PR #677, opened 2026-04-19
fix(kimi): remove CLI dependency from unit tests

### LINKED ISSUE #647: test(agent/kimi): remove hard local dependency on installed kimi CLI
## Problem

The Kimi unit tests currently assume that a `kimi` executable is already installed in the local environment.

After `feat(agent): add Kimi CLI support (#616)` (commit `006718a`), these tests now fail in local full-CI runs on machines that do not have Kimi CLI installed, even though the tests themselves do not need the real CLI implementation.

Typical failures are in:
- `TestAgentNew`
- `TestAgentSetters`
- `TestAgentPermissionModes`
- `TestAgentProviderSwitcher`
- `TestAgentStartSession`
- `TestAgentMemoryAndSkill`
- `TestAgentAvailableModels`

The failure is:

```text
kimi: "kimi" CLI not found in PATH, install with: pip install kimi-cli
```

## Why this is unreasonable

These are unit tests, but they currently depend on a local external binary being preinstalled.

That makes conflict recovery / local full CI verification for unrelated PRs unnecessarily fragile, because every branch now has to provide a real or fake `kimi` command just to get through `go test ./... -race`.

## Suggested fix

One of these would be much better:

1. Make the tests create/use a temporary fake `kimi` executable.
2. Or skip the tests when `kimi` is not available in PATH.
3. Or move the real-CLI assumption into an explicit integration test instead of the default unit-test path.

I think option 1 is the cleanest for current coverage.

chenhg5:
## Summary
- Construct Agent structs directly in unit tests instead of calling New(), which requires the kimi CLI binary on PATH
- Tests for setters, permission modes, providers, memory/skill dirs, and available models now run without the CLI
- Only TestAgentNew and TestAgentStartSession still require the CLI (guarded by skipUnlessKimiAvailable)
- Added TestAgentFields for CLI-independent Name/WorkDir/Mode/Model verification

## Test plan
- [x] go build ./... passes
- [x] go vet ./... passes
- [x] go test ./... passes (all 28 packages OK)
- [x] go test ./agent/kimi/ -v — 6 previously-skipped tests now PASS without kimi CLI

Closes #647

### DIFF
--- a/agent/kimi/kimi_test.go
+++ b/agent/kimi/kimi_test.go
@@ -59,11 +59,26 @@ func TestAgentNew(t *testing.T) {
 	assert.Equal(t, "kimi-k2", a.GetModel())
 }
 
+// TestAgentFields verifies Name/WorkDir/Mode/Model without requiring
+// the kimi CLI on PATH — constructs the struct directly.
+func TestAgentFields(t *testing.T) {
+	a := &Agent{
+		workDir:   "/tmp",
+		model:     "kimi-k2",
+		mode:      "yolo",
+		cmd:       "kimi",
+		activeIdx: -1,
+	}
+	assert.Equal(t, "kimi", a.Name())
+	assert.Equal(t, "Kimi", a.CLIDisplayName())
+	assert.Equal(t, "kimi", a.CLIBinaryName())
+	assert.Equal(t, "/tmp", a.GetWorkDir())
+	assert.Equal(t, "yolo", a.GetMode())
+	assert.Equal(t, "kimi-k2", a.GetModel())
+}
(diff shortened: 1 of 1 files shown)


chenhg5 (commented):
## QA Review — LGTM ✅

Replaces New()-based setup with direct Agent literals for tests so they run without the kimi binary on PATH. Clear split between CLI-required and CLI-optional tests.

---

✅ **Good**: Adds TestAgentFields for CLI-free verification of Name/WorkDir/Mode/Model; improves CI and local ergonomics without changing production code paths.

PR #682, opened 2026-04-19
feat(core): suppress agent replies marked with NO_REPLY

### LINKED ISSUE #681: Feature request: NO_REPLY marker for agent-initiated silent replies
### Problem

In group-chat scenarios — or any scenario where an agent receives a trigger that turns out to warrant no response (e.g. a scheduled cron found nothing worth reporting, an incoming message was clearly directed at another participant, an acknowledgement needs no reaction) — the agent currently has no way to stay silent. Any text it produces is delivered verbatim to the platform, so the agent is forced to reply even when the best action is to say nothing.

Working around this by prompt-engineering the agent to return an empty string does not work reliably: Claude-Code-style agents default to producing user-visible prose, and tool harnesses (CCD, OpenClaw) encourage reasoning narration.

### Proposal

Introduce a `NO_REPLY` marker convention:

- If the agent's cleaned response is exactly `NO_REPLY` (case-insensitive, surrounding whitespace allowed) → suppress platform delivery entirely (no preview, no done emoji, no TTS, no reply footer, no context indicator). History is still written so the agent retains the knowledge of its own decision.
- If the response ends with a trailing `NO_REPLY` marker after some reasoning → strip the marker and deliver the reasoning. (Matches how OpenClaw handles the same marker, and matches real-world agent behaviour in longer turns where the model narrates before concluding "…and this turn should stay silent. NO_REPLY".)
- The injected agent system prompt documents the marker so agents know it is available.

### Prior art

OpenClaw (`openclaw` npm package) ships the same convention. Moving it into `cc-connect` upstream would let Claude-Code-based agents gate their own noise in the same way.

### Environment

- cc-connect v1.3.0-rc.5
- macOS / Lark (but the logic is platform-agnostic — lives entirely in `core/engine.go`)

Cigarrr:
Closes #681

### Summary

Adds a `NO_REPLY` marker convention that lets the agent request silent handling of a turn:

1. **Bare marker** (`isSilentReply`): response is exactly `NO_REPLY` (case-insensitive, surrounding whitespace allowed) → full suppression (no send, no preview, no done emoji, no TTS, no reply footer, no context indicator, no `message.sent` hook emit). History is still written so the agent retains the decision context.
2. **Trailing marker on mixed content** (`stripTrailingSilent`): matches `(?:^|\s+|\*+)NO_REPLY\s*$` → strip and deliver the preceding reasoning. If strip result is empty, treat as bare. This matches the OpenClaw semantic.
3. **Streaming prefix hold** (`couldBeSilentPrefix`): while the first segment chunks are still a prefix of `"NO_REPLY"`, hold the preview flush so the platform never shows a leaked `N` / `NO` / `NO_`. Once the segment can no longer resolve to the bare marker, flush all held chunks at once.
4. **System prompt**: `AgentSystemPrompt()` documents the marker so the agent knows it is available and how the strip semantic works.

`silentHold` is reset at every segment boundary (thinking / tool / user events) so the hold only applies to a fresh segment.

### Test plan

- [x] `go build -tags 'no_web' ./...`
- [x] `go test -tags 'no_web' ./...` — 3 pre-existing test failures (`TestProcessInteractiveEvents_AppendsReplyFooterWhenEnabled`, `TestProcessInteractiveEvents_ReplyFooterPrefersSessionRuntimeState`, `TestResolveLocalDirPath_AcceptsSubdir`) present on upstream/main unchanged; no new failures introduced.
- [x] Live test on Lark: bare `NO_REPLY` response produces no message, no preview, no emoji, no TTS.
- [x] Live test on Lark: reasoning + trailing `NO_REPLY` delivers the reasoning cleanly (marker stripped).
- [x] Live test on Lark: streaming `N`/`NO`/`NO_` prefix does not leak into preview.

### Notes

- Platform-agnostic: all logic lives in `core/engine.go` and `core/interfaces.go`. No platform adapter changes.
- Does not depend on any unmerged PR.


chenhg5 (approved):
## Review Summary
This PR implements NO_REPLY end-to-end in core: bare marker suppresses user-visible delivery and hooks, trailing marker is stripped with OpenClaw-aligned semantics, history keeps the full assistant text before outbound rewriting, and streaming uses a prefix hold so previews do not leak partial "NO_REPLY" tokens. CI is green. No core import-rule violations.

---

✅ **Good**: Silent path skips HookEventMessageSent and TTS; AddHistory happens before rewriting so context stays faithful; streaming hold is the right abstraction for preview-safe behavior.

---

🟡 **Suggestion**: Add table-driven unit tests for isSilentReply, stripTrailingSilent, and couldBeSilentPrefix — PR #684 added tests for a simpler strip helper; porting that coverage here would lock the contract.

---

💭 **Nit**: Coordinate with PR #684 so only one implementation merges to main to avoid duplicate/conflicting changes in engine.go and interfaces.go.

LGTM — approve (recommend merging this over #684 due to more complete behavior).

Cigarrr:
Addressed the table-driven test suggestion in e1c0bac. New cases cover all three helpers:

- `TestIsSilentReply` (13 cases) — exact/lowercase/mixed-case markers, surrounding whitespace/newlines/tabs, vs. empty/mixed-content/suffix-token/partial.
- `TestStripTrailingSilent` (13 cases) — trailing after newline/space/CRLF/trailing whitespace, markdown asterisks, no-marker/mid-string/suffixed/glued negatives.
- `TestCouldBeSilentPrefix` (16 cases) — case-insensitive prefix walk from empty through full marker, whitespace tolerance, negatives for non-N start / wrong letters / overlong / divergent tokens.

All 42 sub-cases pass via `go test ./core/ -run 'TestIsSilentReply|TestStripTrailingSilent|TestCouldBeSilentPrefix' -v`.

chenhg5:
@Cigarrr Please resolve the conflicts before merging

PR #684, opened 2026-04-19
feat(core): add NO_REPLY marker to suppress platform delivery

### LINKED ISSUE #681: Feature request: NO_REPLY marker for agent-initiated silent replies

chenhg5:
## Summary
- Add `NO_REPLY` marker support: when an agent responds with exactly "NO_REPLY" (case-insensitive), platform delivery is suppressed entirely (no message sent, no TTS, no done reaction)
- When "NO_REPLY" appears as a trailing suffix, the marker is stripped and preceding content is delivered normally
- History is always recorded so the agent retains context of the interaction
- Document the marker in `AgentSystemPrompt()` so all agents are aware of the capability

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] Unit tests for `stripTrailingNoReply` cover exact match, trailing newline, trailing whitespace, and negative cases

Closes #681

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -47,6 +47,35 @@ const (
 	replyFooterUsageCacheTTL = 30 * time.Second
 )
 
+// noReplyMarker — when an agent's final response is exactly this string
+// (case-insensitive, trimmed), platform delivery is suppressed. When it
+// appears as a trailing suffix, only the marker is stripped and the preceding
+// content is delivered normally. History is always recorded.
+const noReplyMarker = "NO_REPLY"
+
+// stripTrailingNoReply checks if s ends with the NO_REPLY marker (on its own
+// line or after whitespace) and returns the stripped content if found.
+func stripTrailingNoReply(s string) (string, bool) {
+	trimmed := strings.TrimRight(s, " \t")
+	for _, suffix := range []string{"\n" + noReplyMarker, "\n" + strings.ToLower(noReplyMarker)} {
+		if strings.HasSuffix(trimmed, suffix) {
+			return strings.TrimRight(trimmed[:len(trimmed)-len(suffix)], "\n "), true
+		}
+	}
+	// Also match when trailing on same line after whitespace.
+	lower := strings.ToLower(trimmed)
+	if strings.HasSuffix(lower, strings.ToLower(noReplyMarker)) {
+		before := trimmed[:len(trimmed)-len(noReplyMarker)]
+		if before == "" {
+			return "", false // exact match handled separately
+		}
+		if last := before[len(before)-1]; last == ' ' || last == '\t' || last == '\n' {
+			return strings.TrimRight(before, " \t\n"), true
+		}
+	}
+	return "", false
+}
+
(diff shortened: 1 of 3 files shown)


chenhg5 (commented):
## QA Review — Request Changes

Thanks for the tests and the clear stripTrailingNoReply cases. The noReply branch correctly skips platform send, TTS, and done reaction via sp.discard(). However, there are issues that need addressing:

---

🔴 **Blocker: HookEventMessageSent fires when noReply is true**
**File**: `core/engine.go`
**Why**: hooks.Emit(HookEventMessageSent, ...) executes between AddHistory and the noReply branch, so integrations that listen for message.sent will see a "sent" event when nothing was actually delivered — inconsistent with full suppression.
**Fix**: Gate hooks.Emit on !noReply, or reorder so hooks only fire when content is actually sent.

---

🟡 **Overlap with PR #682**: PR #682 (Cigarrr) implements the same NO_REPLY feature with additional streaming preview suppression and correct hook ordering. These two PRs conflict on the same code regions. Recommend consolidating — either pick one implementation or merge the best parts.

---

✅ **Good**: Unit tests for stripTrailingNoReply; CI green; core-only changes respect the core/ import boundary.

chenhg5:
Closing in favor of #682 which has been merged into main. PR #682 provides a more complete implementation with streaming preview suppression, correct hook ordering, and proper done-reaction gating. Thanks for the contribution!

PR #689, opened 2026-04-19
feat(core): make per-session message queue depth configurable

### LINKED ISSUE #688: Per-session message queue depth is hardcoded at 5 and surfaces misleading error when full
## Environment

- Version: `cc-connect 1.3.0` (also reproduces on earlier 1.3.x)
- OS: macOS 15 (Darwin 25.4.0)
- Install: npm (`@chenhg5/cc-connect`)
- Agent: claudecode
- Platform: Lark / Feishu

## Problem

Since [#198](https://github.com/chenhg5/cc-connect/pull/198), when the agent is busy with the current turn, additional user messages are queued in `interactiveState.pendingMessages` up to a hardcoded cap of 5 (`maxQueuedMessages` in `core/engine.go`). Once that cap is exceeded, further messages are silently rejected with `MsgPreviousProcessing` ("previous request still processing") — the same reply used when no queue path is available at all.

Two concrete issues:

1. **Cap is too small for chatty / long-running use.** A user who dumps several follow-ups while the agent is working on a long-running turn (large refactor, batch of tool calls) easily trips the limit. There is no way to raise it without recompiling.
2. **The reply is misleading.** When the cap is hit, the mechanism in play is queueing (the queue is full), not the "previous request still processing" path. Users reasonably assume their message was received and queued and wait, when it was actually rejected.

## Proposed fix

- Make the cap configurable via `[queue] max_depth` in `config.toml`. Default stays at 5 so behavior is unchanged out of the box.
- When the queue is specifically full, reply with a new `MsgQueueFull` message ("Message queue is full — please try again shortly.") in all 5 supported languages. `MsgPreviousProcessing` still applies when the session is busy AND the queue path is unavailable (e.g. session dead, non-claudecode agent).

Happy to open the PR — it's a small change: new `QueueConfig` struct, one engine setter, one wiring site in `cmd/cc-connect/main.go`, and an i18n key.

Cigarrr:
Closes #688.

## Summary

- Expose the per-session message queue cap as `[queue] max_depth` in `config.toml`. Default remains 5.
- Introduce a distinct `MsgQueueFull` reply so users see "Message queue is full — please try again shortly." when their message is actually rejected due to queue overflow, instead of the misleading "previous request still processing" used for session-busy-with-no-queue paths.

## Changes

| File | Change |
|------|--------|
| `config/config.go` | Add `QueueConfig{ MaxDepth *int }` and wire into `Config` |
| `config.example.toml` | Add commented `[queue]` example |
| `core/engine.go` | Rename `maxQueuedMessages` const → `defaultMaxQueuedMessages`, add `maxQueuedMessages int` field + `SetMaxQueuedMessages(n int)`, swap the overflow reply from `MsgPreviousProcessing` → `MsgQueueFull` |
| `core/i18n.go` | Add `MsgQueueFull` key with EN / ZH / ZH-TW / JA / ES translations |
| `cmd/cc-connect/main.go` | Call `engine.SetMaxQueuedMessages(*cfg.Queue.MaxDepth)` when configured |
| `core/engine_test.go` | Rename references to the renamed const (behavior unchanged) |

`MaxDepth` uses `*int` so `0`/missing is treated as "keep default", and negative values are no-ops (guarded in `SetMaxQueuedMessages`).

## Backward compatibility

No behavior change by default. Existing configs that don't set `[queue]` see the same cap of 5. Only the overflow reply message changes, and only when the queue is actually the bottleneck.

## Test plan

- [x] `go test ./... -tags no_web` — all packages pass (three pre-existing failures in `core` under `TestProcessInteractiveEvents_AppendsReplyFooterWhenEnabled`, `TestProcessInteractiveEvents_ReplyFooterPrefersSessionRuntimeState`, `TestResolveLocalDirPath_AcceptsSubdir` are unrelated and reproduce on plain `main`).
- [x] Manual verification on Lark: with `[queue] max_depth = 20`, can queue >5 messages during a long-running turn; hitting the cap now replies with `MsgQueueFull`.

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -426,6 +426,10 @@ func main() {
 				})
 			}
 		}
+		// Wire per-session message queue depth
+		if cfg.Queue.MaxDepth != nil && *cfg.Queue.MaxDepth > 0 {
+			engine.SetMaxQueuedMessages(*cfg.Queue.MaxDepth)
+		}
 		// Wire outgoing rate limiting
 		{
 			var maxPS float64
--- a/config.example.toml
+++ b/config.example.toml
@@ -174,6 +174,18 @@ level = "info" # debug, info, warn, error
 # window_secs = 60          # Window size in seconds (default: 60) / 窗口时间秒数（默认 60）
 
 # =============================================================================
+# Message Queue / 消息队列
+# When the agent is busy with the current turn, additional incoming messages
+# wait in a per-session queue and are replayed after the turn ends. When the
+# queue is full, further senders receive a "queue full" reply instead of being
+# queued. Increase max_depth for chatty users / long-running turns.
+# 当 agent 正在处理当前轮次时，新消息进入每会话队列，本轮结束后依次回放。
+# 队列满时新发送者会收到"队列已满"提示。聊天频繁或单轮较长时可调大 max_depth。
+# =============================================================================
+# [queue]
+# max_depth = 5             # Max queued messages per session (default: 5) / 每会话最大排队消息数（默认 5）
+
+# =============================================================================
 # Outgoing Rate Limit / 出站速率限制
 # Throttle how fast messages are sent TO platforms.
 # Prevents account bans on platforms with strict API rate limits (e.g. WeChat Work).
--- a/config/config.go
+++ b/config/config.go
@@ -102,6 +102,7 @@ type Config struct {
 	StreamPreview     StreamPreviewConfig     `toml:"stream_preview"`      // real-time streaming preview
 	RateLimit         RateLimitConfig         `toml:"rate_limit"`          // per-session rate limiting
 	OutgoingRateLimit OutgoingRateLimitConfig `toml:"outgoing_rate_limit"` // outgoing message throttling
+	Queue             QueueConfig             `toml:"queue"`               // per-session message queue depth
 	Relay             RelayConfig             `toml:"relay"`               // bot-to-bot relay behavior
 	Cron              CronConfig              `toml:"cron"`
 	Webhook           WebhookConfig           `toml:"webhook"`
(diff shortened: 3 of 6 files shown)


chenhg5 (changes requested):
## Review Summary

This PR cleanly adds `[queue] max_depth` with a sensible default (5), wires it from `main` via `SetMaxQueuedMessages`, and introduces `MsgQueueFull` with full i18n coverage. `core/` does not import `agent/` or `platform/`. Backward compatibility for configs omitting `[queue]` looks good.

---

🟡 **[Correctness / UX]: Queue-full message shown for all `queueMessageForBusySession` failures**

**File**: `core/engine.go` (busy-session branch after `TryLock` fails)

**Why**: `queueMessageForBusySession` returns `false` not only when the queue is at capacity, but also when there is no `interactiveState` (`!hasState || state == nil`) or when the agent session is no longer alive. The PR changes the single fallback reply from `MsgPreviousProcessing` to `MsgQueueFull` for **every** `false`, so users can see "message queue is full" when the problem is not queue capacity.

**Fix**: Only use `MsgQueueFull` when the rejection is due to `len(pendingMessages) >= cap` (e.g. send that reply inside the queue-full branch and return `true` there, or return a reason so `handleMessage` can choose the right message key).

---

🟡 **[Tests]: No coverage for configurable depth**

**Why**: Tests were updated for the renamed default constant only. A short test that sets `SetMaxQueuedMessages(n)` and asserts the `(n+1)`-th message while busy is rejected would lock in the new behavior.

**Fix**: Add a focused test in `core/engine_test.go`.

---

✅ **Good**: Clear `QueueConfig` docs, `*int` for "use default", guard for `n <= 0`, `config.example.toml` guidance, and five-language strings for `MsgQueueFull`.

---

Request changes — please address the mistaken user-facing message for non-queue-full failure paths and add a depth test.

Cigarrr:
Thanks for the review — both points addressed in d7faf44.

**1. Wrong reply for non-queue-full failures**
`queueMessageForBusySession` now returns a typed `queueResult`:
- `queueResultOK` — queued successfully
- `queueResultNoState` — no interactiveState for the key
- `queueResultSessionDead` — agent session existed but is no longer alive
- `queueResultFull` — `len(pendingMessages) >= maxQueuedMessages`

`handleMessage` replies with `MsgQueueFull` **only** on `queueResultFull`; `queueResultNoState` / `queueResultSessionDead` fall back to `MsgPreviousProcessing`, matching the pre-PR behavior for those paths.

**2. Test for configurable depth**
Added `TestQueueMessage_SetMaxQueuedMessages_OverridesDefault`: sets depth to 2, queues 2 messages (both `queueResultOK`), asserts the 3rd returns `queueResultFull`. Also added `TestSetMaxQueuedMessages_IgnoresNonPositive` to lock in the `n <= 0` guard.

Existing queue tests were updated to the new return type; no behavior change for the success path.

`go test -tags no_web ./core/ -run 'TestQueueMessage|TestSetMaxQueuedMessages'` — all 7 pass. The three pre-existing unrelated failures noted in the PR body still reproduce on plain `main`.

Cigarrr:
depreciated by pr#690.

PR #690, opened 2026-04-19
feat(core): make message queue depth configurable

### LINKED ISSUE #688: Per-session message queue depth is hardcoded at 5 and surfaces misleading error when full

chenhg5:
## Summary
- Add `[queue] max_depth` config option to control per-session message queue depth (default: 5)
- Replace misleading `MsgPreviousProcessing` with new `MsgQueueFull` message when queue is full, showing current pending count
- Add `Engine.SetMaxQueuedMessages()` setter and wire through `cmd/cc-connect/main.go`
- Add `MsgQueueFull` i18n translations for all 5 languages (EN, ZH, ZH-TW, JA, ES)

### Config example
```toml
[queue]
max_depth = 10
```

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] Existing queue overflow test updated for new behavior

Closes #688

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,7 +27,13 @@ npm install -g cc-connect
 After installation, the `cc-connect` binary will be available globally.
 
 
-### Option B: Download binary from GitHub Releases
+### Option B: Homebrew (macOS / Linux)
+
+```bash
+brew install cc-connect
+```
+
+### Option C: Download binary from GitHub Releases
 
 Go to https://github.com/chenhg5/cc-connect/releases and download the binary for your platform.
 
@@ -50,7 +56,7 @@ On macOS, you may need to remove the quarantine attribute:
 xattr -d com.apple.quarantine cc-connect
 ```
 
-### Option C: Build from source
+### Option D: Build from source
 
 Requires Go 1.22+.
 
--- a/README.md
+++ b/README.md
@@ -200,9 +200,13 @@ Follow https://raw.githubusercontent.com/chenhg5/cc-connect/refs/heads/main/INST
 **Via npm:**
 
 ```bash
-# Stable version
 npm install -g cc-connect
+```
 
+**Via Homebrew (macOS / Linux):**
+
+```bash
+brew install cc-connect
 ```
 
 **Download binary from [GitHub Releases](https://github.com/chenhg5/cc-connect/releases):**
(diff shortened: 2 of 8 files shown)


chenhg5 (commented):
## QA Review — Almost LGTM

MsgQueueFull correctly scoped to queue-at-max-depth branch only. Fixes the semantic issue from #689. Config wiring and i18n look good.

🟡 Tests: please add a test with SetMaxQueuedMessages(2) that asserts overflow at N+1.
🟡 Docs: queueMessageForBusySession comment and test name no longer match behavior.
✅ Good: proper scoping, backward-compatible default.

Note: #690 and #689 both address #688. Recommend closing #689 in favor of this once test is added.

PR #694, opened 2026-04-20
feat: add env option to agent config for injecting environment variables

silingyuan0:
## Summary

- Add `[projects.agent.options.env]` support in `config.toml` to inject arbitrary environment variables into all agent CLI subprocesses at startup
- Previously only the ACP agent supported this via a local `envPairsFromOpts` helper; now all 10 agents support it
- Add shared `core.AgentEnvFromOpts()` helper, replacing ACP's local copy

## Changes

- **`core/agent_env.go`** (new): shared helper to parse `opts["env"]` (supports `map[string]string` and `map[string]any`)
- **`core/agent_env_test.go`** (new): tests for nil/empty, both map types, order consistency
- **All 10 agents**: add `agentEnv` field, parse in `New()`, append to `extraEnv` in `StartSession()`
  - claudecode, codex, cursor, gemini, iflow, opencode, qoder, kimi, pi, acp
- **`agent/acp/agent.go`**: replace local `envPairsFromOpts` with `core.AgentEnvFromOpts`
- **`config.example.toml`**: add `env` option documentation with proxy example

## Usage

```toml
[projects.agent.options]
work_dir = "/path/to/project"

[projects.agent.options.env]
HTTP_PROXY = "http://127.0.0.1:7890"
HTTPS_PROXY = "http://127.0.0.1:7890"
MY_CUSTOM_VAR = "value"
```

## Test plan

- [x] `go build ./...` — compiles
- [x] `go test ./...` — all tests pass
- [x] `core/agent_env_test.go` covers nil/empty, `map[string]string`, `map[string]any`, order consistency


chenhg5 (approved):
## Review Summary

Centralizing opts["env"] in core.AgentEnvFromOpts and wiring it through agents keeps behavior consistent with the former ACP helper. Tests in core/agent_env_test.go cover the important shapes. core/ does not import agent/ or platform/ — architecture rule satisfied.

---

🟡 **Suggestion**: fmt.Sprintf with map[string]any values is fine for string-like TOML values; document that env values should be strings for clarity.

---

✅ **Good**: Shared helper + example.toml make env injection discoverable; ACP duplication removed cleanly.

LGTM — approve.

PR #696, opened 2026-04-20
fix(feishu): skip bot open_id fetch in webhook mode for private deplo…

zhaomoran:
…yments

In webhook mode with private/self-hosted Feishu deployments, the bot open_id API call may fail due to different auth flows or API support. This failure previously prevented the webhook server from starting.

This change defers bot open_id discovery to WebSocket mode only, allowing webhook mode to start successfully even when the bot info API is unavailable. The webhook server can still receive and process events correctly without the bot open_id (it's only used for group chat filtering).

Fixes compatibility with older private Feishu/Lark deployments that use webhook mode with custom domains.

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -290,11 +290,18 @@ func (p *Platform) KeepPreviewOnFinish() bool {
 func (p *Platform) Start(handler core.MessageHandler) error {
 	p.handler = handler
 
-	if openID, err := p.fetchBotOpenID(); err != nil {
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
## Review Summary

Gating fetchBotOpenID to non-webhook mode avoids startup failures on private/self-hosted Feishu deployments where the bot-info API path is unreliable, while keeping WebSocket behaviour unchanged. Small and localized to platform/feishu.

---

✅ **Good**: Clear comment on why webhook mode skips eager discovery; aligns with existing use of botOpenID for group mention filtering (empty ID disables that path rather than crashing startup).

LGTM — approve.

PR #701, opened 2026-04-20
fix(core): suppress reply footer when only workdir is known

LyInfi:
## Summary

Fixes a regression introduced in v1.3.0 where assistant replies end with a bare `~` on Feishu and Weixin (personal WeChat).

## Root cause

`buildReplyFooter` assembles up to four parts: model, reasoning effort, context remaining / usage, and workdir.

The `claudecode` agent does not implement the optional `GetModel`, `GetReasoningEffort`, or `UsageReporter` interfaces, so those three parts are always empty for it. When the workdir also happens to equal `\$HOME`, `compactReplyFooterPath` compacts it to `"~"`, leaving `parts = ["~"]` and producing the footer string `"~"`.

`appendReplyFooter` then wraps it in Markdown italics: `content + "\n\n*~*"`. On Feishu this is rendered via the rich-text post builder (which strips the italic markers and keeps the content), and on Weixin ilink the `*...*` pair is stripped by the client. In both cases the user sees a lone `~` on its own line below every reply.

## Fix

Only emit the footer when at least one status-carrying part (model, reasoning effort, context remaining, or usage) is populated. A bare workdir is not useful status information, and when combined with the `$HOME`-compaction above it actively creates a visible rendering bug.

## Repro

1. Configure a project that uses the `claudecode` agent and `work_dir = "$HOME"`.
2. Connect the project to a Feishu or Weixin platform.
3. Send any message to the bot.
4. Observe that every reply ends with a standalone `~` on a new line.

After this fix, no footer is emitted in that scenario (the user still sees the plain reply).

## Testing

- Added `TestProcessInteractiveEvents_SuppressesReplyFooterWhenOnlyWorkDir` in `core/engine_test.go` covering an agent that only exposes `GetWorkDir()` and runs in `$HOME`.
- Existing reply-footer tests (`…_AppendsReplyFooterWhenEnabled`, `…_DoesNotAppendReplyFooterWhenDisabled`, `…_ReplyFooterPrefersSessionRuntimeState`) still exercise the full-footer path.
- `go vet ./core/` clean; `gofmt` clean.

Note: two pre-existing reply-footer tests and `TestResolveLocalDirPath_AcceptsSubdir` fail on `main` on macOS due to `/private/var` vs `/var` symlink handling in `t.TempDir()` — unrelated to this change (verified by running the same tests on `main` before applying the fix).

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -4101,21 +4101,26 @@ func (e *Engine) buildReplyFooter(agent Agent, session AgentSession, workspaceDi
 	}
 
 	var parts []string
+	hasStatus := false
 	if model := replyFooterModel(session, agent); model != "" {
 		parts = append(parts, model)
+		hasStatus = true
 	}
 	if effort := replyFooterReasoningEffort(session, agent); effort != "" {
 		parts = append(parts, effort)
+		hasStatus = true
 	}
 	if left := strings.TrimSpace(contextLeft); left != "" {
 		parts = append(parts, left)
+		hasStatus = true
 	} else if usage := e.replyFooterUsageText(session, agent); usage != "" {
 		parts = append(parts, usage)
+		hasStatus = true
 	}
 	if dir := replyFooterWorkDir(session, agent, workspaceDir); dir != "" {
 		parts = append(parts, dir)
 	}
-	if len(parts) == 0 {
+	if !hasStatus {
 		return ""
 	}
 	return strings.Join(parts, " · ")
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## Review Summary
Fixes v1.3.0 regression: suppresses useless bare-tilde footer when only workdir is known. Change in core/, no agent/platform imports. CI green.
✅ Good: Targets root cause in buildReplyFooter.
LGTM — approve.

PR #704, opened 2026-04-20
fix(acp): preserve options for workspace agents

xxb:
## Summary
- preserve ACP agent options when creating per-workspace agents in multi-workspace mode
- include `command`, `args`, `env`, `auth_method`, and `display_name` in the workspace snapshot
- add a regression test covering ACP workspace option propagation

## Problem
ACP projects in multi-workspace mode could fail when a workspace-specific agent was created. The parent ACP agent had the required startup options, but the per-workspace derived agent did not inherit them.

In practice this surfaced as errors like:

`acp: agent option "command" is required (path or name of the ACP agent binary)`

This happened because workspace agent creation rebuilt an options map for the new agent instance, but ACP did not expose its required config through `WorkspaceAgentOptions()`.

## Fix
This change makes the ACP agent snapshot the config it needs to recreate itself in workspace mode:
- `command`
- `args`
- `env`
- `auth_method`
- `display_name`

It also adds a regression test to ensure session-only env does not leak into the workspace snapshot.

## Verification
- `go test ./agent/acp -run "Test(New_DisplayName|WorkspaceAgentOptions)" -v`
- `go test ./core -run "TestGetOrCreateWorkspaceAgent_InheritsSnapshotOptions|TestMultiWorkspaceAgent_PropagatesRunAsUser|TestMultiWorkspaceAgent_NoPropagationWhenParentHasNoRunAs" -v`
- `go build -tags no_web ./cmd/cc-connect`

### DIFF
--- a/agent/acp/agent.go
+++ b/agent/acp/agent.go
@@ -21,6 +21,7 @@ type Agent struct {
 	workDir     string
 	command     string
 	args        []string
+	staticEnv   map[string]string
 	extraEnv    []string
 	sessionEnv  []string
 	authMethod  string // optional, e.g. "cursor_login" for Cursor CLI (see authenticate RPC)
@@ -46,6 +47,7 @@ func New(opts map[string]any) (core.Agent, error) {
 	}
 
 	args := parseStringSlice(opts["args"])
+	staticEnv := envMapFromOpts(opts)
 	extra := envPairsFromOpts(opts)
 	authMethod, _ := opts["auth_method"].(string)
 	authMethod = strings.TrimSpace(authMethod)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## Review Summary

Implements WorkspaceAgentOptionSnapshotter for ACP: snapshots command, args, static env map, auth_method, and display_name for per-workspace agent recreation; session env stays runtime-only. Mutex + defensive copies look correct.

✅ Good: Aligns with engine workspace cloning; excludes SetSessionEnv from snapshot; thread-safe copies. Tests cover the snapshot contract.

LGTM — approve.

Issue #500, opened 2026-04-08
[BUG] /list always empty and /switch fails with Claude Code agent — session agent_session_id cleared on switch


nanjade:
## Bug Description

When using Claude Code as the agent, `/list` always returns "未找到此项目的会话" (no sessions found) and `/switch` fails to switch back to previous sessions.

## Environment

- **cc-connect version**: v1.2.2-beta.5 (commit 28513f3)
- **OS**: Windows 11 Pro
- **Agent**: Claude Code (Opus model)
- **Platform**: Telegram

## Steps to Reproduce

1. Start cc-connect with a Claude Code project
2. Send a message → session `s4` is created with a valid `agent_session_id`
3. Run `/current` → correctly shows current session info (name, ID, message count)
4. Run `/list` → returns "未找到此项目的会话"
5. Run `/new 第二个对话` → new session `s5` created successfully
6. Send a message in the new session → agent responds normally
7. Run `/list` → still returns "未找到此项目的会话"
8. Run `/switch 测试` (previous session name) → fails

## Root Cause Analysis

Inspecting the session file (`~/.cc-connect/sessions/<project>.json`) reveals the issue:

**Before `/new`:**
```json
"s4": {
  "agent_session_id": "ce138e48-6e38-4768-af94-be2ee9bf48c3",
  "history": [ ... 10 messages ... ]
}
```

**After `/new` creates s5:**
```json
"s4": {
  "agent_session_id": "",   // ← CLEARED!
  "history": null            // ← CLEARED!
},
"s5": {
  "agent_session_id": "7740dd68-1e94-4dfb-9023-7e5631d65814",
  "history": [ ... 2 messages ... ]
}
```

When a new session is created, the **old session's `agent_session_id` and `history` are wiped**. Since `/list` appears to filter out sessions with empty `agent_session_id`, and the current active session is excluded from the list, the result is always empty.

## Expected Behavior

- `/list` should display all previous sessions (even those not currently connected to an agent)
- `/switch <name|id>` should reconnect to a previous session, restoring its conversation history
- Old session data (`agent_session_id`, `history`) should be preserved when creating a new session

## Related Issues

This is the same category of bug as #441 (Gemini CLI session sync issues). PR #443 addresses it for Gemini CLI but does not cover Claude Code agent type. The underlying issue in `core/engine.go`'s `cmdSwitch` logic likely affects all agent types.

## Screenshots

`/current` works correctly:
```
📌 当前会话
名称: 测试
[...]

PR #706, opened 2026-04-20
fix(claudecode): encode spaces and "~" in project keys (#500)

silverzhouyinhe:
## Summary

`encodeClaudeProjectKey` in [agent/claudecode/claudecode.go](https://github.com/chenhg5/cc-connect/blob/main/agent/claudecode/claudecode.go#L1094) only collapses `/`, `:`, `_`, and non-ASCII characters to `-`, but Claude Code CLI's on-disk project keys under `~/.claude/projects/` also collapse **spaces** and **`~`** to `-`. As a result, on macOS any work dir that lives inside iCloud Drive — e.g.

```
/Users/<me>/Library/Mobile Documents/com~apple~CloudDocs/<project>
```

produces the candidate key

```
-Users-<me>-Library-Mobile Documents-com~apple~CloudDocs-<project>
```

while Claude Code actually wrote the directory

```
-Users-<me>-Library-Mobile-Documents-com-apple-CloudDocs-<project>
```

The primary lookup misses; all three legacy candidates in `findProjectDir` use only `/ \ : _` replacements so they also miss; and the fallback `ReadDir` scan compares against `encodeClaudeProjectKey(absWorkDir)` too, so it misses as well. `findProjectDir` returns `""`, `ListSessions` returns `nil`, and Feishu / Telegram `/list` permanently shows `(0) 未找到此项目的会话`.

## Fix

Extend the replacement set in `encodeClaudeProjectKey` to also collapse `' '` and `'~'` to `'-'`. The function is the single source of truth — both the primary candidate and the fallback scan go through it, so one edit fixes both paths.

Docstring updated to list the iCloud-path rationale.

## Tests

`go test ./agent/claudecode/ -run 'TestEncodeClaudeProjectKey|TestFindProjectDir' -v` — all pass, including three new `TestEncodeClaudeProjectKey` cases and a new `TestFindProjectDir_ICloudPath` regression test:

- `path with spaces`
- `path with tildes`
- `iCloud path with spaces and tildes`
- `TestFindProjectDir_ICloudPath` — creates a mock `~/.claude/projects/-Users-test-Library-Mobile-Documents-com-apple-CloudDocs-my-project/` and asserts `findProjectDir` resolves the iCloud-style work dir to it

All pre-existing test cases still pass unchanged.

## Scope note

This PR fixes the macOS iCloud variant of the broader "encoded key doesn't match on-disk dir" problem reported in #500. The reporter there is on Windows 11, so their specific case may involve additional characters (drive letters, backslashes) that this PR does not touch — but the root cause ("encoder diverges from Claude Code CLI's real encoding") is shared, and the macOS iCloud case is likely to be reported identically given how common `com~apple~CloudDocs` paths are. Happy to narrow or broaden the scope based on maintainer preference.

Refs #500

## Test plan

- [x] `go test ./agent/claudecode/ -run 'TestEncodeClaudeProjectKey|TestFindProjectDir' -v` — all pass
- [x] `go test ./agent/claudecode/` — full package passes
- [x] Manual: on macOS with work dir under iCloud Drive, confirmed (before this fix) that `/list` returned empty despite `~/.claude/projects/-Users-...-Mobile-Documents-com-apple-CloudDocs-...` existing with valid `.jsonl` session files

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -1096,7 +1096,9 @@ func boolVal(m map[string]any, key string) bool {
 // 1. Replacing path separators (/ or \) with "-"
 // 2. Replacing colons (:) with "-" (Windows drive letters)
 // 3. Replacing underscores (_) with "-"
-// 4. Replacing all non-ASCII characters with "-"
+// 4. Replacing spaces and tildes (~) with "-" (common in macOS iCloud paths like
+//    "/Users/x/Library/Mobile Documents/com~apple~CloudDocs/...")
+// 5. Replacing all non-ASCII characters with "-"
 func encodeClaudeProjectKey(absPath string) string {
 	// First, normalize to forward slashes for consistent processing
 	normalized := strings.ReplaceAll(absPath, "\\", "/")
@@ -1104,7 +1106,7 @@ func encodeClaudeProjectKey(absPath string) string {
 	// Build the encoded key character by character
 	var result strings.Builder
 	for _, r := range normalized {
-		if r == '/' || r == ':' || r == '_' {
+		if r == '/' || r == ':' || r == '_' || r == ' ' || r == '~' {
 			result.WriteRune('-')
 		} else if r < 128 { // ASCII range (0-127)
 			result.WriteRune(r)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## Review Summary

Aligns project key encoding with Claude Code's on-disk layout for paths with spaces and ~ (e.g. iCloud Drive). Minimal change in the right layer (encodeClaudeProjectKey), tests cover both unit encoding and findProjectDir resolution.

✅ Good: Targeted fix; clear doc comment; table tests + regression test for mock ~/.claude/projects/ layout.

🟡 Process: Wait for unit-test CI to pass before merge.

LGTM — approve.

PR #716, opened 2026-04-21
fix(daemon): handle macOS launchd gui and background sessions

xxb:
## Summary

Fix macOS launchd daemon management for both GUI and headless sessions.

- prefer `gui/<uid>` when available
- fall back to `user/<uid>` when no GUI session exists
- keep `restart` GUI-first so a headless-started service can migrate to `gui/<uid>` later
- allow both `Aqua` and `Background` in `LimitLoadToSessionType`
- add regression tests for domain selection and status handling

## Problem

The current macOS daemon implementation always uses `launchctl` with `gui/<uid>`. This fails in headless or remote environments with errors such as:

```text
Bootstrap failed: 125: Domain does not support specified action
```

## Validation

Verified in macOS GUI and headless usage flows:

- reinstall succeeded
- `cc-connect daemon status` worked
- `cc-connect daemon restart` worked
- actual conversation flow remained normal

### DIFF
--- a/daemon/launchd.go
+++ b/daemon/launchd.go
@@ -16,6 +16,12 @@ const (
 	launchdLabel = "com.cc-connect.service"
 )
 
+var runLaunchctl = func(args ...string) (string, error) {
+	cmd := exec.Command("launchctl", args...)
+	out, err := cmd.CombinedOutput()
+	return strings.TrimSpace(string(out)), err
+}
+
 type launchdManager struct{}
 
 func newPlatformManager() (Manager, error) {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## QA Review — #716

✅ **Approve**

Correctly detects GUI vs user domain via `launchctl print`, plist adds `LimitLoadToSessionType` for Aqua+Background. Good test coverage with `runLaunchctl` injection pattern.

No breaking change. macOS-only, medium scope. Architecture clean.

PR #717, opened 2026-04-21
fix(opencode): pass inbound images as files

soaringk:
## Summary
- stage inbound image attachments for OpenCode sessions under the workspace .cc-connect/images directory
- pass staged images to opencode run via repeated --file flags so OpenCode can resolve them as file/image parts
- add unit coverage for image staging and generated run args

## Tests
- go test ./agent/opencode
- go test ./... (fails in this environment: web/embed.go requires missing web/dist; agent/iflow PTY test cannot open /dev/ptmx due sandbox permissions)

PR #718, opened 2026-04-21
fix: forward images to OpenCode agent via local file references

RoyougiShiki:
## Summary

The OpenCode agent's `Send()` method receives `images []core.ImageAttachment` but silently drops it, causing image messages from Lark/Feishu (and other platforms) to be completely ignored when using the OpenCode agent.

Other agents (e.g. Pi) already handle images by saving them to disk and referencing the file paths in the prompt. This PR applies the same pattern to the OpenCode agent.

## Root Cause

In `agent/opencode/session.go`, the `Send` method only handles `files` but ignores `images`:

```go
func (s *opencodeSession) Send(prompt string, images []core.ImageAttachment, files []core.FileAttachment) error {
    if len(files) > 0 {  // only handles files
        filePaths := core.SaveFilesToDisk(s.workDir, files)
        prompt = core.AppendFileRefs(prompt, filePaths)
    }
    // images parameter is completely ignored!
```

The Feishu platform correctly downloads images and generates `ImageAttachment` objects (including images embedded in `post`-type rich text messages via `extractPostParts`), but the OpenCode agent never processes them.

### `core/message.go`
- Add `SaveImagesToDisk()` — saves image attachments to `workDir/.cc-connect/attachments/` with proper MIME type to extension mapping (png/jpg/gif/webp)
- Add `AppendImageRefs()` — appends image file path references to the prompt string
- These mirror the existing `SaveFilesToDisk()` / `AppendFileRefs()` pattern

### `agent/opencode/session.go`
- Update `Send()` to save images to disk and append file path references to the prompt, so OpenCode CLI can read them using its built-in file tools

## Testing

- Pattern is identical to the Pi agent's image handling (`agent/pi/session.go:saveImagesToDisk`)
- Uses the same `.cc-connect/attachments/` directory convention
- File save logic follows the established `SaveFilesToDisk` pattern exactly

## Before/After

**Before:** User sends an image in Lark → Feishu platform downloads it → OpenCode agent drops it → agent only sees text

**After:** User sends an image in Lark → Feishu platform downloads it → OpenCode agent saves to disk → prompt includes file path reference → agent can read and analyze the image

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -1,0 +1,23 @@
+name: Build Windows Binary
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.24'
+
+      - name: Build
+        run: go build -o cc-connect.exe .
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: cc-connect-windows-amd64
+          path: cc-connect.exe
--- a/agent/opencode/session.go
+++ b/agent/opencode/session.go
@@ -58,6 +58,10 @@ func newOpencodeSession(ctx context.Context, cmd, workDir, model, mode, resumeID
 }
 
 func (s *opencodeSession) Send(prompt string, images []core.ImageAttachment, files []core.FileAttachment) error {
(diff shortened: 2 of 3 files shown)


RoyougiShiki:
Closing this in favor of #717 by @soaringk, which addresses the same issue with a better approach:

- Uses OpenCode CLI's native  flag instead of appending paths to the prompt text
- Includes unit tests
- Cleaner code organization ( + )

Great catch on the same root cause. Looking forward to seeing #717 merged.

PR #720, opened 2026-04-21
fix(opencode): prevent empty response when agent process crashes

bingkxu:
## Summary

- Fix empty response when the opencode agent process crashes (e.g. stale session ID causes "Session not found")
- Restructure readLoop termination logic to check stderr before emitting fallback EventResult
- Add `agentErrorHandlers` pattern table for extensible error-to-message mapping with i18n support

## Problem

When the opencode process crashes, `readLoop` sends a fallback `EventResult(Done:true)` before the deferred `cmd.Wait()` can detect the error and emit an `EventError`. The engine receives `EventResult` first, returns from the event loop, and the `EventError` is never consumed. The user sees an empty response with no error information.

## Root Cause

The execution order in `readLoop` is:

1. Scanner loop exits (stdout closed after process crash)
2. Fallback `EventResult` sent (line 182, unconditional)
3. Deferred `cmd.Wait()` detects error and sends `EventError` (but engine already exited)

## Fix

**`agent/opencode/session.go`** - Restructure readLoop:
- Check `stderrBuf` immediately after scanner exits, before sending fallback EventResult
- If stderr has content: emit `EventError` (with "Session not found" detection)
- If stderr is empty: emit fallback `EventResult` as before
- `cmd.Wait()` kept in defer for resource cleanup only

**`core/engine.go`** - Pattern table for error messages:
- `agentErrorHandlers` maps error patterns to localized `MsgKey`
- Matched errors show user-friendly recovery hints (e.g. "Use /new to start a fresh conversation")
- Unmatched errors show raw error message
- Adding new error types requires only one line in the table

**`core/i18n.go`** - Add `MsgSessionNotFound` translations (en/zh/zh-TW/ja/es)

## Testing

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] Manual testing: stale session ID now shows friendly message instead of empty response

### DIFF
--- a/agent/opencode/session.go
+++ b/agent/opencode/session.go
@@ -10,7 +10,8 @@ import (
 	"log/slog"
 	"os"
 	"os/exec"
-	"path/filepath"
+"path/filepath"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
## QA Review — #720

✅ **Approve**

Fixes the main ordering bug where `EventResult` was emitted before `EventError` when stderr had content. Engine-level mapping of "Session not found" to `MsgSessionNotFound` i18n is clean.

Architecture: `core` uses generic `strings.Contains` on error, no `agent/` import — compliant.

💭 Suggestions (non-blocking):
- Consider calling `cmd.Wait()` before the stderr branch for robustness (stderr buffer may be incomplete if process hasn't fully exited)
- A small regression test would be valuable

bingkxu:
Hi @chenhg5, this PR has been approved. Would you mind merging it when you have a chance? Thanks!

bingkxu:
Conflict resolved. Rebased onto latest main.

PR #726, opened 2026-04-21
fix(feishu): parse at/markdown in post content and handle sticker/media messages

### LINKED ISSUE #725: feishu: post message parser drops at/markdown elements; sticker and media types are silently ignored
## Version

cc-connect 1.3.0 (Feishu bot, self-hosted, darwin/arm64)

## Summary

Two related gaps in the Feishu inbound message parser cause content to be silently dropped before the agent sees it:

### Bug 1: post content parser ignores `at` and `markdown` elements

`platform/feishu/feishu.go` — `extractPostParts` and `extractPostPlainText` (quoted-message chain).

Feishu rich-text (`post`) content is a 2D array of elements where each element has a `tag`. The parser switches on `tag` and only handles `text`, `a`, `code_block`, and `img`. Elements with `tag: \"at\"` or `tag: \"markdown\"` are silently dropped.

Consequences:

- **@mentions are lost in post messages.** When a user combines image + text + @mention, Feishu delivers it as `post`, not `text`. `stripMentions` relies on the text containing an `@_user_N` placeholder, but in post format the mention is a structural `<at>` element — no placeholder is ever extracted, so the mention is gone.
- **Posts where the text lives inside `markdown` elements are dropped.** Some Feishu clients / API producers wrap content in `markdown` elements. Our parser returns empty text for those.
- **Quoted-reply context is truncated.** `extractPostPlainText` (used when fetching parent messages for reply chains) handles even fewer tags than the main path — `a`, `at`, `markdown`, `img` all disappear, so reply chain context is missing.

### Bug 2: `sticker` and `media` (video) message types are silently ignored

`dispatchMessage` handles `text`, `image`, `audio`, `post`, `file`, `merge_forward` and logs anything else as `ignoring unsupported message type`. Stickers and videos — both common in daily use — fall through and never reach the agent.

## Reproduction

1. In a Feishu group where cc-connect is bound, send a rich-text message containing an image, some text, and an `@` mention of another user.
2. Expected: the bot/agent sees a message whose text includes `@<name>`.
3. Actual: the bot sees the text without any mention information; in the worst case `stripMentions` returns empty text and the message is dropped entirely (`dropping empty text after mention stripping` in the debug log).

For the sticker/media bug, send any Feishu sticker or short video — the agent receives nothing.

## Proposed fix

Extend both post parsers to handle `at` (resolve `user_id` to `@<name>` via the top-level mentions data or `resolveUserName`, strip the bot's own mention) and `markdown` (append its text). Extend the quoted-message path to also handle `a` and `img` so quoted context is readable. Add `sticker` and `media` cases to `dispatchMessage` with graceful fallbacks (sticker → image attachment or `[sticker]` placeholder; video → `[video: <name>, Xs]` text plus thumbnail image).

Fix in PR linked below.

Cigarrr:
Closes #725.

## Summary

- **post content parser** (`extractPostParts`, `extractPostPlainText`): add `at` and `markdown` cases. `at` resolves the bot's own mention to a no-op, `all` to `@all`, and other users via `user_name` or `resolveUserName(user_id)`; `markdown` passes its text through. Without this, any rich-text message that combines image + text + @mention loses the mention (the `mentions` array carries placeholder keys that never appear in the extracted post text, so `stripMentions` has nothing to replace).
- **quoted-reply path** (`extractPostPlainText`): also add `a` / `img` handling so quoted context preserves link text and an `[image]` marker instead of silently dropping them.
- **sticker / media message types** (`dispatchMessage`): new cases. Stickers are delivered as image attachments (with a `[sticker]` text fallback if download fails). Videos surface as `[video: <name>, Xs]` plus the thumbnail image when `image_key` is present.
- **`postElement` struct**: add `UserId` and `UserName` fields for `at` elements.

## Tests

- `TestExtractPostParts_AtMention` — covers named mention, `@all`, and bot-mention stripping.
- `TestExtractPostParts_Markdown`
- `TestExtractPostPlainText_LinkText` (replaces the old `NonTextTagsIgnored` test, whose assertion is now incorrect)
- `TestExtractPostPlainText_AtMention` (named / `all` / fallback)
- `TestExtractPostPlainText_Markdown`

`go test ./platform/feishu/...` passes locally.

## Notes

- No new dependencies. Bot-mention stripping reuses the existing `botOpenID` field on `Platform`.
- Sticker image download reuses `downloadImage`; media thumbnail download is best-effort with a warn-log on failure.
- Unknown element tags within post content continue to be ignored (existing behavior preserved for `hr`, `emotion`, etc.).

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -924,6 +924,69 @@ func (p *Platform) dispatchMessage(ctx context.Context, msgType, content string,
 		}
 		p.handler(p.dispatchPlatform(), coreMsg)
 
+	case "sticker":
+		var stickerBody struct {
+			FileKey string `json:"file_key"`
+		}
+		if err := json.Unmarshal([]byte(content), &stickerBody); err != nil {
+			slog.Error(p.tag()+": failed to parse sticker content", "error", err)
+			return
+		}
+		slog.Info(p.tag()+": sticker received", "user", userID, "file_key", stickerBody.FileKey)
+		imgData, mimeType, err := p.downloadImage(messageID, stickerBody.FileKey)
+		if err != nil {
+			slog.Warn(p.tag()+": download sticker failed, falling back to placeholder", "error", err)
+			p.handler(p.dispatchPlatform(), &core.Message{
+				SessionKey: sessionKey, Platform: p.platformName,
+				MessageID: messageID,
+				UserID:    userID, UserName: userName, ChatName: chatName,
+				Content: "[sticker]", ExtraContent: quotedPrefix, ReplyCtx: rctx,
+			})
+			return
+		}
+		p.handler(p.dispatchPlatform(), &core.Message{
+			SessionKey: sessionKey, Platform: p.platformName,
+			MessageID: messageID,
+			UserID:    userID, UserName: userName, ChatName: chatName,
+			Images:   []core.ImageAttachment{{MimeType: mimeType, Data: imgData}},
+			ReplyCtx: rctx,
+		})
+
+	case "media":
+		var mediaBody struct {
+			FileKey  string `json:"file_key"`
+			ImageKey string `json:"image_key"`
+			FileName string `json:"file_name"`
+			Duration int    `json:"duration"`
+		}
+		if err := json.Unmarshal([]byte(content), &mediaBody); err != nil {
+			slog.Error(p.tag()+": failed to parse media content", "error", err)
+			return
+		}
+		slog.Info(p.tag()+": media received", "user", userID, "file_key", mediaBody.FileKey, "file_name", mediaBody.FileName)
+		text := "[video"
+		if mediaBody.FileName != "" {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## QA Review — #726

✅ **Approve**

Clean Feishu platform improvements:

1. **Sticker messages**: Downloads sticker image, sends as ImageAttachment. Graceful fallback to "[sticker]" text on download failure.
2. **Media messages**: Converts video to descriptive text "[video: filename, Ns]" with optional thumbnail download. Good approach since IM→agent can't forward video.
3. **Post content parsing**: Now handles `at`, `markdown`, `a` (link), `img` tags that were previously silently dropped.
4. **Bot self-mention filter**: Skips own botOpenID in `extractPostParts` — prevents self-reference noise.

Tests: Good coverage — at mention (table-driven), markdown passthrough, link text extraction. Existing test updated correctly.

No `core/` violations. Platform-scoped, no breaking changes.

PR #728, opened 2026-04-21
feat(feishu): label bot senders by app_id with configurable aliases

### LINKED ISSUE #727: feishu: quoted-reply chain renders all bot senders as bare "Bot", can't distinguish peer bots
### Problem

In Feishu, when a user quotes/replies to a bot message and the chain is fetched via `fetchReplyChain` → `fetchSingleMessage`, every bot sender is rendered as a bare `"Bot"` regardless of which bot actually posted. With multiple bots in a shared chat (e.g. two cc-connect instances, or a peer bot + a third-party bot), the downstream model has no way to tell replies from *itself* apart from replies sent by *another* bot.

This only shows up in the quoted/reply-chain rendering path (`fetchSingleMessage`, around [`platform/feishu/feishu.go:1199-1201`](https://github.com/chenhg5/cc-connect/blob/main/platform/feishu/feishu.go#L1199-L1201)); regular incoming messages are unaffected because the current bot filters its own events upstream.

### Reproduction

1. Two cc-connect bots (A and B) in the same Feishu group chat.
2. User quotes a message from bot B and replies to bot A.
3. Bot A's model sees the reply chain with `[... Bot (assistant): <B's text>]` — no way to know it was B, not A.

Impact scales with number of peer bots sharing the chat.

### Why it happens

Feishu's Open API sets `sender.id` to the **bot's app_id** when `sender.sender_type == "app"` (note app_id is globally stable, unlike user open_ids which are per-app). The existing code ignores the id entirely and hardcodes `senderName = "Bot"`.

### Proposal

Use the app_id. Two tiers:

1. Add a `peer_bots` map in `[platforms.feishu]` mapping `app_id → friendly alias` (e.g. `cli_xxxxxxxx = "Jeeves"`).
2. Resolution:
   - known app_id → configured alias (`Jeeves`, `Ivy`, …)
   - unknown app_id → `Bot[<app_id>]` so operators can copy the id straight into the config
   - empty id → `Bot` (preserves current behavior on degenerate payloads)

Surfacing the raw app_id in the unknown case is deliberate — it beats a silent "Bot" for debugging and lets operators extend the mapping incrementally.

PR coming.

Cigarrr:
Closes #727.

## Problem

In Feishu quoted-reply / reply-chain rendering, every bot sender shows up as a bare `"Bot"`. When multiple bots share a chat (e.g. two cc-connect instances, or a cc-connect bot + a third-party peer), the model can't tell its own prior messages apart from another bot's — issue #727 has the full reproduction.

## Fix

Feishu sets `sender.id` to the **bot's app_id** (globally stable, not an open_id) when `sender.sender_type == "app"`. Use it.

Add a `peer_bots` map in `[platforms.feishu]` config mapping `app_id → friendly alias`, and extract the resolution into a `resolveBotSenderName` helper with a 3-case fallback:

- empty id       → `"Bot"` (preserves current behavior on degenerate payloads)
- known app_id   → configured alias (e.g. `Jeeves`, `Ivy`)
- unknown app_id → `"Bot[<app_id>]"` — surfaces the id so operators can drop it straight into the config rather than guessing which peer posted

### Config example

\`\`\`toml
[platforms.feishu]
peer_bots = { cli_aaaaaaaa = \"Jeeves\", cli_bbbbbbbb = \"Ivy\" }
\`\`\`

## Tests

Table-driven unit tests for `resolveBotSenderName` covering:
- known app_id → alias
- unknown app_id → `Bot[<id>]`
- empty id → `Bot`
- nil `peerBots` map (platform started without the config key)

\`go test -tags 'no_web' -count=1 ./platform/feishu/\` passes locally.

## Back-compat

- Default behavior (no `peer_bots` configured, or unknown app_id) changes from `"Bot"` to `"Bot[<app_id>]"`. This is user-visible text in reply chains, not a schema change — but worth calling out. If strict back-compat is preferred, happy to gate the surfacing behind a config switch.
- Bare `"Bot"` is preserved when `sender.id` is empty (shouldn't happen in practice, but keeps the degenerate path identical).

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -138,9 +138,10 @@ type Platform struct {
 	cancel           context.CancelFunc
 	dedup            core.MessageDedup
 	botOpenID        string
-	userNameCache    sync.Map // open_id -> display name
-	chatNameCache    sync.Map // chat_id -> chat name
-	chatMemberCache  sync.Map // chatID -> *chatMemberEntry
+	peerBots         map[string]string // app_id -> friendly alias, for quoted-reply attribution
+	userNameCache    sync.Map          // open_id -> display name
+	chatNameCache    sync.Map          // chat_id -> chat name
+	chatMemberCache  sync.Map          // chatID -> *chatMemberEntry
 	// Webhook mode fields (for Lark international version)
 	server       *http.Server
 	port         string
@@ -205,6 +206,15 @@ func newPlatform(name, domain string, opts map[string]any) (core.Platform, error
 		noReplyToTrigger = true
 	}
 
+	peerBots := map[string]string{}
+	if raw, ok := opts["peer_bots"].(map[string]any); ok {
+		for k, v := range raw {
+			if s, ok := v.(string); ok && s != "" {
+				peerBots[k] = s
+			}
+		}
+	}
+
 	progressStyle := "legacy"
 	if v, ok := opts["progress_style"].(string); ok {
 		switch strings.ToLower(strings.TrimSpace(v)) {
@@ -258,6 +268,7 @@ func newPlatform(name, domain string, opts map[string]any) (core.Platform, error
 		port:                       port,
 		callbackPath:               callbackPath,
 		encryptKey:                 encryptKey,
+		peerBots:                   peerBots,
 	}
 	if !useInteractiveCard {
 		base.self = base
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
QA Approve: peer_bots config for bot sender alias in quoted replies. Nil-safe, well-tested. Clean.

PR #730, opened 2026-04-21
fix(core): reject /ps when no task is currently running

### LINKED ISSUE #729: core: /ps injects directly into agentSession when session is idle, racing with normal messages
### Problem

`/ps` (introduced in #620 as a rename of `/btw`) is meant as a P.S. to a turn *already in flight* — additional context that gets appended to the running Claude CLI turn via `agentSession.Send`. But when the session is **idle** (no turn running), `cmdPs` still takes the same `Send` path, bypassing `Session.TryLock`.

Claude CLI's stream-json protocol is strictly one-turn-at-a-time on stdin. Direct `Send` from an idle-path `/ps` races with any concurrent normal message that grabs the session lock and starts a new turn — two writes to the same stdin, interleaved.

### Reproduction

1. Session is idle (no turn running).
2. User sends `/ps hello` (goes through `cmdPs` → `agentSession.Send`).
3. User sends a normal message at roughly the same time (goes through the main path → `TryLock` → `Send`).
4. Both writes hit the CLI's stdin unordered; at best the `/ps` text is treated as a separate turn, at worst stdin gets interleaved.

### Why it happens

`cmdPs` in [`core/engine.go:3140`](https://github.com/chenhg5/cc-connect/blob/main/core/engine.go#L3140) only checks `state.agentSession != nil` (\"alive\") before calling `Send`. It does not check whether the session is currently holding the turn lock. The old `/btw` (pre-#620) was not a builtin command — idle `/btw` fell through `handleCommand` as an unrecognized command and ended up on the normal message path, which took `TryLock` correctly. #620 made `/ps` a proper builtin and, in doing so, introduced this idle code path that skips the lock.

### Proposal

Reject `/ps` when the session is idle with the existing `MsgPsNoSession` reply. This matches the documented semantics (\"P.S. to an ongoing turn\") and removes the race. Users who actually want to start a new turn can just send the message without the `/ps` prefix.

Implementation sketch: add `Session.Busy()` (reads the same `s.busy` guarded by `s.mu` as `TryLock`/`Unlock`) and have `cmdPs` check it after the existing alive check.

PR coming (branched off #620 since that's where `/ps` lives).

Cigarrr:
Closes #729.

`/ps` is meant to append a P.S. to a turn already in flight. The idle path used to call `agentSession.Send` directly, which bypassed the session lock and raced with concurrent normal messages on the CLI's stdin (stream-json is strictly one-turn-at-a-time).

### Change

- Add `Session.Busy()` — reads `s.busy` under `s.mu`, same guard as `TryLock`/`Unlock`.
- `cmdPs` checks `Busy()` after the existing alive check. If idle, reply with `MsgPsNoSession` instead of injecting.
- Users who actually want to start a new turn can just send the message without the `/ps` prefix.

### Tests

- `TestSession_Busy` — fresh/locked/unlocked states.
- Four `cmdPs` branches: empty args, no agent session, idle session (new behavior), busy session (unchanged).

`go test ./core/...` passes.

### Note

Stacked on top of #620 (which introduces `/ps` as a builtin). If #620 lands first, this diff becomes just the one fix commit; if this lands first, rebase accordingly.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -3511,6 +3511,15 @@ func (e *Engine) cmdPs(p Platform, msg *Message, args []string) {
 		e.reply(p, msg.ReplyCtx, e.i18n.T(MsgPsNoSession))
 		return
 	}
+	// /ps is only meaningful as a supplement to a turn already in flight.
+	// When the session is idle, injecting via agentSession.Send bypasses the
+	// session lock and races with concurrent normal messages on the CLI's
+	// stdin, so reject instead.
+	_, sessions := e.sessionContextForKey(msg.SessionKey)
+	if session := sessions.GetOrCreateActive(msg.SessionKey); !session.Busy() {
+		e.reply(p, msg.ReplyCtx, e.i18n.T(MsgPsNoSession))
+		return
+	}
 	if err := state.agentSession.Send(text, nil, nil); err != nil {
 		slog.Error("ps: send failed", "error", err)
 		e.reply(p, msg.ReplyCtx, e.i18n.T(MsgPsSendFailed))
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -7103,6 +7103,103 @@ func TestCmdCompress_DrainsQueueAfterSuccess(t *testing.T) {
 	}
 }
 
+// --- cmdPs ---
+
+func TestCmdPs_EmptyArgs_RepliesUsage(t *testing.T) {
+	p := &stubPlatformEngine{n: "test"}
+	e := NewEngine("test", &stubAgent{}, []Platform{p}, "", LangEnglish)
+
+	msg := &Message{SessionKey: "test:user1", Content: "/ps", ReplyCtx: "ctx"}
+	e.cmdPs(p, msg, nil)
+
+	sent := p.getSent()
+	if len(sent) == 0 || !strings.Contains(sent[0], e.i18n.T(MsgPsEmpty)) {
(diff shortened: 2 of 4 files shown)


chenhg5 (approved):
## QA Review — #730

✅ **Approve**

Excellent improvement. Replaces the old `/btw` busy-path injection with a proper `/ps` builtin command. Key improvements:

- **Safety**: `Session.Busy()` check rejects `/ps` when session is idle, preventing stdin race with concurrent normal messages
- **Proper routing**: Goes through `handleCommand` instead of special-casing in `handleMessage`
- **Tests**: Comprehensive — 4 test cases covering empty args, no session, idle session, and busy session injection
- **i18n**: All 5 languages updated, clean MsgKey rename (MsgBtw → MsgPs)
- **Slack manifest**: Updated `/btw` → `/ps`

No breaking changes, no `core/` → `agent/`/`platform/` violations. Architecture clean.

💭 Note: Overlaps with #620 (also by you). Since #730 is more comprehensive (adds Busy() guard + proper builtin), suggest closing #620 and merging this one instead.

chenhg5:
@Cigarrr Please resolve the conflicts before merging, thanks

Cigarrr:
Rebased onto latest `main` (atop `fc79307`). Single commit `9de2422`, no manual conflict resolution needed — the underlying patch applied cleanly once the obsolete #620 base commit was dropped. `go build / vet / test ./core/...` all pass.

PR #742, opened 2026-04-22
feat(platform/max): add MAX messenger bridge with bidirectional file and voice support

### LINKED ISSUE #741: Add support for MAX messenger (Russian Bot API platform)
### Proposal

  Add a new platform adapter for [MAX](https://max.ru/) — a messaging platform from VK group with a public Bot API at
  `https://platform-api.max.ru`. Official API docs: https://dev.max.ru/docs-api

  No existing issues or PRs reference MAX (searched both open and closed).

  ### Why it matters

  - MAX is the primary messaging platform targeted at the Russian market; it has a documented, stable Bot API
  (token-based auth, long-polling `/updates`, `/messages`, `/uploads`).
  - Russian-speaking users of `cc-connect` currently have only Telegram and Discord — both of which have regional access
   issues on that market. Adding MAX fills a real gap without conflicting with existing platforms.
  - API surface is similar to Telegram (long poll + multipart upload), so it fits naturally into the existing
  `core.Platform` + capability-interface model.

  ### Implementation — ready to PR

  I have a working implementation that follows `CLAUDE.md` guidance:

  - `platform/max/max.go` — implements `core.Platform` and these capability interfaces:
    - `InlineButtonSender` (callback buttons)
    - `ImageSender` / `FileSender` / `AudioSender` (bidirectional media)
    - `MessageUpdater` (PUT /messages)
    - `FormattingInstructionProvider` (MAX-supported markdown subset)
    - `ReplyContextReconstructor` (for cron)
    - `TypingIndicator` (no-op — MAX has no equivalent)
  - `cmd/cc-connect/plugin_platform_max.go` with `//go:build !no_max` tag
  - `max` added to `ALL_PLATFORMS` in `Makefile`
  - Example block in `config.example.toml`
  - Two-step attachment upload via `POST /uploads?type=...` → multipart to CDN → token in message attachment
  - Voice input routed to `core.Message.Audio` so the `[speech]` pipeline transcribes it for the agent; TTS replies go
  through `SendAudio`
  - Retry on `attachment.not.ready` with exponential backoff
  - 19 unit/integration tests using `httptest` (including race detector) — `go test ./...` passes

  ### Maintenance

  Happy to take long-term responsibility for the MAX adapter:
  - Respond to MAX-specific bug reports
  - Keep the adapter up to date with future `platform-api.max.ru` API changes
  - Extend it as MAX adds new features (currently missing: reactions, inline actions)

  ### Question

  Are you open to accepting this as a PR? If yes, I'll open one and reference this issue with `Closes #<this issue>`.

  Thank you for `cc-connect` — it's a genuinely well-designed plugin architecture, which is exactly why adding a new
  platform was clean.

alekseybv:
## Summary

Adds support for [MAX](https://max.ru/) messenger — a messaging platform popular in the Russian market, with a documented Bot API at `https://platform-api.max.ru` ([docs](https://dev.max.ru/docs-api)).

## What's included

`core.Platform` plus these capability interfaces:
- `InlineButtonSender` — callback buttons
- `ImageSender` / `FileSender` / `AudioSender` — bidirectional media
- `MessageUpdater` — message editing via `PUT /messages`
- `FormattingInstructionProvider` — MAX-supported markdown subset
- `ReplyContextReconstructor` — cron/proactive replies
- `TypingIndicator` — no-op (MAX has no equivalent action)

Plus:
- `cmd/cc-connect/plugin_platform_max.go` with `//go:build !no_max` tag
- `max` added to `ALL_PLATFORMS` in `Makefile`
- Example block in `config.example.toml` (how to obtain a bot token via @MasterBot)
- 19 unit/integration tests using `httptest`, race-safe (`go test -race`)

## Implementation notes discovered during end-to-end testing

Live testing against the real MAX API surfaced three upstream-facing details worth documenting:

1. **Image CDN response is nested**: `{"photos": {"<photo_id>": {"token": "..."}}}` — not a flat `{"token": "..."}`.
2. **Audio CDN returns XML**, not JSON (`<retval>1</retval>`). The real attachment token lives in the `/uploads` response itself for video/audio, so the code uses `urlInfo.Token` as the source of truth and only scans the CDN body opportunistically.
3. **`/audios/{token}` endpoint does not exist** (returns 404). Incoming voice attachments already carry the download URL directly in `payload.url`, mirroring images/files; the code uses that and falls back to `/videos/{token}` for video where token resolution still applies.

Each of these is handled by `extractCDNToken(kind, body)` + per-kind routing in `fetchAttachments` / `uploadAttachment`.

Also includes retry with exponential backoff on `attachment.not.ready` — MAX occasionally needs a brief moment to index a freshly-uploaded blob before accepting it in a message.

## Minor core tweak

`core/speech.go`: raised the speech HTTP client `Timeout` from 60s to 5m. This covers the first-request cold-start case for local OpenAI-compatible Whisper backends (faster-whisper, whisper.cpp, etc.) that JIT-load their model on first use. Steady-state requests still return in seconds.

## Test plan

- [x] `go build ./...` — clean
- [x] `go test ./...` — all packages pass
- [x] `go test -race ./platform/max/` — pass
- [x] `make build PLATFORMS_INCLUDE=max NO_WEB=1` — selective-build works
- [x] Live test with real MAX bot: send/receive text, images, files, voice; TTS reply via Whisper local; cron scheduling; `cc-connect send --image/--file` from agent

## Config example

```toml
[[projects.platforms]]
type = "max"
[projects.platforms.options]
token = "your-max-bot-token"
allow_from = "*"      # or comma-separated user IDs
api_base = "https://platform-api.max.ru"   # optional override
```

Closes #741

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 *.dylib
 *.test
 *.out
+*.old
 
 # ============================================================================
 # Go workspace & tooling
@@ -36,7 +37,8 @@ vendor/
 # ============================================================================
 # Runtime configuration (contains secrets)
 config.toml
-config.test.toml
+config.*.toml
+!config.example.toml
 .env
 *.local.toml
 
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ PLATFORMS := \
 # ---------------------------------------------------------------------------
 
 ALL_AGENTS    := acp claudecode codex cursor devin gemini iflow kimi opencode pi qoder
-ALL_PLATFORMS := feishu telegram discord slack dingtalk wecom weixin qq qqbot line weibo
+ALL_PLATFORMS := feishu telegram discord slack dingtalk wecom weixin qq qqbot line weibo max
 ALL_EXTRAS    := web
 
 COMMA := ,
--- a/cmd/cc-connect/plugin_platform_max.go
+++ b/cmd/cc-connect/plugin_platform_max.go
@@ -1,0 +1,5 @@
+//go:build !no_max
+
+package main
+
+import _ "github.com/chenhg5/cc-connect/platform/max"
--- a/config.example.toml
+++ b/config.example.toml
@@ -909,6 +909,21 @@ app_secret = "your-feishu-app-secret"
 # share_session_in_channel = false  # If true, all users in a group share one agent session / 群聊共享会话
 # enable_reactions = false  # If true, add ⚡ reaction to incoming messages / 设为 true 收到消息时添加 ⚡ 表情
 
+# MAX messenger (uncomment to enable / 取消注释以启用)
+# 1. Create a bot via @MasterBot in MAX, get the access token
+#    在 MAX 中通过 @MasterBot 创建机器人，获取 access token
+# 2. Paste the token below / 填入下方 token
+# Connection: Long polling via platform-api.max.ru (no public URL needed) / 长轮询，无需公网 IP
+
+# [[projects.platforms]]
+# type = "max"
+#
+# [projects.platforms.options]
+# token = "your-max-bot-token"
+# allow_from = "*"    # Allowed MAX user IDs, e.g. "66624886,12345678"; "*" = all
+#                     # 允许的 MAX 用户 ID，如 "66624886,12345678"；"*" 表示所有
+# api_base = "https://platform-api.max.ru"   # optional override / 可选覆盖默认 API
+
 # Slack (uncomment to enable / 取消注释以启用)
 # 1. Create an app at https://api.slack.com/apps / 创建应用
 # 2. Enable Socket Mode (Settings > Socket Mode) / 开启 Socket Mode
--- a/core/speech.go
+++ b/core/speech.go
@@ -49,7 +49,7 @@ func NewOpenAIWhisper(apiKey, baseURL, model string) *OpenAIWhisper {
 		APIKey:  apiKey,
 		BaseURL: strings.TrimRight(baseURL, "/"),
 		Model:   model,
-		Client:  &http.Client{Timeout: 60 * time.Second},
+		Client:  &http.Client{Timeout: 5 * time.Minute},
 	}
 }
 
@@ -130,7 +130,7 @@ func NewQwenASR(apiKey, baseURL, model string) *QwenASR {
 		APIKey:  apiKey,
 		BaseURL: strings.TrimRight(baseURL, "/"),
 		Model:   model,
-		Client:  &http.Client{Timeout: 60 * time.Second},
+		Client:  &http.Client{Timeout: 5 * time.Minute},
 	}
 }
 
--- a/platform/max/max.go
+++ b/platform/max/max.go
@@ -1,0 +1,1077 @@
+package max
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log/slog"
+	"mime/multipart"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/chenhg5/cc-connect/core"
+)
+
+func init() {
+	core.RegisterPlatform("max", New)
+}
+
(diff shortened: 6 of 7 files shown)


chenhg5 (approved):
## Review Summary

Excellent first contribution — a complete, well-structured MAX messenger platform implementation with comprehensive test coverage. Architecture is clean and follows existing platform patterns.

---

✅ **Architecture**: `platform/max/` properly implements `core.Platform` and optional interfaces (ImageSender, FileSender, AudioSender, InlineButtonSender, MessageUpdater, FormattingInstructionProvider, ReplyContextReconstructor). Does not import `agent/` — correct dependency direction.

✅ **Long polling**: Proper marker-based polling with exponential backoff on errors, reconnect stability detection via `stableConnectionWindow`, graceful shutdown via context cancellation.

✅ **Attachment handling**: Two-step upload (request URL → CDN POST), with per-kind CDN response parsing (JSON for image/file, XML fallback for video/audio). Smart retry for "attachment.not.ready" with exponential backoff. Download capped at 25 MiB.

✅ **Test coverage**: ~20 test functions covering text send, long message splitting, buttons, image/file/audio upload, attachment retry, message editing, poll loop shutdown, audio format detection, and attachment routing. Mock API server is thorough.

✅ **Security**: No hardcoded secrets. Token from config only.

---

🟡 **Cross-cutting change in `core/speech.go`**
HTTP client timeout for OpenAI Whisper and QwenASR increased from 60s → 5 minutes. This affects ALL platforms, not just MAX. The change is reasonable (long audio clips can exceed 60s for transcription), but should be called out in the PR description since it's not MAX-specific.

🟡 **No message deduplication**
Other platforms (e.g. feishu, dingtalk) use `core.Dedup` to filter duplicate message IDs. MAX's marker-based polling should prevent most duplicates, but on reconnect edge cases a message could be reprocessed. Consider adding `p.dedup = core.NewDedup()` + `p.dedup.IsDuplicate(msg.Body.Mid)` in `handleMessage` for robustness.

💭 **Nit: `ReconstructReplyCtx`**
`fmt.Sscanf(sessionKey, "max:%s", &chatID)` works but `strings.TrimPrefix` + `strings.SplitN` is more idiomatic Go for this pattern.

---

✅ **Good**:
- Bilingual config comments (EN + ZH) in `config.example.toml`
- `FormattingInstructions()` properly limits markdown to what MAX supports
- `sniffImageMime()` fallback when CDN returns no Content-Type
- Attachment kind routing (image/* files sent as image type for proper preview)

LGTM — approved.

alekseybv:
Thank you for the thorough review!

Addressed the feedback in commit dddc3ef6:

- **Dedup**: added `core.MessageDedup` to the `Platform` struct and check `msg.Body.Mid` in `handleMessage` + `CallbackID` in `handleCallback`, matching the feishu/dingtalk/wecom pattern. Protects against reprocessing on poll-loop reconnect edge cases.
- **ReconstructReplyCtx style**: replaced `fmt.Sscanf` with `strings.CutPrefix` + `strings.Cut` — cleaner and handles the optional `:userID` suffix without the trailing loop.
- Added `TestHandleMessageDedupsByID` regression test.

### On the cross-cutting `core/speech.go` timeout change

Fair call-out — apologies for not flagging it in the PR body. Raised 60s → 5m specifically to cover the first-request cold-start case for **local OpenAI-compatible STT backends** (faster-whisper, whisper.cpp server): their initial request blocks while the model JIT-loads from disk or downloads weights. Steady-state requests still return in under a second, so this timeout is an upper bound that rarely matters, but the old 60s was too tight for that one cold boot.

If you'd prefer I move it to a separate PR (or make it configurable via `config.toml` rather than a hard-coded constant), happy to split it out — let me know.

alekseybv:
Friendly ping! @chenhg5 — is there anything else you'd like me to address before merge? Happy to iterate or rebase on main if needed. Thanks again for the review!

chenhg5:
@alekseybv Thanks for the pr, all good, merged. 🎉

PR #743, opened 2026-04-23
fix(telegram): increase HTTP client timeout to prevent long-poll EOF

### LINKED ISSUE #738: [Bug] Telegram long-polling disconnects with unexpected EOF every 1-3 minutes since go-telegram/bot migration (v1.2.2-beta.5+)
## Description

After the Telegram platform migrated from `go-telegram-bot-api/v5` to `go-telegram/bot v1.20.0` (PR #321, merged 2026-03-27, shipped since `v1.2.2-beta.5`), the long-polling getUpdates loop regularly experiences `unexpected EOF` every 1–3 minutes in quiet periods. The process stays alive and eventually recovers, but inbound messages from Telegram are delayed in bursts, giving the user the impression that the bot "drops the connection and reconnects" repeatedly.

There are no further Telegram-related commits in the v1.3 series that address this, so the regression affects every v1.3.x release.

## Environment

- cc-connect v1.3.2 (commit `19406df9`, built 2026-04-21)
- Linux container (Docker), Go stdlib HTTP client, no proxy configured
- Single bot instance; no webhook set; no 409 Conflict in logs
- Bot is idle most of the time (long-poll dominates traffic)

## Symptoms (from `/tmp/cc-connect.log`)

```
time=2026-04-22T03:30:43.901-07:00 level=INFO msg="[TGBOT] [ERROR] error get updates, error do request for method getUpdates, Post \"https://api.telegram.org/bot***/getUpdates\": unexpected EOF"
time=2026-04-22T03:30:43.901-07:00 level=INFO msg="[TGBOT] [ERROR] error get updates, error do request for method getUpdates, Post \"https://api.telegram.org/bot***/getUpdates\": unexpected EOF"
time=2026-04-22T03:33:27.972-07:00 level=INFO msg="[TGBOT] [ERROR] error get updates, error do request for method getUpdates, Post \"https://api.telegram.org/bot***/getUpdates\": unexpected EOF"
... (repeats every 1–3 min)
```

Observed pattern:
- Frequent `unexpected EOF` on getUpdates during idle windows.
- Each failure is logged **twice** within the same millisecond (not a duplicate issue in cc-connect — see root cause #4 below).
- Message delivery is not permanently broken; messages sent during the gap are delivered in a batch once the next poll succeeds.
- No 409 Conflict, no auth error, no webhook interference.

### 1. Library migration removed cc-connect's ability to detect a stuck poll loop

PR #321 ("Simplify the reconnect architecture") relies on `bot.Start(ctx)` to manage long polling internally. However, in `go-telegram/bot v1.20.0`:

- `getUpdates` (`get_updates.go:27`) retries forever with a 100 ms → 5 s incremental backoff and only exits on `ctx.Done()`.
- Errors are delivered to `errorsHandler` but never bubbled up to the caller.

As a result, `platform/telegram/telegram.go:301` `startPoll(ctx)` **never returns on connection failure**. The outer `connectLoop()` / `runConnection()` reconnect scaffolding (`telegram.go:229, 275`) is still in place but becomes unreachable once the bot is connected once. If the library enters a degraded-but-not-fatal state, cc-connect has no way to reset it short of a full process restart.

### 2. HTTP client timeout is flush with the server-side long-poll timeout

In `telegram.go:145`:

```go
httpClient := &http.Client{Timeout: 60 * time.Second}
```

And `defaultNewBot` (`telegram.go:214`) calls `tgbot.WithHTTPClient(60*time.Second, httpClient)`, which sets `pollTimeout = 60 s` inside the library. The library then issues `getUpdates` with `Timeout: int((b.pollTimeout - time.Second).Seconds())` = **59 s server-side long poll** (`get_updates.go:51`).

The client-side deadline (60 s total) has only ~1 s of slack over the server-side long poll. Any round-trip latency spike or upstream delay causes the HTTP client to time out just as the response is being returned, surfacing as `unexpected EOF` / `context deadline exceeded`.

### 3. No HTTP/2 keep-alive probing

No `http.Transport` is set when `proxy` is not configured, so Go's default transport is used. The default `http2.Transport.ReadIdleTimeout` is 0, i.e. no PING-based health check on idle HTTP/2 connections. If api.telegram.org or an intermediary silently half-closes an idle connection, the failure only surfaces on the next request — which is exactly the observed `unexpected EOF`.

### 4. Each error is logged twice (library behavior, but worth noting)

`raw_request.go:73-77` in `go-telegram/bot`: when `b.client.Do(req)` fails, the code calls `pr.CloseWithError(errDo)` on the pipe reader. This propagates an error to the pipe-writer goroutine at `raw_request.go:32`, which then calls `b.errorsHandler` a second time with a wrapped "close pipe writer" error. Because cc-connect does not install a custom `errorsHandler`, both messages go through the library's default `log.Printf("[TGBOT] [ERROR] %v", err)` — captured by slog as two entries at the same timestamp.

Less critical than #1–#3, but it inflates log noise and obscures the true failure rate.

## Impact

- Perceived bot unresponsiveness during idle windows: messages can be delayed up to ~5 s while the library backs off and reconnects.
- No automatic recovery path if the library's internal retry gets stuck on a non-fatal error (the outer reconnect loop cannot fire).
- Log noise makes it hard to distinguish a real network problem from normal background noise.

## Related

- #102 — "出现网络波动后，一直显示这个retrying" (same failure class, pre-migration)
- #212 — "[Bug] Telegram getUpdates 409 Conflict with single instance" (outer reconnect loop not observing library state)
- #693 — "cc-connect 进程存活但停止响应飞书消息（僵尸状态）" (same "library stuck, outer loop blind" pattern on Feishu)
- #321 — the migration PR that introduced this regression

## Proposed Fix (in order of impact/risk)

1. **Relax the HTTP client timeout.** Set `httpClient.Timeout = 0` (or `pollTimeout + buffer`, e.g. 75 s) so the long-poll deadline is controlled solely by the server-side `Timeout` parameter. One-line change, eliminates the client/server deadline race.
2. **Install a custom `errorsHandler` via `tgbot.WithErrorsHandler` and add an outer watchdog.** Bridge library errors to slog (including classification), and when N consecutive errors or M seconds without a successful poll are observed, cancel the ctx and let `connectLoop` reconnect the whole bot — restoring the recovery path the migration removed.
3. **Configure an explicit `http.Transport` with HTTP/2 PING-based keep-alive probing** (`ReadIdleTimeout` ~30 s, `PingTimeout` ~15 s) so silent idle half-closes are detected proactively instead of showing up as `unexpected EOF` on the next request.
4. (Optional) Consider suppressing the duplicate error log in the custom handler added in (2).

Happy to send a PR once there is agreement on the direction.

chenhg5:
## Summary
- Increase HTTP client timeout from 60s to 90s in `platform/telegram/telegram.go`
- The go-telegram/bot library sends `getUpdates` with `Timeout=59s` (pollTimeout − 1s)
- With HTTP client at 60s, network latency causes the client to timeout before the server responds → `unexpected EOF` every 1-3 minutes
- 90s gives 30s headroom while keeping the Telegram API long-poll at 59s

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes (all 28 packages)

Closes #738

### DIFF
--- a/platform/telegram/telegram.go
+++ b/platform/telegram/telegram.go
@@ -141,8 +141,10 @@ func New(opts map[string]any) (core.Platform, error) {
 	allowFrom, _ := opts["allow_from"].(string)
 	core.CheckAllowFrom("telegram", allowFrom)
 
-	// Build HTTP client with optional proxy support
-	httpClient := &http.Client{Timeout: 60 * time.Second}
+	// Build HTTP client with optional proxy support.
+	// Timeout must exceed the server-side long-poll duration (pollTimeout − 1s = 59s)
+	// to avoid the HTTP client racing with Telegram's response. 90s gives 30s headroom.
+	httpClient := &http.Client{Timeout: 90 * time.Second}
 	if proxyURL, _ := opts["proxy"].(string); proxyURL != "" {
 		u, err := url.Parse(proxyURL)
 		if err != nil {


chenhg5 (commented):
LGTM. Telegram HTTP client timeout 60s to 90s fixes long-poll EOF race condition. CI 5/5 green. Minimal change, correct reasoning, no breaking changes. Recommend merge.

kfhk:
我是用Codex，也遇到这个问题了，怎么改？

chenhg5:
> 我是用Codex，也遇到这个问题了，怎么改？

@kfhk 问题可以描述下吗，看看一不一样

PR #746, opened 2026-04-23
fix: handle compaction_continue to prevent engine early exit

bingkxu:
## Summary

OpenCode CLI sends `compaction_continue` (synthetic text event with `metadata.compaction_continue=true`) after auto-compaction to signal that it wants to continue working. Previously, cc-connect's engine would exit after receiving `EventResult` from the first turn, missing subsequent events from the continuation turn.

## Changes

- Add `expectingContinue` flag to `opencodeSession` to track compaction state
- Detect `compaction_continue` in `handleText`, set flag and skip `EventText` emission
- Check flag in `readLoop` before sending fallback `EventResult`, skip when expecting continuation
- Add `Metadata` and `Synthetic` fields to `core.Event` for future extensibility

## Testing

Manual testing with OpenCode CLI confirmed that:
- `compaction_continue` is correctly detected and logged
- Engine continues processing subsequent events (permissions, tool calls) after compaction
- No early exit occurs when OpenCode wants to continue

## Related

This fix addresses the auto-compaction continuation flow documented in OpenCode source at `packages/opencode/src/session/compaction.ts` line 442.


chenhg5 (approved):
## Review Summary

Clean, well-targeted fix for the OpenCode compaction_continue signal handling.

✅ **Correctness**: Logic is sound — detecting synthetic+metadata.compaction_continue to suppress premature EventResult is the right approach.
✅ **Concurrency**: atomic.Bool for expectingContinue; handleText and readLoop tail-check run in the same goroutine — no race.
✅ **Backward compatible**: New Metadata/Synthetic fields on core.Event are additive.
✅ **CI**: 5/5 green.

💭 Minor note: if the process crashes after compaction_continue but before the next step, EventResult would be skipped and the session could hang. This is an edge case and acceptable given the fix solves the much more common premature-exit bug.

LGTM — approved.

bingkxu:
Hi @chenhg5, this PR has been approved. Would you mind merging it when you have a chance? Thanks!

chenhg5:
@bingkxu Please resolve the conflicts before merging.

PR #766, opened 2026-04-24
fix(kimi): capture session UUID from stderr instead of stdout

### LINKED ISSUE #758: [Bug] Kimi agent cannot resume sessions: UUID read from wrong stream
### cc-connect Version

1.3.2

### Operating System

macOS

### Agent Type

Other

### Platform

N/A

### Installation Method

None

### Bug Description

When using type = "kimi" as the agent, every message is handled by a fresh kimi --print subprocess without --resume <session-id>, so Kimi has no memory of earlier messages in the same conversation.
The persisted agent_session_id in ~/.cc-connect/sessions/<project>_*.json remains an empty string forever.                                                                                          
  
Root cause: agent/kimi/session.go expects to capture Kimi's session UUID from the To resume this session: kimi -r <uuid> line on stdout, but kimi-cli writes that line to stderr. The stdout scanner therefore never sees it, and ks.sessionID is never set.

### Steps to Reproduce

1. Standalone proof that kimi-cli writes the resume hint to stderr (one command, no cc-connect involved):
                                                                                                                                                                                                       
  kimi --print --output-format stream-json --prompt "say hello" 2>stderr.txt 1>stdout.txt
                                                                                                                                                                                                       
  cat stdout.txt                                            
  # {"role":"assistant","content":[{"type":"think",...},{"type":"text","text":"Hello!..."}]}                                                                                                           
                                                                                                                                                                                                       
  cat stderr.txt
  #                                                                                                                                                                                                    
  # To resume this session: kimi -r 5663b2f2-6c1c-41be-ab51-f624cae29b97
                                                                                                                                                                                                       
  The UUID required for --resume is emitted on stderr, never on stdout.                                                                                                                                
                                                                                                                                                                                                       
  2. End-to-end reproduction inside cc-connect:                                                                                                                                                        
                                                            
  Minimal ~/.cc-connect/config.toml:                                                                                                                                                                   
  [[projects]]                                                                                                                                                                                         
    name = "kimi-repro"                                                                                                                                                                                                                         
    [projects.agent]
      type = "kimi"                                                                                                                                                                                                                                                                                                                                      
      [projects.agent.options]
        mode = "default"                                                                                                                                                                               
        work_dir = "/tmp/kimi-repro-work"    # any writable dir                                                                                                                                               
  [log]
    level = "debug"                                                                                                                                                                                                                                                                                                                                                         
  [bridge]
    enabled = true                                                                                                                                                                                     
    port = 9810                                             
  [management]
    enabled = true                                                                                                                                                                                     
    port = 9820
                                                                                                                                                                                                       
  Run cc-connect, open the Web UI, send two messages:                                                                                                                                                  
  
  1. "Please remember the number 42."                                                                                                                                                                  
  2. "What number did I just ask you to remember?"          
                                                                                                                                                                                                       
  Kimi cannot recall "42" on the second message. 

### Expected Behavior

- On turn 1, cc-connect captures Kimi's session UUID (which kimi-cli prints on exit) and stores it.
- On turn 2, the subprocess is launched with --resume <uuid>, so Kimi loads the turn-1 context and answers "42".                                                                                     
- ~/.cc-connect/sessions/<project>_*.json → agent_session_id holds the Kimi UUID so resume survives a cc-connect restart.  

### Actual Behavior

- Session UUID is never captured. ks.sessionID stays empty.
- Every turn launches Kimi without --resume, which Kimi treats as a brand-new session.                                                                                                               
- agent_session_id on disk stays "" forever.                                                                                                                                                         
- Kimi answers turn 2 with something like "You haven't told me a number yet."

```toml

```

### Logs / Error Output

```shell

```

### Additional Context

  The existing code already captures stderr into a buffer (var stderrBuf bytes.Buffer; cmd.Stderr = &stderrBuf) but only reads it on a non-zero exit. Under normal exit the buffer is discarded, dropping the "To resume this session:" line with it.                                                                                                                                                   
                                                                                                                                                                                                                                                                   
  Only one line of code needs to change: where the scan happens. Suggested minimal patch in agent/kimi/session.go, readLoop function, immediately after waitErr := cmd.Wait():                         
                                                                                                                                                                                                       
  // Kimi writes "To resume this session: kimi -r <uuid>" to stderr (not stdout),                                                                                                                      
  // so the stdout scanner above never sees it. Extract it from the captured                                                                                                                           
  // stderr buffer before emitting EventResult so the next turn can pass --resume.                                                                                                                     
  for _, line := range strings.Split(stderrBuf.String(), "\n") {                                                                                                                                       
      if strings.HasPrefix(strings.TrimSpace(line), "To resume this session:") {                                                                                                                       
          if id := extractResumeSessionID(line); id != "" {                                                                                                                                            
              ks.sessionID.Store(id)                                                                                                                                                                   
              slog.Debug("kimiSession: session id from stderr", "session_id", id)                                                                                                                      
          }                                                                                                                                                                                            
          break                                                                                                                                                                                        
      }                                                                                                                                                                                                
  }                                                                                                                                                                                                    
                                                            
  I've tested this locally against kimi-cli 1.31.0 and confirmed:                                                                                                                                      
  - Session UUID is captured (kimiSession: session id from stderr session_id=<uuid>)
  - Next turn's args include --resume <uuid>                                                                                                                                                           
  - Kimi correctly recalls prior turn content               
  - agent_session_id is persisted to disk                                                                                                                                                              
                                                                                                                                                                                                       
  Happy to turn this into a PR with a regression test if you'd like — just let me know.

Li-Zhouzi:
Fixes #758

Kimi CLI writes "To resume this session: kimi -r <uuid>" to stderr, but the scanner in readLoop only reads stdout, so the UUID was never captured. As a result, every turn spawned a fresh kimi process without --resume, and conversation context was lost between messages.

Scan the already-captured stderrBuf after cmd.Wait() to extract the UUID. All downstream plumbing (sessionID.Store, EventResult writeback, engine disk persistence, next-turn --resume arg) is unchanged.

Verified locally (against kimi-cli 1.31.0 on macOS arm64).

### DIFF
--- a/agent/kimi/session.go
+++ b/agent/kimi/session.go
@@ -242,6 +242,19 @@ func (ks *kimiSession) readLoop(ctx context.Context, cmd *exec.Cmd, stdout io.Re
 	// never sees EventError after EventResult from the same turn.
 	waitErr := cmd.Wait()
 
+	// Kimi writes "To resume this session: kimi -r <uuid>" to stderr (not stdout),
+	// so the scanner above never sees it. Extract it from the captured stderr
+	// buffer before emitting EventResult so the next turn can pass --resume.
+	for _, line := range strings.Split(stderrBuf.String(), "\n") {
+		if strings.HasPrefix(strings.TrimSpace(line), "To resume this session:") {
+			if id := extractResumeSessionID(line); id != "" {
+				ks.sessionID.Store(id)
+				slog.Debug("kimiSession: session id from stderr", "session_id", id)
+			}
+			break
+		}
+	}
+
 	if scanErr != nil {
 		slog.Error("kimiSession: scanner error", "error", scanErr)
 		evt := core.Event{Type: core.EventError, Error: fmt.Errorf("read stdout: %w", scanErr)}


chenhg5 (approved):
## Review Summary

Clean, surgical fix for Kimi session persistence. The bug was real — kimi-cli writes the resume UUID to stderr, not stdout, so the scanner never captured it.

---

✅ **Good**:
- Correctly scans `stderrBuf` (already captured) after `cmd.Wait()` — no new I/O or timing issues
- Reuses existing `extractResumeSessionID()` — no duplicated parsing logic
- Placed at the right point in the lifecycle: after `cmd.Wait()` but before `EventResult` emission
- Minimal change (+13 lines, 1 file)
- Verified locally against kimi-cli 1.31.0

---

LGTM. Fixes #758. Logic correct, CI green.

PR #777, opened 2026-04-25
fix(core): keep streaming preview as one in-place card under fully-quiet display

### LINKED ISSUE #776: fix(core): single-card streaming preview lost paragraph separators between text segments under fully-quiet display
## Repro config

```toml
[stream_preview]
enabled = true

[display]
thinking_messages = false
tool_messages = false

[[projects]]
# ... 

[[projects.platforms]]
type = "lark"   # or feishu, any platform that implements MessageUpdater + PreviewStarter
[projects.platforms.options]
progress_style = "card"
```

In other words: stream preview on, both thinking and tool messages hidden, and a platform that supports streaming-preview cards (Feishu/Lark, Telegram, etc.).

## Expected

A single card is in-place edited from the first streamed text right through to the final reply. Text segments that are bounded by tool calls are kept on the same card with a blank line between them so they don't visually merge.

## Observed

The turn is split into N+1 messages instead. Every `EventThinking` and `EventToolUse` boundary triggers `sp.freeze() + sp.detachPreview()` in `core/engine.go` (the `if !e.display.ThinkingMessages` / `if !e.display.ToolMessages` blocks), which commits the current preview as a permanent card and disables further updates. Subsequent text events fall back to fresh `p.Send` messages. Result: one frozen card per text segment plus a separate final reply, with no visual separator between segments.

## Why

The freeze + detach was inherited from a "compact" display mode design that emitted each text segment as its own card. The earlier three-mode `display.mode` enum (full / compact / quiet) had a separate "quiet" branch that called `sp.appendSeparator("\n\n")` instead of freezing — so a fully-quiet turn stayed on one card with paragraph breaks between segments.

When `display.mode` was replaced by the two booleans `display.thinking_messages` and `display.tool_messages`, the freeze branch survived but the appendSeparator branch was dropped. Today there is no path that exercises the "quiet" UX even though both knobs being false is exactly the configuration that should produce it.

## Fix shape

Restore `streamPreview.appendSeparator(sep) bool` and call it from the `!ThinkingMessages` / `!ToolMessages` branches when `sp.canPreview()` is true, mirroring the separator into `textParts` so the final response text matches the live preview. Keep the degraded-preview fallback (send the accumulated segment as a standalone message) so platforms without message-edit support keep their previous behavior. Pure UX fix — no semantic changes to other display modes or to tool/thinking handling.

PR with full implementation, tests, and regression coverage to follow.

Cigarrr:
Closes #776.

## Problem

With `stream_preview.enabled = true`, `progress_style = "card"`, `display.thinking_messages = false`, and `display.tool_messages = false`, a turn that includes tool calls is rendered as multiple messages instead of a single edit-updated card. Every `EventThinking` and `EventToolUse` boundary triggers `sp.freeze() + sp.detachPreview()`, which commits the current preview as a permanent card and disables further updates. Subsequent text events fall back to fresh `p.Send` messages. Adjacent text segments also visually merge because nothing inserts a separator between them.

## Fix

Restore the "quiet" branch that the earlier three-mode `display.mode` enum used: at hidden segment boundaries, append a paragraph separator into the live preview instead of freezing it.

* `core/streaming.go`: add `streamPreview.appendSeparator(sep) bool`. It appends the separator into `fullText` without sending an update. Returns `false` when the preview is degraded, disabled, or has no content yet so we never produce a leading separator on an otherwise empty card.

* `core/engine.go`: in the `!ThinkingMessages` and `!ToolMessages` branches, replace `sp.freeze() + sp.detachPreview()` with `sp.appendSeparator("\n\n")` when `sp.canPreview()` is true, and mirror the separator into `textParts` so the final response text matches the live preview. The degraded-preview fallback (send the accumulated segment as a standalone message) is unchanged so platforms without message-edit support keep their previous behavior.

This is purely a streaming-preview UX fix; tool/thinking semantics, segmentation logic for visible events, and the other display modes are untouched.

## Compatibility

* `display.thinking_messages = true` or `display.tool_messages = true` (default): no behavior change. The visible-event branches still flush the segment, freeze, and emit the thinking/tool message as before.
* Platforms without `MessageUpdater` (legacy preview-degraded path): no behavior change. The accumulated segment is still flushed via `p.Send` at the boundary.

## Tests

`go test -count=1 -tags no_web ./...` passes (3 pre-existing failures in `core/` are unrelated — they fail on the unmodified `main` branch as well on macOS due to a `/var` ↔ `/private/var` symlink in the path-resolution tests: `TestProcessInteractiveEvents_AppendsReplyFooterWhenEnabled`, `_ReplyFooterPrefersSessionRuntimeState`, `TestResolveLocalDirPath_AcceptsSubdir`).

New tests:

* `core/streaming_test.go`:
  * `TestStreamPreview_AppendSeparator_AppendsAfterText`
  * `TestStreamPreview_AppendSeparator_SuppressedOnEmptyPreview`
  * `TestStreamPreview_AppendSeparator_SuppressedAfterFreeze`
  * `TestStreamPreview_AppendSeparator_DoesNotTriggerFlush`
* `core/engine_test.go`:
  * `TestProcessInteractiveEvents_FullyQuietKeepsSinglePreviewAcrossToolCalls` — exercises the engine event loop end-to-end with the bug repro (text → tool → text → tool → text → result) and asserts a single preview start, an in-place final update containing every segment, and a paragraph break between segments.
  * `TestProcessInteractiveEvents_FullyQuietInsertsBlankLinesBetweenSegments` — focused regression for the separator restoration.

## Test plan

- [x] `go vet -tags no_web ./...`
- [x] `go test -count=1 -tags no_web ./...`
- [x] Live testing on Lark / Feishu under the repro config: a multi-tool turn now stays on a single in-place updated card and adjacent text segments are separated by a blank line.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2731,11 +2731,25 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 
 		switch event.Type {
 		case EventThinking:
-			// In quiet mode, still split text segments so they don't merge.
+			// When thinking messages are hidden, this event won't surface a
+			// visible message of its own. To keep the streaming preview
+			// rendering one in-place updated card across the whole turn
+			// (rather than freezing it into a permanent message and
+			// emitting subsequent text as fresh sends), insert a paragraph
+			// separator into the live preview instead of freezing. The
+			// separator is mirrored into textParts so the final response
+			// text includes the same break that appeared on screen.
+			//
+			// When the preview is degraded (platform doesn't support edits,
+			// or a prior visible thinking/tool event already froze it),
+			// fall back to flushing the accumulated text segment as a
+			// standalone message — matches the legacy behavior for those
+			// platforms.
 			if !e.display.ThinkingMessages && len(textParts) > segmentStart {
 				if sp.canPreview() {
-					sp.freeze()
-					sp.detachPreview()
+					if sp.appendSeparator("\n\n") {
+						textParts = append(textParts, "\n\n")
+					}
 				} else {
 					// Preview degraded — send accumulated text directly
 					segment := strings.Join(textParts[segmentStart:], "")
@@ -2744,9 +2758,9 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 							sendWorkspace(p, replyCtx, chunk)
 						}
 					}
+					segmentStart = len(textParts)
+					silentHold = false
 				}
-				segmentStart = len(textParts)
-				silentHold = false
 			}
 			if e.display.ThinkingMessages && event.Content != "" {
 				// Flush accumulated text segment before thinking display
@@ -2776,11 +2790,17 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 
 		case EventToolUse:
 			toolCount++
-			// When tool messages are hidden, split text segments.
+			// When tool messages are hidden, this event won't surface a
+			// visible message of its own. To keep the streaming preview
+			// rendering one in-place updated card across the whole turn,
+			// insert a paragraph separator into the live preview instead
+			// of freezing it. See the symmetric EventThinking case above
+			// for rationale and the degraded-preview fallback.
 			if !e.display.ToolMessages && len(textParts) > segmentStart {
 				if sp.canPreview() {
-					sp.freeze()
-					sp.detachPreview()
+					if sp.appendSeparator("\n\n") {
+						textParts = append(textParts, "\n\n")
+					}
 				} else {
 					// Preview degraded — send accumulated text directly
 					segment := strings.Join(textParts[segmentStart:], "")
@@ -2789,9 +2809,9 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 							sendWorkspace(p, replyCtx, chunk)
 						}
 					}
+					segmentStart = len(textParts)
+					silentHold = false
 				}
-				segmentStart = len(textParts)
-				silentHold = false
 			}
 			if e.display.ToolMessages {
 				// Flush accumulated text segment before tool display
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -1192,6 +1192,121 @@ func TestProcessInteractiveEvents_SuppressesReplyFooterWhenOnlyWorkDir(t *testin
 	}
 }
 
+// TestProcessInteractiveEvents_FullyQuietKeepsSinglePreviewAcrossToolCalls is
+// the regression test for the bug where stream_preview + progress_style=card
+// + ThinkingMessages=false + ToolMessages=false produced multiple messages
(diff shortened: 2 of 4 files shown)


chenhg5 (approved):
## Review Summary

Targeted fix for a real UX regression. The new `appendSeparator` approach is clean and well-tested.

---

✅ **Good**:
- Correctly replaces `freeze() + detachPreview()` with `appendSeparator("\n\n")` at hidden event boundaries
- Preview card survives across tool/thinking boundaries in fully-quiet mode — single card instead of N+1 messages
- Symmetric handling for EventThinking and EventToolUse (same pattern, consistent fallback)
- `appendSeparator` correctly suppressed when: preview is degraded, preview disabled, or no content yet
- 2 integration tests (multi-tool, blank-line separator) + 4 unit tests for appendSeparator
- `segmentStart`/`silentHold` now correctly scoped inside the else branch (only for degraded fallback)

---

LGTM. Logic correct, well-tested, CI green. Closes #776.

happyTonakai:
it seems to be duplicate with #655, i think this was introduced by #642

Cigarrr:
@happyTonakai you're right — #655 and this PR fix the same UX bug (#642 introduced `freeze()+detachPreview()` at thinking/tool boundaries in fully-quiet mode, which split a single turn into multiple cards). After reading #655 carefully:

**Engine layer is essentially identical** — both PRs add `streamPreview.appendSeparator()` and replace the freeze+detach with it at hidden segment boundaries. The runtime behavior in single-card mode is the same.

**Schema layer differs**:
- This PR: keeps `thinking_messages`/`tool_messages` booleans; when both are `false`, automatically restores single-card behavior (the implicit pre-#642 semantics).
- #655: introduces explicit `display.mode = full | compact | quiet` enum. Only `mode="quiet"` triggers single-card; `mode="compact"` keeps the post-#642 multi-card behavior under a named mode.

**#655 is the better choice IMO** because:
1. Explicit naming for the two distinct UX intents ("compact: hide chatter, still segment cards" vs "quiet: hide chatter, single card") is clearer than reading boolean combinations.
2. Backward-compat for legacy `quiet=true` is handled in `EffectiveDisplay` mode resolution.
3. Avoids two equivalent config paths (`mode=quiet` vs `thinking_messages=false && tool_messages=false`).

@chenhg5 — recommend merging #655 over this PR. I'll keep #777 open until you decide; if you go with #655 I'll close this. If you prefer the no-schema-change approach, this one's ready.

chenhg5:
关闭此 PR，功能已通过 #655 合并进 main（commit 9989fdc4）。

PR #655 引入了完整的三模式展示系统（full/compact/quiet），其中的 `appendSeparator` 实现覆盖了此 PR 的修复目标：在 `display.mode = "quiet"` 时，thinking/tool 边界使用 `appendSeparator("\n\n")` 代替 `freeze+detachPreview()`，从而保持所有文本在同一张 in-place 更新卡片中。

感谢贡献！

PR #787, opened 2026-04-26
fix(codex): handle app-server approvals and idle completion

0KABE:
This PR fixes codex app-server session handling in cc-connect by adding support for server-initiated approval requests and by improving turn completion detection for newer Codex versions.

Previously, the app-server adapter mainly handled client-initiated RPC requests and notifications. With newer Codex behavior, the server may also send approval requests that require a response, and turn completion may be signaled via thread status changes instead of only turn/completed. This patch updates the adapter to handle those cases correctly.

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -135,6 +135,9 @@ type appServerSession struct {
 	pendingMu sync.Mutex
 	pending   map[int64]chan rpcResponseEnvelope
 
+	approvalsMu     sync.Mutex
+	pendingApprovals map[string]chan core.PermissionResult
+
 	threadID atomic.Value
 	alive    atomic.Bool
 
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
## Review Summary

Solid implementation of server-initiated approval requests and improved turn completion detection for Codex app-server. The code is well-structured with proper concurrency handling.

---

✅ **Good**:

1. **Approval flow**: Clean separation of `pendingApprovals` map with proper mutex protection. The goroutine-based timeout pattern (5min) with select on ctx/timer is correct.

2. **Message type discrimination**: The `readLoop` switch on `hasID && !hasMethod` / `hasID && hasMethod` / `default` is a clean way to distinguish responses, server requests, and notifications.

3. **Turn completion deduplication**: `completeTurn()` checks `currentTurn` and clears it atomically, preventing duplicate `EventResult` emissions.

4. **`thread/status/changed` handling**: Correctly detects codex 0.125+ idle status as turn completion signal.

5. **Cleanup**: `rejectPendingApprovals()` properly drains all pending channels on connection close.

6. **Tool name mapping**: `handleApprovalRequest` translates `item/commandExecution/requestApproval` → "Bash", `item/fileChange/requestApproval` → "Patch" for user-friendly display.

---

💭 **Observations** (non-blocking):

- `connect()` now hardcodes `--listen stdio://` instead of conditionally adding `--listen` when URL is set. This is correct for the stdio transport use case.
- `handleDynamicToolCall` returns a stub error response — appropriate for now since dynamic tools aren't supported.

---

LGTM. CI green, well-tested pattern, addresses real Codex protocol evolution.

PR #792, opened 2026-04-27
feat(config): per-project [projects.display] override

Cigarrr:
## Summary

Allow each `[[projects]]` entry to override individual fields of the global `[display]` block, on a per-field basis. Useful when you want one project to run quiet while others stay verbose, or vice versa, without splitting into separate cc-connect instances.

## Schema

```toml
[display]
thinking_messages = true
tool_messages = true

[[projects]]
name = "noisy-project"
[projects.display]
thinking_messages = false  # overrides only this field
tool_messages = false
```

Each sub-field is independently optional; unset fields fall back to the global `[display]` value, then to the built-in default.

## Resolution order (per field)

1. `[projects.display].<field>`  — highest precedence
2. `[display].<field>`
3. built-in default

All four `DisplayConfig` fields participate (`ThinkingMessages`, `ToolMessages`, `ThinkingMaxLen`, `ToolMaxLen`).

## Backward compatibility

- Projects with no `[projects.display]` block see no behavior change.
- Legacy `quiet` (root or per-project) is preserved: if quiet is in effect AND neither layer set `thinking_messages` / `tool_messages`, they default to `false` — same as before, just extended to honor project-level overrides when present.

## Tests

7 new sub-tests in `TestEffectiveDisplay_ProjectOverride` cover:

- project overrides global thinking_messages
- project unset falls back to global
- both unset falls back to default
- project overrides max-len fields
- project quiet still respected when project display unset
- project display.thinking_messages true overrides project quiet
- nil project display behaves like before

`go test ./...` passes (the 4 pre-existing failures on `main` — `TestProcessInteractiveEvents_AppendsReplyFooterWhenEnabled`, `TestProcessInteractiveEvents_ReplyFooterPrefersSessionRuntimeState`, `TestResolveLocalDirPath_AcceptsSubdir`, `TestGetModelAndReasoningEffort_FromRuntimeConfigWhenUnset` — are unrelated to this change).

## Docs

`config.example.toml` updated with a commented-out `[projects.display]` example next to the global `[display]` section.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -92,6 +92,20 @@ level = "info" # debug, info, warn, error
 # tool_max_len = 500       # Max chars for tool use messages (default: 500) / 工具调用消息最大字符数（默认 500）
 # tool_messages = true     # Show/hide tool progress messages (default: true) / 是否显示工具进度消息（默认 true）
 
+# Per-project override: each [[projects]] entry may have its own [projects.display]
+# block that overrides individual fields of the global [display] block above.
+# Resolution order is per-field — set fields win, unset fields fall back to global,
+# then to built-in defaults. Useful when one project should be quiet while others
+# stay verbose, or vice versa.
+# 每个 [[projects]] 条目可以有自己的 [projects.display] 块，逐字段覆盖全局
+# [display]。已设字段优先，未设字段回落到全局值，再回落到默认值。
+#
+# [[projects]]
+# name = "noisy-project"
+# [projects.display]
+# thinking_messages = false  # Override only this field; others inherit from global
+# tool_messages = false
+
 # Agent idle timeout: max minutes between consecutive agent events before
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
## Review Summary

Clean implementation of per-project `[projects.display]` override. The per-field resolution logic is well-designed and backward-compatible.

---

✅ **Good**:

1. **Per-field resolution**: `pickBool` / `pickInt` helpers cleanly implement the precedence: project → global → default.

2. **Legacy quiet preserved**: When `quiet=true` AND neither layer explicitly sets `thinking_messages` / `tool_messages`, they default to false. Project-level `[projects.display]` takes precedence over both global display and quiet.

3. **Comprehensive tests**: 7 test cases in `TestEffectiveDisplay_ProjectOverride` cover all resolution paths.

4. **Documentation**: Clear config example showing the per-field override pattern.

---

LGTM. CI green.

PR #793, opened 2026-04-27
fix(core): reset replyCtx on inline queued-message dequeue

Cigarrr:
## Summary

When a queued message is dequeued mid-loop in `processInteractiveEvents`, the per-turn locals (`textParts`, `segmentStart`, `toolCount`, `turnStart`, etc.) get reset for the new turn — but the function-scope `replyCtx` parameter was left pointing at the original turn's trigger context.

**Symptom:** when msg2 arrives while msg1 is still being processed, msg2's final reply ends up `Send`/`Reply`'d against msg1's reply context. On platforms that derive the parent message_id from `replyCtx` for the reply quote (e.g. feishu), msg2's reply visibly quotes msg1's bubble instead of msg2's.

## Repro

1. Send msg1 to a busy session
2. While the agent is still processing msg1, send msg2
3. Wait for the agent to finish msg1, then start replying to msg2
4. The reply card / message for msg2 will quote msg1 in its parent reference

## Fix

`state.replyCtx` is correctly updated to `queued.replyCtx` in the same dequeue block — only the local function-scope variable was missing. Adding the reassignment alongside the other per-turn resets is sufficient.

`drainPendingMessages` (the cleanup path called when the loop exits cleanly) was already correct because it re-enters `processInteractiveEvents` with `queued.replyCtx` as the parameter — only the inline mid-loop dequeue path was affected.

## Tests

Added `TestProcessInteractiveEvents_QueuedMessageUsesItsOwnReplyCtx` using a recording stub platform that captures the `(replyCtx, content)` pair for each `Send`/`Reply` call and asserts:

- turn1's response uses `ctx-turn1`
- turn2's response uses `ctx-turn2`

Verified the test catches the bug (fails before the fix, passes after).

`go test ./...` passes (the 4 pre-existing failures on `main` are unrelated to this change).

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -3248,6 +3248,13 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 				turnStart = time.Now()
 				firstEventLogged = false
 				waitStart = time.Now()
+				// Reassign the local replyCtx parameter to the queued message's
+				// trigger context. state.replyCtx was updated above, but the
+				// function-scope replyCtx is what gets passed to p.Send / p.Reply
+				// further down — and platforms derive the parent message_id from
+				// it for the reply quote. Without this reassignment, msg2's
+				// reply would quote msg1's bubble.
+				replyCtx = queued.replyCtx
 				queuedRenderer := func(content string) string {
 					return e.renderOutgoingContentForWorkspace(queued.platform, content, workspaceDir)
 				}
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -6405,6 +6405,114 @@ func TestProcessInteractiveEvents_DrainsQueuedMessages(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
## Review Summary

Surgical one-line fix for a subtle but impactful bug. When a queued message is dequeued mid-loop, the function-scope `replyCtx` was still pointing at the original turn's trigger context, causing msg2's reply to quote msg1's bubble.

---

✅ **Good**:

1. **Minimal change**: Single line `replyCtx = queued.replyCtx` after the queued message dequeue reset block.

2. **Clear comment**: Explains why the reassignment is needed (platforms derive parent message_id from replyCtx for reply quote).

3. **Strong test**: `TestProcessInteractiveEvents_QueuedMessageUsesItsOwnReplyCtx` records each Send/Reply call's replyCtx and asserts turn2's reply uses `ctx-turn2`, not `ctx-turn1`.

---

LGTM. CI green.

Issue #421, opened 2026-04-02
Support Telegram forum topics: include message_thread_id in session key


johnil:
## Problem

When cc-connect is used with a Telegram group that has **forum topics** enabled, all messages from different topics are routed to the same session. This means the bot cannot distinguish which topic a message came from, and replies may go to the wrong topic.

## Root Cause

The current `CC_SESSION_KEY` format is `telegram:<chat_id>:<user_id>`, which does not include Telegram's `message_thread_id`. In forum-enabled groups, each topic has a unique `message_thread_id`, but cc-connect ignores it.

Example:
- Current: `telegram:-1003883543771:1064655557`
- Expected: `telegram:-1003883543771:1064655557:<message_thread_id>`

## Expected Behavior

- When a Telegram group has forum topics enabled, `message_thread_id` should be included in the session key
- Messages from different topics should be routed to separate sessions
- Replies should be sent back to the correct topic

## Additional Context

Telegram Bot API provides `message_thread_id` and `is_topic_message` fields in the `Message` object for forum groups. These fields can be used to differentiate topics.

Reference: https://core.telegram.org/bots/api#message

chenhg5:
This was already implemented in PR #321 (commit 1104739c). Session keys include `message_thread_id` for forum topics, and all send methods use `MessageThreadID` from the reply context. Closing as already resolved.

PR #804, opened 2026-04-28
fix(telegram): preserve private topic session keys

pjpjq:
## Summary

- Preserve Telegram `message_thread_id` whenever Telegram provides it, instead of only when `chat.is_forum` is true.
- This lets private chat topics and group forum topics map to distinct Telegram session keys.
- Update the Telegram regression test and docs for topic-based sessions.

## Related discussion

- Related to #421, which added/covered group forum topic session isolation.
- I searched existing issues and PRs for `Telegram topic message_thread_id session`; I did not find an open duplicate for private chat topics.

## Reproduction

Environment tested locally:

- Version before patch: `cc-connect v1.3.2`
- OS: macOS arm64
- Install method: npm global package
- Agent/platform: Codex + Telegram

Steps:

1. Start cc-connect with a Telegram bot.
2. Open a Telegram private chat topic with that bot.
3. Send messages in different topics.

Before this change, private chat topic messages were handled with the same session key because the code only used `MessageThreadID` when `Chat.IsForum` was true.

## Expected behavior

Messages from different Telegram topics should use different session keys when Telegram includes `message_thread_id`, so each topic has an independent conversation context.

## Actual behavior before this PR

Private chat topics could be collapsed into the base private chat session key, for example:

```text
telegram:680002712
```

After applying this change locally, a private topic message used a topic-aware session key, for example:

```text
telegram:680002712:34923
```

## Breaking changes

None expected. Messages without `message_thread_id` still use the existing non-topic session key formats.

## Validation

Passed:

```bash
pnpm --dir web install --frozen-lockfile
pnpm --dir web build
go test ./platform/telegram
go build -tags 'no_web' ./cmd/cc-connect
```

Also verified against a real local Telegram bot: a private chat topic message created/used `telegram:680002712:34923` instead of the base `telegram:680002712` session.

Attempted full validation as requested by CONTRIBUTING:

```bash
go test ./...
```

On my local macOS machine this still fails in packages unrelated to this patch:

- `agent/codex`: `TestGetModelAndReasoningEffort_FromRuntimeConfigWhenUnset` failed once in the full suite, but passes when run directly.
- `core`: existing macOS path expectation failures involving `/var/...` vs `/private/var/...`, plus reply footer path truncation expectations.

The Telegram package touched by this PR passes.

### DIFF
--- a/docs/telegram.md
+++ b/docs/telegram.md
@@ -185,6 +185,13 @@ level=INFO msg="cc-connect is running" projects=1
 3. Search and add your bot
 4. Send messages in the group
 
+### 6.3 Topic Sessions
+
+Telegram topics include a `message_thread_id`. cc-connect uses that thread ID
+as part of the Telegram session key, so each topic has its own independent
+conversation context. This applies to forum topics in groups and private chat
+topics when Telegram includes `message_thread_id`.
+
 ---
 
 ## Usage Example
--- a/platform/telegram/telegram.go
+++ b/platform/telegram/telegram.go
@@ -333,10 +333,7 @@ func (p *Platform) handleMessage(ctx context.Context, msg *models.Message) {
 		userName = strings.TrimSpace(msg.From.FirstName + " " + msg.From.LastName)
 	}
 
-	threadID := 0
-	if msg.Chat.IsForum {
-		threadID = msg.MessageThreadID
-	}
+	threadID := msg.MessageThreadID
 	sessionKey := p.buildSessionKey(msg.Chat.ID, threadID, msg.From.ID)
 	channelKey := buildChannelKey(msg.Chat.ID, threadID)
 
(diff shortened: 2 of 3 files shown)


chenhg5 (approved):
## Review Summary

Clean fix for private chat topic session isolation. Previously `message_thread_id` was only used when `chat.is_forum` was true, ignoring private chat topics.

---

✅ **Good**:

1. **Minimal change**: Removes the `if msg.Chat.IsForum` condition, directly using `msg.MessageThreadID` (defaults to 0 when not present).

2. **Applied to both paths**: `handleMessage` and `handleCallbackQuery` both updated consistently.

3. **Test updated**: `TestHandleMessageNonForumIgnoresThreadID` renamed to `TestHandleMessagePrivateTopicUsesThreadID`, now expects `telegram:100:55:7` session key with thread ID preserved.

4. **Docs updated**: New section 6.3 explains topic session behavior for forum and private chat topics.

5. **Zero default is safe**: When `MessageThreadID` is 0 (no topic), session key remains unchanged (`telegram:100:7`).

---

LGTM. CI green.

PR #811, opened 2026-04-29
fix(claudecode): propagate cli_path and other options to per-workspace agents

0xsegfaulted:
## Summary

Two related multi-workspace fixes for the claudecode + Discord stack.

### 1. claudecode: propagate options to per-workspace agents

claudecode agent didn't implement `core.WorkspaceAgentOptionSnapshotter`, so when multi-workspace mode lazily spawned per-workspace agents via `core.Engine.getOrCreateWorkspaceAgent`, every claudecode-specific opts field — `cli_path`, `cli_args_flag`, `allowed_tools`, `disallowed_tools`, `reasoning_effort`, `max_context_tokens`, `router_url`, `router_api_key` — was silently dropped. Only `model` / `mode` / `run_as_user` / `run_as_env` survived (each via a dedicated accessor path).

Concrete user-visible symptom: `cli_path = "claude --add-dir /shared"` worked on the project-level agent but workspace agents reverted to bare `claude` with none of the extra args.

`WorkspaceAgentOptions()` now snapshots all of these. `cli_path` is rebuilt from `cliBin + cliExtraArgs`. `run_as_user` / `run_as_env` stay out of the snapshot because the engine has dedicated `GetRunAsUser` / `GetRunAsEnv` propagation paths (cc-connect#496).

### 2. Discord: make `thread_isolation` compatible with multi-workspace auto-bind

When `thread_isolation = true`, Discord rewrites a guild message's `SessionKey` from `discord:<channelID>:<userID>` to `discord:<threadID>` so each thread carries its own conversation. The platform never set `Message.ChannelKey`, so `engine.effectiveChannelID(msg)` extracted the **thread** ID from `SessionKey`. Multi-workspace auto-bind by convention (`<base_dir>/<channel-name>` → workspace) then matched the **thread name** — which never matches a workspace folder. Auto-bind silently broke for every thread-isolated channel.

`Message.ChannelKey` is now set to the parent channel ID in all three Discord message paths (regular messages, slash commands, component callbacks). `resolveThreadReplyContext` returns the parent channel alongside the thread session key.

The workspace binding is now stored under the parent channel, but the interactive state is still keyed under the thread-keyed sessionKey. `interactiveKeyForSessionKey` and `sessionContextForKey` now use a unified precedence:

1. **Exact** sessionKey match in `interactiveStates` (raw placeholders win).
2. **Channel-binding-derived** workspace — deterministic; the current binding wins over stale state from a re-bound channel.
[...]

PR #812, opened 2026-04-29
feat(claudecode): support project-level environment variables

dlutsniper:
Add parsing of [projects.agent.options.env] in config.toml to allow per-project environment variable overrides (e.g. ANTHROPIC_BASE_URL, ANTHROPIC_AUTH_TOKEN, ANTHROPIC_MODEL) instead of requiring a single global configuration in ~/.claude/settings.json.

- Parse env map from opts["env"] in New(), supporting both map[string]string and map[string]any TOML type variants
- Merge sessionEnv into the existing three-layer env resolution chain (provider < sessionEnv < router) via core.MergeEnv()
- Add example config with Chinese comments in config.example.toml
- Add 4 unit tests in project_env_test.go

This enables different projects to use different LLM providers (e.g. Kimi, DeepSeek) without conflicting with each other.

Made-with: Claude Code


chenhg5 (changes requested):
## Review Summary

Good feature, but CI is failing because the new tests call `New()` which requires `claude` CLI to be on PATH.

---

🔴 **Blocker: CI unit-test failure**

```
--- FAIL: TestNew_ParsesProjectEnvFromOpts (0.00s)
    project_env_test.go:58: New() error: claudecode: "claude" CLI not found in PATH
--- FAIL: TestNew_ParsesProjectEnvFromMapStringAny (0.00s)
--- FAIL: TestNew_NoEnvOpts (0.00s)
--- FAIL: TestNew_ProjectEnvOverridesProviderEnv (0.00s)
```

**Fix options:**

1. **Use fake CLI** (preferred): See `TestWorkspaceAgentOptions_RoundTripsThroughNew` in #811 — set `opts["run_as_user"] = "skip-lookpath"` to bypass LookPath check on Unix.

2. **Test parsing logic directly**: Instead of calling `New()`, test the env parsing helper function in isolation.

3. **Skip on CI**: Use build constraint or environment check, but this is less preferred.

---

The feature implementation looks reasonable — will approve once CI is green.

chenhg5 (approved):
## Review Summary

Good feature for per-project environment variables. CI is now green after the test fix.

---

✅ **Good**:

1. **Env parsing**: Handles both `map[string]string` and `map[string]any` TOML variants.

2. **Test fix**: Uses `run_as_user: "skip-lookpath"` to bypass LookPath check on Unix (as suggested).

3. **Config example**: Clear documentation of `[projects.agent.options.env]` usage.

4. **Comprehensive tests**:
   - `TestNew_ParsesProjectEnvFromOpts` — map[string]string variant
   - `TestNew_ParsesProjectEnvFromMapStringAny` — map[string]any variant
   - `TestNew_NoEnvOpts` — no env in opts
   - `TestNew_ProjectEnvOverridesProviderEnv` — override behavior

---

LGTM. CI green.

PR #815, opened 2026-04-29
fix(feishu): notify user when media download fails

kylin19860916:
## Summary
- When image, audio, or file downloads fail due to network errors (e.g. `connection reset by peer`), the message was **silently dropped** with only a server-side log entry
- Users had no indication their message wasn't received and would wait indefinitely for a response
- Now sends a notification back to the user (via `p.Send`) so they know to resend

## Changes
Three locations in `platform/feishu/feishu.go` where `downloadImage` / `downloadResource` errors caused silent `return`:
- **Image download failure** (line ~894) — now notifies user
- **Audio/voice download failure** (line ~918) — now notifies user  
- **File download failure** (line ~963) — now notifies user

## Context
Encountered this in production: a Feishu voice message download failed due to a TCP connection reset. The bot went silent for 25 minutes with no feedback to the user. The error was only visible in server logs.

## Test plan
- [x] Compiles successfully (`go build ./cmd/...`)
- [ ] Trigger media download failure (e.g. network interruption) and verify notification is sent
- [ ] Verify normal media processing still works unchanged

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -892,6 +892,9 @@ func (p *Platform) dispatchMessage(ctx context.Context, msgType, content string,
 		imgData, mimeType, err := p.downloadImage(messageID, imgBody.ImageKey)
 		if err != nil {
 			slog.Error(p.tag()+": download image failed", "error", err)
+			if sendErr := p.Send(ctx, rctx, "⚠️ Image download failed (network error). Please resend."); sendErr != nil {
+				slog.Error(p.tag()+": failed to notify user about image download failure", "error", sendErr)
+			}
 			return
 		}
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
## Review Summary

Simple UX improvement. When media downloads fail, users now get a notification instead of silent drops.

---

✅ **Good**:

1. **Consistent pattern**: All three locations (image, audio, file) now notify user on download failure.

2. **Friendly messages**: Clear English text with emoji and "Please resend" guidance.

3. **Error handling**: If `p.Send` fails, logs the error but doesn't propagate (best-effort notification).

4. **Minimal diff**: +9 lines, no unnecessary changes.

---

LGTM. CI green.

PR #822, opened 2026-04-30
fix: use question text as answers key for AskUserQuestion

SXongin:
## Problem

`AskUserQuestion` tool results show empty answers when used through cc-connect.

The `buildAskQuestionResponse` function uses numeric string indices as keys in the `answers` map, but Claude Code appears to expect question text as key.

**Evidence (binary analysis, not official docs):** Used `strings` on Claude Code 2.1.112 and 2.1.123 binaries. Both show:

```
answers: E.record(E.string(), E.string())
  .describe("The answers provided by the user (question text -> answer string)")
```

And the rendering function uses `Object.entries(answers)` to iterate, displaying each entry as `key → value`.

> **Note for reviewers:** This analysis is based on reverse-engineering the Claude Code binary, not official protocol documentation. I could not find public docs for the `--permission-prompt-tool stdio` protocol. If you have access to internal protocol specs or can confirm the expected format, please verify before merging.

## Fix

Change `buildAskQuestionResponse` to use `questions[idx].Question` as the answers key instead of `strconv.Itoa(idx)`:

```go
// Before
answers[strconv.Itoa(idx)] = ans

// After
answers[questions[idx].Question] = ans
```

## Backward Compatibility

Observed the same `E.record(E.string(), E.string())` schema in both 2.1.112 and 2.1.123 binaries, suggesting the format has not changed between versions. However, this is also based on binary analysis — please verify if you have access to version-specific protocol documentation.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2049,7 +2049,9 @@ func buildAskQuestionResponse(originalInput map[string]any, questions []UserQues
 	}
 	answers := make(map[string]any)
 	for idx, ans := range collected {
-		answers[strconv.Itoa(idx)] = ans
+		if idx >= 0 && idx < len(questions) {
+			answers[questions[idx].Question] = ans
+		}
 	}
 	result["answers"] = answers
 	return result
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -4815,16 +4815,16 @@ func TestBuildAskQuestionResponse(t *testing.T) {
 		"questions": []any{map[string]any{"question": "Which?"}},
 	}
 	collected := map[int]string{0: "PostgreSQL", 1: "Gin"}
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
LGTM.

This fixes the AskUserQuestion response shape by keying `answers` with the question text expected by Claude Code instead of numeric indices. The change is narrowly scoped, preserves the original tool input, and updates the relevant tests for single and multi-question flows.

CI is green: lint, unit-test, smoke-test, regression-test, and performance-test all passed.

One non-blocking note: if Claude Code ever emits duplicate question text in one AskUserQuestion payload, the map key format would collapse answers. That appears to be a protocol-level ambiguity rather than an issue introduced by this PR.

PR #824, opened 2026-04-30
fix(feishu): send audio and video files as native messages

CodeEagle:
## Summary
- send uploaded opus/audio files with Feishu msg_type=audio instead of file
- send uploaded mp4 files with Feishu msg_type=media instead of file
- detect common audio MIME types/extensions and convert non-opus audio to opus before sending
- include Feishu upload duration for native audio/video messages so clients display the correct length

## Testing
- go test -tags no_web ./cmd/cc-connect ./core ./platform/feishu
- git diff --check
- manually deployed the branch and sent both a 3.4s voice file and a 4.2s mp4 through Feishu; cc-connect send returned success for both


chenhg5 (changes requested):
## Review Summary

Thanks for the focused Feishu media fix. The native `audio`/`media` routing and upload duration handling match Feishu's API shape, and the added unit tests cover the new helper logic well. I found one compatibility regression that should be fixed before merge.

🔴 **Blocker: non-opus audio attachments now fail when ffmpeg conversion is unavailable**

`SendFile` now classifies common audio attachments such as `.mp3`, `.wav`, `.m4a`, etc. as Feishu `opus` uploads and calls `core.ConvertAudioToOpus` before sending. If `ffmpeg` is not installed, or conversion fails for any supported-looking file, the whole file send returns an error. Before this PR those attachments were still deliverable as regular files.

Please preserve the existing generic file-send behavior as a fallback: try native audio only when conversion succeeds, otherwise upload/send as `stream` + `file` like the previous implementation. Please also add a test for the fallback path.

CI: all GitHub checks are green.

chenhg5 (approved):
LGTM — the requested change (fallback when ffmpeg conversion fails) has been addressed:

✅ **ffmpeg fallback added**: when audio conversion fails, code now falls back to `larkim.FileTypeStream` (generic file download) with a `slog.Warn` log, preserving the original behavior for the user rather than breaking the file delivery entirely.

✅ **Extended audio format detection**: `detectFeishuAudioFormat` now covers opus/ogg/mp3/wav/m4a/aac/flac/webm/amr/silk — comprehensive format support.

✅ **CI all green** (all 5 checks pass including unit-test). PR also incorporates the `NewBridgeServerInsecure` test fix from #851 — note this means #824 and #851 both modify `core/bridge_capabilities_snapshot_test.go`. Whichever merges second may get a trivial conflict on that file; easy to resolve by keeping one copy.

**APPROVED.**

PR #826, opened 2026-04-30
fix(core): drop stale platform permission callbacks instead of forwarding to agent

chenhg5:
## Summary

When a user taps an old **Allow / Deny** inline button on Telegram (or the equivalent action button on a Feishu permission card) **after** the corresponding pending permission request has already been resolved — for example, the session was reset, the bot restarted, the user retried the prompt, or another user in the group already responded — the platform layer would synthesize a `core.Message` with `Content = "allow" / "deny" / "allow_all"` and dispatch it through the normal message handler.

Previously this synthetic click fell through to **regular user-message handling**:

1. `handlePendingPermission` returned `false` (no pending request found for the session).
2. The engine treated the message as a normal user prompt.
3. If a turn was already running, the engine replied with the `"message received, will process after the current turn"` toast.
4. The literal string `"allow"` / `"deny"` was eventually delivered to the agent as user input — leaking a UI control word into the agent's prompt stream.

This PR fixes that by tagging permission-callback messages with a new flag on `core.Message` and teaching the engine to drop them silently when no matching pending request exists.

### User-visible change

Reported by users on Telegram:

> Clicking the approval menu after the request had already been answered turned into a regular message and the bot replied with `"message received, will process after the current turn"`. The agent then sometimes saw `"allow"` / `"deny"` in its history.

After this PR: stale taps are silently ignored (logged at DEBUG only). The agent never sees `"allow"` / `"deny"` from a button click.

### `core/message.go`

Add `IsPermissionResponse bool` field. Documented invariant:

> Plain text `"allow"` / `"deny"` typed by the user must NOT set this flag.

This flag is **only** set by inline-button / card-action paths in platforms.

### `core/engine.go` — `handlePendingPermission`

When `IsPermissionResponse == true` and either:

- there is no `interactiveState` for the session, **or**
- the state has no `pending` permission,

the message is dropped silently (DEBUG log) and `handlePendingPermission` returns `true`, short-circuiting all downstream handling (no queue toast, no agent forwarding).

Plain-text `"allow"` typed by a real user (`IsPermissionResponse == false`) continues to fall through to normal message handling exactly as before, so users who type the words still get the previous behavior.

### `platform/telegram/telegram.go` — `handleCallbackQuery`

Set `IsPermissionResponse: true` on the synthesized `core.Message` for `perm:allow`, `perm:deny`, and `perm:allow_all` callbacks.

### `platform/feishu/feishu.go` — `onCardAction`

Set `IsPermissionResponse: true` on the synthesized `core.Message` for `perm:*` card actions.

### `core/engine_test.go` (new tests)

- `TestHandlePendingPermission_StaleCallback_NoState_Dropped` — asserts a permission-response message is dropped when no `interactiveState` exists for the session.
- `TestHandlePendingPermission_StaleCallback_NoPending_Dropped` — asserts a permission-response message is dropped when state exists but `pending == nil`.
- `TestHandlePendingPermission_UserTextAllow_FallsThrough` — regression: plain text `"allow"` (no flag set) still falls through to normal handling, so the existing UX of typing `allow`/`deny` is preserved.

### `platform/telegram/telegram_test.go` (new test)

- `TestHandleCallbackQuery_PermissionMarksIsPermissionResponse` — exercises `perm:allow`, `perm:deny`, `perm:allow_all` and asserts the synthesized `core.Message` has `IsPermissionResponse == true`, `Content` set correctly, and the session/channel keys propagated.

## Test Plan

- [x] `go build ./...`
- [x] `go test ./core/ ./platform/telegram/ ./platform/feishu/`
- [ ] Manual: tap a stale Allow/Deny button on Telegram after `/reset` — verify no toast, no message in chat, no `"allow"` in agent transcript.
- [ ] Manual: type `allow` as a normal message during a permission prompt — verify it still resolves the pending request as before.
- [ ] Manual: tap a current Allow/Deny button on Telegram — verify the permission resolves normally.
- [ ] Manual: same three checks for Feishu cards.

## Risk

- **Backward compat:** The new field defaults to `false`; existing callers that don't touch it preserve current behavior.
- **Scope:** Engine-side change is gated entirely on `IsPermissionResponse == true`; user-typed `allow`/`deny` paths are untouched (covered by `TestHandlePendingPermission_UserTextAllow_FallsThrough`).
- **Other platforms:** Discord, Slack, DingTalk, etc. that don't (yet) emit `perm:*` callbacks via the `core.Message` channel are unaffected. When they add inline-button permission flows, they should also set this flag.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1937,6 +1937,15 @@ func (e *Engine) handlePendingPermission(p Platform, msg *Message, content strin
 	state, ok := e.interactiveStates[iKey]
 	e.interactiveMu.Unlock()
 	if !ok || state == nil {
+		// Stale platform-callback permission click (e.g. user tapped an old
+		// "Allow"/"Deny" button after the session was reset or the bot
+		// restarted). Drop silently so the synthesized "allow"/"deny" payload
+		// is not forwarded to the agent or queued as a normal user message.
+		if msg.IsPermissionResponse {
+			slog.Debug("dropping stale permission callback (no interactive state)",
+				"session", msg.SessionKey, "content", content)
+			return true
+		}
 		return false
 	}
 
@@ -1944,6 +1953,11 @@ func (e *Engine) handlePendingPermission(p Platform, msg *Message, content strin
 	pending := state.pending
 	state.mu.Unlock()
 	if pending == nil {
+		if msg.IsPermissionResponse {
+			slog.Debug("dropping stale permission callback (no pending request)",
+				"session", msg.SessionKey, "content", content)
+			return true
+		}
 		return false
 	}
 
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -2297,6 +2297,73 @@ func TestHandlePendingPermission_MultiWorkspaceLookup(t *testing.T) {
 	}
 }
 
+// Stale permission callback (no interactive state for the session, e.g. clicked
+// after bot restart). Must be dropped silently and never reach the agent or
+// queue, regardless of session-busy state.
+func TestHandlePendingPermission_StaleCallback_NoState_Dropped(t *testing.T) {
+	e := newTestEngine()
+
+	p := &stubPlatformEngine{n: "telegram"}
+	msg := &Message{
+		SessionKey:           "telegram:1:2",
+		ReplyCtx:             "ctx",
+		IsPermissionResponse: true,
+	}
+
+	if !e.handlePendingPermission(p, msg, "allow") {
+		t.Fatal("expected stale permission callback to be handled (dropped)")
+	}
+	if got := p.getSent(); len(got) != 0 {
+		t.Fatalf("expected no reply for dropped stale callback, got %v", got)
+	}
+}
+
+// Stale permission callback when state exists but pending is nil (e.g. previous
+// permission was already resolved). Must also be dropped silently.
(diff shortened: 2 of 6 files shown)


chenhg5 (commented):
## Review Summary

REQUEST_CHANGES (owner-authored PR, so leaving this as a blocking review comment rather than approval/request-changes state).

The direction is good: `IsPermissionResponse` can prevent stale Telegram/Feishu permission callbacks from falling through as ordinary user messages. The current patch still leaves permission callback paths uncovered.

🔴 **Blocker: bridge card permission callbacks still dispatch as normal messages**

`core/bridge.go` still handles `perm:*` card actions through `dispatchAsMessage`, which constructs a normal `Message` without `IsPermissionResponse=true`. That means bridge/bot-gateway/web card paths can still forward stale `allow` / `deny` callbacks to the agent as ordinary user input, bypassing the engine protection added here.

Please update that path and add a focused test.

🔴 **Blocker: stale callbacks can target a newer pending permission**

The callback payload is still only `perm:allow` / `perm:deny`, with no permission request ID. If a user clicks an old button while the same session has a newer pending permission, the old callback can be interpreted as approval/denial for the current request. Please either carry and validate a request ID, or add tests/documentation proving this behavior is intentionally acceptable.

CI is green.

PR #837, opened 2026-05-02
fix(codex): app_server backend should honor model/effort/provider config + add stdio sentinel

AaronZ345:
## Problem

Two issues with the `app_server` backend that I hit while trying to use it as a long-lived alternative to `exec` per-turn cold starts:

### 1. Per-project config silently ignored at spawn time

The `exec` backend already passes the per-project `model`, `reasoning_effort`, `model_provider` and `openai_base_url` on the codex CLI command line (see `buildExecArgs` in `agent/codex/session.go`). The `app_server` backend launches `codex app-server` with **none** of these — `connect()` only ever appended `--listen <url>`.

As a result, the codex CLI falls back to whatever is in the user's `~/.codex/config.toml`, and the per-project values configured in cc-connect are silently overridden. Concretely: setting `reasoning_effort = \"medium\"` in the project block had no effect — the spawned process used the home-config global (`xhigh` in my case), which I could verify both by inspecting the running process command line and by the `reasoningEffort` field returned by `thread/start`.

### 2. \`--listen\` is forced — no way to pick pure stdio

\`appServerURL\` defaults to \`ws://127.0.0.1:3845\` and there was no way to opt out of that. \`connect()\` would always spawn the codex process in WebSocket-listen mode while cc-connect kept writing JSON-RPC requests to its **stdin**. On the codex CLI versions I tested (0.128.0), this leaves the RPC channel silently stalled — the client never receives a response and the session hangs.

A pure stdio mode (\`codex app-server\` with no \`--listen\`) does work end-to-end against the existing NDJSON reader in \`appserver_session.go:readLoop\`, but there was no way to ask for it from config.

## Fix

- \`appServerSession\` gains \`baseURL\` and \`modelProvider\` fields; the constructor signature is updated and the call site in \`Agent.StartSession\` is updated to pass the values it was already computing.
- \`connect()\` now appends \`-c model=...\`, \`-c model_reasoning_effort=...\`, \`-c model_provider=...\`, \`-c openai_base_url=...\` whenever the corresponding value is set, mirroring \`buildExecArgs\`.
- \`app_server_url = \"stdio\"\` is treated as a sentinel: it resolves to an empty URL internally, which makes \`connect()\` skip \`--listen\` and run the app server in pure stdio mode. The default behaviour (\`ws://127.0.0.1:3845\`) is unchanged for users who don't set the option, so this is fully backward compatible.

Net diff: +38 / -20 across two files, no new dependencies, no protocol changes.

## Verification

Tested against \`@openai/codex\` 0.128.0 on macOS arm64.

**Before** (the bug): the spawned process is

\`\`\`
codex app-server --listen ws://127.0.0.1:3845
\`\`\`

…and the JSON-RPC client times out with no response, so the agent never replies.

**After** the fix, with \`backend = \"app_server\"\` and \`app_server_url = \"stdio\"\`:

\`\`\`
codex app-server -c model=\"gpt-5.5\" -c model_reasoning_effort=\"medium\"
\`\`\`

\`thread/start\` now returns \`\"reasoningEffort\":\"medium\"\` instead of the \`xhigh\` global default, confirming the per-project config is honoured.

End-to-end turn timings on a hot session (same project, same prompts, gpt-5.5 reasoning effort=medium):

| turn | duration |
|------|----------|
| 1 (cold start: spawn + initialize + thread/start + first OpenAI prompt cache build) | 37.78s |
| 2 | 14.96s |
| 3 | 6.84s |
| 4 | 3.90s |

Compared to the \`exec\` backend on the same project, which paid the cold-start cost on **every** turn (~30–45s consistently), this brings the app_server backend to roughly the same steady-state latency as the long-lived \`claudecode\` stdio process.

## Config example

\`\`\`toml
[[projects]]
  name = \"my-project\"

  [projects.agent]
    type = \"codex\"

    [projects.agent.options]
      mode = \"yolo\"
      model = \"gpt-5.5\"
      reasoning_effort = \"medium\"
      backend = \"app_server\"
      app_server_url = \"stdio\"   # opt in to pure stdio; omit for default WebSocket
      work_dir = \"...\"
\`\`\`

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -112,13 +112,15 @@ type appServerCreditsSnapshot struct {
 }
 
 type appServerSession struct {
-	url       string
-	workDir   string
-	model     string
-	effort    string
-	mode      string
-	extraEnv  []string
-	codexHome string
+	url           string
+	workDir       string
+	model         string
+	effort        string
+	mode          string
+	baseURL       string
+	modelProvider string
+	extraEnv      []string
+	codexHome     string
 
 	events chan core.Event
 
@@ -135,7 +137,7 @@ type appServerSession struct {
 	pendingMu sync.Mutex
 	pending   map[int64]chan rpcResponseEnvelope
 
-	approvalsMu     sync.Mutex
+	approvalsMu      sync.Mutex
 	pendingApprovals map[string]chan core.PermissionResult
 
 	threadID atomic.Value
@@ -158,7 +160,7 @@ const (
 	appServerUsageRefreshTimeout = 1500 * time.Millisecond
 )
 
-func newAppServerSession(ctx context.Context, url, workDir, model, effort, mode, resumeID string, extraEnv []string, codexHome string) (*appServerSession, error) {
+func newAppServerSession(ctx context.Context, url, workDir, model, effort, mode, resumeID, baseURL, modelProvider string, extraEnv []string, codexHome string) (*appServerSession, error) {
 	sessionCtx, cancel := context.WithCancel(ctx)
 	s := &appServerSession{
 		url:              url,
@@ -166,6 +168,8 @@ func newAppServerSession(ctx context.Context, url, workDir, model, effort, mode,
 		model:            model,
 		effort:           effort,
 		mode:             mode,
+		baseURL:          baseURL,
+		modelProvider:    modelProvider,
 		extraEnv:         append([]string(nil), extraEnv...),
 		codexHome:        strings.TrimSpace(codexHome),
 		events:           make(chan core.Event, 128),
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. app_server startup passes model/reasoning_effort/model_provider/openai_base_url as -c flags; stdio sentinel normalizes appServerURL. CI green.

chenhg5:
@AaronZ345 Please resolve the conflicts before merging, thanks.

AaronZ345:
@chenhg5 Conflicts have been resolved.

I rebased this branch onto the latest upstream main (5a328919), force-pushed the updated branch, and verified it is mergeable locally with `git merge-tree origin/main HEAD`.

Validation: go test -count=1 ./agent/codex

PR #707, opened 2026-04-20
feat(core): message collect mode

xukp20:
## Summary

This PR implements issue #665 by adding an **opt-in project-level message collection mode** for pre-execution message aggregation.

The goal is to support chat workflows where users naturally send a request in several consecutive messages, without changing the default `cc-connect` behavior for everyone.

This PR adds:

- `messages.mode = immediate | collect | manual`
- `/flush` as the explicit submit command for buffered messages
- configurable `messages.collect_wait_ms` (default `5000ms`)
- configurable `messages.collect_max_messages`
- configurable `messages.collect_max_bytes`
- `/config messages.*` runtime updates with project persistence

The default remains `immediate`, so existing deployments keep the current "send one message, start immediately" behavior.

## Why This PR

Issue `#665` asks for a debounce / collect mode so that multiple rapid consecutive messages can be treated as a single logical request instead of triggering multiple agent runs.

There is an older related PR (#464), but that implementation is narrower and mainly focuses on a fresh-session debounce path. This PR instead introduces:

- an explicit project-level mode switch
- a clearer separation between:
  - **pre-execution buffering** (`immediate` / `collect` / `manual`)
  - **in-flight continuation** (`/steer`)
- a manual submission path via `/flush`
- runtime configuration via `/config messages.*`

### Config layer

- Adds `[projects.message_queue]` to `config.ProjectConfig`
- Supports:
  - `mode`
  - `collect_wait_ms`
  - `collect_max_messages`
  - `collect_max_bytes`
- Adds config validation and persistence helpers
- Exposes these fields through `/config`

### Engine behavior

- `immediate`
  - unchanged current behavior
- `collect`
  - buffers pre-execution messages
  - automatically flushes after an idle window
  - flushes early if count/byte limits are reached
- `manual`
  - buffers pre-execution messages
  - only submits when the user sends `/flush`

Important boundaries:

- buffering only applies **before** an execution starts
- once a session is already busy, later messages go back to existing busy/queue behavior
- `/steer` remains the in-flight continuation mechanism and is not changed by this PR

### `/flush`

- added as a builtin command
- does **not** forward `/flush` itself to the agent
- instead, it submits all buffered messages as one merged request
- when flushing manually, the `/flush` message becomes the typing/progress anchor
- when collect mode auto-flushes, the final buffered user message becomes the anchor

## Scope

- aggregate messages only before an execution starts
- keep `/steer` as the in-flight continuation mechanism
- use the final triggering message as the typing / processing anchor
[...]

PR #839, opened 2026-05-02
[codex] steer busy text by default

juanfradb:
## Summary

- default busy plain-text messages to same-turn steering instead of enqueueing for later
- add `display.busy_input_mode = "queue"` as an opt-out for the previous queue behavior
- wire Codex app-server sessions to native `turn/steer` with the active `expectedTurnId`

## Behavior

Plain text sent while a session is busy now uses the active turn when steering is available. Empty messages, media/file messages, unavailable sessions, failed steering, and explicit `busy_input_mode = "queue"` continue to use the existing busy queue fallback.

Related but not duplicate: #523 adds an opt-in busy action mode, #669 adds an explicit `/steer` command, and #707 handles pre-execution message collection while leaving in-flight continuation unchanged.

## Validation

- `go test ./config ./core ./agent/codex ./cmd/cc-connect`

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -64,6 +64,10 @@ type turnStartResponse struct {
 	} `json:"turn"`
 }
 
+type turnSteerResponse struct {
+	TurnID string `json:"turnId"`
+}
+
 type turnNotification struct {
 	ThreadID string `json:"threadId"`
 	Turn     struct {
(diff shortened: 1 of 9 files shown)


chenhg5 (approved):
LGTM. Codex steer busy text by default，CI pass，测试覆盖。

ternary-chen:
Thanks for the earlier work here. I opened #1652 as a Draft based on the current `main` after testing the same busy-input path with Feishu and the Codex app-server backend.

The draft keeps `queue` as the backward-compatible default and makes `steer` opt-in. It also adds a capability-based interface, `expectedTurnId` protection, fallback to the existing queue, queue-only handling for attachments, history/watermark persistence, CUJ and targeted race coverage, and safer `/ps` behavior.

This clearly overlaps #839. I am not treating #1652 as an automatic replacement — happy to align, split, or close it if the author or maintainers prefer reviving this PR.

PR #844, opened 2026-05-03
feat(slack): support Slack Assistant API (Agent toggle) — replies land in Chat tab not History

svjozi:
## Problem

When a Slack app has the Agent/Assistant toggle on (Features → Agents & AI Apps), users see a **Chat** tab in the app sidebar instead of the standard Messages tab. Messages typed in Chat arrive as `message.im` events with `ThreadTimeStamp` set to the assistant_thread's root ts.

Before this patch, cc-connect's `Reply()` and `Send()` ignore that `thread_ts`:
- `MsgOptionTS` is a no-op on `chat.postMessage` (it's only for `chat.update`/`chat.delete`)
- `Send()` doesn't pass `thread_ts` at all

So replies go to the DM root and surface in the **History** tab feed instead of the Chat tab — breaking the conversational UX for any bot configured as an Agent/Assistant.

## Changes

- New helper `assistantOrThreadTS()` returns `ev.ThreadTimeStamp` if present, else empty
- `MessageEvent` handler builds `replyContext` with `assistantOrThreadTS(ev)` instead of `ev.TimeStamp`
- `Reply()` switched from `MsgOptionTS` (no-op for postMessage) to `MsgOptionPostMessageParameters{ThreadTimestamp}` (real thread reply)
- `Send()` also includes `ThreadTimestamp` when present (the engine falls back to `Send` when streaming preview is unavailable, so both paths must support threading)
- New `AssistantThreadStartedEvent` case logs the event + calls `SetAssistantThreadsStatus` (informational; the actual routing happens via `assistantOrThreadTS` in MessageEvent)

## Backward compatibility

For non-Assistant apps and for true top-level DM messages, \`ThreadTimeStamp\` is empty → `assistantOrThreadTS()` returns empty → no `thread_ts` passed → identical behavior to before. AppMention handler is untouched (channel @mentions still thread under the parent message, which is correct).

## Required scope (already common)

- \`assistant:write\` — for SetAssistantThreadsStatus
- \`assistant_thread_started\` event subscription (auto-adds the scope when added in app config)

## Tested against

Live Donna Slack app with Agent toggle on. Confirmed: messages typed in Chat tab now receive replies in the same tab (not History).


chenhg5 (approved):
LGTM. Slack Assistant thread routing uses thread_ts via PostMessageParameters; AssistantThreadStarted handled; slack-go bumped to v0.16. CI green.

PR #860, opened 2026-05-06
Fix cron prompt expansion and delivery status

### LINKED ISSUE #856: Cron slash skill prompts are not expanded
## Problem

Cron prompt jobs accept a raw prompt string, but a prompt that starts with a skill slash command is sent to the agent as literal text. For example, a cron configured with `/daily-brief for today` does not expand the `daily-brief` skill body the way an interactive `/daily-brief` message does.

## Expected

Cron prompt jobs should resolve known slash skills before calling the agent, so scheduled jobs can reuse the same skill prompts as chat-triggered invocations. Unknown slash commands should keep the existing behavior and be sent unchanged.

## Why it matters

Scheduled jobs are a natural place to use reusable skills, especially for daily reports and other recurring workflows. Without expansion, the agent only sees the slash text and may ignore the intended skill instructions.

## Suggested fix

In `Engine.ExecuteCronJob`, detect prompt strings beginning with `/`, resolve them through the engine skill registry, and replace the cron prompt with `BuildSkillInvocationPrompt(...)` when a skill is found. Add a regression test covering successful expansion and non-slash prompts.

### LINKED ISSUE #857: Cron jobs should fail when the agent produces no deliverable response
## Problem

A cron prompt job can complete successfully even when the agent produces no user-visible output. In practice this can send an `(empty response)` placeholder, or with context/footer metadata enabled it can look like a non-empty response even though the actual agent answer is empty.

## Expected

For non-muted cron prompt jobs, `ExecuteCronJob` should verify that at least one deliverable response was actually sent. If the only output is the localized empty-response placeholder, the run should be marked failed with a clear error such as `cron produced an empty response`.

Muted cron jobs should keep their current behavior because they intentionally suppress delivery.

## Why it matters

Cron success/failure state is used for scheduled workflows. Reporting success after an empty answer hides broken prompts, failed agent sessions, or prompt-too-long failures behind a green run.

## Suggested fix

Track delivery for non-muted cron runs around the interactive processing path. Count successful final sends or successful stream-preview finalization as delivery, suppress empty-response placeholders, and return an error if the cron turn finishes without any delivered message. Add regression coverage for empty output and for empty output plus context-indicator/footer metadata.

### LINKED ISSUE #858: Expose cron silent mode in the CLI
## Problem

The cron data model and API support suppressing the start notification through the `silent` field, but the CLI does not expose it. Users can set `mute`, `mode`, `timeout_mins`, etc. from `cc-connect cron add/edit`, but cannot set `silent` without editing the store/API payload externally.

## Expected

The CLI should support:

```bash
cc-connect cron add --silent ...
cc-connect cron edit <id> silent true
cc-connect cron edit <id> silent false
```

This should only suppress the cron start notice, while leaving the final agent response behavior unchanged.

## Why it matters

For recurring scheduled reports, the start notification is noisy. `silent` is different from `mute`: users often want the final report delivered, just not the `cron started` message.

## Suggested fix

Add `--silent` parsing to `runCronAdd`, include `silent: true` in the add payload, and allow `silent` in the boolean field list for `runCronEdit`. Add CLI parser tests for add/edit behavior.

zzl360:
## Summary

- expand cron prompt jobs that use slash skill syntax before sending them to the agent
- track non-muted cron delivery so empty/no-output runs fail instead of appearing successful
- expose cron `silent` through `cron add --silent` and `cron edit <id> silent true|false`

Fixes #856.
Fixes #857.
Fixes #858.

## Tests

- `go test ./core -run 'TestEngine_ExpandCronSkillPrompt|TestExecuteCronJob_EmptyAgentResponse|TestProcessInteractiveEvents_CronDeliveryDoesNotSuppressQueuedTurn|TestExecuteCronJob_ResolvesCronReplyTarget|TestExecuteCronJob_WorkspacePrefixedSessionKey'`\n- `go test -tags no_web ./cmd/cc-connect -run 'TestParseCron|TestCron|TestParseSendArgs'`\n\nNote: `go test -tags no_web ./cmd/cc-connect ./core` currently fails in `core.TestBridgeBuildCapabilitiesSnapshotIncludesProjectCatalog` on latest main because Bridge now requires a token and the test dereferences a nil server. The targeted cron/core tests above pass.

### DIFF
--- a/cmd/cc-connect/cron.go
+++ b/cmd/cc-connect/cron.go
@@ -42,6 +42,7 @@ func runCron(args []string) {
 func runCronAdd(args []string) {
 	var project, sessionKey, cronExpr, prompt, execCmd, desc, dataDir, sessionMode string
 	var timeoutMins *int
+	var silent bool
 
 	var positional []string
 	for i := 0; i < len(args); i++ {
@@ -96,6 +97,8 @@ func runCronAdd(args []string) {
 				}
 				timeoutMins = &n
 			}
+		case "--silent":
+			silent = true
 		case "--help", "-h":
 			printCronAddUsage()
 			return
@@ -151,6 +154,9 @@ func runCronAdd(args []string) {
 	}
 	if timeoutMins != nil {
 		body["timeout_mins"] = *timeoutMins
+	}
+	if silent {
+		body["silent"] = true
 	}
 	payload, _ := json.Marshal(body)
 
@@ -452,7 +458,7 @@ func runCronEdit(args []string) {
 	// Parse value based on field type
 	var value any
 	switch field {
-	case "enabled", "mute":
+	case "enabled", "mute", "silent":
 		v, err := strconv.ParseBool(valueStr)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "Error: %s must be true or false\n", field)
--- a/core/cron.go
+++ b/core/cron.go
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
-	"reflect"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -396,13 +396,13 @@ func toExportedFieldName(s string) string {
 
 // CronScheduler runs cron jobs by injecting synthetic messages into engines.
 type CronScheduler struct {
-	store         *CronStore
-	cron          *cron.Cron
-	engines       map[string]*Engine // project name → engine
-	mu            sync.RWMutex
-	entries       map[string]cron.EntryID // job ID → cron entry
-	defaultSilent      bool   // global default for suppressing cron start notifications
-	defaultSessionMode string // global default session mode; "" = reuse, "new_per_run" = fresh session each run
+	store              *CronStore
+	cron               *cron.Cron
+	engines            map[string]*Engine // project name → engine
+	mu                 sync.RWMutex
+	entries            map[string]cron.EntryID // job ID → cron entry
+	defaultSilent      bool                    // global default for suppressing cron start notifications
+	defaultSessionMode string                  // global default session mode; "" = reuse, "new_per_run" = fresh session each run
 }
 
 func NewCronScheduler(store *CronStore) *CronScheduler {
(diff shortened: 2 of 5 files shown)


zzl360:
Addressed Copilot review in 1687991b: cron delivery is now counted only from final reply paths. Intermediate thinking/tool/progress sends no longer mark the tracker delivered, and a regression test covers progress before an empty final response.

chenhg5 (approved):
LGTM — three well-targeted cron improvements, all CI green.

**1. Cron skill prompt expansion** ✅
- `expandCronSkillPrompt` correctly checks for `/` prefix, resolves skill name, injects SKILL.md content with args — matching the same UX as interactive `/skill-name` invocations
- Test `TestEngine_ExpandCronSkillPrompt` validates expand/no-expand paths

**2. Delivery tracking** ✅
- `cronDeliveryTracker` correctly uses a mutex (no data race)
- `checkCronDelivery` distinguishes 'empty response' vs 'no response at all' — useful error differentiation for monitoring
- `TestExecuteCronJob_EmptyAgentResponseFails` + `CronDeliveryIgnoresProgressBeforeEmptyFinal`: correct regression coverage for progress-only runs that produce no final content
- `sendFinalWorkspaceWithError` correctly marks delivery only on actual final sends, not progress messages
- `cronDelivery` field properly reset between queued turns (`cronDelivery` only attached to the first turn in the queue)

**3. Silent flag** ✅
- `--silent` for `cron add` and `cron edit <id> silent true|false` — clean additive change

**APPROVED.**

chenhg5:
ℹ️ **Overlap notice**: The repo owner has created two focused PRs that cover the core features of this PR:

- **#867** — cron empty-response detection (same as this PR's `cronDeliveryTracker`, simpler implementation via platform wrapper)
- **#868** — cron skill command expansion (same as this PR's `expandCronSkillPrompt`)

**The only feature in this PR not covered by #867+#868 is the `--silent` CLI flag.**

These PRs will conflict if all three are merged. Possible paths:
1. Merge #867 + #868 → close this PR (or extract just the `--silent` flag as a follow-up)
2. Merge this PR → close #867/#868

The decision is the owner's. If this PR is the preferred one, please let us know and we'll close the others. The code quality here is solid.

chenhg5:
✅ **Update**: All three features in this PR are now covered by separate owner PRs:
- **#867** — cron empty-response detection (delivery tracking)
- **#868** — cron skill command expansion
- **#875** — `--silent` CLI flag

Recommend closing this PR once #867+#868+#875 are merged. Your implementation has been appreciated — the separate focused PRs are cleaner for review history.

PR #867, opened 2026-05-06
feat(cron): fail when non-muted cron produces empty response

### LINKED ISSUE #857: Cron jobs should fail when the agent produces no deliverable response

chenhg5:
## Summary

- Add `deliveryTrackingPlatform` wrapper that tracks when `Reply`/`Send` is successfully called on the platform
- For non-muted cron jobs, wrap the platform with this tracker after the "⏰" notification is sent
- After `processInteractiveMessageWith` completes, check if any delivery occurred
- If no delivery occurred, return error `"cron job produced an empty response"`

This prevents silent failures where cron jobs appear successful but the agent actually returned nothing meaningful to the user (e.g., when the agent produces only `(empty response)` placeholder).

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes  
- [x] `go test ./...` passes (all 89 test cases)
- [x] New tests for `deliveryTrackingPlatform` added in `cron_test.go`
- [ ] Manual smoke test: verify cron job failure is logged when agent returns empty

Closes #857

### DIFF
--- a/core/cron.go
+++ b/core/cron.go
@@ -678,6 +678,41 @@ type mutePlatform struct {
 func (m *mutePlatform) Reply(_ context.Context, _ any, _ string) error { return nil }
 func (m *mutePlatform) Send(_ context.Context, _ any, _ string) error  { return nil }
 
+// deliveryTrackingPlatform wraps a Platform and tracks whether any message
+// was successfully delivered via Reply or Send. Used by ExecuteCronJob to
+// detect empty responses.
+type deliveryTrackingPlatform struct {
+	Platform
+	mu       sync.Mutex
+	delivered bool
+}
+
+func (d *deliveryTrackingPlatform) Reply(ctx context.Context, replyCtx any, content string) error {
+	err := d.Platform.Reply(ctx, replyCtx, content)
+	if err == nil {
+		d.mu.Lock()
+		d.delivered = true
+		d.mu.Unlock()
+	}
+	return err
+}
+
+func (d *deliveryTrackingPlatform) Send(ctx context.Context, sessionKey any, content string) error {
+	err := d.Platform.Send(ctx, sessionKey, content)
+	if err == nil {
+		d.mu.Lock()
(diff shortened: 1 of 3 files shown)


chenhg5 (commented):
LGTM — clean, focused implementation of cron empty-response detection.

The `deliveryTrackingPlatform` wrapper approach is preferable to passing a tracker through `processInteractiveMessageWith`'s signature (as #860 does): it's less invasive, doesn't require updating all call sites, and correctly tracks delivery at the platform boundary.

✅ Mutex usage correct  
✅ Wrapper only created *after* the ⏰ notification (so the start message doesn't count as a delivery)  
✅ Shell jobs bypass tracking (they have their own output model)  
✅ Test coverage: delivery/non-delivery, Name() delegation

⚠️ **Overlap with #860** (zzl360): #860 implements this same feature plus skill expansion and `--silent` CLI. See comment on #860. Recommend merging this and closing the overlapping parts of #860 — but that's your call.

PR #872, opened 2026-05-06
feat(slack): add reply_in_thread config for threading responses

### LINKED ISSUE #862: Slack: add thread-based replies for channel messages
When the bot is @mentioned in a Slack channel, all replies are posted as top-level channel messages. This floods the channel, especially for multi-step agent tasks that produce several messages.

**Expected behavior:** Bot replies should be posted as a thread off the triggering message (using Slack's `thread_ts` parameter on `chat.postMessage`).

**What I've tried:**
- `thread_isolation = true` — accepted in config but has no effect on Slack (docs mention Feishu only)
- `reply_to_trigger = true` — same, no effect on Slack

**Suggested implementation:** When responding to an `app_mention` or `message.channels` event in a Slack channel, set `thread_ts` to the triggering message's `ts` value. This would thread all replies under the original message. Could be gated behind a config option like `reply_in_thread = true` in Slack platform options.

Discord and Feishu both have `thread_isolation` support — this would bring Slack to parity.

**Environment:**
- cc-connect v1.3.2
- Platform: Slack (Socket Mode)
- OS: Linux

chenhg5:
## Summary
- Add `reply_in_thread` config option (default: true) to control whether bot responses in channels are threaded under the triggering message
- Add `threadTSForMessage()` method to determine thread_ts based on message context and config
- Update AppMentionEvent handler to use thread_ts for replies
- Add comprehensive tests for thread behavior and config defaults

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes  
- [x] `go test ./...` passes
- [x] Manual verification: `TestThreadTSForMessage_ReplyInThreadConfig` covers all scenarios
- [x] Manual verification: `TestNew_ReplyInThreadDefault` confirms default is true
- [x] Manual verification: `TestNew_ReplyInThreadExplicit` confirms explicit config works

Fixes #862

### DIFF
--- a/platform/slack/slack.go
+++ b/platform/slack/slack.go
@@ -34,6 +34,7 @@ type Platform struct {
 	appToken              string
 	allowFrom             string
 	shareSessionInChannel bool
+	replyInThread         bool // thread replies under the triggering message (default: true)
 	client                *slack.Client
 	socket                *socketmode.Client
 	handler               core.MessageHandler
@@ -49,6 +50,12 @@ func New(opts map[string]any) (core.Platform, error) {
 	allowFrom, _ := opts["allow_from"].(string)
 	core.CheckAllowFrom("slack", allowFrom)
 	shareSessionInChannel, _ := opts["share_session_in_channel"].(bool)
+	// reply_in_thread defaults to true (thread replies under triggering message)
+	// Set to false to send replies as top-level channel messages instead.
+	replyInThread := true
+	if v, ok := opts["reply_in_thread"].(bool); ok {
+		replyInThread = v
+	}
 	if botToken == "" || appToken == "" {
 		return nil, fmt.Errorf("slack: bot_token and app_token are required")
 	}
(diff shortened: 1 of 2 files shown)


chenhg5 (commented):
🔴 **Lint failure: unused function `assistantOrThreadTS`**

The `assistantOrThreadTS` function is now dead code — it is no longer called anywhere. Keeping it "for reference" with a Deprecated comment causes the linter to fail, blocking all other CI checks.

**Fix**: Remove the `assistantOrThreadTS` function entirely. It's superseded by `threadTSForMessage` and the logic is already well-documented there.

---

Otherwise the change is correct:

✅ `replyInThread` config option with sensible default (`true` = existing behaviour preserved)  
✅ Resolution order in `threadTSForMessage`: already-in-thread → channel respects config → DM top-level empty  
✅ AppMentionEvent handler also updated correctly  
✅ Tests cover all 5 scenarios

Once the dead function is removed and lint passes, this is LGTM.

PR #873, opened 2026-05-06
fix(telegram): handle message too long errors from HTML table expansion

### LINKED ISSUE #866: Telegram reply is dropped when Markdown table rendering expands below-limit responses past Telegram size limit
## Summary

cc-connect can complete an agent turn but fail to send any Telegram reply when the assistant response contains a Markdown table with wide cells.

The original assistant response can be below cc-connect's generic message splitting threshold, but cc-connect's Telegram renderer expands the Markdown table into an aligned HTML `<pre>` table. During that conversion, shorter cells are padded with spaces to match the widest cell in each column. With a moderately wide table, this padding can expand the final Telegram payload beyond Telegram's message size limit.

When this happens, Telegram rejects the send with `Bad Request: message is too long`, and the user receives no final reply.

## Version

Observed on:

- cc-connect: `v1.3.2`
- commit: `19406df9`
- build time: `2026-04-21T00:42:17Z`
- platform: Telegram

## Observed Behavior

The agent turn completes successfully in cc-connect, but Telegram receives no final message.

Observed log pattern:

```text
turn complete ... response_len=<under split threshold>
platform send failed platform=telegram error="telegram: send: bad request, Bad Request: message is too long"
```

There is no automatic fallback, truncation, or post-render split after the Telegram HTML payload becomes too large.

chenhg5:
## Summary
- Handle "message is too long" error from Telegram when HTML table expansion exceeds 4096-char limit
- Reply/Send: chunk the HTML content into multiple messages using `SplitMessageCodeFenceAware`
- SendWithButtons: chunk with buttons attached only to the first message
- SendPreviewStart: fall back to plain text (preview messages shouldn't be chunked)
- Add `sendChunked` and `sendChunkedWithButtons` helper functions

## Root Cause
The engine splits messages at 4000 chars based on Markdown length. But `MarkdownToSimpleHTML()` expands tables with column padding, causing the HTML output to exceed Telegram's actual 4096-char limit. Previously, this error was not handled, resulting in message drops.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes  
- [x] `go test ./...` passes
- Manual test needed: send a message with a wide Markdown table and verify it's chunked properly

Fixes #866

### DIFF
--- a/platform/telegram/telegram.go
+++ b/platform/telegram/telegram.go
@@ -964,16 +964,25 @@ func (p *Platform) Reply(ctx context.Context, rctx any, content string) error {
 	}
 
 	if _, err := bot.SendMessage(ctx, params); err != nil {
-		if strings.Contains(err.Error(), "can't parse") {
+		errMsg := err.Error()
+		// Handle HTML parsing errors by falling back to plain text
+		if strings.Contains(errMsg, "can't parse") {
 			slog.Warn("telegram: HTML rejected by Telegram, sending as plain text",
 				"method", "Reply",
-				"error", err.Error(),
+				"error", errMsg,
 				"html_prefix", truncateForLog(html, 200),
 				"html_len", len(html),
 			)
 			params.Text = content
 			params.ParseMode = ""
 			_, err = bot.SendMessage(ctx, params)
+		} else if strings.Contains(errMsg, "message is too long") {
+			// Handle message too long by splitting and sending as multiple messages
+			slog.Warn("telegram: message too long, splitting into chunks",
+				"method", "Reply",
+				"html_len", len(html),
+			)
+			return p.sendChunked(ctx, bot, rc, html)
 		}
 		if err != nil {
 			return fmt.Errorf("telegram: send: %w", err)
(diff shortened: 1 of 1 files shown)


chenhg5 (commented):
LGTM — targeted, correct fix for the Telegram 4096-char overflow edge case.

**Root cause is correctly diagnosed**: `MarkdownToSimpleHTML()` expands table columns with padding, so a message that fits Markdown's 4000-char budget can still exceed Telegram's actual 4096 HTML limit.

✅ **`sendChunked`/`sendChunkedWithButtons`** — reuse existing `core.SplitMessageCodeFenceAware`, which already respects code block boundaries; correct choice.

✅ **`sendChunkedWithButtons`** — buttons only on first chunk; avoids confusing "Submit" buttons appearing after the content they relate to.

✅ **Preview messages** → plain-text fallback (not chunked). Correct: preview handles are tied to a single message ID; chunking would break the edit-in-place model.

✅ **Nested `can't parse`** handler inside `sendChunked` — good defensive coding for malformed HTML in a chunk.

✅ **Consistent `errMsg` local variable** across all 4 send methods (small style improvement).

CI all green.

chenhg5:
LGTM. Good "message too long" fix with chunking. CI green.

PR #868, opened 2026-05-06
feat(cron): expand skill commands in cron job prompts

chenhg5:
## Summary

- When a cron job's prompt starts with "/", parse it as a slash command
- Resolve through the skill registry to find matching skill
- Use `BuildSkillInvocationPrompt` to expand into full skill instructions
- Unknown slash commands pass through unchanged to the agent

This allows cron jobs to invoke skills like `/daily-brief for today`, which will be expanded into full skill instructions before being sent to the agent.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes (all tests)
- [x] Added regression test `TestExecuteCronJob_ExpandsSkillCommand`
- [x] Added regression test `TestExecuteCronJob_UnknownSlashCommandPassthrough`

Closes #856

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #875, opened 2026-05-06
feat(cli): expose --silent flag for cron add and edit

### LINKED ISSUE #858: Expose cron silent mode in the CLI

chenhg5:
## Summary
- Add `--silent` flag to `cc-connect cron add` to suppress cron start notification
- Add "silent" as a valid boolean field in `cc-connect cron edit`
- Update CLI help documentation

## Root Cause
The cron data model and API already support the `silent` field, but it was not exposed in the CLI.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes  
- [x] `go test ./cmd/cc-connect/...` passes
- Manual: `cc-connect cron add --silent --cron "*/5 * * * *" --exec "date"` should create a silent cron job
- Manual: `cc-connect cron edit <id> silent true` should update the silent field

Fixes #858

### DIFF
--- a/cmd/cc-connect/cron.go
+++ b/cmd/cc-connect/cron.go
@@ -42,6 +42,7 @@ func runCron(args []string) {
 func runCronAdd(args []string) {
 	var project, sessionKey, cronExpr, prompt, execCmd, desc, dataDir, sessionMode string
 	var timeoutMins *int
+	var silent bool
 
 	var positional []string
 	for i := 0; i < len(args); i++ {
@@ -96,6 +97,8 @@ func runCronAdd(args []string) {
 				}
 				timeoutMins = &n
 			}
+		case "--silent":
+			silent = true
 		case "--help", "-h":
 			printCronAddUsage()
 			return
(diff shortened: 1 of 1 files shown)


chenhg5 (commented):
LGTM — minimal, correct CLI exposure of the existing `silent` field.

✅ `--silent` flag only sets `body["silent"] = true` when explicitly provided (not sent when false — correct for optional fields)  
✅ `cron edit` boolean parsing now includes `"silent"` alongside `"enabled"` and `"mute"`  
✅ Help text updated with example  
✅ CI all green

**Context**: Together with #867 (delivery tracking) and #868 (skill expansion), this PR completes the owner's implementation of all three features in #860 (zzl360). Suggest merging #867+#868+#875 and closing #860 as superseded.

PR #886, opened 2026-05-08
fix(relay): preserve thread/user suffix in relay session key

KorenKrita:
parseSessionKeyParts now returns fullChatID (chatID + suffix) so that HandleRelay constructs per-thread or per-user relay sessions instead of sharing one session across all topics/users in the same chat.

Previously, "feishu:chat:root:threadID" was truncated to chatID="chat", causing relaySessionKey="relay:proj:chat" for all threads. With this fix, fullChatID="chat:root:threadID" produces distinct relay sessions per thread, matching the isolation semantics of reply sessions.

Binding lookup and groupSessionKey still use plain chatID.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -12632,7 +12632,7 @@ func (e *Engine) cmdBind(p Platform, msg *Message, args []string) {
 		return
 	}
 
-	_, chatID, err := parseSessionKeyParts(msg.SessionKey)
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
LGTM. Good fix to preserve thread/user suffix in relay session key.

PR #846, opened 2026-05-04
fix(claudecode): add dot (.) to encodeClaudeProjectKey replacement set

chenhg5:
## Summary
- Added `.` to the character replacement set in `encodeClaudeProjectKey()` to match Claude Code 2.1.91's behavior
- Fixes `/list` returning 0 sessions when macOS usernames contain dots (e.g. `yinglong.li`)
- Added test cases for single and multiple dots in paths

## Root Cause
cc-connect encoded `/Users/yinglong.li` → `-Users-yinglong.li`  
Claude Code actual directory: `-Users-yinglong-li`  
This mismatch caused `findProjectDir()` to return empty string.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./agent/claudecode/...` passes
- [ ] Manual: user with dot in username runs `/list` and sees sessions

Closes #842

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #820, opened 2026-04-30
fix(claudecode): encode dots in project key to match Claude Code CLI behavior

Re-TinyLin:
## Summary

The `encodeClaudeProjectKey` function currently collapses `/`, `:`, `_`, space, `~`, and non-ASCII to `-`, but the Claude Code CLI also collapses **dots (`.`)** to `-` when creating on-disk project keys under `~/.claude/projects/`. This causes a mismatch for home directory paths containing dotted usernames like:

`/home/user.name/project`

The encoded candidate key becomes:

`-home-user.name-project`

while the actual on-disk directory is:

`-home-user-name-project`

The primary lookup, all legacy candidates, and the fallback scan all miss because they all derive from the same `encodeClaudeProjectKey` output. As a result, `findProjectDir` returns `""`, `ListSessions` returns `nil`, and `/list` and `/switch` show "No sessions found".

## Fix

Extend the replacement set in `encodeClaudeProjectKey` to also collapse `'.'` to `'-'`. The function is the single source of truth for path encoding, so this one edit fixes both the primary candidate and the fallback scan. The docstring has been updated to reflect the new step (dot handling is inserted as step 4, renumbering the others).

This complements PR #706 (which fixed the space and `~` variants of issue #500) by covering the remaining dot character case.

## Test plan

- [x] `go test ./agent/claudecode/ -run 'TestEncodeClaudeProjectKey|TestFindProjectDir' -v` — all pass
- [x] One new `TestEncodeClaudeProjectKey` case: "path with dots (e.g. username in home dir)"
- [x] New `TestFindProjectDir_DotInPath` regression test — creates a mock `~/.claude/projects/-home-user-name-project/` directory and asserts `findProjectDir` resolves the dotted work dir
- [x] All pre-existing test cases still pass unchanged
- [x] Full suite: `go test ./...` passes
- [x] Manual confirmation on Linux with a dotted-username home dir: before the fix `/list` returned "No sessions found"; after the fix sessions are visible

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #889, opened 2026-05-08
fix(claudecode): replace dots in encodeClaudeProjectKey to match Claude Code 2.1.91+

### LINKED ISSUE #842: encodeClaudeProjectKey missing dot (.) replacement — /list returns 0 sessions
## Bug

`/list` always returns 0 sessions when the macOS username contains a dot (e.g. `yinglong.li`).

## Root Cause

`encodeClaudeProjectKey()` in `agent/claudecode/claudecode.go:1103` does NOT replace `.` with `-`, but Claude Code 2.1.91 does.

Result:
- cc-connect encodes `/Users/yinglong.li` → `-Users-yinglong.li`
- Claude Code actual directory: `-Users-yinglong-li`
- `findProjectDir()` returns `""` → `ListSessions()` returns nil → `/list` shows 0

The fallback scan also fails because all candidates preserve the dot.

## Fix

Add `.` to the replacement set in `encodeClaudeProjectKey`:

```go
if r == '/' || r == ':' || r == '_' || r == ' ' || r == '~' || r == '.' {
    result.WriteRune('-')
}
```

## Environment

- cc-connect v1.3.2 (19406df9)
- Claude Code 2.1.91
- macOS Darwin 25.3.0 (arm64)
- Username: `yinglong.li` (contains a dot)

## Workaround

```bash
ln -s ~/.claude/projects/-Users-yinglong-li ~/.claude/projects/-Users-yinglong.li
```

kevinWangSheng:
## Summary

- `encodeClaudeProjectKey` did not replace `.` with `-`, but Claude Code 2.1.91+ does. This caused `/list` to always return 0 sessions whenever the absolute work-dir path contained a dot — most commonly macOS usernames like `yinglong.li`.
- Add `'.'` to the replacement set so the encoded key matches Claude Code's actual on-disk project directory.
- Add two test cases covering a dotted username and a dotted directory name.

## Root cause

For `/Users/yinglong.li`:

| Source | Encoded key |
| --- | --- |
| cc-connect (before) | `-Users-yinglong.li` |
| Claude Code 2.1.91+ | `-Users-yinglong-li` |

`findProjectDir()` therefore failed every candidate and the fallback scan also missed (every candidate kept the dot), so `ListSessions()` returned nil → `/list` showed 0 sessions.

## Fix

Single-character extension to the replacement set in `agent/claudecode/claudecode.go:1217`:

```go
if r == '/' || r == ':' || r == '_' || r == ' ' || r == '~' || r == '.' {
    result.WriteRune('-')
}
```

No existing test inputs contained dots, so no test expectations needed updating. Two new cases pin the new behaviour.

Closes #842

## Test plan

- [x] `go test ./agent/claudecode/...` — all existing tests pass plus the two new cases
- [x] Confirmed the only test failures in the wider suite (`tests/release_local/engine_matrix` race, `agent/codex` flake) reproduce on `main` without this change and are unrelated to the dot-replacement edit

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -1213,7 +1213,9 @@ func boolVal(m map[string]any, key string) bool {
 //  3. Replacing underscores (_) with "-"
 //  4. Replacing spaces and tildes (~) with "-" (common in macOS iCloud paths like
 //     "/Users/x/Library/Mobile Documents/com~apple~CloudDocs/...")
-//  5. Replacing all non-ASCII characters with "-"
+//  5. Replacing dots (.) with "-" (e.g. macOS usernames like "yinglong.li";
+//     matches Claude Code 2.1.91+ behavior)
+//  6. Replacing all non-ASCII characters with "-"
 func encodeClaudeProjectKey(absPath string) string {
 	// First, normalize to forward slashes for consistent processing
 	normalized := strings.ReplaceAll(absPath, "\\", "/")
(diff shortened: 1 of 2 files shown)


kevinWangSheng:
Duplicate of #820 (Re-TinyLin's already-approved fix) and #846 (chenhg5's own). Closing this to avoid review duplication.

PR #890, opened 2026-05-08
feat(pi): implement AvailableModels via pi --list-models

CudtMFrag:
## Summary

Pi agent's `AvailableModels()` previously returned `nil`, causing the `/model` command to show an empty list. This fix calls `pi --list-models` to dynamically discover available models and parse the output.

## Problem

When users run `/model` with the Pi agent, they see no available models because `AvailableModels()` just returned `nil`.

## Solution

- **`AvailableModels`** now calls `pi --list-models` and parses the table output
- **`parseModelsOutput`** (new pure function) parses the column-based table:
  ```
  provider     model                  context  max-out  thinking  images
  opencode-go  deepseek-v4-flash      1M       384K     yes       no
  ```
- Returns models in `provider/model` format
- Deduplicates and sorts results alphabetically
- Filters out rows containing `Use /login` (providers requiring auth)
- Gracefully handles errors, returning `nil` on failure

## Changes

| File | Change |
|------|--------|
| `agent/pi/pi.go` | Replace `return nil` with `discoverModels` + `parseModelsOutput` |
| `agent/pi/pi_test.go` | Add 7 unit tests covering normal output, empty, header skip, `"No models"` skip, `"Use /login"` filter, dedup, sorting, malformed lines |


chenhg5 (approved):
LGTM. Good fix, well-tested.

happyTonakai:
After #1178 Pi reads `~/.pi/agent/settings.json`. I think it can get a cleaner model list than `pi --list-models`.

PR #796, opened 2026-04-27
feat(feishu): cardkit-v1 typewriter + rich card UX bug fixes (split from original)

Cigarrr:
## Summary

Lark/Feishu rich card UX improvements — **typewriter only** (footer concerns split into a separate PR; see "Stacked PRs" below).

After @cigarrr and the maintainer aligned on splitting the original #796, this PR has been narrowed to platform UX work that's independent of footer rendering:

- **cardkit-v1 typewriter** (3-step API): `POST /cards` → `POST /messages` → `PUT /elements/main_text/content`. Lark renders the per-element streaming text update natively as a typewriter effect; the engine routes `EventText` through `RichCardTextStreamer.StreamRichCardText` when the platform implements it, falls back to full-card `Im.Message.Patch` otherwise.
- **Reply API also uses cardkit-v1**, so `@`-mention flows get the same typewriter rendering (was disabled on every `@` reply previously).
- **Path C: forced final cardkit-v1 flush** in `EventResult` rich path. The 200ms / 20-char throttle in the streaming branch may skip the last small chunk; this catch-up flush guarantees the typewriter renders all the way to Done.
- **`updateCardEntity` via cardkit-v1 PUT** for messages bound to a card-entity. `Im.Message.Patch` is silently no-op for entity-referenced messages, which previously caused two visible bugs: (1) status header stuck in `Pondering...` / `Working...` forever, (2) two cards rendered per turn (the stuck-working one + the EventResult fall-through Send). Routes `cardID != ""` to `PUT /open-apis/cardkit/v1/cards/{card_id}` and shares the per-handle monotonic sequence with `streamRichCardText` so card+element updates stay totally ordered.
- **Throttle bifurcation**: 200ms / 20 chars on the streaming text path (smoother typewriter), 1500ms / 30 chars on the full-card Patch fallback (preserves existing behavior).
- **Per-turn state reset on queue dequeue** for rich mode — fixes a bug where queued messages reused the previous turn's `cardMessageID` / `lastRichCardLen`.
- `chore`: gitignore `cc-connect-new` build artifact.

## What this PR does NOT include

Footer-related changes (multi-line `statusFooter`, CCD-statusline-style `out N · in N cw N cr N · ctx N% left` rendering, `StatusFooterSender`/`StatusFooterUpdater` interfaces, `text_size: "notation"` styling) have been **moved to a separate stacked PR** to keep review surface small. This PR focuses purely on the typewriter / streaming UX.

## Stacked PRs

[...]

PR #892, opened 2026-05-08
fix(rich): gate rich card creation/stream by silentHold (NO_REPLY no-trace)

### LINKED ISSUE #891: rich card: NO_REPLY in card_mode=rich leaves '撤回了一条消息' gray bar
## Summary

In rich card mode (`[display].card_mode = "rich"` after #655 lands; equivalent to `[display].mode = "rich"` on the #796 branch), Lark/Feishu, when an agent emits `NO_REPLY` to indicate a silent turn, Lark renders the resulting card lifecycle (Send → Delete) as a **「Bot 撤回了一条消息」** gray bar. This breaks the no-trace semantics that NO_REPLY (#682 / #684) was introduced to provide.

The most common trigger in real deployments is **tool calls followed by NO_REPLY** (e.g. a heartbeat cron that scans something, decides nothing is worth reporting, and emits `NO_REPLY` as the final assistant text). Both pure-text NO_REPLY and tool-then-NO_REPLY hit the same recall.

## Reproduction

1. Lark/Feishu platform with `card_mode = "rich"` (or pre-#655: `mode = "rich"`)
2. Trigger an agent turn whose final assistant content is `NO_REPLY`. Easiest:
   - via cron: `cc-connect cron add --cron "<near-future>" --prompt "Reply with exactly: NO_REPLY"`
   - via DM: any prompt that the agent decides to drop (mail-scan with "nothing today", etc.)
3. Observe in the Lark chat: a "Bot 撤回了一条消息" gray bar appears on top of any tool-step card from that turn.

## Root cause

`processInteractiveEvents` (core/engine.go) handles `EventText` with two parallel branches:

- **Legacy** (non-rich): consults `silentHold` + `couldBeSilentPrefix` to defer streaming while the accumulated segment could still resolve to bare `NO_REPLY`. If it never resolves to NO_REPLY, the held chunks are flushed to the streaming preview.
- **Rich card** (`hasRichCard`): always
  - creates the preview card on first text chunk (`SendPreviewStart`)
  - streams every chunk into the card via `StreamRichCardText` / `UpdateMessage`
  - has no silentHold consultation

At end of stream, the silent path (`if isSilent { … DeletePreviewMessage(cardMessageID) … }`) deletes the just-created card. Lark surfaces that delete as "撤回了一条消息".

Edge cases:
- **tool-then-NO_REPLY**: tool events create the card before any text. The silentHold gate alone can't stop creation. Same delete → recall trace.
- **text-then-NO_REPLY in same agent block**: agent emits `Hello\nNO_REPLY` and an event integration that exposes `EventResult.Content` as the trailing block (`"NO_REPLY"`) makes `isSilent=true`. If a card already exists (e.g. from a tool earlier), the silent finalize blanks it — the user's just-seen `Hello` "disappears" when the card flips to Done.

## Proposed fix

Sketched in this issue's PR (will follow up after confirmation):

1. **Pre-compute silentHold transition** once per `EventText` chunk before any card creation or streaming. Both rich and legacy paths share the single transition. `couldBeSilentPrefix` is monotonically decreasing as segments grow, so the held → released transition happens at most once per segment.
2. **Gate rich card creation and streaming** on `!silentHold`. Lazy-create the card if held → released (so accumulated content lands in body).
3. **In silent end-of-stream path**: when `cardMessageID != nil`,
   - If `partialText` (sans trailing `NO_REPLY` marker) or `toolSteps` is non-empty → finalize card to `Done` with that body. Mirrors legacy + full mode where prior tool messages remain visible even when the final reply is silent.
   - Otherwise (nothing visible to keep) → `DeletePreviewMessage`. Rare path under the new gate; covers transient/race conditions.

## Related

- #682 introduced the `NO_REPLY` marker
- #684 follow-up that suppresses platform delivery
- #774 introduced the per-turn reply footer (touches the rich card BuildRichCard surface)
- #777 / `9989fdc` (replaced by upstream #655) handled quiet mode single-card behavior
- #796 (open) integrates rich-card streaming via cardkit-v1 — this fix builds on top
- #882 fixes the same "撤回了一条消息" UX symptom under a different trigger (`/stop` flow); orthogonal code path

Cigarrr:
Closes #891

## Summary

Rich card mode (`card_mode = "rich"`) didn't gate preview creation or streaming on the `silentHold` mechanism that legacy mode uses to defer messages until the agent's intent (NO_REPLY or real reply) is known. As a result, a `NO_REPLY` turn went through Send-then-Delete on the card, which Lark renders as a "Bot 撤回了一条消息" gray bar. See #891 for full root-cause analysis.

## What this changes

**`processInteractiveEvents` (`core/engine.go`)**

Pre-compute the `silentHold` transition once per `EventText` chunk before any card creation or streaming. Both rich and legacy paths share that single transition. `couldBeSilentPrefix` is monotonically decreasing as segments grow, so the held → released transition happens at most once per segment.

Rich `EventText` path:
- Card creation gated on `!silentHold`.
- Lazy creation when held → released and no card exists yet, using the accumulated `partialText` so the card emerges with post-prefix content already in the body (not as an empty shell).
- Throttle / streaming block runs only when `!silentHold`.

Silent end-of-stream path:
- When the card already exists (e.g. from earlier tool events) and the turn resolves silent, finalize the card to `Done` with body = `stripTrailingSilent(partialText)`. This preserves any pre-`NO_REPLY` text the user already saw streaming, plus any `toolSteps` panel content, and avoids the recall-stamp UX. Mirrors legacy + full mode where prior tool messages remain visible even when the final reply is silent.
- Only call `DeletePreviewMessage` when there is genuinely nothing visible worth keeping (no body, no toolSteps) — a transient/race fallback under the new gate.

Legacy path simplification: the `silentHold`/release transition is already done above, so the inner `if`/`else` collapses to "if `!silentHold` flush; `peekSegment` when `releasedNow` else `event.Content`".

## Tests

Adds 5 cases in `core/engine_test.go`:

- `TestProcessInteractiveEvents_RichCard_NoReplySingleChunk` — single-chunk `NO_REPLY` produces no `SendPreviewStart`/`StreamRichCardText`/`UpdateMessage`/`DeletePreviewMessage`.
- `TestProcessInteractiveEvents_RichCard_NoReplyChunked` — same invariant when `NO_REPLY` is split across two chunks.
- `TestProcessInteractiveEvents_RichCard_PrefixThenContent` — `"N"` then `"ote that..."`: card lazy-created after release with full accumulated content.
- `TestProcessInteractiveEvents_RichCard_TextThenNoReply_PreservesBody` — visible text plus trailing `NO_REPLY` marker (engine sees `EventResult.Content="NO_REPLY"`): final card preserves the pre-marker text instead of being blanked.
- `TestProcessInteractiveEvents_RichCard_ToolThenNoReply` — turn issues a tool call (creating the card with steps) then resolves to `NO_REPLY`: card finalized to `Done`, no `DeletePreviewMessage`.

New stub `stubRichCardSilentPlatform` implements `RichCardSupporter`, `PreviewStarter`, `MessageUpdater`, `RichCardTextStreamer`, and `PreviewCleaner`, and tracks call counts so each test asserts the no-trace invariant directly.

## Test plan

- [x] `go test ./core/... -run "RichCard"` (8 cases incl. existing rich-card tests, all pass)
- [x] Local Feishu/Lark e2e validated: `card_mode = rich` × `mode = {full, quiet, compact}` (post-#655) including tool-then-NO_REPLY — no recall trace, tool steps preserved on Done card

## Stacked on / context

This PR is **stacked on #796** (cardkit-v1 typewriter + rich-card statusFooter integration); it references the `BuildRichCard(..., statusFooter string)` signature and `RichCardTextStreamer` introduced there. Once #796 lands, this can rebase on `main` directly. The fix is small enough that the silentHold gate is also easy to back-port to the older `BuildRichCard(..., elapsed time.Duration)` signature if the maintainer prefers landing this on `main` first.

Adjacent: #882 fixes the same "Bot 撤回了一条消息" UX symptom under a different trigger (`/stop` flow) — orthogonal code path. Refs: #682, #684, #774, #882.


Cigarrr:
Closing this PR for now to observe upstream's footer evolution before re-submitting.

Context:
- This PR is stacked on #796 (cardkit-v1 typewriter), which itself is rooted in older `main` (pre-#655).
- `main` HEAD `34af331a890d` ("fix(core): suppress ellipsis-only events and handle context indicator in footer", 2026-05-08, direct-push without a PR) reworks the same `buildReplyFooter` / `EventResult` send-path region that #774 (now CLOSED, content folded into #796) reworks. The two designs are independently coherent but collide on the same code.
- Rebasing #796 (and this PR by extension) onto current `main` requires a manual semantic merge of `contextText`-into-footer (`34af331a`) with `statusFooter`-via-Platform-interface (#774). Doable but non-trivial; better to wait for the right moment.

Plan:
- Hold this PR until upstream rebases #796 OR clarifies the footer direction (e.g. unifies `34af331a`'s ctx-in-footer with #796's structured statusFooter).
- The fix in this PR (silentHold gate on rich card path + tool-then/text-then-NO_REPLY card finalize) is unchanged in spirit; will re-submit once base picture stabilizes.
- Issue #891 remains open as the canonical record of the bug.
- Local deployment continues running the equivalent fix on top of our b023 patch stack.

Apologies for the noise. Will re-engage at the right base.

PR #894, opened 2026-05-08
fix(telegram): unbound large file uploads from the long-poll timeout

### LINKED ISSUE #539: [Bug] Telegram file upload fails for larger attachments due to hardcoded 60s HTTP timeout
### cc-connect Version

v1.2.2-beta.5   commit: 28513f3   built: 2026-03-31T01:53:43Z

### Operating System

Other

### Agent Type

Claude Code

### Platform

Telegram

### Installation Method

### Bug Description

Sending larger files to Telegram via `cc-connect send --file` fails, while small files work normally.

  In my case:
  - text messages work
  - small file attachments work
  - larger file attachments fail

  For a generated APK (~46 MB), `cc-connect send --file` fails consistently.

  After testing with different file sizes, this appears to be related to a hardcoded Telegram HTTP client timeout rather than session selection, path format, or command arguments.

### Steps to Reproduce

 1. Configure Telegram platform normally in `config.toml`
  2. Start `cc-connect`
  3. Ensure file sending is enabled:
     `attachment_send = "on"`
  4. Send a small file with:
     `cc-connect send -s "telegram:<user_id>:<chat_id>" --file /tmp/probe.txt`
     -> succeeds
  5. Send a larger file (for example ~30 MB or ~46 MB APK):
     `cc-connect send -s "telegram:<user_id>:<chat_id>" --file /path/to/app-release.apk`
     -> fails

### Expected Behavior

 Large Telegram file uploads should either:
  1. succeed if they are within Telegram's allowed file size limit, or
  2. fail with a clear timeout/error message and ideally allow timeout configuration.

  I expected a 46 MB APK (below the 50 MB limit enforced by `cc-connect`) to upload successfully, or at least produce a direct Telegram upload timeout error consistently.

### Actual Behavior

The behavior depends on file size:

  - small text messages: success
  - small file attachments: success
  - 10 MB test file: success
  - 30 MB test file: failure with:
    `telegram: send file: error do request for method sendDocument, Post "https://api.telegram.org/bot***/sendDocument": context deadline exceeded (Client.Timeout exceeded while awaiting headers)`
  - 46 MB APK: often fails with:
    `invalid JSON: unexpected EOF`
  - 55 MB file: correctly rejected by `cc-connect` size limit:
    `attachment ... exceeds size limit (50 MB)`

  So the real issue seems to be that larger Telegram uploads are timing out, and in some cases the CLI surfaces that failure as `invalid JSON: unexpected EOF`.

```toml

```

### Logs / Error Output

```shell

```

### Additional Context

 I checked the source and found that Telegram uses a hardcoded HTTP client timeout:

  `platform/telegram/telegram.go`
  ```go
  httpClient := &http.Client{Timeout: 60 * time.Second}

  And file sending goes through:
  bot.SendDocument(ctx, params)

  So it looks like Telegram uploads are limited by a fixed 60-second timeout, with no config option exposed in config.toml.

  Possible improvement ideas:
  - expose a Telegram request/upload timeout config option
  - improve error propagation so large upload failures don’t surface as invalid JSON: unexpected EOF

kevinWangSheng:
## Summary

The Telegram platform builds a single `*http.Client{Timeout: 90 * time.Second}` and passes it to the bot SDK. The comment on the existing constant explains why 90s is right for the long-poll (server-side `getUpdates` timeout is 59s, +30s headroom so a stuck connection self-recovers). What it misses: the SDK uses the *same* client for every other Bot API call too — including `SendDocument` / `SendPhoto` / `SendVoice`.

cc-connect already enforces a 50 MB attachment cap, so a perfectly legitimate 30–50 MB upload over a slow uplink exceeds 90s and is aborted with `context deadline exceeded (Client.Timeout exceeded while awaiting headers)`. The same ceiling also bounded direct attachment downloads via `p.httpClient.Get`.

## Fix

Split the single client into two with different timeout profiles:

| Client | Timeout | Used for |
|---|---|---|
| `pollHTTPClient` | 90s | `getUpdates` only |
| `sendHTTPClient` | none | every other Bot API call + direct attachment downloads |

The SDK only accepts one `HttpClient` though, so I route by URL via a tiny `telegramHTTPClient` wrapper that selects `pollHTTPClient` for paths ending in `/getUpdates` and falls through to `sendHTTPClient` for the rest. Proxy transport is applied to both clients.

The `botFactory` type widens from `*http.Client` to `tgbot.HttpClient` (the SDK interface) so the wrapper can satisfy it; three test stubs need a one-token type swap (`_ *http.Client` → `_ tgbot.HttpClient`).

`p.httpClient` is now the unbounded `sendHTTPClient`, so the existing `p.httpClient.Get(link)` call site for downloading user attachments also escapes the 90s ceiling at no extra cost.

## Why this approach (vs. alternatives)

- **Just bump the timeout to e.g. 10 min** — works for files but a hung poll then takes 10 min to self-recover instead of 90s. The whole reason the original 90s exists is to bound that. Worse trade-off.
- **Drop `Client.Timeout` entirely** — same problem amplified; relies entirely on TCP keepalives for poll resilience.
- **Bypass the SDK for file ops** — substantial change, fragmenting the call surface.
- **Per-call `context.WithTimeout` for sends** — the engine context (`e.ctx`) reaching `SendFile` has no deadline; we'd need a configurable per-send timeout. Possible but more API surface than this fix.

The wrapper keeps both timeout profiles on their respective code paths with no behaviour change for callers.

## Test plan

- [x] `go build -tags=no_web ./...` and `GOOS=windows go build ./platform/telegram/...` both succeed
- [x] `go test -race -tags=no_web ./platform/telegram/...` — all existing tests pass
- [x] New `TestTelegramHTTPClient_RoutesByPath` pins the routing table — `getUpdates` → poll, `sendDocument`/`sendMessage`/`sendPhoto`/`getMe` → send

Closes #539

### DIFF
--- a/platform/telegram/telegram.go
+++ b/platform/telegram/telegram.go
@@ -101,8 +101,28 @@ type stdlibTypingTicker struct {
 
 func (t *stdlibTypingTicker) C() <-chan time.Time { return t.Ticker.C }
 
+// telegramHTTPClient routes Telegram Bot API requests to one of two underlying
+// http.Clients depending on the API method. The long-poll (getUpdates) keeps a
+// 90s ceiling so a hung connection self-recovers; everything else (sendMessage,
+// sendDocument, sendPhoto, …) honours only the caller's context, so large file
+// uploads — up to cc-connect's 50 MB attachment cap — can outlast the poll
+// window on slow uplinks. Without this split, the 90s Client.Timeout applies to
+// SendDocument as well and aborts uploads with `context deadline exceeded`
+// (see issue #539).
+type telegramHTTPClient struct {
+	poll *http.Client
+	send *http.Client
+}
+
+func (c *telegramHTTPClient) Do(req *http.Request) (*http.Response, error) {
+	if strings.HasSuffix(req.URL.Path, "/getUpdates") {
+		return c.poll.Do(req)
+	}
+	return c.send.Do(req)
+}
+
 // botFactory creates a bot, returns it plus self user info and a blocking poll function.
-type botFactory func(token string, onUpdate func(context.Context, *models.Update), httpClient *http.Client) (telegramBot, *models.User, func(context.Context), error)
+type botFactory func(token string, onUpdate func(context.Context, *models.Update), httpClient tgbot.HttpClient) (telegramBot, *models.User, func(context.Context), error)
 
 type Platform struct {
 	token                 string
@@ -110,7 +130,10 @@ type Platform struct {
 	groupReplyAll         bool
 	shareSessionInChannel bool
 	enableReactions       bool
-	httpClient            *http.Client
+	// httpClient is used for direct attachment downloads (no fixed timeout).
+	httpClient *http.Client
+	// pollHTTPClient bounds the SDK long-poll request (90s); see telegramHTTPClient.
+	pollHTTPClient *http.Client
 
 	mu                  sync.RWMutex
 	bot                 telegramBot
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Good fix, well-tested.

PR #895, opened 2026-05-09
fix(cron): preserve explicit session_mode=reuse override over global default

kevinWangSheng:
## Summary

When the global `cron.session_mode` default is `"new_per_run"`, a per-job override of `session_mode = "reuse"` is silently lost: the job runs in `new_per_run` mode anyway, contradicting the user's explicit setting.

## Root cause

`NormalizeCronSessionMode` collapses both `""` (unspecified) and `"reuse"` to a single canonical value `""`:

```go
case "", "reuse":
    return ""
```

`AddJob` then runs `job.SessionMode = NormalizeCronSessionMode(job.SessionMode)` — so an inbound `"reuse"` is stored as `""`. At dispatch time, `UsesNewSession` does:

```go
if job.SessionMode != "" {
    return job.UsesNewSessionPerRun()
}
return cs.defaultSessionMode == "new_per_run"
```

Since `SessionMode == ""`, the per-job branch is skipped and the global default wins — exactly the case the user was trying to override.

The existing `TestCronScheduler_UsesNewSession_GlobalDefault` test passes only because it bypasses `AddJob` and assigns `job.SessionMode = \"reuse\"` directly:

```go
// Test 2: per-job \"reuse\" overrides global \"new_per_run\"
job.SessionMode = \"reuse\"          // ← raw assignment, skips Normalize
if cs.UsesNewSession(job) { ... }
```

So the unit test demonstrates the *intended* behaviour, while the production code path silently loses the override.

## Fix

Make `\"reuse\"` a third canonical value alongside `\"\"` and `\"new_per_run\"`:

| Stored value | Meaning |
|---|---|
| `\"\"` | unspecified — fall back to global default |
| `\"reuse\"` | explicit opt-out of new_per_run |
| `\"new_per_run\"` | explicit opt-in to fresh session per run |

`UsesNewSession` already routes through `job.UsesNewSessionPerRun()` whenever `SessionMode` is non-empty, and `UsesNewSessionPerRun` calls `NormalizeCronSessionMode(...) == \"new_per_run\"`. With the normalization preserving `\"reuse\"`, the chain returns `false` correctly. No call-site changes needed.

Also widen `validateCronJob` to accept the new canonical `\"reuse\"` and update the `SessionMode` JSON tag comment.

## Migration

Existing jobs on disk with `SessionMode = \"\"` keep their previous (buggy) behaviour: they continue to follow the global default. No data migration. Users whose explicit `\"reuse\"` was being clobbered just need to re-issue `cc-connect cron add --session-mode reuse …` (or `cron edit … session_mode reuse`) once the patch is in.

## Test plan

- [x] New `TestCronScheduler_AddJob_PreservesReuseOverride` regression test that goes through `AddJob` and asserts both storage and `UsesNewSession` semantics. Confirmed to **fail on main** and **pass on this branch**.
- [x] `go test -race -tags=no_web ./core/ -run \"TestCron|Cron\"` — all cron tests pass.
- [x] `go vet -tags=no_web ./core/...` clean.
- [x] `go build -tags=no_web ./...` succeeds.

### DIFF
--- a/core/cron.go
+++ b/core/cron.go
@@ -31,7 +31,7 @@ type CronJob struct {
 	Enabled     bool      `json:"enabled"`
 	Silent      *bool     `json:"silent,omitempty"`       // suppress start notification; nil = use global default
 	Mute        bool      `json:"mute,omitempty"`         // suppress ALL messages (start + result); job runs silently
-	SessionMode string    `json:"session_mode,omitempty"` // "" or "reuse" = share active session; "new_per_run" = fresh session each run
+	SessionMode string    `json:"session_mode,omitempty"` // "" = use global default; "reuse" = share active session; "new_per_run" = fresh session each run
 	Mode        string    `json:"mode,omitempty"`         // permission mode override for this job; "" = use project default
 	TimeoutMins *int      `json:"timeout_mins,omitempty"` // nil = default 30m wait; 0 = no limit; >0 = minutes
 	CreatedAt   time.Time `json:"created_at"`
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Clear regression test for the reuse override bug.

PR #901, opened 2026-05-09
fix(rich): gate rich card creation/stream by silentHold (NO_REPLY no-trace)

Cigarrr:
## Summary

Stacks on **#796**. The bug: in rich card mode, when the agent emits `NO_REPLY` to indicate a silent turn, Lark renders the resulting card lifecycle (Send → Delete) as a "Bot 撤回了一条消息" gray bar — breaking the no-trace semantics that `NO_REPLY` (#682) was introduced to provide.

This PR was previously submitted as #892 (now CLOSED) on top of the un-rebased original #796; resubmitting on the rebased typewriter-only stack.

## Reproduction

1. Lark/Feishu platform with `card_mode = "rich"`
2. Trigger an agent turn whose final assistant content is `NO_REPLY`. Easiest: cron heartbeat or DM to claudecode with a request that resolves to "nothing to say".
3. Lark UI shows a "Bot 撤回了一条消息" gray bar instead of staying silent.

## Root cause

The rich card path in `processInteractiveEvents` (`core/engine.go` `EventText` case) created the preview card unconditionally on the first text chunk and streamed every subsequent chunk, regardless of whether the text could resolve to a bare `NO_REPLY` marker. The legacy (non-rich) path correctly held streaming via `silentHold` + `couldBeSilentPrefix`.

When the agent replied with `NO_REPLY` in rich card mode (or did tool calls then `NO_REPLY`):
1. `SendPreviewStart` created the card (or it was created earlier by a tool event handler).
2. `StreamRichCardText` pushed the `NO_REPLY` text into the card body.
3. End-of-stream `isSilent` path called `DeletePreviewMessage` to clear it.
4. Lark rendered steps 1+3 as a "撤回了一条消息" gray bar.

## Fix

Pre-compute the silentHold transition (`peekSegment` = accumulated segment + this chunk) before any card creation or streaming. Both rich and legacy paths share that single transition. `couldBeSilentPrefix` is monotonically decreasing as segments grow, so the held → released transition happens at most once per segment.

**Rich `EventText` path**:
- Card creation gated on `!silentHold`.
- Lazy creation when held → released and no card exists yet, using accumulated `partialText` (so the card emerges with post-prefix content already in body, not as an empty shell).
- Throttle / streaming block runs only when `!silentHold`.

**Silent end-of-stream path**:
- When the card already exists (e.g. from earlier tool events) and the turn resolves silent, finalize the card to `Done` with body = `stripTrailingSilent(partialText)`. Preserves any pre-`NO_REPLY` text the user already saw streaming, plus any `toolSteps` panel content. Mirrors legacy + full mode where prior tool messages remain visible even when the final reply is silent.
- Only call `DeletePreviewMessage` when there is genuinely nothing visible worth keeping (no body, no toolSteps) — a transient/race fallback under the new gate.

**Legacy path simplification**: the silentHold/release transition is already done above, so the inner if/else collapses.

## Tests

5 cases in `core/engine_test.go`:

- `TestProcessInteractiveEvents_RichCard_NoReplySingleChunk` — single-chunk `NO_REPLY` produces no `SendPreviewStart`/`StreamRichCardText`/`UpdateMessage`/`DeletePreviewMessage`.
- `TestProcessInteractiveEvents_RichCard_NoReplyChunked` — same invariant when `NO_REPLY` is split across two chunks.
- `TestProcessInteractiveEvents_RichCard_PrefixThenContent` — `"N"` then `"ote that..."`: card lazy-created after release with full accumulated content.
- `TestProcessInteractiveEvents_RichCard_TextThenNoReply_PreservesBody` — visible text plus trailing `NO_REPLY` marker (engine sees `EventResult.Content="NO_REPLY"`): final card preserves the pre-marker text.
- `TestProcessInteractiveEvents_RichCard_ToolThenNoReply` — turn issues a tool call (creating the card with steps) then resolves to `NO_REPLY`: card finalized to `Done`, no `DeletePreviewMessage`.

New stub `stubRichCardSilentPlatform` implements `RichCardSupporter`, `PreviewStarter`, `MessageUpdater`, `RichCardTextStreamer`, `PreviewCleaner`.

## Test plan

- [x] `go test ./core/... -run "RichCard"` (8 cases incl. existing rich-card tests, all pass)
- [x] Local Lark e2e: `card_mode = rich` × `mode = {full, quiet, compact}` including tool-then-NO_REPLY — no recall trace, tool steps preserved on Done card

## Stacking note

This PR is stacked on **#796** (rich card typewriter). Diff against `main` includes #796's commits because the branch is built on top. Once #796 lands, this PR will rebase to land on top cleanly.

Supersedes #892 (closed). The fix is the same; this branch is rebased onto the typewriter-only #796.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -141,3 +141,4 @@ secrets/
 credentials/
 
 .codex
+cc-connect-new
--- a/core/engine.go
+++ b/core/engine.go
@@ -4211,6 +4211,15 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 				// it for the reply quote. Without this reassignment, msg2's
 				// reply would quote msg1's bubble.
 				replyCtx = queued.replyCtx
+				// Rich-mode per-turn state must reset too — otherwise EventText for
+				// the queued message would StreamRichCardText against the previous
+				// turn's cardID, overwriting that card's body with the new turn's
+				// content. Same risk for partialText/toolSteps leaking across turns.
+				cardMessageID = nil
+				toolSteps = nil
+				partialText = ""
+				lastRichCardUpdate = time.Time{}
+				lastRichCardLen = 0
 				queuedRenderer := func(content string) string {
 					return e.renderOutgoingContentForWorkspace(queued.platform, content, workspaceDir)
 				}
--- a/core/engine.go
+++ b/core/engine.go
@@ -4039,6 +4039,17 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 				}
 				finalCard := richCardSupporter.BuildRichCard(CardStatusDone, "", toolSteps, parts[0], false, time.Since(turnStart))
 				if cardMessageID != nil {
+					// Forced final flush via cardkit-v1 streaming text update before
+					// flipping status to Done via full-card Patch. The throttle in the
+					// EventText path may have skipped the last <200ms / <20 chars; this
+					// catch-up keeps the typewriter rendering smooth all the way to the
+					// end. ErrNotSupported (no cardID) and any error are silent — the
+					// subsequent UpdateMessage will rewrite the body anyway.
+					if streamer, ok := p.(RichCardTextStreamer); ok {
+						if err := streamer.StreamRichCardText(e.ctx, cardMessageID, parts[0]); err != nil && !errors.Is(err, ErrNotSupported) {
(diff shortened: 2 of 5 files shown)


chenhg5:
Hi @kevinWangSheng! This PR has a merge conflict. Please rebase onto the latest main branch:

```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```

Once resolved, CI will automatically re-run. Let me know if you need any help!

PR #902, opened 2026-05-09
fix(markdown): stop StripMarkdown corrupting snake_case and dunder identifiers

kevinWangSheng:
## Summary

`core.StripMarkdown` is used to flatten Markdown to plain text for platforms that don't render Markdown (LINE `Reply` / WeCom `Reply`) and for TTS synthesis (`engine.go:12431`). It strips `**bold**` / `*italic*` / `~~strike~~` / links / headings / code fences — fine. It also strips the **underscore** variants:

```go
reBoldUnd   = regexp.MustCompile(`__(.+?)__`)
reItalicUnd = regexp.MustCompile(`_(.+?)_`)
```

Both happily match identifier fragments. With these enabled, `StripMarkdown` silently rewrites:

| Input | Output (before) | Output (after) |
|---|---|---|
| `call my_func_name() to start` | `call myfuncname() to start` | `call my_func_name() to start` |
| `use module_name_var here` | `use modulenamevar here` | `use module_name_var here` |
| `Python's __init__ method` | `Python's init method` | `Python's __init__ method` |
| `the __version__ string` | `the version string` | `the __version__ string` |
| `_italic_ text` | `italic text` | `_italic_ text` |
| `__bold__ text` | `bold text` | `__bold__ text` |

That is, exactly the things an agent writes whenever it talks about code. The corrupted text then ships out to LINE / WeCom messages and TTS audio, where the underscores are gone and the user can no longer copy-paste or pronounce the identifier correctly.

## Why removal, not a smarter regex

Go's `regexp` package has no lookbehind, so `_italic_` can't be narrowed to "not adjacent to a word char" with a one-line regex change. And the dunder case (`__init__` flanked by spaces) is shape-identical to `__bold__` flanked by spaces — there's no structural way to distinguish them. Any context-aware fix needs `FindAllStringSubmatchIndex` plus a manual flank check, which still can't tell `__init__` from `__bold__`.

Real-world agent output overwhelmingly uses `**bold**` / `*italic*` (asterisk form). The underscore forms are mostly a hazard with little upside in this codebase, so the safe minimal fix is to drop them. Inputs containing literal `_italic_` now keep the underscores — small cosmetic loss vs. losing identifier integrity on TTS/LINE/WeCom.

## Test plan

- [x] Added `core/markdown_test.go` (no test file previously existed) covering both the preserved asterisk paths and every regression case in the table above. Confirmed to **fail on main** and **pass on this branch**.
- [x] `go test -tags=no_web ./core/ -run \"TestStripMarkdown|TestMarkdown\"` passes
- [x] `go test -tags=no_web ./platform/line/... ./platform/wecom/...` (the two non-TTS callers) — both packages green

### DIFF
--- a/core/markdown.go
+++ b/core/markdown.go
@@ -9,9 +9,7 @@ var (
 	reCodeBlock   = regexp.MustCompile("(?s)```[a-zA-Z]*\n?(.*?)```")
 	reInlineCode  = regexp.MustCompile("`([^`]+)`")
 	reBoldAst     = regexp.MustCompile(`\*\*(.+?)\*\*`)
-	reBoldUnd     = regexp.MustCompile(`__(.+?)__`)
 	reItalicAst   = regexp.MustCompile(`\*(.+?)\*`)
-	reItalicUnd   = regexp.MustCompile(`_(.+?)_`)
 	reStrike      = regexp.MustCompile(`~~(.+?)~~`)
 	reLink        = regexp.MustCompile(`\[([^\]]+)\]\(([^)]+)\)`)
 	reHeading     = regexp.MustCompile(`(?m)^#{1,6}\s+`)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Simple, targeted fix — passing prompt via stdin avoids Windows .cmd newline truncation. No breaking changes, small footprint. Approving.

PR #695, opened 2026-04-20
fix(gemini): pass prompt via stdin to preserve newlines

chenhg5:
## Summary
- Fix newline truncation when using Gemini CLI with Telegram (or any platform sending multi-line messages)
- Root cause: `-p <prompt>` CLI flag truncates at newline characters in some Gemini CLI versions
- Fix: pass prompt via stdin using `-p -` flag, matching the pattern used by the Codex agent (`cmd.Stdin = strings.NewReader(prompt)`)

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes (all 25 packages)
- [ ] Manual: send multi-line message via Telegram to Gemini agent, verify full content is received

Closes #687

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #903, opened 2026-05-09
fix(opencode): pass prompt via stdin to avoid Windows .cmd newline truncation

bingkxu:
## Summary
- Fix Windows .cmd newline truncation issue when passing prompt as positional argument
- Pass prompt via stdin instead of command-line argument to preserve newlines and special characters
- Update test to match new args format (remove `--` separator and prompt from args)

## Problem
On Windows, when opencode CLI is invoked via `.cmd` wrapper, passing prompt as positional argument causes newline truncation. This is because Windows command-line parsing differs from Unix, and newlines in arguments are often stripped or incorrectly handled by cmd.exe.

## Solution
Pass prompt via stdin instead of command-line argument:
- Remove `args = append(args, --, prompt)` from buildRunArgs
- Add `cmd.Stdin = strings.NewReader(prompt)` before launching process

This approach is already used in Gemini agent (PR #695) for similar reasons (preserve newlines in prompts).

## Test Plan
- Updated TestOpencodeSessionBuildRunArgsIncludesImagesAsFiles to expect args without prompt
- Manual testing on Windows confirms newlines are preserved in stdin-passed prompts
- All existing tests pass

## References
- Similar approach in Gemini agent: commit 1c25dd42 (PR #695)
- No existing issue/PR for OpenCode stdin prompt in upstream

### DIFF
--- a/agent/opencode/session.go
+++ b/agent/opencode/session.go
@@ -95,6 +95,7 @@ func (s *opencodeSession) Send(prompt string, images []core.ImageAttachment, fil
 
 	var stderrBuf bytes.Buffer
 	cmd.Stderr = &stderrBuf
+	cmd.Stdin = strings.NewReader(prompt)
 
 	if err := cmd.Start(); err != nil {
 		return fmt.Errorf("opencodeSession: start: %w", err)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):

PR #904, opened 2026-05-09
[codex] Capture runtime metadata and configurable Slack mentions

eSlashie:
## Summary

Capture runtime model and token/context metadata from Claude Code and Gemini CLI stream events so reply footers can show actual session state without requiring hardcoded configured model values. Also add a configurable Slack mention policy so DMs work directly while channel/private-channel/thread messages can require or skip @mention by global, per-channel, or per-thread config.

## Root Cause

Claude Code and Gemini CLI both emit runtime metadata in their stream-json outputs, but cc-connect was not caching the session-level values that the reply footer reads. Claude Code emitted the active model and context-window usage, while Gemini CLI emitted the active model plus token stats but no context window in headless stream-json mode.

Slack subscribed to message events and treated ordinary channel/thread messages like direct messages, so any channel message could trigger an agent response. The first fix made channel/thread mentions mandatory globally, but that was too coarse: some channels or threads intentionally need no-mention replies.

## Changes

- Store Claude Code runtime model from system and assistant stream-json payloads.
- Parse Claude Code result.modelUsage into core.ContextUsage for footer context percentage.
- Store Gemini CLI runtime model from init events.
- Parse Gemini CLI result.stats into event token counts and session context usage.
- Only expose Gemini context-window percentages when a context-window field is actually present in the stream data.
- Add Slack `require_mention` default policy for non-DM messages.
- Add Slack `require_mention_channels` and `require_mention_threads` overrides, with thread override taking precedence over channel override.
- Keep Slack DMs and Slack Assistant Chat working without mention.
- Keep Slack `app_mention` as an always-valid trigger and reply into the existing thread when the mention happens inside a thread.
- Deduplicate Slack `message`/`app_mention` delivery for the same channel timestamp and strip only this bot's mention in the ordinary message path.
- Update Slack config example and docs for `app_mention`, `message.im`, and optional `message.channels`/`message.groups` subscriptions.
- Return defensive copies of context usage to avoid external mutation.
- Add unit tests for Claude Code, Gemini CLI, Slack mention policy, and Slack event dedup behavior.

## Validation

- go test ./agent/claudecode
- go test ./agent/gemini ./agent/claudecode
- go test ./platform/slack
- go test ./platform/slack ./agent/gemini ./agent/claudecode
- go test ./config ./platform/slack ./agent/gemini ./agent/claudecode
- make build

## Notes

Gemini CLI interactive footer can show context percentage because it has access to internal TUI state and tokenLimit(config.getModel()). Headless stream-json currently emits simplified token stats but does not include a context window, so cc-connect does not hardcode Gemini context limits.

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -49,6 +50,10 @@ type claudeSession struct {
 	// Stop hook timeout. The wait ends as soon as the process exits,
 	// so typical shutdowns take seconds, not the full timeout.
 	gracefulStopTimeout time.Duration
+
+	runtimeMu    sync.RWMutex
+	runtimeModel string
+	contextUsage *core.ContextUsage
 }
 
 func newClaudeSession(ctx context.Context, workDir, cliBin string, cliExtraArgs []string, cliArgsFlag string, model, effort, sessionID, mode, systemPrompt string, allowedTools, disallowedTools []string, extraEnv []string, platformPrompt string, disableVerbose bool, spawnOpts core.SpawnOptions, maxContextTokens int) (*claudeSession, error) {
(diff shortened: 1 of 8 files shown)


chenhg5 (approved):
LGTM. Runtime metadata capture from Claude Code / Gemini stream events + configurable Slack mentions. No breaking changes, clean extension. Approving.

PR #905, opened 2026-05-09
fix(markdown_html): use decimal placeholder keys to avoid HTML-escape collision

kevinWangSheng:
## Summary

`convertInlineHTML` (used by `MarkdownToSimpleHTML`, the Telegram outbound rendering path) protects code/link content as placeholders before running `escapeHTML` across the whole string, then restores them at the end. The placeholder key was:

```go
key := \"\x00PH\" + string(rune('0' + phIdx)) + \"\x00\"
```

That single rune rolls past `'9'` once `phIdx` reaches 10:

| phIdx | char |
|---|---|
| 9 | `'9'` |
| 10 | `':'` |
| 11 | `';'` |
| **12** | **`'<'`** |
| 13 | `'='` |
| **14** | **`'>'`** |

So once a single message contains 13+ inline-code spans (or links), the placeholder key includes `<`. The whole-string `escapeHTML` pass then rewrites it to `&lt;` *inside* the key, and the restore pass at the end can no longer find the original key. The final HTML leaks literal `\x00PH<\x00` / `\x00PH>\x00` fragments and silently drops the original code/link content.

### Reproduction (before the fix)

Input: 15 inline-code spans in one message, `\` `f0` `\`, `\` `f1` `\`, …, `\` `fE` `\`:

```
in:  \"Functions are `f0`, `f1`, ..., `fC`, `fD`, `fE` here.\"
out: \"Functions are <code>f0</code>, ..., <code>fB</code>,
       \\x00PH&lt;\\x00, <code>fD</code>, \\x00PH&gt;\\x00 here.\"
```

`fC` and `fE` are gone, replaced with junk that Telegram renders as raw text (NUL chars and all).

## Fix

Switch the placeholder key from `string(rune('0'+phIdx))` to `strconv.Itoa(phIdx)`. Decimal digits stay in the safe ASCII range regardless of `phIdx`, never overlap with HTML-special chars, and the keys remain unique.

```go
key := \"\x00PH\" + strconv.Itoa(phIdx) + \"\x00\"
```

Single-character change in the live path; +1 import.

## Test plan

- [x] New `TestMarkdownToSimpleHTML_ManyInlineCodePlaceholders` regression test fires 15 inline-code spans through `MarkdownToSimpleHTML` and asserts the output contains no NUL byte, no escaped placeholder fragment, and that every code span is preserved. Confirmed to **fail on main** (catches the leaked `\\x00PH&lt;\\x00` / `\\x00PH&gt;\\x00`) and **pass on this branch**.
- [x] `go test -race -tags=no_web ./core/ -run TestMarkdown` — all markdown tests pass
- [x] `go test -tags=no_web ./platform/telegram/... ./platform/discord/...` — both downstream packages green

### DIFF
--- a/core/markdown_html.go
+++ b/core/markdown_html.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"regexp"
+	"strconv"
 	"strings"
 	"unicode/utf8"
 )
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):

PR #907, opened 2026-05-09
fix(feishu): prevent card action callback timeout on slow operations

bingkxu:
## Summary

- Wrap `cardNavHandler` calls in `onCardAction` with a goroutine + 2.5s deadline to prevent exceeding Feishu's 3-second card callback timeout
- Fast operations return the card inline as before (no behavior change)
- Slow operations (e.g. `agent.ListSessions()`) return a loading toast immediately, then async-refresh the card via the Feishu Patch API once complete
- Add 3 unit tests covering fast response, slow+refresh, and slow+nil-card scenarios

## Test Plan

- [x] `TestCardAction_NavFast_ReturnsCard` — fast handler returns card synchronously, no toast
- [x] `TestCardAction_NavSlow_ReturnsToastThenRefreshes` — slow handler triggers toast, then `RefreshCard` is called
- [x] `TestCardAction_NavSlow_NilCard_NoRefresh` — slow handler returning nil does not call `RefreshCard`
- [x] All 85 existing feishu platform tests pass with no regression

### DIFF
--- a/agent/opencode/opencode.go
+++ b/agent/opencode/opencode.go
@@ -674,13 +674,13 @@ func querySessionMessageCounts() map[string]int {
 
 func opencodeDBPath() string {
 	if xdg := os.Getenv("XDG_DATA_HOME"); xdg != "" {
-		return filepath.Join(xdg, "opencode", "opencode-local.db")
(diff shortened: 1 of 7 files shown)


chenhg5 (approved):

PR #908, opened 2026-05-09
fix(core): show thinking content when agent produces no text response

AndreLYL:
## Problem

When Claude Code (or other agents with extended thinking) produces only thinking blocks without any text output, cc-connect shows a bare `(empty response)` / `(空响应)` message. This gives users zero context about what happened.

**Real-world scenario**: When a user pastes sensitive content (e.g., API tokens), Claude's safety layer may only produce internal reasoning (thinking blocks) without generating a visible text response. The user gets `(空响应)` repeatedly and the session becomes unresponsive.

## Fix

- Track the last thinking block content via `lastThinkingContent` in the interactive event loop
- When `fullResponse` is empty but thinking content exists, format it as a user-visible message using a new `MsgThinkingOnlyResponse` i18n key (with translations for EN, ZH, ZH-TW, JA, ES)
- Log a `thinking-only response` warning for debugging
- Reset tracking state on new turn boundaries

## Changes

- `core/engine.go`: Add `lastThinkingContent` tracking in `EventThinking` handler, use it as fallback in `EventResult` handler
- `core/i18n.go`: Add `MsgThinkingOnlyResponse` message key with 5 language translations

## Test plan

- [ ] Verify thinking-only responses now show thinking content instead of `(空响应)`
- [ ] Verify normal responses (text + thinking) are unaffected
- [ ] Verify thinking content resets between turns

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -3337,6 +3337,7 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 
 	var textParts []string
 	var segmentStart int // index into textParts: text before this has been sent/displayed
+	var lastThinkingContent string // tracks last thinking block for thinking-only fallback
 	silentHold := false  // true while accumulated segment text could still resolve to a bare NO_REPLY marker
 	toolCount := 0
 	waitStart := time.Now()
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):

PR #914, opened 2026-05-10
fix(core/setup): handle url.Parse error in weixin poll handler to avoid nil-pointer panic

kevinWangSheng:
## Summary

`handleSetupWeixinPoll` (`core/setup.go:350`) does:

```go
u, _ := url.Parse(apiBase + \"/\")
u = u.JoinPath(\"ilink\", \"bot\", \"get_qrcode_status\")
```

For inputs like `\"://\"`, `\"://malformed\"`, or `\"%zz\"`, `url.Parse` returns a **nil `*URL`** along with a non-nil error. The discarded error means the next line dereferences a nil pointer and crashes the management server:

```
runtime error: invalid memory address or nil pointer dereference
```

Reproduced (test included): a single POST to `/api/v1/setup/weixin/poll` with `{\"qr_key\": \"x\", \"api_url\": \"://\"}` panics the handler goroutine.

The management API is authenticated, but an operator pasting a typo into the WeChat setup form (or any web client posting a malformed URL) is enough to bring the daemon down. The sibling handler `handleSetupWeixinBegin` already validates the same parse and returns HTTP 400 on error — this PR mirrors that handling in the poll handler.

## Fix

```go
u, err := url.Parse(apiBase + \"/\")
if err != nil {
    mgmtError(w, http.StatusBadRequest, \"invalid api_url\")
    return
}
```

Plus an inline comment so the next reader doesn't think this is defensive paranoia.

## Test plan

- [x] New `TestMgmt_SetupWeixinPoll_RejectsMalformedAPIURL` covers three malformed inputs (`\"://\"`, `\"://malformed\"`, `\"%zz\"`). The test wraps the handler call in a `defer/recover` so a regression to the panicking behaviour produces a clear test failure rather than crashing the test binary, then asserts the response is HTTP 400. Confirmed to **fail on `main`** with `handleSetupWeixinPoll panicked on api_url=\"://\": runtime error: invalid memory address or nil pointer dereference` and **pass on this branch**.
- [x] `go test -tags=no_web ./core/ -run TestMgmt` — all management tests pass

### DIFF
--- a/core/management_test.go
+++ b/core/management_test.go
@@ -2615,3 +2615,45 @@ func TestMgmt_CCSwitchProviders_MethodNotAllowed(t *testing.T) {
 		t.Fatal("expected DELETE on cc-switch to fail")
 	}
 }
+
+// TestMgmt_SetupWeixinPoll_RejectsMalformedAPIURL is a regression test for a
+// nil-pointer panic in handleSetupWeixinPoll. The handler did
+// `u, _ := url.Parse(apiBase + "/")` and then immediately called
+// `u.JoinPath(...)`. For inputs like "://" or "%zz", url.Parse returns a nil
+// URL plus an error; the discarded error meant the next line crashed the
+// management server with `runtime error: invalid memory address or nil
+// pointer dereference`. handleSetupWeixinBegin already validated this same
+// field; this test pins the symmetric handling here.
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. url.Parse error 处理避免 nil-pointer panic。

PR #915, opened 2026-05-10
fix(cron): send `silent` edit value as bool so management API accepts it

kevinWangSheng:
## Summary

`cc-connect cron edit <id> silent true` is documented in `printCronEditUsage` as a supported bool edit, but `runCronEdit`'s type-switch only coerced `enabled`, `mute`, and `timeout_mins` from strings — `silent` fell through to the string default.

That meant the JSON body shipped to the management API had `{\"value\": \"true\"}` (string) for silent. The server-side `updateJobField` (`core/cron.go:333`) does:

```go
case \"silent\":
    if v, ok := value.(bool); ok {
        job.Silent = &v
        return nil
    }
```

so the type-assertion failed, and the field then took the string-fallback path that uses `reflect.FieldByName(...).Kind() == reflect.String`. `CronJob.Silent` is `*bool`, not String, so the fallback also failed and the user got back the misleading error:

```
Error: failed to update field \"silent\" (may be read-only or invalid type)
```

## Reproducer

Direct against the server (no CLI/API loop needed):

```go
cs.UpdateJob(\"edit-test\", \"silent\", \"true\")
//  -> err = failed to update field \"silent\" (may be read-only or invalid type)
cs.UpdateJob(\"edit-test\", \"silent\", true)
//  -> err = <nil>
```

So the server is correct — it's the CLI that's sending the wrong JSON type for `silent`.

## Fix

Add `silent` to the bool case alongside `enabled` and `mute`:

```go
case \"enabled\", \"mute\", \"silent\":
```

While here, extract the parse logic into a small helper `parseCronEditValue` so it can be unit-tested without standing up a fake socket server.

## Test plan

- [x] New `cmd/cc-connect/cron_edit_test.go` covers all three bool fields, `timeout_mins`, the string-fallthrough fields, and an invalid-bool case (13 sub-tests). All pass.
- [x] **Verified regression-catching**: temporarily reverting `silent` out of the bool case (`sed -i.bak 's/case \"enabled\", \"mute\", \"silent\":/case \"enabled\", \"mute\":/' cmd/cc-connect/cron.go`) makes the three `silent_*` sub-tests fail with the original symptom — `parseCronEditValue(\"silent\", \"true\") = true (string), want true (bool)`. Restoring the line makes them green.
- [x] `go test -tags=no_web ./cmd/cc-connect/...` all pass

### DIFF
--- a/cmd/cc-connect/cron.go
+++ b/cmd/cc-connect/cron.go
@@ -449,26 +449,10 @@ func runCronEdit(args []string) {
 		os.Exit(1)
 	}
 
-	// Parse value based on field type
-	var value any
-	switch field {
-	case "enabled", "mute":
-		v, err := strconv.ParseBool(valueStr)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Error: %s must be true or false\n", field)
-			os.Exit(1)
-		}
-		value = v
-	case "timeout_mins":
-		v, err := strconv.Atoi(valueStr)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Error: timeout_mins must be an integer\n")
-			os.Exit(1)
-		}
-		value = v
-	default:
-		// String fields: project, session_key, cron_expr, prompt, exec, work_dir, description, session_mode
-		value = valueStr
+	value, err := parseCronEditValue(field, valueStr)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. cron edit value 解析发送正确 JSON 类型到 API。

PR #538, opened 2026-04-10
feat(cron): add manual run support for management API and Web UI

Aokoooooo:
## Summary

Add manual cron run support across the remaining management surfaces so saved cron jobs can be triggered immediately from the daemon management API, CLI, and Web UI.

This keeps the execution path unified by reusing the existing scheduler/job configuration instead of introducing a separate ad-hoc runner.

Closes #537

## Context

There is already an existing feature request for manual cron execution in #263 and an overlapping PR #357 focused on the core CLI/chat execution flow.

This branch extends the feature through the management surface area and admin UI so the capability is available consistently for daemon/API/Web users as well.

## What Changed

- add `cc-connect cron run <id>` in the CLI help and command handling
- add `CronScheduler.RunJobNow(id)` and explicit not-found / project-unavailable errors
- add `/cron run <id>` handling in the engine with localized success and error messages
- add management API support for `POST /api/v1/cron/{id}/run`
- document the new management API endpoint in English and Chinese docs
- add a Run button in the Web cron list and wire it to the cron API client
- update i18n strings for API/UI feedback across supported locales
- add regression coverage for scheduler, engine command handling, API, management API, and CLI usage

## Why

Today the manual cron trigger flow is incomplete for users who manage cc-connect through the daemon API or Web admin page.

Without this, users have to wait for the next schedule, temporarily re-enable a disabled job, or switch over to another entrypoint just to run one saved task once.

This PR makes manual triggering available from the same saved cron job definition while preserving project, session, timeout, and silent/session-mode behavior.

## Validation

- [x] `go test ./core`
- [x] `go test ./cmd/cc-connect`
- [x] `go test ./....` 

## Notes

- No breaking config change.
- Docs were updated because the management API surface changed.
- This PR overlaps conceptually with #357 but additionally covers management API and Web UI parity.

PR #357, opened 2026-03-29
feat(cron): add manual exec command for existing jobs

xukp20:
## Summary

  Add a manual `cron exec` entrypoint so an existing cron job can be triggered immediately by job ID.

  This reuses the existing cron execution path instead of introducing a separate ad-hoc runner.

  Fixes #263

  ## What changed

  - add `CronScheduler.TriggerJob(id)` for manual execution of existing jobs
  - add `/cron exec <id>` in chat command handling
  - add cron card `Run Now` action button
  - add local API endpoint `POST /cron/exec`
  - add CLI command `cc-connect cron exec <id>`
  - update cron help text and i18n strings
  - add regression tests for scheduler, card action, text command, API, and CLI usage

  ## Why

  Today cron jobs can be created, listed, edited, enabled, disabled, muted, and deleted, but they cannot be run immediately.

  This makes testing and one-off triggering awkward because users have to wait for the next scheduled time.

  `cron exec` solves that by letting an existing job run once on demand while preserving its configured project, session, prompt/exec mode, timeout, and
  other execution settings.

  ## Validation

  - [x] `go test ./core -run 'Cron|ExecuteCronJob|HandleCronExec' -v`
  - [x] `go test ./cmd/cc-connect -run 'Cron' -v`
  - [x] `$(go env GOPATH)/bin/staticcheck ./core/... ./cmd/cc-connect/...`
  - [x] `go build ./...`
  - [x] `go test ./...`

PR #917, opened 2026-05-10
feat(cron): align manual trigger command with exec

AaronZ345:
## Summary

This PR builds on the manual cron trigger work from #538 and aligns its naming with the official `cron exec` direction from #357, while keeping compatibility aliases so both paths can coexist cleanly.

Compared with #538, this PR makes the feature more complete by:
- using `cron exec` as the primary command in CLI help, chat help, cron cards, docs, and Web API calls
- keeping `cron run` and `cron trigger` as compatibility aliases instead of breaking existing usage
- keeping `/api/v1/cron/{id}/run` as a management API compatibility alias while documenting `/api/v1/cron/{id}/exec` as the primary endpoint
- adding setup-injected agent instructions via `AgentSystemPrompt()` / install docs, so agents learn that existing scheduled jobs can be triggered with `cc-connect cron exec <job-id>`
- explicitly distinguishing `cron exec <job-id>` from the `--exec <command>` flag used when creating shell-command cron jobs

In short: #538 adds the broader management/API/Web surface; #357 establishes the `exec` naming; this PR combines those choices and adds the setup documentation needed for agents to actually know the command exists.

## Tests

- `go test ./cmd/cc-connect -run 'Cron|PrintUsage'`
- `go test ./core -run 'Cron|Management|AgentSystemPrompt|HandleCardNav'`
- `cd web && pnpm build` (passes; Vite reports the existing large chunk warning)
- `go test ./...` currently has unrelated existing failures in `agent/codex`, `core` path-display assertions on `/private/var`, and `daemon` launchd status


chenhg5 (approved):
LGTM. cron exec/run/trigger 命令支持手动触发定时任务。

chenhg5:
@AaronZ345 Thanks for the work on aligning manual cron triggers with `exec`. I reviewed the PR and the CI is green, but I think one issue should be fixed before merge.

The new `/cron exec|run|trigger <id>` chat path does not check admin permissions before calling `RunJobNow`. That means a non-admin chat user who knows a shell cron job ID can manually trigger an existing shell cron job. `ExecuteCronJob` will then route shell jobs into `executeCronShell()` and run `sh -c job.Exec` / PowerShell.

This bypasses the existing permission model: `/cron addexec` correctly requires admin because it creates a shell command, and custom exec commands also require admin to execute. Manual execution of a shell cron job should have the same protection.

Suggested fix: in `cmdCronExec`, look up the job first, and if `job.IsShellJob()` and `!e.isAdmin(msg.UserID)`, return `MsgAdminRequired` before calling `RunJobNow`. Prompt-based cron jobs can still be allowed for regular users if that is the intended behavior.

A couple of non-blocking cleanup items:
- Please remove `web/tsconfig.tsbuildinfo` from the PR unless the repo intentionally tracks generated TS build cache files.
- It would be good to add small alias regression tests for CLI `cron run/trigger`, chat `/cron run|trigger`, and local socket `/cron/run`.

I would hold merge until the shell cron manual-trigger permission check is fixed.

AaronZ345:
@chenhg5  Fixed the shell cron manual-trigger admin check, removed `web/tsconfig.tsbuildinfo` from the PR, and added regression coverage for `exec`/`run`/`trigger` aliases; CI is green.

AaronZ345:
Superseded by fresh rebased PRs against the latest upstream main.

PR #918, opened 2026-05-10
fix(core): sanitize attachment FileName to prevent path traversal in SaveFilesToDisk

kevinWangSheng:
## Summary

`core.SaveFilesToDisk` takes the user-supplied `FileAttachment.FileName` and joins it directly into `workDir/.cc-connect/attachments/`:

```go
fpath := filepath.Join(attachDir, f.FileName)
os.WriteFile(fpath, f.Data, 0o644)
```

`FileName` originates from IM / HTTP upload metadata — Telegram document filenames, Slack `file_share`, Feishu file events, etc. — and is fully attacker-controlled. `filepath.Join` collapses `..` segments, so an upload whose filename contains `../` writes outside the attachments directory.

### Reproducer

```go
workDir := t.TempDir()
SaveFilesToDisk(workDir, []FileAttachment{
    {FileName: \"../../escape.txt\",       Data: []byte(\"x\")},  //  -> workDir/escape.txt
    {FileName: \"../../../way-up.txt\",     Data: []byte(\"x\")},  //  -> parent of workDir
    {FileName: `..\\..\\winescape.txt`,    Data: []byte(\"x\")},  //  -> workDir on Windows
})
```

`escape.txt` lands two levels above the intended attachments directory; `way-up.txt` lands above `workDir` entirely.

## Fix

Add a single `sanitizeAttachmentFileName` helper, called once before the join. It:

1. Normalizes `\\` to `/` so `filepath.Base` strips Windows-style separators on Linux too (an attacker doesn't get a free pass just because cc-connect runs on Linux).
2. Takes `filepath.Base` to drop directory components.
3. Rejects the `\"\"`, `\".\"`, `\"..\"` results — they would still join to attachDir or its parent — so the existing empty-name fallback kicks in and a generated `file_<ts>_<i>` name is used instead.

After the fix:

| Input | Old output | New output |
|---|---|---|
| `image.png` | attachments/image.png | attachments/image.png |
| `subdir/file.txt` | attachments/subdir/file.txt (mkdir not done → write fails silently) | attachments/file.txt |
| `../../escape.txt` | **workDir/escape.txt** (escapes!) | attachments/escape.txt |
| `../../../way-up.txt` | **parent of workDir/way-up.txt** (escapes!) | attachments/way-up.txt |
| `..\\..\\winescape.txt` | **workDir/winescape.txt** on Windows | attachments/winescape.txt |
| `..` | attempts to overwrite `attachments/..` | falls back to `file_<ts>_<i>` |

All eight call sites — `agent/{claudecode,codex,cursor,iflow,opencode,qoder,acp,pi}/session.go` — go through this central function and inherit the fix; no per-agent change needed.

## Test plan

- [x] New `core/message_test.go` covers:
  - 12 `sanitizeAttachmentFileName` unit cases including the four traversal vectors, Windows-style paths, `..` / `.` / `\"\"` edge cases.
  - `TestSaveFilesToDisk_RejectsPathTraversal` is a real end-to-end repro: walks the `workDir` tree afterwards and asserts no file landed outside `attachDir`, plus that the legitimate `ok.txt` was still written.
- [x] **Verified regression-catching**: temporarily reverting the live call back to `fname := f.FileName` (sanitization off) makes the integration test fail with the exact escape symptoms — `escape.txt` and `way-up.txt` landing outside `attachDir`. Restoring the sanitization makes it green.
- [x] `go test -tags=no_web ./core/ -run \"TestSanitize|TestSaveFilesToDisk\"` all pass

### DIFF
--- a/core/message.go
+++ b/core/message.go
@@ -81,6 +81,12 @@ type FileAttachment struct {
 // SaveFilesToDisk saves file attachments to workDir/.cc-connect/attachments/
 // and returns the list of absolute file paths. Agents can reference these paths
 // in their prompts so the CLI can read them with built-in tools.
+//
+// The attachment FileName is treated as untrusted user input (it comes from
+// the IM/HTTP upload metadata) and is sanitized to a basename before being
+// joined into attachDir. Without this, FileName="../../escape.txt" was
+// written to workDir/escape.txt — outside the intended attachments
+// directory.
 func SaveFilesToDisk(workDir string, files []FileAttachment) []string {
 	if len(files) == 0 {
 		return nil
@@ -92,7 +98,7 @@ func SaveFilesToDisk(workDir string, files []FileAttachment) []string {
 
 	var paths []string
 	for i, f := range files {
-		fname := f.FileName
+		fname := sanitizeAttachmentFileName(f.FileName)
 		if fname == "" {
 			fname = fmt.Sprintf("file_%d_%d", time.Now().UnixMilli(), i)
 		}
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. 安全修复 - path traversal 漏洞，sanitizeAttachmentFileName 正确处理，测试覆盖充分。

PR #919, opened 2026-05-10
fix(core): workspace init flow rejects absolute paths and tilde paths

happyTonakai:
## Problem

The workspace init flow fails to bind local directories in several cases:

1. **Absolute paths rejected as commands**: `looksLikeLocalDir` matched everything that wasn't a URL, but `/Users/foo` was intercepted by the slash-command guard and forwarded to the agent as an unknown command instead of being bound as a workspace.

2. **Two-message requirement**: The first message always showed the hint, even if it was already a valid path or git URL — requiring a redundant second message.

3. **Tilde paths fail**: `~/project` was not expanded, and bare `~` was not handled. Even after initial tilde support, `resolveLocalDirPath`'s escape check used `filepath.IsAbs` on the original target (`"~/project"`), which returns `false` in Go, causing the expanded absolute path to be rejected as "escaping" baseDir.

4. **Wrong success message**: Binding a local directory showed "repository clone success" instead of a directory-specific success message.

## Fix

- Rewrite `looksLikeLocalDir` to check `/word` against `builtinCommands`, distinguishing `/dir` (command) from `/root` (path).
- Process the first message immediately if it's a valid path or URL.
- Handle bare `~` in both `looksLikeLocalDir` and `resolveLocalDirPath`.
- Skip the escape check for tilde-prefixed paths since they expand to absolute paths under the user's home.
- Use `MsgWsBindSuccess` for local dir binding instead of `MsgWsCloneSuccess`.
- Replace hardcoded English messages with i18n calls.

## Changes

- `core/engine.go`: Rewrite `looksLikeLocalDir`, fix init flow first-message handling, add tilde expansion in `resolveLocalDirPath`, skip escape check for `~` paths, fix success message for local dir binding
- `core/multi_workspace_test.go`: Add test for absolute path handling

## Test plan

- [x] `go test ./core/ -run TestMulti` passes
- [x] Send a dir path directly → binds immediately without hint roundtrip
- [x] `~/project` binds correctly
- [x] `/absolute/path` binds correctly
- [x] Local dir binding shows "bind success" instead of "clone success"

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -13483,7 +13483,7 @@ func resolveLocalDirPath(target, baseDir string) (string, error) {
 	if err != nil {
 		resolved = cleaned
 	}
-	if baseDir != "" && !filepath.IsAbs(target) {
+	if baseDir != "" && !filepath.IsAbs(target) && !strings.HasPrefix(target, "~") {
 		cleanBase := filepath.Clean(baseDir)
 		if evalBase, err := filepath.EvalSymlinks(cleanBase); err == nil {
 			cleanBase = evalBase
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. workspace init flow 拒绝绝对路径和 tilde 路径，安全改进。

PR #920, opened 2026-05-10
fix(pi): sanitize attachment FileName to prevent path traversal in saveImagesToDisk

kevinWangSheng:
## Summary

`agent/pi/saveImagesToDisk` has the same path-traversal pattern that PR #918 fixed in `core.SaveFilesToDisk`, just on a different code path (pi's image-attachment writer). The attacker-controlled `ImageAttachment.FileName` (sourced from IM upload metadata) flows straight into `filepath.Join`:

```go
fname := img.FileName
if fname == "" { fname = \"image_<ts>_<i>.png\" }
fpath := filepath.Join(attachDir, fname)
os.WriteFile(fpath, img.Data, 0o644)
```

`filepath.Join` collapses `..` segments, so an upload whose filename contains `../` writes outside the attachments directory.

### Reproducer (against the real `saveImagesToDisk`)

```go
imgs := []core.ImageAttachment{
    {MimeType: \"image/png\", Data: ..., FileName: \"../../escape.png\"},
    {MimeType: \"image/png\", Data: ..., FileName: \"../../../way-up.png\"},
    {MimeType: \"image/png\", Data: ..., FileName: `..\\..\\winescape.png`},
}
```

| Filename | Where it actually lands | Where it should land |
|---|---|---|
| `../../escape.png` | **workDir/escape.png** (escapes!) | attachments/escape.png |
| `../../../way-up.png` | **parent of workDir/way-up.png** (escapes!) | attachments/way-up.png |
| `..\\..\\winescape.png` | **workDir/winescape.png** on Windows | attachments/winescape.png |

## Fix

Mirror the same sanitization as `core.SaveFilesToDisk` (PR #918): a tiny `sanitizePiAttachmentName` helper that:

1. Normalizes `\\` to `/` so `filepath.Base` strips Windows-style separators on Linux.
2. Takes `filepath.Base` to drop directory components.
3. Rejects the resulting `\"\"`, `\".\"`, `\"..\"` values so the existing empty-name fallback substitutes a generated `image_<ts>_<i>.<ext>`.

The two PRs are independent — they touch different files. This one stands alone if #918 hasn't merged yet.

## Test plan

- [x] `TestSaveImagesToDisk_RejectsPathTraversal` is an end-to-end repro: walks `workDir` afterwards and asserts no file landed outside `attachDir`, plus that the legitimate `ok.png` was still saved.
- [x] `TestSanitizePiAttachmentName` covers 8 unit cases including Windows-style paths and the `..` / `.` / `\"\"` edge cases.
- [x] **Verified regression-catching**: temporarily reverting the live call back to `fname := img.FileName` makes `TestSaveImagesToDisk_RejectsPathTraversal` fail with the exact escape symptoms (`escape.png` and `way-up.png` landing outside `attachDir`). Restoring the sanitization makes it green.
- [x] All pre-existing pi tests still pass.

### DIFF
--- a/agent/pi/pi_test.go
+++ b/agent/pi/pi_test.go
@@ -363,6 +363,81 @@ func TestSaveImagesToDisk_Empty(t *testing.T) {
 	}
 }
 
+// TestSaveImagesToDisk_RejectsPathTraversal is a regression test for a path
+// traversal vulnerability in saveImagesToDisk: the user-supplied
+// ImageAttachment.FileName (sourced from IM upload metadata) was passed
+// directly to filepath.Join, so a malicious uploader could escape the
+// attachments directory by using `../` segments. This mirrors the same
+// issue and fix in core.SaveFilesToDisk.
+func TestSaveImagesToDisk_RejectsPathTraversal(t *testing.T) {
+	workDir := t.TempDir()
+	attachDir := filepath.Join(workDir, ".cc-connect", "attachments")
+
+	images := []core.ImageAttachment{
+		// Two levels up — escapes attachments/ and .cc-connect/.
+		{MimeType: "image/png", Data: []byte("payload"), FileName: "../../escape.png"},
+		// Three levels up — would land outside workDir entirely.
+		{MimeType: "image/png", Data: []byte("payload"), FileName: "../../../way-up.png"},
+		// Windows-style separators must also be stripped on Linux.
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. 安全修复 - path traversal 漏洞，与 core.SaveFilesToDisk 同样的处理逻辑，测试覆盖充分。

PR #922, opened 2026-05-10
fix(core/i18n): protect lang/detected fields with mutex to fix data race

kevinWangSheng:
## Summary

`I18n` had no synchronization on its mutable state (`lang`, `detected`, `saveFunc`), but the type is shared across every concurrent platform message handler in the engine:

| Method | Field access | Caller |
|---|---|---|
| `DetectAndSet` | **writes** `i.detected` | `processInteractiveMessageWith` + queued-message dequeue (3 sites in `core/engine.go`) |
| `T` / `CurrentLang` / `IsZhLike` | **reads** `i.lang`, `i.detected` | dozens of reply / footer / error / mode-display paths |
| `SetLang` | **writes** both | management API |

Two messages arriving simultaneously from different platforms means two unsynchronized writes to the same `detected` field, and the read paths run alongside.

## Reproducer

```go
func TestI18n_DetectAndSet_Race(t *testing.T) {
    i := NewI18n(LangAuto)
    var wg sync.WaitGroup
    for n := 0; n < 100; n++ {
        wg.Add(2)
        go func() { defer wg.Done(); i.DetectAndSet(\"hello world\") }()
        go func() { defer wg.Done(); i.DetectAndSet(\"你好世界\") }()
    }
    wg.Wait()
}
```

`go test -race` flags:

```
WARNING: DATA RACE
Read at 0x...: core.(*I18n).DetectAndSet
Previous write at 0x...: core.(*I18n).DetectAndSet
...
testing.go:1617: race detected during execution of test
```

## Fix

Add `sync.RWMutex` to `I18n`. Lock policy:

- `DetectAndSet` uses an RLock fast-path so the common case (lang already detected and unchanged) is read-only. Only the mutating branch takes the write lock, and re-checks under it to avoid losing a concurrent update.
- `T` / `CurrentLang` / `IsZhLike` take `RLock` for the duration of the read.
- `SetLang` / `SetSaveFunc` take the write lock.
- `currentLang` stays unexported and assumes the caller holds a lock; every public caller now does.

## Test plan

- [x] New `TestI18n_ConcurrentAccess`: 16 goroutines × 200 iterations fanning `DetectAndSet` / `T` / `CurrentLang` / `IsZhLike` across four input languages (English / Chinese / Japanese / Spanish).
- [x] **Verified regression-catching**: `git stash` of just `core/i18n.go` (keeping the test) and re-running under `-race` reproduces `race detected during execution of test` on main. Restoring the fix makes it green.
- [x] All pre-existing `TestI18n_*` and `TestDetectLanguage` / `TestIsChinese` / `TestIsJapanese` tests still pass.
- [x] Confirmed unrelated pre-existing failures in `core/` race run (`TestProcessInteractiveEvents_AppendsReplyFooterWhenEnabled` cosmetic footer string, `TestResolveLocalDirPath_AcceptsSubdir` macOS `/private/var` symlink) reproduce on `main` without this change.

### DIFF
--- a/core/i18n.go
+++ b/core/i18n.go
@@ -1,6 +1,9 @@
 package core
 
-import "fmt"
+import (
+	"fmt"
+	"sync"
+)
 
 // Language represents a supported language
 type Language string
@@ -14,8 +17,15 @@ const (
 	LangSpanish            Language = "es"
 )
 
-// I18n provides internationalized messages
+// I18n provides internationalized messages.
+//
+// All exported methods are safe to call from multiple goroutines: cc-connect
+// fans out platform message handlers concurrently, all of which can call
+// DetectAndSet (writes `detected`) and T / CurrentLang (read `lang`/`detected`)
+// at the same time. Without the mutex `go test -race` flags real data races
+// on the language fields.
 type I18n struct {
+	mu       sync.RWMutex
 	lang     Language
 	detected Language
 	saveFunc func(Language) error
@@ -26,6 +36,8 @@ func NewI18n(lang Language) *I18n {
 }
 
 func (i *I18n) SetSaveFunc(fn func(Language) error) {
+	i.mu.Lock()
+	defer i.mu.Unlock()
 	i.saveFunc = fn
 }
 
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. i18n lang/detected 字段 mutex 保护，修复 data race。

PR #925, opened 2026-05-11
fix(claudecode): isolate spawn into process group to reap MCP grandchildren

### LINKED ISSUE #924: Child MCP server processes become zombies and consume 100% CPU when cc-connect exits abnormally
### Environment

- **cc-connect version**: 1.3.2 (Homebrew bottle, built 2026-04-21)
- **OS**: macOS Darwin 25.3.0 (arm64 / Apple Silicon)
- **Installed via**: Homebrew
- **Bridged platform**: Telegram (via Claude Code `telegram@claude-plugins-official` v0.0.6)
- **Child runtime**: Bun 1.3.11 running `server.ts`

### Summary

When `cc-connect` exits abnormally (force-killed, crashed, or restarted via `brew upgrade`), the child MCP server processes it spawned **do not exit** — they lose their parent's stdio but enter a tight CPU spin-loop instead of detecting EOF and shutting down. Each leaked child consumes ~95–100% of a single CPU core indefinitely.

### Symptoms

On my machine I discovered **5 leaked `bun server.ts` processes**, each pinned to ~95–100% CPU, with cumulative CPU time of ~3500 minutes:

```
PID    %CPU   ELAPSED       COMMAND
32086  96.1   12:30:40      /opt/homebrew/bin/bun server.ts
33642  95.7   12:25:06      /opt/homebrew/bin/bun server.ts
34193  96.5   12:23:52      /opt/homebrew/bin/bun server.ts
35607  94.7   12:19:38      /opt/homebrew/bin/bun server.ts
88173  94.2   07:03:03      /opt/homebrew/bin/bun server.ts
```

All five had identical `cwd`:

```
/Users/<me>/.claude/plugins/cache/claude-plugins-official/telegram/0.0.6
```

All five held ESTABLISHED TCP connections to Telegram MTProto endpoints (`149.154.166.x:443`) — i.e., they still believed they were the active bridge instance.

The start timestamps were clustered around moments when cc-connect was restarted or interrupted, consistent with one leak per abnormal restart.

### Root Cause (hypothesis)

`cc-connect` communicates with the spawned MCP server child over stdio (typical MCP transport). When the parent dies:

1. The OS closes the stdio pipes.
2. The child's stdin `read()` should return EOF.
3. The child should treat EOF on its MCP transport as "parent gone → exit."
4. **Instead**, the child appears to ignore EOF and enters a busy loop (likely an unhandled async-iterator / event-loop tight retry, given the 100% single-core CPU signature).

This means cc-connect leaks one runaway process per abnormal exit, and the leaks survive across cc-connect restarts. Over a few days, this can saturate multiple CPU cores and drain battery / heat the laptop.

### Reproduction

1. Install cc-connect 1.3.2 and configure the Telegram bridge per the README.
2. Start cc-connect (foreground or daemon).
3. Verify a `bun server.ts` child process is running under the Claude Code Telegram plugin path.
4. **Forcibly terminate cc-connect**: `kill -9 $(pgrep cc-connect)` (or simulate a crash).
5. Observe the child `bun server.ts` is **still running** and now consuming ~100% CPU.
6. Restart cc-connect → a new child spawns; the old one remains.
7. Repeat → CPU usage stacks linearly.

### Expected Behavior

When cc-connect exits (clean or unclean), its spawned MCP server children should detect the broken transport and exit within a few seconds.

### Suggested Fixes

Several layers could help — ideally all of them:

1. **Process-level safety net (cc-connect side)**: when starting a child, set `prctl(PR_SET_PDEATHSIG, SIGTERM)` on Linux or use the equivalent `kqueue`/`NOTE_EXIT` on macOS so the child is reaped automatically when the parent dies. Alternatively, pass a parent-PID and have the child poll it.
2. **Heartbeat / health-check**: send a periodic ping over the MCP transport; the child exits after N missed heartbeats.
3. **Explicit EOF handling in the spawned MCP server**: not strictly cc-connect's bug, but cc-connect could ship a small wrapper that wraps the stdio child and exits on EOF detection. (The underlying MCP server may not handle EOF correctly — this is a defense-in-depth measure.)
4. **Startup reaper**: on cc-connect start, scan for orphaned children matching its known spawn signature and reap them before starting new ones. (`--force` exists for the cc-connect main process, but not for its children.)
5. **Documentation**: until fixed, document the issue and recommend `cc-connect daemon stop` over `kill` / SIGKILL.

### Mitigations I Used

- `kill -9 <pid>` on each leaked child.
- Backed up `~/.claude/settings.json` and `~/.claude/plugins/installed_plugins.json`.
- Uninstalled the Telegram Claude Code plugin to break the chain until this is fixed.

### Severity

I'd rate this **high**:
- Silent, accumulates over time.
- Drains laptop battery and heats the machine.
- The only signal to the user is "fan is loud" — most users won't connect it back to cc-connect.

Happy to provide further debugging info (process samples, `lsof` output, etc.) if helpful. Thanks for building cc-connect!

meikocho1:
## Summary

Fixes #924 — `claudecode` agent leaks MCP grandchildren (e.g. the Telegram bridge `bun server.ts`) that spin at ~95–100% CPU when cc-connect shuts down.

The leaked children were the MCP servers Claude Code itself spawns. cc-connect's `claudeSession.Close()` only signaled the direct child, not the descendants — so MCP servers like the Telegram bridge became orphans, and at least one of them entered a busy-loop on EOF instead of exiting. Across a few abnormal restarts the leaks stacked; in my case 5 leaked `bun server.ts` processes had burned a combined ~3500 minutes of CPU time before I noticed the fan noise.

## Approach

Port the **same process-group isolation already used by the codex agent** (`agent/codex/proc_unix.go`, `agent/codex/proc_windows.go`):

- New `agent/claudecode/proc_unix.go` and `agent/claudecode/proc_windows.go` provide `prepareCmdForKill`, `signalProcessGroup`, and `forceKillCmd`.
- `prepareCmdForKill(cmd)` is called right after `core.BuildSpawnCommand` so the child inherits a fresh process group (`Setpgid` on unix, `CREATE_NEW_PROCESS_GROUP` on Windows).
- `claudeSession.Close()` Phase 2 (SIGTERM) and Phase 3 (SIGKILL) now signal the whole group via `syscall.Kill(-pid, …)` / `taskkill /T`, reaping descendants.

The shutdown ladder is unchanged in shape:

```
Phase 1: stdin close   → clean EOF, lets Stop hooks (e.g. claude-mem) run
Phase 2: SIGTERM group → was: SIGTERM direct child
Phase 3: SIGKILL group → was: SIGKILL direct child
```

## Scope / Limitations

This is **defense-in-depth, not a complete fix**:

| Scenario | Before | After |
|---|---|---|
| Normal cc-connect shutdown | leaks grandchildren | reaped via process group |
| `brew upgrade` / daemon restart | leaks grandchildren | reaped via process group |
| Config reload | leaks grandchildren | reaped via process group |
| `kill -9 <cc-connect>` (Close never runs) | leaks grandchildren | **still leaks** — needs PDEATHSIG (Linux-only) or a child-side parent-PID poll |

The `kill -9` case is the original trigger in #924 but the fix that covers it lives on the *child* side, not in cc-connect. The process-group change here at least makes every code path that *does* call `Close()` reliable, which I believe covers most leaks in practice.

A follow-up could add `Pdeathsig: syscall.SIGTERM` in a Linux-only `proc_linux.go` to also handle the `kill -9` supervisor case on Linux. I left that out to keep this PR focused and to match the codex pattern exactly.

## Testing

- `go test ./agent/claudecode/...` — passes, including a new regression test that spawns a backgrounded grandchild and verifies `forceKillCmd` reaps the whole group.
- `go test -race ./agent/claudecode/...` — passes.
- `go vet ./...` — clean.
- `go test ./...` — all packages pass; one flaky `TempDir RemoveAll` failure in `tests/release_local/engine_matrix/TestUnknownSlashCommandNotifiesThenFallsThroughToAgent` was confirmed unrelated (passes on isolated re-runs and exists on `main`).

## Test plan

- [ ] On Linux/macOS: build cc-connect, run the Telegram + Claude Code config, exercise a session, then issue `cc-connect daemon stop` and confirm no `bun server.ts` (or other MCP server) child survives.
- [ ] On macOS: kill cc-connect's main PID with `kill -TERM`, then verify all descendants are gone within ~5s.
- [ ] On Windows: confirm `taskkill /T /F` is invoked and the spawned process tree exits.

Closes #924

### DIFF
--- a/agent/claudecode/proc_unix.go
+++ b/agent/claudecode/proc_unix.go
@@ -1,0 +1,48 @@
+//go:build unix
+
+package claudecode
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+// prepareCmdForKill puts the spawned child into its own process group so that
+// the entire descendant tree can be terminated with a single signal aimed at
+// the negative PID. Without this, cc-connect can only signal the direct
+// child (e.g. the `claude` CLI), leaving any grandchildren (MCP server
+// processes such as the Telegram bridge) as orphans that may spin at 100%
+// CPU when their parent disappears.
+//
+// Mirrors the pattern used by agent/codex/proc_unix.go.
+func prepareCmdForKill(cmd *exec.Cmd) {
+	if cmd == nil {
+		return
+	}
+	if cmd.SysProcAttr == nil {
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
+	}
+	cmd.SysProcAttr.Setpgid = true
+}
+
+// signalProcessGroup sends sig to the entire process group rooted at cmd.
+// Returns nil if the group is already gone.
+func signalProcessGroup(cmd *exec.Cmd, sig syscall.Signal) error {
+	if cmd == nil || cmd.Process == nil {
+		return nil
+	}
+	if err := syscall.Kill(-cmd.Process.Pid, sig); err != nil &&
+		!errors.Is(err, os.ErrProcessDone) &&
+		!errors.Is(err, syscall.ESRCH) {
+		return err
+	}
+	return nil
+}
+
+// forceKillCmd SIGKILLs the entire process group rooted at cmd. Use this
+// as the last-resort escalation when graceful shutdown has timed out.
+func forceKillCmd(cmd *exec.Cmd) error {
+	return signalProcessGroup(cmd, syscall.SIGKILL)
+}
--- a/agent/claudecode/proc_unix_test.go
+++ b/agent/claudecode/proc_unix_test.go
@@ -1,0 +1,113 @@
+//go:build unix
+
+package claudecode
+
+import (
+	"os/exec"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestPrepareCmdForKill_SetsSetpgid(t *testing.T) {
+	cmd := exec.Command("/bin/true")
+	prepareCmdForKill(cmd)
+	if cmd.SysProcAttr == nil {
+		t.Fatal("SysProcAttr is nil after prepareCmdForKill")
+	}
+	if !cmd.SysProcAttr.Setpgid {
+		t.Fatal("Setpgid not set after prepareCmdForKill")
+	}
+}
+
+func TestPrepareCmdForKill_PreservesExistingSysProcAttr(t *testing.T) {
+	cmd := exec.Command("/bin/true")
+	cmd.SysProcAttr = &syscall.SysProcAttr{Foreground: false}
+	prepareCmdForKill(cmd)
+	if !cmd.SysProcAttr.Setpgid {
+		t.Fatal("Setpgid not set when SysProcAttr was pre-populated")
+	}
+}
+
+func TestPrepareCmdForKill_NilCmd(t *testing.T) {
+	// Must not panic on a nil *exec.Cmd.
+	prepareCmdForKill(nil)
+}
+
+func TestForceKillCmd_NoProcess(t *testing.T) {
+	cmd := exec.Command("/bin/true")
+	// cmd has not been Start()ed, so cmd.Process is nil.
+	if err := forceKillCmd(cmd); err != nil {
+		t.Errorf("expected no error on un-started cmd, got %v", err)
+	}
+}
+
+func TestForceKillCmd_NilCmd(t *testing.T) {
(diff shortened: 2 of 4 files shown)


chenhg5 (approved):
LGTM. claudecode process group isolation，清理 MCP grandchildren orphan。

PR #926, opened 2026-05-11
fix(core): dir card and status card show wrong workDir in multi-workspace mode

happyTonakai:
## Problem

In multi-workspace mode, `/dir` successfully switches the working directory (confirmed via `/shell pwd`), but:

1. The Feishu interactive card highlights the original workspace init path instead of the currently active one
2. The displayed "current working directory" in both the `/dir` text output and card shows the initial workspace path, not the `/dir` override
3. The `/status` card also displays the wrong working directory

## Root Cause

Two separate issues:

**Issue 1**: `sessionContextForKey` resolves the workspace agent using the binding's original path without checking `projectState` dir overrides. `renderDirCard` calls `sessionContextForKey` → `GetWorkDir()`, which returns the init path instead of the `/dir` override.

**Issue 2**: `dirApply` stores overrides in `projectState` without updating the agent's `workDir` in multi-workspace mode (by design — the agent is shared). But `renderDirCard`, `renderStatusCard`, and `cmdDir` all used `switcher.GetWorkDir()` directly, which always returns the initial workspace path.

## Fix

1. Apply `resolveChannelWorkDir` in `sessionContextForKey` before `getOrCreateWorkspaceAgent`, matching the same pattern used by `workspaceContext`
2. Add `resolveCurrentWorkDir` helper that checks `projectState` for workspace dir overrides before falling back to the agent's `workDir`
3. Use the helper in `renderDirCard`, `renderStatusCard`, and `cmdDir`

## Changes

- `core/engine.go`:
  - `sessionContextForKey`: add `resolveChannelWorkDir` call before `getOrCreateWorkspaceAgent`
  - Add `resolveCurrentWorkDir(sessionKey, switcher)` helper
  - `renderDirCard`: use `resolveCurrentWorkDir` for display and highlight
  - `renderStatusCard`: use `resolveCurrentWorkDir` for workDir display
  - `cmdDir`: use `resolveCurrentWorkDir` for text output

## Test plan

- [x] `go test ./core/ -count=1` — no new failures
- [x] `/dir <path>` in multi-workspace mode → card correctly highlights the new directory
- [x] `/dir` text output shows the correct current directory
- [x] `/status` card shows the correct working directory
- [x] Without override → behavior unchanged

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -6219,7 +6219,7 @@ func (e *Engine) dirApply(agent Agent, sessions *SessionManager, interactiveKey,
 }
 
 func (e *Engine) cmdDir(p Platform, msg *Message, args []string) {
-	agent, sessions, interactiveKey, err := e.commandContext(p, msg)
+	agent, sessions, interactiveKey, _, err := e.commandContextWithWorkspace(p, msg)
 	if err != nil {
 		e.reply(p, msg.ReplyCtx, e.i18n.Tf(MsgWsResolutionError, err))
 		return
@@ -6230,7 +6230,7 @@ func (e *Engine) cmdDir(p Platform, msg *Message, args []string) {
 		return
 	}
 
-	currentDir := switcher.GetWorkDir()
+	currentDir := e.resolveCurrentWorkDir(msg.SessionKey, switcher)
 
 	if len(args) == 0 {
 		if supportsCards(p) {
@@ -6871,8 +6871,8 @@ func (e *Engine) renderStatusCard(sessionKey string, userID string) *Card {
 	}
 
 	workDirStr := ""
-	if wd, ok := agent.(interface{ GetWorkDir() string }); ok {
-		workDirStr = strings.TrimSpace(wd.GetWorkDir())
+	if wd, ok := agent.(WorkDirSwitcher); ok {
+		workDirStr = strings.TrimSpace(e.resolveCurrentWorkDir(sessionKey, wd))
 	}
 	if workDirStr == "" {
 		workDirStr, _ = os.Getwd()
@@ -10363,7 +10363,7 @@ func (e *Engine) renderDirCard(sessionKey string, page int) (*Card, error) {
 	if !ok {
 		return nil, fmt.Errorf("%s", e.i18n.T(MsgDirNotSupported))
 	}
-	currentDir := switcher.GetWorkDir()
+	currentDir := e.resolveCurrentWorkDir(sessionKey, switcher)
 	var history []string
 	if e.dirHistory != nil {
 		history = e.dirHistory.List(e.name)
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. dir card 正确高亮 multi-workspace 模式下的 workspace。

chenhg5 (approved):
Conclusion: Approve

Overall assessment: Re-review confirms the dir-card / status-card workDir fix integrates cleanly with current main. The new commit extends the original fix to cover the `/dir` text output and `/status` card display — same root cause, complementary surfaces.

Review scope:
- Reviewed core/engine.go (PR's actual change): +30/-6 across `cmdDir`, `renderStatusCard`, `renderDirCard`, `sessionContextForKey`, plus a new `resolveCurrentWorkDir` helper.
- Verified `WorkDirSwitcher` interface (core/interfaces.go:470), `WorkspaceDirOverride` (core/projectstate.go:43), `commandContextWithWorkspace` (core/engine.go:13132), and `resolveChannelWorkDir` (core/engine.go:2741) all unchanged in current main.
- Confirmed no interface drift between prior-review commit e9993dd2 and current head 3f4e19c9.

✅ What looks good:
- The new `resolveCurrentWorkDir` helper centralizes the "what directory should this card display?" logic — checks `projectState.WorkspaceDirOverride(interactiveKey)` (multi-workspace mode only), validates the override via `os.Stat` + `IsDir()` (silently falls back if override path no longer exists), then falls back to `switcher.GetWorkDir()`. Single source of truth for the three render sites.
- `cmdDir` now uses `commandContextWithWorkspace` instead of `commandContext` — consistent with the prior fix's pattern in `sessionContextForKey` and `workspaceContext`. The discarded `effectiveDir` return value (`_`) is fine here because the helper already returns the resolved dir when needed.
- `renderStatusCard` switched from inline `interface{ GetWorkDir() string }` assertion to the named `WorkDirSwitcher` type. Cleaner, more consistent with the rest of engine.go (other call sites at lines 5376/5594/6106/6227 already use the named type).
- The `os.Stat` guard in `resolveCurrentWorkDir` defends against stale overrides pointing to deleted directories — important since `dirApply` only sets the override and `ClearWorkspaceDirOverride` only runs on error paths.
- Tests: `go test -tags no_web -count=1 ./core/...` PASS (41.6s); `go vet -tags no_web ./core/...` clean.
- CI: 5/5 green (lint 1m20s, perf 44s, regression 29s, smoke 25s, unit 3m49s) on SHA 3f4e19c9.
- Mergeable: clean.

- The new commit completes the fix: prior review (e9993dd2) addressed `sessionContextForKey` for `renderDirCard` agent lookup, but `/dir` text output and `/status` card still showed the stale init path. This commit closes that gap.
- ~20 commits landed on main since the prior review (PR #1204, #1230, #1246, #1280, #1276, etc.) — none of them changed the function signatures or interfaces PR #926 depends on. Cross-checked: `WorkDirSwitcher` (core/interfaces.go:470), `ProjectStateStore.WorkspaceDirOverride` (core/projectstate.go:43), `commandContextWithWorkspace` (core/engine.go:13132), `resolveChannelWorkDir` (core/engine.go:2741) all unchanged.

🔵 Optional (not blocking):
1. `resolveCurrentWorkDir` calls `os.Stat` on every `/dir`, `/status`, and `renderDirCard` invocation. For interactive use (where users run these commands manually) the overhead is negligible, but if any high-frequency code path starts using the helper in a tight loop, consider memoizing the stat result keyed by `interactiveKey + override`. Pure performance nit; current usage is fine.
2. The helper is only invoked when `e.multiWorkspace && e.projectState != nil`. Single-workspace mode (the default for most deployments) bypasses the multi-workspace branch entirely — fallback path is `switcher.GetWorkDir()`. Worth a comment near `multiWorkspace` declaration linking to this helper so future readers understand the two-mode distinction. Cosmetic.
3. The PR description's test plan is manual (Feishu interactive card highlight). A unit test for `resolveCurrentWorkDir` covering (a) multi-workspace with override, (b) multi-workspace without override, (c) override points to deleted dir, (d) single-workspace mode would pin the semantics. Not blocking — manual testing is sufficient for the scope — but a small regression-test addition would harden future refactors.

Testing / Risk:
- Verified locally: `go test -tags no_web -count=1 ./core/...` PASS, `go vet -tags no_web ./core/...` clean.
- CI run 25783785694 (5/5 green on SHA 3f4e19c9).
- Risk: Low. The fix is additive: `resolveCurrentWorkDir` is a new helper, three call sites switch from `switcher.GetWorkDir()` to the helper, and `sessionContextForKey` adds the `resolveChannelWorkDir` call (same pattern as `workspaceContext`).
- Backwards compatible: single-workspace mode is a no-op pass-through to `switcher.GetWorkDir()`. Multi-workspace mode only changes display — the underlying agent behavior (`/shell pwd`, file ops, etc.) is unchanged.
- Multi-platform: `cmdDir`, `renderStatusCard`, `renderDirCard` are platform-agnostic engine paths; the fix applies uniformly to Feishu, Telegram, Discord, Slack, DingTalk, etc.

Next step: OK to merge. The dir-card highlight, `/dir` text output, and `/status` card display will all correctly reflect the `/dir` override in multi-workspace mode. Optional follow-up: add a unit test for `resolveCurrentWorkDir` covering the 4 cases listed above.

PR #928, opened 2026-05-11
fix(cron): expand /skill slash prompts via skill registry

### LINKED ISSUE #856: Cron slash skill prompts are not expanded

kevinWangSheng:
## Summary

A cron job configured with a slash-prefixed prompt like `/daily-brief today` was sent to the agent as literal text — the skill's prompt body was never resolved, so scheduled jobs could not reuse the same skills as chat-triggered invocations.

This change makes ` Engine.ExecuteCronJob` mirror the interactive command path: when `job.Prompt` starts with `/`, look up the leading token in the engine skill registry and, if found, replace the message content with `BuildSkillInvocationPrompt(skill, args)`. Unknown slash commands and non-slash prompts pass through unchanged.

Fixes #856.

## What changed

- `core/engine.go` — Resolve `/skill` slash prompts through `e.skills.Resolve` before constructing the cron `Message`. Only the message body sent to the agent is rewritten; `job.Prompt` itself is left untouched so the cron run notice and listing keep showing what the user configured.
- `core/engine_test.go` — New table-driven test `TestExecuteCronJob_ExpandsSlashSkillPrompt` covering four cases:
  1. registered skill expands with args
  2. registered skill expands with no args
  3. unknown slash command passes through literally
  4. non-slash prompt passes through unchanged

The patch is ~10 lines plus the new test, scoped strictly to the cron flow.

## Test plan

- [x] `go test -race -tags no_web -run TestExecuteCronJob_ ./core/` — all cron subtests pass
- [x] `go test -race -tags no_web ./core/... ./agent/... ./platform/... ./config/... ./cmd/...` — full suite green on the changed packages (one pre-existing `daemon/launchd` failure on `TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable` reproduces on `main` too — unrelated environment-dependent test).

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1103,12 +1103,23 @@ func (e *Engine) ExecuteCronJob(job *CronJob) error {
 		return e.executeCronShell(effectivePlatform, replyCtx, job)
 	}
 
+	content := job.Prompt
+	if strings.HasPrefix(content, "/") {
+		parts := strings.Fields(content)
+		if len(parts) > 0 {
+			cmd := strings.ToLower(strings.TrimPrefix(parts[0], "/"))
+			if skill := e.skills.Resolve(cmd); skill != nil {
+				content = BuildSkillInvocationPrompt(skill, parts[1:])
+			}
+		}
+	}
+
 	msg := &Message{
 		SessionKey:   sessionKey,
 		Platform:     platformName,
 		UserID:       "cron",
 		UserName:     "cron",
-		Content:      job.Prompt,
+		Content:      content,
 		ReplyCtx:     replyCtx,
 		ModeOverride: job.Mode,
 	}
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -10766,6 +10766,91 @@ func TestExecuteCronJob_WorkspacePrefixedSessionKey(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
LGTM. cron job 中 /skill slash prompt 通过 skill registry 扩展。

PR #930, opened 2026-05-11
fix(tts): honor ctx cancellation in EspeakTTS and PicoTTS subprocesses

kevinWangSheng:
## Summary

`EspeakTTS.Synthesize` and `PicoTTS.Synthesize` accept a `context.Context` parameter but spawn the underlying `espeak` / `pico2wave` processes via `exec.Command`, so the context is silently ignored. When the caller cancels (session stop, request timeout, etc.) the subprocess keeps running to natural completion, leaking CPU and a process slot per cancelled request.

Sibling `EdgeTTS.Synthesize` already correctly uses `exec.CommandContext`.

## Fix

Two-line change: switch both call sites in `core/tts.go` to `exec.CommandContext(ctx, ...)`.

## Test

Added `TestEspeakTTS_HonorsContextCancellation` and `TestPicoTTS_HonorsContextCancellation` in `core/tts_test.go`. Each writes a temp shell script that sleeps 30 s, points the corresponding TTS struct at it, calls `Synthesize` with a pre-cancelled context, and asserts the call returns in well under 5 s with a non-nil error.

Verified the new tests catch the bug:
- With the fix → both tests pass in `0.00s`.
- After reverting only the `EspeakTTS` change → `TestEspeakTTS_HonorsContextCancellation` waits the full 30 s and fails. Restored the fix afterwards.

Tests skip on Windows since the fake binary uses `/bin/sh`.

## Test plan

- [x] `go test -race -tags no_web -run 'TTS|Tts|tts' ./core/` — all TTS tests pass
- [x] New regression tests pass under `-race`
- [x] Five pre-existing test failures on `core/` (footer ctx-indicator and resolve-local-dir-path) reproduce on `main` and are unrelated to this change (tilde-expansion / `/private` symlink on macOS).

### DIFF
--- a/core/tts.go
+++ b/core/tts.go
@@ -429,7 +429,7 @@ func (e *EspeakTTS) Synthesize(ctx context.Context, text string, opts TTSSynthes
 
 	// Execute espeak command
 	// Use Output() instead of CombinedOutput() to avoid mixing stderr warnings with audio data
-	cmd := exec.Command(e.Path, args...)
+	cmd := exec.CommandContext(ctx, e.Path, args...)
 	output, err := cmd.Output()
 	if err != nil {
 		return nil, "", fmt.Errorf("espeak: voice=%s text=%q: %w", voice, text, err)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. TTS subprocess 使用 exec.CommandContext 响应 ctx cancellation，测试覆盖。

PR #932, opened 2026-05-11
fix(claudecode): handle array content in scanSessionMeta

### LINKED ISSUE #931: [Bug] claudecode scanSessionMeta skips messages with array content, causing /list to show incorrect message counts
## Bug Description

The `scanSessionMeta` function in `agent/claudecode/claudecode.go` uses a struct with `Content string` to parse Claude Code JSONL entries. When Claude Code stores messages with array-type `content` (e.g., assistant messages with thinking blocks, or user messages with tool results), `json.Unmarshal` returns an error, causing the entire entry to be skipped via `continue`.

This leads to `/list` showing significantly fewer messages than actually exist in the session.

## Environment

- cc-connect version: latest main (commit 003d96d)
- Agent: claudecode (Claude Code)
- OS: macOS/Linux (affects all platforms)

## Reproduction Steps

1. Start a cc-connect session with claudecode agent
2. Have a multi-turn conversation that includes tool calls (e.g., file reads, bash commands)
3. Run `/list`
4. Observe that the message count is much lower than expected

## Expected vs Actual

**Expected:** `/list` shows the total number of user + assistant message turns in the session.

**Actual:** `/list` only counts entries where `content` is a plain string. All assistant messages (which use array content with thinking+text blocks) and tool_result user messages are skipped.

## Root Cause

In `scanSessionMeta` (line ~491):

```go
var entry struct {
    Type    string `json:"type"`
    Message struct {
        Content string `json:"content"`  // <-- Assumes string, fails on array
    } `json:"message"`
}
if err := json.Unmarshal(scanner.Bytes(), &entry); err != nil {
    continue  // Skips the entire entry when content is array
}
```

Claude Code changed its JSONL format around April 2026 to use array content for assistant messages. The `scanSessionMeta` code was written in February 2026 before this format change.

## Proposed Fix

Change `Content string` to `Content json.RawMessage` so that unmarshaling succeeds regardless of content type. The message count logic only needs `entry.Type` (which is still parsed correctly), not the content itself.

For summary extraction, check if the RawMessage is a JSON string before extracting.

## References

- Original commit introducing `scanSessionMeta`: 92164b3 (2026-02-28)
- Other agents (cursor, codex, opencode) handle this correctly by using flexible content parsing

---

**Note:** This affects both `MessageCount` accuracy in `/list` and summary extraction (when the last user message has array content, the summary falls back to empty).

YongmaoLuo:
Claude Code changed its JSONL format around April 2026 to use array-type content for assistant messages (thinking+text blocks) and user messages with tool results. The previous `Content string` field caused `json.Unmarshal` to fail on these entries, making `scanSessionMeta` skip them entirely.

This led to `/list` showing severely deflated message counts — e.g., a session with 27 user+assistant entries would display as only 2 messages.

## Changes

- Change `Content string` to `Content json.RawMessage` so unmarshaling succeeds regardless of content type
- Add `extractStringContent` helper to safely extract string content for summary generation
- Add regression tests for array content handling

## Verification

- `go test ./agent/claudecode/` passes
- `go build ./...` passes

Closes #931

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -509,6 +509,19 @@ func (a *Agent) DeleteSession(_ context.Context, sessionID string) error {
 	return os.Remove(path)
 }
 
+// extractStringContent attempts to extract a plain string from a json.RawMessage.
+// Returns empty string if the raw message is not a JSON string.
+func extractStringContent(raw json.RawMessage) string {
+	if len(raw) == 0 {
+		return ""
+	}
+	var s string
+	if err := json.Unmarshal(raw, &s); err != nil {
+		return ""
+	}
+	return s
+}
+
 func scanSessionMeta(path string) (string, int) {
 	f, err := os.Open(path)
 	if err != nil {
@@ -526,7 +539,7 @@ func scanSessionMeta(path string) (string, int) {
 		var entry struct {
 			Type    string `json:"type"`
 			Message struct {
-				Content string `json:"content"`
+				Content json.RawMessage `json:"content"`
 			} `json:"message"`
 		}
 		if err := json.Unmarshal(scanner.Bytes(), &entry); err != nil {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. scanSessionMeta 处理 array content，回归测试覆盖。

PR #939, opened 2026-05-11
fix(core): permission response deadlock in multi-workspace mode

happyTonakai:
## Problem

In multi-workspace mode, users hit a complete deadlock when interacting with the agent:

1. Agent enters plan mode → user clicks "Allow" or "Deny" on the permission card → message is received but queued with "📬 消息已收到，将在当前任务完成后处理"
2. Agent asks a question via AskUserQuestion → user clicks any answer or types a reply → same queued message
3. User types `/stop` → responds "没有正在执行的任务"

The session is completely stuck — the permission response will never be processed because the session is blocked waiting for it, and `/stop` can't find the running task to kill it. The only recovery is restarting cc-connect.

This affects **all permission flows** in multi-workspace mode: ExitPlanMode, tool approval, AskUserQuestion, etc.

## Root Cause

The deadlock is caused by an `interactiveKey` resolution inconsistency between `handleMessage` and the permission handler.

In multi-workspace mode, `interactiveKey` is computed as `resolvedWorkspace + ":" + sessionKey`. But the computation happens at **two independent points** with **different lookup logic**:

- `handleMessage` resolves the workspace via `resolveWorkspace()`, which has a convention-match fallback (e.g. binding `proj` matches session `proj:user1`)
- `handlePendingPermission` computes its own key via `interactiveKeyForSessionKey()`, which does a strict binding lookup without the fallback

When these two resolve to different keys, the permission handler looks up `interactiveStates[differentKey]` — finds nothing — returns `false`. The engine then calls `queueMessageForBusySession()`, which queues the response for a session that's blocked waiting for exactly this response. Classic deadlock.

The event loop holds the session lock (`TryLock`) while blocking on `<-pending.Resolved`. Permission responses bypass the lock check (they happen before `TryLock`), but since the key mismatch prevents them from reaching the pending request, the lock is never released.

## Fix

1. **Single key computation** — move workspace resolution and `interactiveKey` computation to the top of `handleMessage`, before any permission or command checks. Pass the pre-computed key to `handlePendingPermission` and `handlePendingProviderAdd` so they use the same key as the event loop.

2. **Suffix scan fallback** in `cmdStop` and `cmdCompress` — if `interactiveKeyForSessionKey()` resolves a different key than the one actually storing the session state (due to the same binding lookup inconsistency), fall back to `findInteractiveKeyForSession()` which scans all keys by suffix match.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes (no new failures)
- [x] Manual: ExitPlanMode allow/deny resolves immediately in multi-workspace mode
- [x] Manual: AskUserQuestion answers resolve immediately
- [x] Manual: `/stop` correctly finds and stops running sessions
- [x] Manual: Single-workspace mode unaffected

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1991,6 +1991,16 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 		}
 	}
 
+	// Select session manager and agent based on workspace mode
+	sessions := e.sessions
+	agent := e.agent
+	interactiveKey := msg.SessionKey
+	if e.multiWorkspace && wsSessions != nil {
+		sessions = wsSessions
+		agent = wsAgent
+		interactiveKey = resolvedWorkspace + ":" + msg.SessionKey
+	}
+
 	if len(msg.Images) == 0 && strings.HasPrefix(content, "/") {
 		if e.handleCommand(p, msg, content) {
 			return
@@ -1998,8 +2008,9 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 		// Unrecognized slash command — fall through to agent as normal message
 	}
 
-	// Permission responses bypass the session lock
-	if e.handlePendingPermission(p, msg, content) {
+	// Permission responses bypass the session lock.
+	// Must be after workspace resolution so interactiveKey is correct.
+	if e.handlePendingPermission(p, msg, content, interactiveKey) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. 多 workspace 模式下 permission response 死锁修复，提前选择 sessions/agent 和 interactiveKey，逻辑正确。

PR #940, opened 2026-05-11
fix(weibo): serialize writeWS with wsMu to honor gorilla/websocket single-writer contract

kevinWangSheng:
## Summary

`platform/weibo`'s `writeWS` acquired `wsMu` only long enough to read `p.ws`, released the mutex, and then called `ws.WriteJSON` unlocked. Every outbound message goes through `writeWS` (`sendMessage`, `SendImage`, `SendFile`), and the engine routinely fires these concurrently with the heartbeat `pingLoop` (which already correctly held the mutex through its own `WriteJSON`). gorilla/websocket explicitly requires single-writer access and panics on overlap:

```
panic: concurrent write to websocket connection
github.com/gorilla/websocket.(*messageWriter).flushFrame ...
github.com/gorilla/websocket.(*Conn).beginMessage ...
github.com/gorilla/websocket.(*Conn).NextWriter ...
github.com/chenhg5/cc-connect/platform/weibo.(*Platform).writeWS
```

## Fix

Hold `wsMu` for the full `WriteJSON` call in `writeWS`, matching the pattern `pingLoop` already uses. Three-line change (lock acquisition style + `defer`).

## Test

Added `TestWriteWS_ConcurrentSendsSerialized` in `platform/weibo/weibo_test.go`. It spins up an httptest websocket echo server, dials it, fires **50 concurrent** `sendMessage` calls, and asserts the server received 50 distinct, well-formed JSON frames.

Verified deterministically:

- **With the fix** (`go test -race -run TestWriteWS_ConcurrentSendsSerialized ./platform/weibo/`): PASS in ~0.0s.
- **After reverting only the production code**, leaving the test in place: the same `-race` run reliably panics inside `gorilla/websocket.(*Conn).beginMessage` — the exact stack the bug produces in production.

Fix restored before commit.

## Test plan

- [x] `go test -race -tags no_web ./platform/weibo/` — full suite passes
- [x] New regression test deterministically catches the bug when production code is reverted

### DIFF
--- a/platform/weibo/weibo.go
+++ b/platform/weibo/weibo.go
@@ -621,13 +621,16 @@ func (p *Platform) SendFile(_ context.Context, rctx any, file core.FileAttachmen
 }
 
 func (p *Platform) writeWS(data any) error {
+	// gorilla/websocket only allows one concurrent writer; wsMu must guard the
+	// full WriteJSON call (pingLoop already follows this pattern), otherwise
+	// concurrent sendMessage / SendImage / SendFile calls interleave frames
+	// on the wire.
 	p.wsMu.Lock()
-	ws := p.ws
-	p.wsMu.Unlock()
-	if ws == nil {
+	defer p.wsMu.Unlock()
+	if p.ws == nil {
 		return fmt.Errorf("weibo: not connected")
 	}
-	if err := ws.WriteJSON(data); err != nil {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. gorilla/websocket 单写入者限制，wsMu 保护 WriteJSON 调用，并发测试覆盖。

PR #942, opened 2026-05-11
fix(api): /send rejects unknown project name instead of silent fallback

kevinWangSheng:
## Summary

`APIServer.handleSend` resolved the target engine in two steps:

```go
engine, ok := s.engines[req.Project]
if !ok {
    // If only one engine, use it by default
    if len(s.engines) == 1 { ... pick the lone engine ... }
}
```

The second step fired whenever the first lookup missed — including when the caller passed a **typo'd** project name. With a single engine loaded, `project="projectB"` against `{"projectA": ...}` returned `200 OK` and delivered the message to `projectA`, the exact opposite of "project not found".

This is the only resolver in the codebase with that loose semantics; `webhook.go`'s `resolveEngine` and `api.go`'s own `handleCronAdd` both fall back **only** when the project name is empty.

## Fix

Rewrite resolution to gate the single-engine fallback on `req.Project == ""`:

| `Project` field | engines loaded | behavior |
|---|---|---|
| `""` | 1 | auto-pick the lone engine (convenience preserved) |
| `""` | >1 | `400 project is required (multiple projects configured)` |
| `"name"` | matches | use it |
| `"name"` | does not match | `404 project %q not found` |

## Tests

`core/api_test.go` adds three tests beside the existing `TestHandleSend_AllowsAttachmentOnly`:

1. `TestHandleSend_UnknownProjectReturns404` — typo'd project name → `404` (formerly `200` with silent misrouting).
2. `TestHandleSend_EmptyProjectFallsBackToSingleEngine` — pins the preserved convenience path.
3. `TestHandleSend_EmptyProjectMultipleEnginesRequiresName` — multi-engine, empty project → `400` (formerly `404 project \"\" not found`).

Verified the new tests fail against the pre-fix production code with the diff reverted in place; restored before commit.

## Test plan

- [x] `go test -race -tags no_web -run 'TestHandleSend|TestApi|TestAPI' ./core/` green
- [x] Reverted fix locally, confirmed `TestHandleSend_UnknownProjectReturns404` and `TestHandleSend_EmptyProjectMultipleEnginesRequiresName` fail with the buggy behavior, then restored the fix.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -147,22 +147,27 @@ func (s *APIServer) handleSend(w http.ResponseWriter, r *http.Request) {
 	}
 
 	s.mu.RLock()
-	engine, ok := s.engines[req.Project]
+	var engine *Engine
+	var ok bool
+	if req.Project != "" {
+		engine, ok = s.engines[req.Project]
+	} else if len(s.engines) == 1 {
+		// No project specified and only one engine: use it by default.
+		// Do NOT silently fall back when a non-empty project name is unknown —
+		// that misroutes the message to the wrong engine. Mirrors the resolve
+		// pattern in webhook.go and handleCronAdd.
+		for _, e := range s.engines {
+			engine = e
+			ok = true
+		}
+	}
 	s.mu.RUnlock()
 
 	if !ok {
-		// If only one engine, use it by default
-		s.mu.RLock()
-		if len(s.engines) == 1 {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. /send API 在未知 project name 时返回错误而非 silent fallback，逻辑正确。

PR #944, opened 2026-05-11
Pass Feishu quoted images to agents

realraelrr:
## Summary

Passes images from Feishu quoted/replied messages through to agents.

When a user replies to or quotes a Feishu image message and asks the bot a text or rich-text question, cc-connect now fetches the quoted image, keeps the quoted text context, and includes the image in `core.Message.Images`.

## Problem

Feishu already exposed reply context through `parent_id`, and cc-connect already fetched quoted message text. However, quoted attachments were flattened into textual markers such as `[image]`.

That made these cases inconsistent:

- Direct rich-text image message: image reached the agent.
- Text question quoting an image: only quoted text reached the agent.
- Rich-text question quoting an image: only quoted text reached the agent.

## Changes

- Replaces string-only quoted context with a `quotedMessage` value carrying text and images.
- Fetches quoted `image` messages by parent message id and image key.
- Reuses rich-text post parsing so quoted `post` messages can contribute text and embedded images.
- Adds quoted images to `core.Message.Images` for text replies.
- Merges quoted images with current-message images for rich-text replies.
- Adds regression coverage for text and rich-text replies to quoted image messages.

## Verification

Automated:

```bash
go test -count=1 ./platform/feishu
# also checked targeted quote/image parser tests and git diff whitespace
```

Manual:

- Verified in a Feishu group that quoting an image and asking the bot about it succeeds.
- Verified direct rich-text image handling still works.
- Verified outbound Feishu image sending still works.

## Compatibility

- Existing text-only quote behavior is preserved through `ExtraContent`.
- If quoted image download fails, the user message still proceeds with textual quoted context.
- The change is scoped to Feishu quoted/replied message handling.

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -1049,9 +1049,9 @@ func (p *Platform) dispatchMessage(ctx context.Context, msgType, content string,
 	// Skip quote injection when thread_isolation is enabled and the message is
 	// inside a thread — the thread already provides conversational context, and
 	// long quoted prefixes can drown out the user's actual text (issue #764).
-	quotedPrefix := ""
+	var quoted quotedMessage
 	if parentID != "" && !(p.threadIsolation && isThreadSessionKey(sessionKey)) {
-		quotedPrefix = p.fetchQuotedMessage(ctx, parentID)
+		quoted = p.fetchQuotedMessage(ctx, parentID)
 	}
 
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. 飞书引用消息中的图片传递给 agent，逻辑正确。

PR #945, opened 2026-05-11
fix(agent): capture workDir under mutex in qoder + cursor StartSession

kevinWangSheng:
## Summary

`StartSession` in both `agent/qoder` and `agent/cursor` reads `a.workDir` **outside** the agent's own mutex, while `SetWorkDir` (and `GetWorkDir`) consistently take the lock. Since the `/dir` command lets users hot-swap the workspace at runtime, these two paths can race in production. `go test -race` confirms the race deterministically.

### qoder (`agent/qoder/qoder.go`)

```go
func (a *Agent) StartSession(ctx context.Context, sessionID string) (core.AgentSession, error) {
    a.mu.Lock()
    mode := a.mode
    model := a.model
    extraEnv := append([]string{}, a.sessionEnv...)
    a.mu.Unlock()

    return newQoderSession(ctx, a.workDir /* unlocked read */, model, mode, sessionID, extraEnv)
}
```

### cursor (`agent/cursor/cursor.go`)

Same shape — every other field is captured under `a.mu.Lock()`, but `a.workDir` is read in the call to `newCursorSession` after the unlock.

Sibling agents already do this correctly: `gemini.go`, `opencode.go`, `kimi.go`, and `acp/agent.go` all capture `workDir := a.workDir` before unlocking. PR #157 (`fix/pi-thinking-data-race`) applied the same pattern to `agent/pi` for the `thinking` field.

## Fix

One added line per agent: bind `workDir := a.workDir` inside the existing critical section, then thread the local through to the session constructor.

## Tests

Adds `TestAgent_StartSessionWorkDirRace` in both `agent/qoder/qoder_test.go` and a new `agent/cursor/cursor_concurrent_test.go`. Each test launches 50 pairs of goroutines: one running `SetWorkDir`, the other running `StartSession`.

`newQoderSession` and `newCursorSession` only initialise their session structs (the CLI is not spawned until `Send`), so the tests don't need `qodercli` or the Cursor agent binary on `PATH`.

Verified deterministically:

- **With the fix** (`go test -race -run TestAgent_StartSessionWorkDirRace ./agent/qoder/ ./agent/cursor/`): both pass.
- **Reverting only the production code** (kept the tests in place) and re-running `-race`: each agent fails with `testing.go:1617: race detected during execution of test`.

Fix restored before commit.

## Test plan

- [x] `go test -race -tags no_web ./agent/qoder/ ./agent/cursor/` — full suites pass
- [x] Race regression: reverted, re-ran, observed `race detected`, restored.

### DIFF
--- a/agent/cursor/cursor.go
+++ b/agent/cursor/cursor.go
@@ -193,6 +193,7 @@ func (a *Agent) StartSession(ctx context.Context, sessionID string) (core.AgentS
 	model := a.model
 	mode := a.mode
 	cmd := a.cmd
+	workDir := a.workDir
 	extraEnv := a.providerEnvLocked()
 	extraEnv = append(extraEnv, a.sessionEnv...)
 	if a.activeIdx >= 0 && a.activeIdx < len(a.providers) {
@@ -202,7 +203,7 @@ func (a *Agent) StartSession(ctx context.Context, sessionID string) (core.AgentS
 	}
 	a.mu.Unlock()
 
-	return newCursorSession(ctx, cmd, a.workDir, model, mode, sessionID, extraEnv)
+	return newCursorSession(ctx, cmd, workDir, model, mode, sessionID, extraEnv)
 }
 
 // ListSessions reads sessions from ~/.cursor/chats/<workspace_hash>/.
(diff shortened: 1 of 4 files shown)


chenhg5 (approved):
LGTM. data race fix，workDir 在 mutex 内捕获，测试覆盖。

PR #946, opened 2026-05-12
Fix(dingtalk): accept 'picture' msgtype for image messages

sj136270426-ai:
DingTalk enterprise bots send `msgtype: "picture"` (not `"image"`) when users upload photos. This change accepts both `"image"` and `"picture"` to ensure image messages are processed correctly.


chenhg5 (approved):
LGTM. 钉钉 picture msgtype 支持 + file message handler，逻辑正确。

PR #949, opened 2026-05-12
fix(cron): validate UpdateJob enabled value type before removing schedule

kevinWangSheng:
## Summary

`CronScheduler.UpdateJob` pre-validates the value type for `cron_expr`, `mode`, and `session_mode` before touching anything, but the `enabled` field is missed — even though `enabled` is also in the `needsReschedule` set. Today the flow on a bad type is:

1. `needsReschedule := field == \"cron_expr\" || field == \"enabled\"` → `true`
2. Remove the cron entry from the in-memory scheduler.
3. Call `store.Update(id, \"enabled\", value)`.
4. `updateJobField` switch on `case \"enabled\"` requires `value.(bool)`; a JSON string `\"true\"` falls through, the reflection fallback can't set a bool field via a string, and `store.Update` returns `false`.
5. `UpdateJob` returns an error.

Net effect: the cron entry is gone but the job is still marked `Enabled: true` in the store, so it never fires again until the daemon restarts (or the user toggles `disable` then `enable` to round-trip through `EnableJob`/`DisableJob`).

The CLI (`cmd/cc-connect/cron.go`) coerces to bool before sending, but the HTTP API in `core/api.go` and the management PATCH endpoint in `core/management.go` both accept arbitrary JSON, so a misbehaving client or test can trigger this.

## Fix

Mirror the existing up-front type validation for `cron_expr` and reject non-bool `enabled` values **before** removing the cron entry. Minimal change, no behavior change for callers that already send the correct type.

## Test plan

- [x] New regression test `TestCronScheduler_UpdateJob_EnabledNonBoolPreservesSchedule` asserts the schedule entry is still present after a rejected update and the stored `Enabled` flag is unchanged. The test fails on `main` and passes with the fix.
- [x] `go test -race -tags no_web -run TestCron ./core/...` — all green.
- [x] `go vet -tags no_web ./core/...` clean.
- [x] `go build -tags no_web ./...` clean.

### DIFF
--- a/core/cron.go
+++ b/core/cron.go
@@ -554,6 +554,17 @@ func (cs *CronScheduler) UpdateJob(id string, field string, value any) error {
 		}
 	}
 
+	// Validate enabled type up-front. Without this, a non-bool value (e.g. a
+	// JSON string "true" from a misbehaving API client) reaches updateJobField
+	// only after we've already removed the cron entry below, and store.Update
+	// then fails on the type mismatch — leaving the job marked Enabled in the
+	// store but never firing again until the daemon restarts.
+	if field == "enabled" {
+		if _, ok := value.(bool); !ok {
+			return fmt.Errorf("enabled must be a boolean")
+		}
+	}
+
 	// Check if reschedule is needed
 	needsReschedule := field == "cron_expr" || field == "enabled"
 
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. 验证 enabled 类型避免 cron job 意外停火，测试覆盖充分。

PR #950, opened 2026-05-12
Handle WeCom display mentions before commands

realraelrr:
## Summary

Fixes WeCom group commands when the user types the display mention before the command, such as `@bot /whoami` or `＠bot !pwd`.

## Problem

In some WeCom group messages, the webhook payload does not include the bot id form that `stripWeComAtMentions` can remove. The visible text may still begin with the display mention, so command detection sees `@bot /whoami` instead of `/whoami` and does not route it as a command.

## Changes

- Strips a leading display-style mention only when the remaining text begins with `/` or `!`.
- Preserves normal text such as `@bot hello` to avoid over-stripping user content.
- Adds regression tests for slash commands, bang commands, and normal text preservation.

## Verification

```bash
go test -count=1 ./platform/wecom
```

## Compatibility

- Scoped to WeCom mention stripping.
- Does not alter regular mention removal when the bot id is present.
- Does not strip display mentions from non-command text.

### DIFF
--- a/platform/wecom/mention_strip.go
+++ b/platform/wecom/mention_strip.go
@@ -17,7 +17,7 @@ func stripWeComAtMentions(s string, botIDs ...string) string {
 	for strings.Contains(s, "  ") {
 		s = strings.ReplaceAll(s, "  ", " ")
 	}
-	return strings.TrimSpace(s)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. WeCom display mention 前置处理，/command 前的 @mention 正确剥离。

PR #951, opened 2026-05-12
Use CDN HTTP client for Weixin media downloads

realraelrr:
## Summary

Uses the configured CDN HTTP client when downloading inbound Weixin media.

## Problem

The Weixin platform has separate clients for API calls and CDN media downloads, but inbound media collection used the API HTTP client. In deployments where the CDN client carries different proxy, TLS, or transport settings, image downloads can fail even though the CDN client is configured correctly.

## Changes

- Uses `cdnHttpClient` for inbound media downloads.
- Falls back to `httpClient` when `cdnHttpClient` is nil.
- Adds a regression test proving image downloads use the CDN client rather than the API client.

## Verification

```bash
go test -count=1 ./platform/weixin
```

## Compatibility

- Existing behavior is preserved when no CDN HTTP client is configured.
- The change is scoped to Weixin inbound media download.

### DIFF
--- a/platform/weixin/media_inbound.go
+++ b/platform/weixin/media_inbound.go
@@ -40,7 +40,10 @@ func (p *Platform) collectInboundMedia(ctx context.Context, items []messageItem)
 	}
 	dlCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
 	defer cancel()
-	client := p.httpClient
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. CDN HTTP client for Weixin media downloads，测试覆盖。

PR #953, opened 2026-05-12
fix(claudecode): handle array content in scanSessionMeta

### LINKED ISSUE #931: [Bug] claudecode scanSessionMeta skips messages with array content, causing /list to show incorrect message counts

chenhg5:
## Summary
- Change `Content` field from `string` to `json.RawMessage` in scanSessionMeta
- Add `extractContentString` to handle both string and array content formats
- Claude Code now uses array content for messages with thinking blocks or tool results

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes

Fixes #931

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -526,7 +526,7 @@ func scanSessionMeta(path string) (string, int) {
 		var entry struct {
 			Type    string `json:"type"`
 			Message struct {
-				Content string `json:"content"`
+				Content json.RawMessage `json:"content"`
 			} `json:"message"`
 		}
 		if err := json.Unmarshal(scanner.Bytes(), &entry); err != nil {
@@ -534,8 +534,8 @@ func scanSessionMeta(path string) (string, int) {
 		}
 		if entry.Type == "user" || entry.Type == "assistant" {
 			count++
-			if entry.Type == "user" && entry.Message.Content != "" {
-				summary = entry.Message.Content
+			if entry.Type == "user" && len(entry.Message.Content) > 0 {
+				summary = extractContentString(entry.Message.Content)
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


chenhg5:
先关闭这个旧 PR。\n\n目标 issue #931 已经关闭；这个 PR 当前也和 main 冲突。ClaudeCode scanSessionMeta array content 如果 current main 仍有回归，应基于 latest main 重开一个干净小 PR，而不是继续这个旧冲突分支。\n\n---\n\nClosing this old PR for now.\n\nThe target issue #931 is already closed, and this PR now conflicts with main. If ClaudeCode scanSessionMeta array-content handling still regresses on current main, please open a fresh minimal PR based on latest main instead of continuing this old conflicting branch.

PR #702, opened 2026-04-20
feat(dingtalk): add proactive messaging and reply-quote support

sinberCS:
### Proactive Messaging
- Implement `ReplyContextReconstructor` interface for proactive message sending via DingTalk REST API
- Support both group messages (`/v1.0/robot/groupMessages/send`) and direct messages (`/v1.0/robot/oToMessages/batchSend`)
- Enable CLI send, Cron, Webhook, and restart notification scenarios

### Reply/Quote Message Support
- Switch from `RegisterChatBotCallbackRouter` to `RegisterRouter` to access raw callback JSON
- Parse `text.isReplyMsg` and `text.repliedMsg` from raw JSON to recover quoted message content that the SDK's `BotCallbackDataTextModel` silently drops
- Prepend quoted text to user messages in format: `引用: "original text"\n\nuser message`

### Technical Details
- The DingTalk Stream SDK's `BotCallbackDataModel.Content` field is `nil` for text messages, so we bypass the SDK's typed deserialization and parse the raw JSON ourselves
- Added `richTextContent`, `repliedMessage`, `repliedTextContent` types for structured parsing
- Added `onRawMessage` as the new entry point, which parses raw JSON then delegates to `onMessage`
- Added `formatReplyContent` to replace the old `parseTextContent` with strongly-typed struct access

PR #955, opened 2026-05-12
fix(core): extract channel/user IDs correctly for 3-segment type-tagged session keys

kevinWangSheng:
## Summary

\`extractChannelID\` and \`extractUserID\` in \`core/engine.go\` only handled the 4-segment session key form \`platform:t:channelID:userID\` (where \`t\` is a single-char type tag). Three platforms emit a 3-segment form when a single session is shared across a whole group/channel:

- \`dingtalk:g:cid\` — DingTalk with \`share_session_in_channel = true\` (added by #702)
- \`qq:g:groupID\` — QQ group session (\`platform/qq/qq.go:207\`)
- \`qqbot:g:groupOpenID\` — QQ Bot group session (\`platform/qqbot/qqbot.go:988\`)

For these keys:

- \`extractChannelID\` returned the tag \`\"g\"\` / \`\"d\"\` instead of the conversation ID.
- \`extractUserID\` returned the conversation ID as if it were a user ID.

Callers that take \`extractChannelID(sessionKey)\` directly (rather than going through \`effectiveChannelID(msg)\` which prefers \`msg.ChannelKey\`) silently used the wrong key:

- \`core/engine.go:1123\` — multi-workspace resolution (\`extractChannelID\` → \`resolveWorkspace\`)
- \`core/observer.go:98\` — observer channel id from \`observeSessionKey\`
- \`core/engine.go:13015\` — \`buildSenderPrompt\` fallback when \`channelKey\` is empty
- \`core/engine.go:9254/9280/9296/9311\` — \`extractUserID\` for status/cron card ownership

## Fix

Treat any 3+ segment key whose \`parts[1]\` is a single character as a type-tagged form and use \`parts[2]\` as the channel. The 3-segment form has no per-user ID, so \`extractUserID\` returns \`\"\"\`. Existing 4-segment behavior is preserved.

## Test plan

- [x] \`go test ./core/ -run \"TestExtractChannelID|TestExtractSessionKeyParts\"\` — new rows for \`dingtalk:g:cidZZZ\`, \`qq:g:12345\`, \`qqbot:g:openid_abc\` pass.
- [x] \`go test ./core/ -run \"TestMultiWorkspace|TestObserve|TestWorkspaceBinding|TestExtract|TestEffective|TestBuildSenderPrompt\"\` — all related suites pass.
- [x] \`go build ./...\` (excluding the \`web/embed.go\` \`all:dist\` pattern which requires the JS bundle).
- [x] Pre-existing \`TestProcessInteractiveEvents_*Footer*\` and \`TestResolveLocalDirPath_AcceptsSubdir\` failures verified to also fail on \`main\` (macOS HOME-tilde / \`/var\`↔\`/private/var\` symlink env issues, unrelated to this change).

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -13024,11 +13024,11 @@ func (e *Engine) buildSenderPrompt(content, userID, userName, platform, sessionK
 func extractChannelID(sessionKey string) string {
 	// Format: "platform:channelID:userID" or "platform:channelID"
 	// Some platforms encode a short type tag as an extra segment, e.g.
-	// "platform:t:channelID:userID" where t is a single-char tag.
-	// When 4+ segments exist and parts[1] is a single char, treat parts[2]
-	// as the real channel ID.
+	// "platform:t:channelID:userID" or "platform:t:channelID" (shared session)
+	// where t is a single-char tag. When parts[1] is a single char and at
+	// least one more segment follows, treat parts[2] as the real channel ID.
 	parts := strings.SplitN(sessionKey, ":", 4)
-	if len(parts) >= 4 && len(parts[1]) == 1 {
+	if len(parts) >= 3 && len(parts[1]) == 1 {
 		return parts[2]
 	}
 	if len(parts) >= 2 {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. 3-segment type-tagged session keys 正确提取 channel/user IDs，测试覆盖。

PR #573, opened 2026-04-11
fix(core): use accumulated text when tool messages hidden

chenhg5:
## Summary
- When `tool_messages=false`, `EventResult.Content` only contains the last assistant turn text (e.g. 198 chars), not the full accumulated response (e.g. 822 chars across multiple turns)
- The engine now compares accumulated `textParts` against `event.Content` and uses whichever is longer
- This ensures the final card/message and history contain the complete multi-turn response

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] New regression test: `TestProcessInteractiveEvents_HiddenToolMessages_AccumulatesAllText`
- [ ] Manual: set `tool_messages=false`, send a message that triggers multiple tool calls, verify final response contains all text segments

Closes #549

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #508, opened 2026-04-08
fix(feishu): extract link, hr, and code_block in quoted post messages

chenhg5:
## Summary
- Extract `"a"` (link) tag text in `extractPostPlainText` so hyperlinks in quoted posts are visible to the agent
- Add `"hr"` tag support, rendered as markdown `---`
- Improve `code_block` formatting: skip empty language prefix when language field is unset

Based on work from PR #472 by @Cigarrr — the main feature (reply-quote context) was already merged into main via earlier PRs; this captures the remaining improvements from #472's second commit.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] Added tests: `TestExtractPostPlainText_LinkTag`, `TestExtractPostPlainText_CodeBlockNoLang`, `TestExtractPostPlainText_HrTag`

Related: #472, #470

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #957, opened 2026-05-12
feat(core): add /cancel command to interrupt and reset session

### LINKED ISSUE #938: feat: support /cancel command to interrupt running session
When Claude Code is executing a long-running command (e.g., mvn install),   
the session is blocked until the command completes. All subsequent messages   
from the chat platform are queued and cannot interrupt.

Proposed solution: Add a /cancel or /stop slash command that:  
1. Kills the current AI agent process (SIGTERM)  
2. Starts a fresh session  
3. Confirms "Session cancelled, ready for new instructions"

Claude Code's own message queue already has a 3-level priority system   
(now, next, later), but cc-connect submits all messages at the same   
priority level and cannot leverage this.

chenhg5:
## Summary
- Add `/cancel` command that stops current execution and creates a fresh session
- Unlike `/stop` which only halts execution, `/cancel` also clears session history
- Add `MsgSessionCancelled` i18n key with 5 language translations
- Add `/cancel` to help card session section

## Use Case
When the agent is blocked on a long-running task (e.g., `mvn install`), `/cancel` lets the user interrupt and start fresh without waiting. This addresses the limitation that cc-connect submits all messages at the same priority level and cannot leverage Claude Code's 3-level priority queue.

## Behavior
- If execution is in progress: stops it (SIGTERM to agent process)
- Creates a new session (like `/new`)
- Clears old session's agent session ID and history
- Returns "Session cancelled. Ready for new instructions."

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes

Fixes #938

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -4616,6 +4616,7 @@ var builtinCommands = []struct {
 	{[]string{"heartbeat", "hb"}, "heartbeat"},
 	{[]string{"compress", "compact"}, "compress"},
 	{[]string{"stop"}, "stop"},
+	{[]string{"cancel"}, "cancel"},
 	{[]string{"help"}, "help"},
 	{[]string{"version"}, "version"},
 	{[]string{"commands", "command", "cmd"}, "commands"},
@@ -4803,6 +4804,8 @@ func (e *Engine) handleCommand(p Platform, msg *Message, raw string) bool {
 		e.cmdCompress(p, msg)
 	case "stop":
 		e.cmdStop(p, msg)
+	case "cancel":
+		e.cmdCancel(p, msg)
 	case "help":
 		e.cmdHelp(p, msg)
 	case "start":
@@ -7184,6 +7187,7 @@ func helpCardGroups() []helpCardGroup {
 			titleKey: MsgHelpSessionSection,
 			items: []helpCardItem{
 				{command: "/new", action: "act:/new"},
+				{command: "/cancel", action: "cmd:/cancel"},
 				{command: "/list", action: "nav:/list"},
 				{command: "/current", action: "nav:/current"},
 				{command: "/switch", action: "nav:/list"},
@@ -7936,6 +7940,37 @@ func (e *Engine) cmdStop(p Platform, msg *Message) {
 		return
 	}
 	e.reply(p, msg.ReplyCtx, e.i18n.T(MsgExecutionStopped))
+}
+
+// cmdCancel stops the current execution and starts a fresh session.
+// Unlike /stop which only halts execution, /cancel also resets the session
+// so the user can immediately continue with new instructions.
+func (e *Engine) cmdCancel(p Platform, msg *Message) {
+	_, sessions, interactiveKey, err := e.commandContext(p, msg)
+	if err != nil {
+		e.reply(p, msg.ReplyCtx, e.i18n.Tf(MsgWsResolutionError, err))
+		return
+	}
+
+	slog.Info("cmdCancel: stopping execution and creating new session", "session_key", msg.SessionKey)
+
+	// Stop the current execution (like /stop)
+	stopped := e.stopInteractiveSession(interactiveKey, p, msg.ReplyCtx)
+	if !stopped {
+		// No execution in progress, but still create a new session
+		slog.Debug("cmdCancel: no execution to stop, proceeding with new session", "session_key", msg.SessionKey)
+	}
+
+	// Clear old session's agent session ID so it cannot be resumed
+	old := sessions.GetOrCreateActive(msg.SessionKey)
+	old.SetAgentSessionID("", "")
+	old.ClearHistory()
+	sessions.Save()
+
+	// Create a new session (like /new)
+	sessions.NewSession(msg.SessionKey, "")
+
+	e.reply(p, msg.ReplyCtx, e.i18n.T(MsgSessionCancelled))
 }
 
 func (e *Engine) stopInteractiveSession(sessionKey string, quietPlatform Platform, quietReplyCtx any) bool {
--- a/core/i18n.go
+++ b/core/i18n.go
@@ -129,6 +129,7 @@ const (
 	MsgToolResultFmtOk           MsgKey = "tool_result_fmt_ok"
 	MsgToolResultFmtFailed       MsgKey = "tool_result_fmt_failed"
 	MsgExecutionStopped          MsgKey = "execution_stopped"
+	MsgSessionCancelled          MsgKey = "session_cancelled"
 	MsgNoExecution               MsgKey = "no_execution"
 	MsgPreviousProcessing        MsgKey = "previous_processing"
 	MsgQueueFull                 MsgKey = "queue_full"
@@ -643,6 +644,13 @@ var messages = map[MsgKey]map[Language]string{
 		LangTraditionalChinese: "⏹ 執行已停止。",
 		LangJapanese:           "⏹ 実行を停止しました。",
 		LangSpanish:            "⏹ Ejecución detenida.",
+	},
+	MsgSessionCancelled: {
+		LangEnglish:            "Session cancelled. Ready for new instructions.",
+		LangChinese:            "会话已取消。可以继续新的对话。",
+		LangTraditionalChinese: "會話已取消。可以繼續新的對話。",
+		LangJapanese:           "セッションをキャンセルしました。新しい指示を受け付けます。",
+		LangSpanish:            "Sesion cancelada. Listo para nuevas instrucciones.",
 	},
 	MsgNoExecution: {
 		LangEnglish:            "No execution in progress.",


chenhg5:
LGTM. Clean /cancel implementation with proper session reset and i18n coverage.

chenhg5 (commented):
QA review (comment — same account as PR author, so using `--comment` instead of `--approve`):

> **Note:** My GitHub CLI is authenticated as `@chenhg5`, the PR author. GitHub blocks `--approve` on a PR you opened yourself, so I'm filing this as a comment. The role-doc conclusion for this PR (no blockers found) would normally be `Approve`.

Conclusion: Comment (functionally approve, see note above)

Overall assessment:
- This PR adds the `/cancel` command as a hybrid of `/stop` (halt current execution) and `/new` (create a fresh session). The implementation is small, well-scoped, and i18n-complete. CI is green; `go vet` is clean. No blockers found. I'd recommend this lands — the command is a real user-facing need (linked P1 #938) and the implementation matches the engine's existing patterns.

Review scope:
- Reviewed `core/engine.go` and `core/i18n.go`.
- Cross-referenced with `cmdNew` (`core/engine.go:5463`), `cmdStop` (`core/engine.go:8592`), `stopInteractiveSession` (`core/engine.go:8616`), `cleanupInteractiveState` (`core/engine.go:3074`), and the `Session` API (`core/session.go:155-206`).
- Ran `go test -count=1 ./core/ -run "TestCmd|TestHandle"` on the PR branch — all pass (2.082s).

What looks good:
- The 3-step flow — stop execution → clear old session ID + history → create new session — is exactly what users mean by "cancel". The order is right: stop first so no new writes hit the old session, then clear, then create.
- Falls back gracefully when there's no execution in progress (logs `cmdCancel: no execution to stop, proceeding with new session` and still creates a new session). This is the right behavior: `/cancel` is essentially `/new` when idle.
- `old.SetAgentSessionID("", "")` correctly routes through the `PastAgentSessionIDs` recorder (see `core/session.go:170-172`), so the old session ID is preserved for `filterOwnedSessions` and `/list` won't list it as a fresh orphan.
- i18n is complete: `MsgSessionCancelled` translated in all 5 supported languages (en, zh, zh-TW, ja, es). The translations look natural, not auto-translated-looking.
- Help card entry uses `cmd:/cancel` prefix, consistent with the majority of help-card commands.
- Branch is **MERGEABLE** on main (the other big-blocker PRs like #581, #573, #508 are CONFLICTING; this one is the clean one of the bunch).

1. **No tests** for the new `cmdCancel` function. The engine has `TestCmdStop_*` (3 tests in `core/engine_test.go:6883,9143,9427`) and an implicit `/new` test surface, but no `TestCmdCancel_*`. Suggest at minimum: (a) one test for "no execution in progress, creates new session"; (b) one test for "execution in progress, stops + creates new session"; (c) one test for the PastAgentSessionIDs recording on the old session. This is a public-facing command — worth the test coverage.
2. **Action prefix inconsistency** with `/stop`. `/stop` uses `act:/stop` (line 7871), `/cancel` uses `cmd:/cancel`. Both have "halt current state" semantics; for consistency, `/cancel` should arguably use `act:`. That said, `cmd:` is defensible (it's primarily a slash command), and the PR's choice doesn't break anything. Mentioning for awareness.
3. **`stopInteractiveSession` uses `closeAgentSessionAsync`**, not the synchronous `closeAgentSessionWithTimeout` that `/new` uses (via `cleanupInteractiveState`). The async close is fine for `/cancel` semantics (we want the new session created immediately, not after a 130s graceful shutdown ceiling), but worth a one-line comment in `cmdCancel` explaining why we chose the async path. Without it, a future reader may "fix" it to use `cleanupInteractiveState` and introduce a 130s blocking reply.
4. **`SetAgentSessionID("", "")` after `stopInteractiveSession`** — the old `state.agentSession` in `interactiveStates` is gone, but the old `Session` in the `SessionManager` still has its `AgentSessionID` from when the old agent started. Clearing it here is correct, but the sequence is: state-map delete (sync) → `closeAgentSessionAsync` (async, can take 130s) → SessionManager mutation (sync, here). If `closeAgentSessionWithTimeout` writes back to the `Session` during teardown (e.g., recording past IDs on its own), there could be a race. The existing code seems to avoid this, but it's worth a sanity check. If no race exists, please add a brief comment; if it does, the fix is to take `sessions.mu` in `closeAgentSessionWithTimeout` consistently.

Optional (P3):
- The reply text for `/cancel` is "Session cancelled. Ready for new instructions." — this is the success path. Consider also replying with a brief explanation that the old session's history was cleared, so users understand why the next message is a clean slate. Cosmetic.
- `/cancel` doesn't take args, but if a user does `/cancel something`, the `something` is silently dropped (the command handler ignores extra args). Consider replying with a usage hint if extra args are present. Trivial.

Questions:
- For users without an active execution, `/cancel` is functionally identical to `/new`. Should `/cancel` reject when there's no execution? (Probably no — being forgiving is better.) Just confirming the intent.
- Should the help card `/cancel` be in the "Session" group (where it is now) or also in the "Action" group alongside `/stop`? Current placement is correct IMO; just checking.

Testing / Risk:
- Verified in this review:
  - `go vet ./core/...` — clean.
  - `go test -count=1 ./core/ -run "TestCmd|TestHandle"` — all pass.
  - `gh pr checks 957` — lint, unit-test, smoke-test, regression-test, performance-test all pass.
- Not covered:
  - No live test of `/cancel` in a real interactive session (no Claude/Codex/OpenCode sandbox in CI). Highest-value manual test: start a long-running tool call in a session, send `/cancel` mid-execution, verify: (a) the tool call is halted, (b) the new session starts fresh, (c) `/list` doesn't show the cancelled session as a "fresh orphan".
  - No concurrent-cancel test (two `/cancel` messages arriving back-to-back).
- Regression risk: low. The change is contained to `core/engine.go` (one new function) and `core/i18n.go` (one new MsgKey + 5 translations). No changes to existing function signatures.

Next step:
- From QA's perspective, this is **ready to merge** after (1) adding at least one unit test, ideally (2) confirming one live P2P interaction manually.
- I lean toward landing this without blocking on tests, since the change is small and the engine already has the relevant `TestCmdStop_*` patterns to copy from. But the P2 above on test coverage is real.

chenhg5 (commented):
## QA Verification (P1 #938 follow-up)

结论: Comment — verification complete. Implementation matches spec for 4/5 points; 1 point is based on a behavior assumption that does not hold in the current design.

> Note: This is a verification follow-up, separate from my earlier review (2026-06-06T13:55:59Z). Same author as PR, so filing as comment.

### Verification matrix

| # | Point | Status | Evidence |
|---|-------|--------|----------|
| 1a | `/cancel` 终止 agent 进程 | **PASS** | `core/engine.go:7948` `cmdCancel` → `stopInteractiveSession` → `closeAgentSessionAsync` → `closeAgentSessionWithTimeout` (3-phase: stdin EOF → SIGTERM → SIGKILL, 130s ceiling). Same chain as `/stop` (`core/engine.go:7871`). |
| 1b | `/list` 不再显示该 session | **DOES NOT MATCH** — by design | `cmdCancel` calls `old.SetAgentSessionID("", "")` (line 7972), which triggers `recordPastAgentSessionID` (`core/session.go:139` → 80-90). The cleared ID is preserved in `PastAgentSessionIDs`, so `KnownAgentSessionIDs` (line 453-465) and `filterOwnedSessions` (line 5172) still recognise the session. Cancelled session **will** appear in `/list` with its last agent ID. This is **intentional and consistent with `/new`** (the same pattern), so users see a "cancelled" session rather than a fresh orphan. If the spec wanted true exclusion, `cmdCancel` would need to call `SetAgentSessionID` differently or the filter would need a "status" flag. |
| 2 | 下一次用户消息触发新 session,session_id 变化 | **PASS** | `core/engine.go:7975` `cmdCancel` calls `e.sessions.NewSession(msg.SessionKey, ...)` which calls `createLocked` → `nextID()` returns `s%d` with monotonic counter (`core/session.go:253-281`). Every call increments the counter, so the new session has a strictly greater ID than the old one. |
| 3 | 后台 tasks 不残留 | **PASS** | Same teardown chain as `/stop` (point 1a): `stopUnsolicitedReader` cancels context + waits done channel with `unsolicitedReaderStopTimeout` fallback (line 3105), `markStopped` closes the singleton `stopCh` (line 3135), `closeAgentSessionAsync` spawns a goroutine for the 3-phase close (line 5190). The `Engine` itself is not reset (intentional — the same `Engine` serves all sessions), so no per-session goroutine leaks. |
| 4 | `go test ./core -run TestCancel / TestSessionCancelled` 全 pass | **FAIL — no tests exist** | `go test -count=1 -timeout 180s -run "TestCancel\|TestSessionCancelled" ./core/...` returns `ok ... [no tests to run]`. `core/engine_test.go` has 3 `TestCmdStop_*` tests and 0 `TestCmdCancel_*`. This is the P2 test-coverage gap from my earlier review, now elevated to a verification finding. |
| 5 | worktree 验证 + 5 个 CI job 全 pass | **PASS** | Worktree at `/tmp/qa-pr957` (branch `pr-957`, HEAD `89d1e7b0`): `go vet ./core/...` clean, `go build -tags no_web ./core/ ./cmd/cc-connect/` clean. CI: lint ✅, unit-test ✅, smoke-test ✅, regression-test ✅, performance-test ✅ (5/5 green, all from 2026-05-12 run on this exact commit). |

### Summary

- **4/5 points PASS** as specified.
- **Point 4** (test coverage) **FAILS** — no `TestCmdCancel_*` or `TestSessionCancelled` tests exist. The P2 from my earlier review is real and should be addressed before this lands. The PR is small enough that a single test ("`/cancel` in progress: stops + creates new session, old session ID preserved in PastAgentSessionIDs") would close it.
- **Point 1b** is not a bug — it's a spec mismatch. The current design treats `/cancel` as "the old session is now closed" (visible in `/list` as past tense) rather than "the old session is gone" (excluded from `/list`). This mirrors `/new`. If users actually want true exclusion, that's a separate UX/feature change, not a `/cancel` bug. **Recommend confirming with PM whether 1b is a requirement or just a description of the expected behavior.**

### Recommendation

- **Ready to merge** from a correctness standpoint.
- **Should add tests** before merge to address verification point 4. The existing `TestCmdStop_*` patterns (`core/engine_test.go:6883,9143,9427`) are direct templates — copy and adapt.
- **Point 1b** is a UX question, not a bug. Suggest a brief confirmation from the spec author.

— cc-connect/qa-claudecode

chenhg5 (commented):
## Code Review: PR #957 (full review, t-20260606-i7gqet)

结论: **Comment — Approve ready for maintainer merge** (functionally approve; same author as PR, so using `--comment` per the standard same-author convention).

This is a full code review (separate from the verification review I posted earlier today at 2026-06-06T17:02:19Z for the 5-point PM verification task).

### Overall assessment

`cmdCancel` is a 4-step composite of well-understood primitives (`stopInteractiveSession` + `SetAgentSessionID("", "")` + `ClearHistory` + `NewSession`), and each primitive is already battle-tested elsewhere in the codebase. The new code is 33 lines, contained to one function, and reuses existing patterns. **No correctness, concurrency, or persistence blockers found.** Test coverage is the only real gap (P2 from prior review, still open).

### Correctness

- **Sequence is safe**: `stopInteractiveSession` is sync up to `closeAgentSessionAsync` (line 8015), then control returns. `SetAgentSessionID("", "")` runs immediately, which internally calls `recordPastAgentSessionID()` so the old ID is preserved in `PastAgentSessionIDs`. `ClearHistory` resets the in-memory `History` slice. `Save` persists the SessionManager state to disk. `NewSession` calls `createLocked` then `saveLocked` (`core/session.go:273-279`) — confirmed persistence-safe per dev-codex's prior check.
- **Fallback when no execution in progress**: correct. When `stopInteractiveSession` returns `false` (no `interactiveStates[interactiveKey]`), the function logs and proceeds. This is the right UX — `/cancel` should be forgiving and behave like `/new` when idle.
- **`SetAgentSessionID("", "")` is idempotent on the ID-empty path**: per `core/session.go:138-140`, if `s.AgentSessionID` is already empty, no `recordPastAgentSessionID` is called. So a `/cancel` on an already-idle session won't bloat `PastAgentSessionIDs`.
- **i18n complete**: `MsgSessionCancelled` translated in all 5 supported languages (`core/i18n.go:648-654`); translations look natural.

### Concurrency / session state

- **`interactiveStates[interactiveKey]` is deleted before the async close**: this is correct. Once the entry is removed, any new event delivery to the old session key short-circuits to "no state" instead of trying to mutate the deleted state. The agent's own Close() runs in the background but writes back via the agent's session handle, not cc-connect's `Session` struct.
- **`closeAgentSessionWithTimeout` does NOT mutate cc-connect's `Session`**: confirmed at `core/engine.go:3023-3054`. It only calls `agentSession.Close()` (the agent runtime's own handle) and waits up to 130s. So the P2 race concern I raised earlier ("could `closeAgentSessionWithTimeout` write back to the old `Session`?") — **no, it cannot**. The two state objects are disjoint.
- **P2 (latent)**: after `/cancel` returns, the old agent process may still be alive for up to 130s. If the user immediately sends a new message, a new agent process starts in parallel. If both write to the same `work_dir`, there is a brief contention window. In practice this is rare (different session, often different workspace) and the old process is exiting, so writes are minimal. **Not blocking, worth a comment in the function to acknowledge the async-close semantics.**

### Multi-workspace safety

- **`commandContext` returns the workspace-bound `*SessionManager`**: confirmed at `core/engine.go:13139-13167`. In multi-workspace mode, `sessions` is the workspace's `SessionManager`, and `interactiveKey` is the workspace-prefixed key. The function uses `sessions` (correct) and `msg.SessionKey` as the user key into that workspace's `SessionManager` (also correct — `commandContext` already factored this in).
- **No cross-workspace bleed**: each workspace has its own `SessionManager` with its own `mu`, `sessions`, `userSessions`, `activeSession`, `counter`, and on-disk store. `cmdCancel` operates entirely on the resolved workspace's manager. The other workspaces' state is untouched.
- **Edge case**: if a user has the same `msg.SessionKey` (e.g., Discord channel) bound to two different workspaces over time (binding change), each workspace sees its own history independently. `/cancel` on workspace A doesn't affect workspace B. Correct.

### Semantic differences vs `/stop` and `/new`

| Behavior                          | `/stop` | `/new` | `/cancel` (this PR) |
|-----------------------------------|---------|--------|---------------------|
| Halts current execution           | ✅      | ❌     | ✅                  |
| Creates new session (new ID)      | ❌      | ✅     | ✅                  |
| Old session's history kept        | ✅      | ✅     | ❌ (cleared)        |
| Old `AgentSessionID` preserved in `PastAgentSessionIDs` | ❌ | ✅ | ✅ |
| Async vs sync close of old agent  | async   | sync via `cleanupInteractiveState` | async (matches `/stop`) |
| Action prefix in help card        | `act:`  | `cmd:` | `cmd:`              |

**Observation**: `/cancel` is the most aggressive of the three — it discards old history. This matches the user's mental model ("cancel = wipe the slate clean"), and the comment in the function explicitly says so. Worth a one-line note in the user-facing help text that says "history of the previous session is cleared" so users know to `/list` for the old session ID if they need to resume it.

### Persistence safety (dev-codex's concern, verified)

- `NewSession` (`core/session.go:272-279`) takes `sm.mu`, calls `createLocked` (allocates ID, populates `sessions`/`activeSession`/`userSessions`), then `saveLocked` (persists to disk). So a new session is never "active in memory but not on disk" — they're atomic under `sm.mu`.
- `GetOrCreateActive` + `SetAgentSessionID` + `ClearHistory` + `Save` sequence: each takes `Session.mu` or `sm.mu` as needed. The `Save` call between `ClearHistory` and `NewSession` is a deliberate "checkpoint old as cleared" before allocating the new one. This is a safe ordering — if the engine crashes between `Save` and `NewSession`, on restart the user sees "session was cancelled, no new session" rather than "session is mid-cancel with inconsistent state". Reasonable.

### Test coverage (P2 from prior review, still open)

- **No `TestCmdCancel_*` tests** in `core/engine_test.go`. Three `TestCmdStop_*` tests (lines 7893, 8177, plus the `TestCmdStop_ReturnsWhileCloseBlockedAndStopsEventLoop`) are direct templates — copy-paste and adapt for `/cancel`:
  - `TestCmdCancel_HasExecution_StopsAndCreatesNewSession` — verifies the 3-step composition.
  - `TestCmdCancel_NoExecution_CreatesNewSession` — fallback path.
  - `TestCmdCancel_PreservesOldAgentIDInPastAgentSessionIDs` — verifies the `SetAgentSessionID("", "")` → `recordPastAgentSessionID` chain.
  - `TestCmdCancel_ClearsOldSessionHistory` — verifies `ClearHistory` actually cleared.
- These are 4 short tests, ~30 lines each, well within scope of this PR. I'd block on adding at least the first one before merge. Not P1 — the change is small and the primitives are tested — but a public-facing command deserves at least one happy-path test.

### Other observations

- **P3 (nit)**: PastAgentSessionIDs grows with each `/cancel` (and each `/new`). Bounded by the number of unique `AgentSessionID` values per session lifecycle, not unbounded, but over a long-lived user with many cancels, the list could grow. Worth a future cleanup pass; not for this PR.
- **P3 (nit)**: `cmd:` action prefix instead of `act:` (matches `/stop`). Trivial inconsistency; the PR's choice is defensible (it's primarily a slash command).
- **P3 (nit)**: no comment explaining why `stopInteractiveSession`'s async close is preferred over `cleanupInteractiveState`'s sync close. Future readers may "fix" it and introduce a 130s blocking reply.

### Testing / Risk

- Verified locally on the worktree at `/tmp/qa-pr957-i7gqet` (branch `pr-957`, HEAD `89d1e7b0`):
  - `go vet ./core/...` — clean.
  - `go build -tags no_web ./core/ ./cmd/cc-connect/` — clean.
  - `go test -count=1 -timeout 180s -run "TestCmd|TestHandle" ./core/...` — all pass (2.271s).
- CI: lint, unit-test, smoke-test, regression-test, performance-test — all 5/5 green (from 2026-05-12 run on the same commit).
- Not covered: live test in a real interactive session (no sandbox in CI); concurrent-cancel test (two `/cancel` messages back-to-back); multi-workspace cancel test.

### Recommendation

- **Ready to merge after at least one `TestCmdCancel_*` test is added** (ideally 2: with-execution and no-execution paths). The PR is otherwise correct, well-scoped, and persistence-safe. The composite is built from well-tested primitives.
- **Maintainer decision** (not QA's call): whether to land before tests, given the small change size, or to require tests as a precondition.

— cc-connect/qa-claudecode

PR #628, opened 2026-04-15
feat(core): add native /interrupt support for Codex and Claude

xukp20:
## Summary

- add `/interrupt` as a new builtin command
- keep `/stop` semantics unchanged
- use native interrupt for Codex app-server and Claude Code
- provide start/done feedback instead of only acknowledging request dispatch

## Changes

- add `SessionInterrupter` optional session capability
- wire `/interrupt` through `core.Engine`
- implement `turn/interrupt` for Codex app-server sessions
- implement `control_request(subtype=interrupt)` for Claude sessions
- wait for native interrupt completion before reporting success
  - Codex app-server: `turn/completed(status=interrupted)`
  - Claude: `control_response(success)` + final `result.terminal_reason`
- add regression tests and command docs

## Notes

- `/interrupt` does not replace `/stop`
- semantics split:
  - `/interrupt`: try to natively interrupt the current in-flight turn and keep the session alive
  - `/stop`: stop the current execution/session with the existing stronger cleanup semantics
- support matrix in this PR:
  - Codex `app_server`: supported
  - Codex `exec`: unsupported
  - Claude Code: supported
  - other CLIs: unchanged / unsupported
- verified with real Feishu single-workspace tests
  - Codex app-server:
    - thinking/streaming interrupt works
    - tool execution interrupt works
  - Claude Code:
    - thinking/streaming interrupt works
    - tool execution interrupt works
- known minor UX issue:
  - the interrupted original request may still end up as `(empty response)`
  - intentionally left out of this PR to avoid broadening scope

## Screenshots

image

image

image

image

PR #961, opened 2026-05-12
feat(core): add latest-wins interrupt queue mode

realraelrr:
## Summary

- Add an opt-in `[queue].mode = "interrupt"` busy-session mode where the latest normal follow-up message supersedes stale output from the active turn and becomes the next processed input.
- Keep default FIFO queue behavior unchanged, including `max_depth`; interrupt mode uses a single latest-wins pending slot.
- Add optional `core.SessionInterrupter` support and implement Codex app-server `turn/interrupt` without killing or recreating backend processes.
- Expand `[instant_reply]` into `initial`, `superseded`, and `queued`, while preserving `content` as a legacy alias for `initial`.

## Design Notes

- This is distinct from `/btw` or steer-style injection: follow-up messages do not get sent into the active turn.
- This is distinct from pre-execution collection/debounce: it applies only after a session is already busy.
- Stale user-visible output is suppressed using turn generation tracking.
- If a backend supports native interruption, the engine calls it opportunistically; otherwise the old turn is allowed to finish and the latest pending input runs next.

## Testing

- `go test -count=1 ./core ./config ./agent/codex`
- `go test -race -count=1 ./core -run 'TestQueueMessageForBusySession_InterruptModeLatestWins|TestProcessInteractiveEvents_InterruptModeSuppressesOldResultAndProcessesLatest|TestProcessInteractiveEvents_InterruptModeSkipsInitialAckWhenSupersededAckWasSent|TestProcessInteractiveEvents_InterruptPermissionRaceSuppressesOldResult|TestProcessInteractiveEvents_InterruptModeChannelClosedSuppressesOldPartial|TestAdoptPendingFromPlaceholderCopiesSupersedeMarker'`
- `go test -race -count=1 ./agent/codex -run 'TestAppServerSession_Interrupt|TestAppServerSession_ResolveInterrupt|TestAppServerSession_TurnCompleted|TestAppServerSession_RejectInterrupt'`
- `git diff --check`

## Notes

- `go test -count=1 -tags no_web ./...` was also run. It exposed two unrelated local/full-suite failures outside this change: a flaky/isolated `agent/codex` runtime-config test that passed when rerun directly, and a daemon launchd status test that fails in this local macOS environment.

Relates to #938 and is compatible with the native interrupt direction from #628.


chenhg5 (approved):
LGTM. latest-wins interrupt queue mode，逻辑清晰，CI pass。

realraelrr:
@chenhg5 This PR has been refreshed on top of the latest upstream main (`5a7dde44`) and is mergeable again. It keeps the interrupt queue behavior focused and tested; could you please take another look and merge it when convenient?

PR #962, opened 2026-05-12
fix(speech): honor context cancellation in ConvertAudioToMP3

kevinWangSheng:
## Summary

\`ConvertAudioToMP3\` in \`core/speech.go\` is the only ffmpeg helper in that file that does not use \`exec.CommandContext\`. All its siblings — \`ConvertAudioToOpus\`, \`ConvertAudioToAMR\`, \`ConvertMP3ToOGG\`, \`ConvertMP3ToAMR\` — already take a \`context.Context\` and use \`exec.CommandContext\` so the caller can abort the subprocess. \`ConvertAudioToMP3\` used \`exec.Command\` and ignored its caller's context entirely.

The caller \`TranscribeAudio\` accepts a \`context.Context\` and forwards it to the STT provider, but the conversion step that runs first was unbounded. If ffmpeg stalls on malformed input or its stdin/stdout pipes get wedged, a cancelled \`TranscribeAudio\` call leaks a running ffmpeg process until ffmpeg's own internal timeout (if any) fires.

## Fix

- Add \`ctx context.Context\` as the first parameter of \`ConvertAudioToMP3\`.
- Switch both \`exec.Command(ffmpegPath, ...)\` call sites to \`exec.CommandContext(ctx, ffmpegPath, ...)\`.
- Thread the caller's \`ctx\` through \`TranscribeAudio → ConvertAudioToMP3\`.

\`ConvertAudioToMP3\` is exported but its only call site in this repo is \`TranscribeAudio\` in the same file (grepped). The API surface change is contained to the conversion helper itself.

## Test plan

- [x] \`TestConvertAudioToMP3_HonorsContextCancellation\` — calls \`ConvertAudioToMP3\` with an already-cancelled context and asserts an error is returned. Skips cleanly when ffmpeg is not in \`PATH\` (the \"ffmpeg not found\" check runs before the context is consulted, so the test cannot otherwise assert plumbing).
- [x] \`go test ./core/ -run TestGeminiSTT\` and other existing tests pass.
- [x] Pre-existing macOS-specific test failures (\`TestProcessInteractiveEvents_*Footer*\`, \`TestResolveLocalDirPath_AcceptsSubdir\`) reproduce on \`main\` without this change — unrelated.

### DIFF
--- a/core/speech.go
+++ b/core/speech.go
@@ -298,7 +298,9 @@ func (g *GeminiSTT) Transcribe(ctx context.Context, audio []byte, format string,
 
 // ConvertAudioToMP3 uses ffmpeg to convert audio from unsupported formats to mp3.
 // Returns the mp3 bytes. If ffmpeg is not installed, returns an error.
-func ConvertAudioToMP3(audio []byte, srcFormat string) ([]byte, error) {
+// The ctx is honored: cancellation kills the ffmpeg subprocess, matching the
+// behavior of the other Convert* helpers in this file.
+func ConvertAudioToMP3(ctx context.Context, audio []byte, srcFormat string) ([]byte, error) {
 	ffmpegPath, err := exec.LookPath("ffmpeg")
 	if err != nil {
 		return nil, fmt.Errorf("ffmpeg not found in PATH: install ffmpeg to enable voice message support")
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. ConvertAudioToMP3 使用 exec.CommandContext 响应 ctx cancellation，测试覆盖。

PR #963, opened 2026-05-12
fix(wecom): split long messages at semantic boundaries

realraelrr:
## Summary

- Replace WeCom hard byte slicing with deterministic backward-window semantic chunking.
- Search from the byte limit backward using explicit fallback windows:
  - paragraph (`\n\n`) in the final 30% window (`70%..100%`)
  - line break (`\n`) in the final 20% window (`80%..100%`)
  - sentence punctuation (`。！？；.!?;`) in the final 15% window (`85%..100%`)
  - comma/space (`，`, `,`, space, tab) in the final 10% window (`90%..100%`)
- Remove enumeration comma (`、`) as a split boundary because splitting inside lists reads poorly.
- Keep both WeCom HTTP and WebSocket send paths using the same `splitByBytes` helper.

## Why

WeCom messages were already split by cc-connect at 2000 bytes. The previous splitter only avoided invalid UTF-8, so long replies could leave awkward partial clauses as separate messages. This version fills each chunk as much as possible, then backs up to the nearest acceptable semantic boundary.

## Testing

- `go test -count=1 ./platform/wecom`
- `go build -tags no_web -o /tmp/cc-connect-wecom-semantic-chunking-pr ./cmd/cc-connect`
- `git diff --check`

## Notes

- Also ran `go test -count=1 ./core ./config ./platform/wecom`; `config` and `platform/wecom` passed, while `core` failed on existing local reply-footer path expectations unrelated to this change.


chenhg5 (approved):
LGTM. WeCom 长消息语义边界分割，测试覆盖，CI pass。

chenhg5:
Hi @realraelrr! This PR has merge conflicts with the latest main branch. Please rebase onto the latest main:

```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```

Once resolved, CI will automatically re-run. Let me know if you need any help!

realraelrr:
@chenhg5 This PR has been refreshed on top of the latest upstream main (`5a7dde44`) and is mergeable again. It is the WeCom semantic message chunking fix; could you please take another look and merge it when convenient?

chenhg5 (approved):
Conclusion: Approve

Overall assessment: Re-review confirms PR #963 is still mergeable and the semantic-message-chunking implementation is well-designed and well-tested. The two merge commits (fa1466f9 + 3e020b2d) brought in main updates without touching `platform/wecom/`, so the feature itself is unchanged from the prior approval.

Review scope:
- Reviewed `platform/wecom/message_split.go` (new, 125 lines), `platform/wecom/websocket_test.go` (new, 108 lines / 12 splitByBytes tests), and `platform/wecom/wecom.go` (-24 lines: removed old splitByBytes).
- Cross-checked the call site at `platform/wecom/wecom.go:471` and `platform/wecom/websocket.go:509` — both pass `maxBytes=2000` matching WeChat Work's 2048-byte limit with safety margin.
- Verified the algorithm via test coverage (12 dedicated tests cover all boundary windows).

✅ What looks good:
- `semanticByteCut` uses a 4-tier degressive priority: paragraph (70%) > line (80%) > sentence (85%) > soft (90%) > hard cut. This avoids the common mistake of splitting at the first available boundary (which would produce tiny first chunks); the windows ensure each split only triggers when the candidate is reasonably close to `maxBytes`.
- `hardByteCut` at message_split.go:117 uses `!utf8.RuneStart(s[end])` to back off continuation bytes — more idiomatic and correct than the old approach (which manually walked backward through `s[end]>>6 == 0b10`). The fallback `utf8.DecodeRuneInString` for the all-continuation edge case is correct.
- `splitCandidates` does a single pass over the prefix building all 4 candidate lists simultaneously, with `i >= maxBytes` early break — O(maxBytes) per call, not O(len(s)).
- Enumeration comma `、` is explicitly excluded via the soft-boundary list (only `，`/`,`/space/tab). The test `TestSplitByBytes_DoesNotSplitAtEnumerationComma` pins this — important for Chinese text where `、` is the natural list separator.
- 12 dedicated tests cover the algorithm's edge cases: short string, exact boundary, ASCII split, UTF-8 no-mid-rune invariant, empty string, large content round-trip, paragraph/line/sentence/soft boundary window thresholds, enumeration comma exclusion, soft boundary min window (90%), hard cut fallback. Reassembly invariant (`strings.Join(parts, "") == original`) is asserted in 4 tests.
- Additional tests in the file cover adjacent functionality that the prior review may have skimmed over: `handleMsgCallback` (3 tests: single-chat chatID fallback, group chat chatID preserved, bot mention strip), `ReconstructReplyCtx` (3 tests: valid/invalid prefix/too few parts), `writeAndWaitAck` (4 tests: success/error/timeout/context cancel), `handleFrame` (3 tests: ack dispatch/with error/ping resets missedPong), `generateReqID` (3 tests: monotonic/format/concurrent safety), `SendImage` (1 end-to-end frame flow), `newWebSocket` (2 tests: missing creds/valid config).
- Tests: `go test -tags no_web -count=1 ./platform/wecom/...` PASS (0.160s).
- `go vet -tags no_web ./platform/wecom/...` clean.
- CI run 26743242012: lint 2m23s / unit 4m01s / smoke 28s / regression 27s / performance 49s — 5/5 green on SHA 3e020b2d.
- Mergeable: clean.

- PR rebased onto current main via two merge commits (fa1466f9 then 3e020b2d). Neither merge touched `platform/wecom/` files — `git log 249921ac~1..3e020b2d -- platform/wecom/` returns only the feature commit itself. The feature is bit-identical to what was previously approved.
- The 16 main commits brought in by the rebase (PR #1154 feishu href forward, #1142 maintenance path, #1160 Windows instance lock, #1145 dingtalk spaceId, etc.) don't affect the wecom platform code or the message-split algorithm.

🔵 Optional (not blocking):
1. The `selectWindowSplit` function returns 0 when no candidate is in the window (causing the caller to fall through to the next priority tier). This is correct but subtle — a brief comment explaining the 0-means-keep-looking contract would help future maintainers.
2. `splitCandidates` doesn't dedupe candidates at the same position. In practice this doesn't matter (each `\n` only triggers once for paragraphs vs lines), but if a future refactor adds overlapping boundary types, dedup will need to be added.
3. The `splitByBytes` early-return at line 8 has `maxBytes <= 0` returning `[]string{s}` — but a maxBytes of 0 is semantically nonsensical (no chunk can fit). Consider returning an error or empty slice for `maxBytes <= 0` to surface caller bugs early. Defensive but currently dead-code since both call sites pass `2000`.

Testing / Risk:
- Verified locally: `go test -tags no_web -count=1 ./platform/wecom/...` PASS (0.160s), `go vet -tags no_web ./platform/wecom/...` clean.
- CI run 26743242012 (5/5 green on SHA 3e020b2d).
- Risk: Low. The algorithm is purely text-manipulation; no external dependencies, no API surface change, no schema change. The only behavior change is that long messages are now split at semantic boundaries (paragraph > line > sentence > soft) instead of mid-UTF-8-rune-safe hard cuts.
- Backwards compatible: any caller passing `maxBytes > 0` and a non-empty string gets semantically equivalent or better chunks. Hard cut fallback preserves the previous UTF-8-safe guarantee.
- Multi-platform relevance: this only affects WeCom. PR #971 (`fix(weixin): split long messages at semantic boundaries`) and similar platform-specific forks can be considered as follow-ups.

Next step: OK to merge. The WeCom 2048-byte limit previously produced awkward mid-word splits for Chinese content; this PR makes splits land at natural boundaries while preserving the byte-limit guarantee. After merge, consider whether to upstream the same semantic-chunking helper to `core/message_split.go` for reuse across WeChat (PR #971) and any other byte-limited platform (Telegram caption limit, Discord 2000-char limit).

PR #973, opened 2026-05-13
fix(cron): reject empty session_key in validateCronJob so management API doesn't persist unrunnable jobs

kevinWangSheng:
## Summary

\`CronJob.SessionKey\` is the platform anchor that \`Engine.ExecuteCronJob\` uses to find the target platform — it derives \`platformName\` from the prefix before the first \`:\`, walks \`e.platforms\`, and bails with

\`\`\`
platform \"\" not found for session \"\"
\`\`\`

if \`session_key\` is empty.

\`validateCronJob\` checks \`session_mode\`, \`mode\`, and \`timeout_mins\`, but **not** \`session_key\`. The constraint is only enforced by the unix-socket handler in \`core/api.go\` (\`handleCronAdd\`), which both auto-detects from active sessions and returns 400 when nothing resolves.

The management API path skips that pre-flight entirely:

\`\`\`go
// core/management.go handleCron POST
job := &CronJob{ ... SessionKey: req.SessionKey, ... }
if err := m.cronScheduler.AddJob(job); err != nil { ... }
mgmtJSON(w, http.StatusOK, job)
\`\`\`

\`AddJob\` calls \`validateCronJob\` (passes, since \`session_key\` isn't checked), persists the job to \`jobs.json\`, and \`scheduleJob\` attaches a cron entry to it. The job lives forever in visible-but-unrunnable state — every fire-time logs the cryptic \`platform \"\" not found for session \"\"\` and \`MarkRun\` writes that as \`last_error\`, but the operator never sees a 400 at create time.

## Fix

Add the \`session_key\` check inside \`validateCronJob\` so every \`AddJob\` caller — management API today, any future caller tomorrow — rejects the bad input at write time instead of leaking it into the store and the scheduler. \`api.go\`'s pre-flight remains: callers that want the auto-detect convenience still get it; callers that go straight to \`AddJob\` now get a clean error.

## Tests

- \`TestCronScheduler_AddJob_EmptySessionKey\` covers empty + whitespace-only \`session_key\` (subtests). Stash-revert just \`core/cron.go\` on this branch and the test fails with \`expected error for empty session_key, got nil\`, confirming the regression guard.
- The pre-existing \`TestMgmt_CronPatch\` was building its fixture by POSTing a job without \`session_key\` (i.e., exercising the buggy management path). Updated to pass \`session_key=\"test:chan:user\"\` so it goes through the canonical entry shape and still exercises the PATCH flow.

## Test plan

- [x] \`go test ./core/ -run \"TestCron|TestMgmt_Cron\" -count=1\` — all cron + management cron tests pass.
- [x] Pre-existing macOS-specific failures (\`TestProcessInteractiveEvents_*Footer*\`, \`TestResolveLocalDirPath_AcceptsSubdir\`) reproduce on \`main\` without this change — unrelated.

### DIFF
--- a/core/cron.go
+++ b/core/cron.go
@@ -81,6 +81,15 @@ func NormalizeCronSessionMode(s string) string {
 }
 
 func validateCronJob(j *CronJob) error {
+	// SessionKey anchors the cron execution to a platform (ExecuteCronJob
+	// derives platformName from the prefix before ":"). Without it the job
+	// is persisted but fails at fire-time with the unhelpful
+	// `platform "" not found for session ""`. Reject it up front so the
+	// caller (management API, /cron/add, /cron edit) sees an immediate
+	// 400 instead of a job that silently never runs.
+	if strings.TrimSpace(j.SessionKey) == "" {
+		return fmt.Errorf("session_key is required")
+	}
 	mode := NormalizeCronSessionMode(j.SessionMode)
 	if mode != "" && mode != "new_per_run" {
 		return fmt.Errorf("invalid session_mode %q (want reuse, new_per_run, or new-per-run)", j.SessionMode)
--- a/core/cron_test.go
+++ b/core/cron_test.go
@@ -443,6 +443,42 @@ func TestCronScheduler_AddJob_NegativeTimeoutMins(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


chenhg5 (approved):

PR #978, opened 2026-05-13
fix(relay): parseSessionKeyParts must skip single-char type tag to get the real chatID

kevinWangSheng:
## Summary

\`parseSessionKeyParts\` in \`core/relay.go\` splits the session key on \`\":\"\` and returns \`parts[1]\` as \`chatID\`, treating the format as \`platform:chatID:userID\`. That ignores the single-char conversation type tag that **dingtalk**, **qq**, and **qqbot** put between the platform name and the real chat identifier:

\`\`\`
dingtalk:g:cidXXX            (group, shared session)
dingtalk:g:cidXXX:staff1     (group, per-user session)
dingtalk:d:cidYYY:staff2     (direct/1:1 session)
qq:g:12345                   (group shared session)
qqbot:g:groupOpenID          (group shared session)
\`\`\`

For all of those, \`parts[1]\` is the literal tag \`\"g\"\` / \`\"d\"\` — not a channel ID. The returned \`chatID\` collapses **every shared-session group on those platforms to the same string** (\"g\" or \"d\").

The relay layer keys its state by \`chatID\`:

\`\`\`go
rm.bindings[chatID] = &RelayBinding{...}
\`\`\`

So \`/bind\` in dingtalk **group A** and \`/bind\` in dingtalk **group B** both end up under \`rm.bindings[\"g\"]\` — the second bind silently overwrites the first. Any subsequent relay message from group B reads group A's binding, and its

\`\`\`go
groupSessionKey := platform + \":\" + chatID + \":relay\"
\`\`\`

produces \`\"dingtalk:g:relay\"\`, which dingtalk's \`ReconstructReplyCtx\` then routes to whichever chat the binding points at — a real cross-group leak.

\`cmdBind\`, \`cmdBindStatus\`, \`RelayManager.Send\`, and the relay group-echo path all go through the same parser, so this fixes routing for every flow that uses it.

## Fix

Apply the same single-char tag heuristic that \`extractChannelID\` in \`core/engine.go\` (PR #955) already uses: when \`parts[1]\` is exactly one character and a third segment is present, treat \`parts[2]\` as the real \`chatID\`.

\`\`\`go
parts := strings.SplitN(sessionKey, \":\", 4)
// ...
if len(parts) >= 3 && len(parts[1]) == 1 {
    return parts[0], parts[2], nil
}
return parts[0], parts[1], nil
\`\`\`

## Tests

- \`TestParseSessionKeyParts_TypeTaggedKeys\` exercises feishu / slack / telegram / relay (must not regress) plus dingtalk-shared, dingtalk-per-user, dingtalk-direct, qq, and qqbot shapes (must now return the real chat id).
- \`TestParseSessionKeyParts_DistinctGroupsDontCollide\` asserts that two distinct dingtalk groups (\`cid_groupA\`, \`cid_groupB\`) produce two distinct \`chatID\`s — directly pinning the cross-group routing regression.

Stash-reverting just \`core/relay.go\` on this branch makes both new cases fail, confirming the regression guard.

## Test plan

- [x] \`go test ./core/ -run \"TestParseSessionKeyParts\" -count=1\` — all new + existing parser tests pass.
- [x] \`go test ./core/ -count=1\` — only the pre-existing macOS-specific failures (\`TestProcessInteractiveEvents_*Footer*\`, \`TestResolveLocalDirPath_AcceptsSubdir\`) reproduce on \`main\` without this change — unrelated.

### DIFF
--- a/core/relay.go
+++ b/core/relay.go
@@ -290,12 +290,21 @@ func (rm *RelayManager) relayContext(ctx context.Context) (context.Context, cont
 func parseSessionKeyParts(sessionKey string) (platform, chatID string, err error) {
 	// Format: "platform:chatID:userID"
 	// Relay session format: "relay:sourceProject:chatID"
-	parts := strings.SplitN(sessionKey, ":", 3)
+	// Some platforms (dingtalk, qq, qqbot) encode a single-char conversation
+	// type tag as an extra segment, e.g. "dingtalk:g:cidXXX", "qq:g:groupID",
+	// "qqbot:g:openid". For those, the real chatID is parts[2], not parts[1].
+	// Without this handling, every shared-session key on those platforms maps
+	// to chatID="g", so /bind bindings collide across different groups and
+	// relay routing leaks across chats.
+	parts := strings.SplitN(sessionKey, ":", 4)
 	if len(parts) < 2 {
 		return "", "", fmt.Errorf("invalid session key format: %q", sessionKey)
 	}
-	if parts[0] == "relay" && len(parts) == 3 {
+	if parts[0] == "relay" && len(parts) >= 3 {
 		// For relay sessions, chatID is the third part: "relay:sourceProject:chatID"
+		return parts[0], parts[2], nil
+	}
+	if len(parts) >= 3 && len(parts[1]) == 1 {
 		return parts[0], parts[2], nil
 	}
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):

PR #984, opened 2026-05-14
fix(observe): match Claude project key encoding for workdirs containing _, space, ~, non-ASCII

kevinWangSheng:
## Summary

`resolveClaudeProjectDir` in `cmd/cc-connect/main.go` builds the path to `~/.claude/projects/<key>/` by replacing only `os.PathSeparator` with `-`. However, the on-disk project key written by the editor encodes a broader set of characters as `-`: `/`, `:`, `_`, ` `, `~`, and any non-ASCII rune. The same encoding lives in `agent/claudecode/encodeClaudeProjectKey` already.

For any user running `--observe` with a workdir that contains those characters, `os.Stat` misses the real directory and observe mode silently disables itself:

```
WARN observe: could not find Claude Code project directory workDir=...
```

Common cases this hits:

- Workdirs with `snake_case` directory names (`/Users/alice/my_project`).
- Paths under iCloud / Mobile Documents (`/Users/x/Library/Mobile Documents/com~apple~CloudDocs/...`) — spaces and `~`.
- Workspaces with non-ASCII directory names.

## Fix

- Replace the simplistic encoding in `resolveClaudeProjectDir` with the same per-character encoding rules as `encodeClaudeProjectKey` (helper duplicated locally because the agent helper is unexported).
- Keep a fallback to the legacy separator-only encoding so on-disk keys produced before this change still resolve.

## Tests

New `cmd/cc-connect/observe_path_test.go` covers:

- `encodeClaudeProjectKey` table tests for underscores, spaces, `~`, non-ASCII, and Windows-style backslashes.
- `resolveClaudeProjectDir` end-to-end resolution with `t.TempDir()` mocks of `~/.claude/projects/<encoded>/` for underscored and spaced workdirs.
- A "missing dir → empty" guard.
- A legacy-encoding fallback case (on-disk key with `~` intact).

I also confirmed the regression guard pins the bug by stash-reverting `cmd/cc-connect/main.go`: the resolver returns `""` instead of the expected encoded path, just like the WARN in the real `--observe` startup.

## Test plan

- [x] `go test -tags no_web ./cmd/cc-connect/`
- [x] `go test -tags no_web -count=1 ./...` (core / daemon failures observed on plain `main` on macOS; HOME tilde shortening and `/var` → `/private/var` symlink — unrelated to this change)

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -1226,14 +1226,47 @@ func resolveClaudeProjectDir(workDir string) string {
 	if err != nil {
 		return ""
 	}
-	// Claude Code encodes paths by replacing os.PathSeparator with "-"
-	// e.g. /home/leigh/workspace/cc-connect -> -home-leigh-workspace-cc-connect
-	encoded := strings.ReplaceAll(workDir, string(os.PathSeparator), "-")
-	dir := filepath.Join(homeDir, ".claude", "projects", encoded)
-	if info, err := os.Stat(dir); err != nil || !info.IsDir() {
-		return ""
+	projectsBase := filepath.Join(homeDir, ".claude", "projects")
+	// Claude Code encodes paths by replacing '/', ':', '_', ' ', '~' and any
+	// non-ASCII rune with '-'. Replacing only os.PathSeparator misses common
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):

PR #991, opened 2026-05-14
fix(max): wire async webhook handler ctx to platform Stop()

kevinWangSheng:
## Summary

`webhookHandler` in `platform/max/max.go` spawns a goroutine for each incoming MAX update with a comment promising that \`Stop()\` will short-circuit in-flight handler work:

\`\`\`go
go func() {
    p.mu.RLock()
    ctx := context.Background()
    if p.cancel != nil {
        // Use the platform's own cancellation context so a Stop() also
        // short-circuits in-flight handler work.
        ctx2, cancel := context.WithCancel(ctx)
        _ = cancel
        ctx = ctx2
    }
    p.mu.RUnlock()
    p.handleUpdate(ctx, &upd)
}()
\`\`\`

The derived \`ctx2\` is a child of \`context.Background()\`, not of the parent context that \`p.cancel\` cancels. So:

- \`Stop()\` calling \`p.cancel()\` does **not** cancel the goroutine's ctx.
- The \`cancel\` returned by \`WithCancel\` is captured and immediately discarded (\`_ = cancel\`).
- Downstream cancellable work that ctx feeds — \`sendChatAction\`, \`fetchAttachments\`, \`p.client.Do(req)\` via \`NewRequestWithContext\` — keeps running until each request's HTTP client timeout instead of being short-circuited by shutdown.

## Fix

- Store the parent context the platform creates in \`Start\` on a new \`Platform.ctx\` field, alongside the existing \`Platform.cancel\`.
- Extract \`webhookCtx()\` which returns \`p.ctx\` when the platform is running and \`context.Background()\` as a unit-test fallback for callers that invoke \`webhookHandler\` directly.
- Replace the broken \`WithCancel + discarded cancel\` block in the webhook goroutine with a single \`p.webhookCtx()\` call.

The long-polling path (\`go p.pollLoop(ctx)\` in \`Start\`) was already correctly wired to \`p.cancel\` and is unchanged.

## Tests

Added to \`platform/max/max_test.go\`:

- \`TestWebhookCtx_FallsBackToBackgroundBeforeStart\` — pins the unit-test fallback so existing webhook tests that exercise \`webhookHandler\` without \`Start\` keep working.
- \`TestWebhookCtx_CanceledByStop\` — starts a platform against the mock API, captures \`webhookCtx()\`, calls \`Stop()\`, and asserts the context's \`Done\` channel closes within 2s. Before the fix this asserts on a field/method that doesn't exist; after the fix it observes real cancellation propagation.

Stash-revert of \`platform/max/max.go\` confirms the regression guard pins the bug (test file fails to compile against the un-fixed production code).

## Test plan

- [x] \`go test -tags no_web -count=1 ./platform/max/\`
- [x] \`go vet -tags no_web ./platform/max/\`
- [x] \`go test -tags no_web -count=1 ./...\` (pre-existing failures in \`core\` / \`daemon\` / \`agent/codex\` reproduce on plain \`main\` on macOS; unrelated to this change)

### DIFF
--- a/platform/max/max.go
+++ b/platform/max/max.go
@@ -70,6 +70,7 @@ type Platform struct {
 
 	mu           sync.RWMutex
 	handler      core.MessageHandler
+	ctx          context.Context
 	cancel       context.CancelFunc
 	stopping     bool
 	client       *http.Client // general API calls — httpTimeout
@@ -160,6 +161,7 @@ func (p *Platform) Start(handler core.MessageHandler) error {
 	p.handler = handler
 
 	ctx, cancel := context.WithCancel(context.Background())
+	p.ctx = ctx
 	p.cancel = cancel
 
 	// Verify token at startup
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):

PR #995, opened 2026-05-14
fix(daemon/launchd): xml-escape user-supplied paths in the plist template

kevinWangSheng:
## Summary

\`daemon/launchd.go:buildPlist\` embeds \`cfg.BinaryPath\`, \`cfg.WorkDir\`, \`cfg.LogFile\`, and \`cfg.EnvPATH\` into the launchd \`.plist\` template using plain \`fmt.Sprintf(\"%s\", …)\`. None of them is XML-escaped.

If any of those values contains one of XML's five reserved characters — \`&\`, \`<\`, \`>\`, \`\"\`, \`'\` — the resulting plist is malformed and \`launchctl bootstrap\` refuses to load it with a parse error. \`cc-connect daemon install\` then fails before the daemon is ever registered.

Concrete triggers users hit in real environments:

- A working directory like \`/Users/jane/Projects/dev & test/cc-connect\` (the \`&\` makes the plist invalid XML).
- A log file path under a home directory that contains an apostrophe.
- Any path that has been pre-quoted with embedded \`\"\`.

The Windows daemon manager already escapes user values via \`powerShellLiteral\`. The launchd manager was the inconsistent path.

## Fix

- Pass each user-supplied \`%s\` placeholder through a new \`xmlEscape\` helper that delegates to \`encoding/xml.EscapeText\` (the canonical escaper, handles all five reserved chars correctly).
- Leave \`launchdLabel\` (compile-time constant) and \`cfg.LogMaxSize\` (\`%d\`) unchanged — neither needs escaping.

## Tests

\`TestBuildPlist_EscapesXMLSpecialCharsInPaths\` asserts that:

1. Output is well-formed XML (parses with \`encoding/xml\`).
2. The original path values round-trip through \`xml.Decoder\` — confirming escaped entities decode back to the literal characters.
3. There are no bare \`&\` characters in the output that are not part of a valid entity reference (the precise condition launchctl rejects on).

Regression guard verified by stash-revert: without the production fix, the test fails at the XML-parse step.

## Test plan

- [x] \`go test -tags no_web -count=1 ./daemon/\` (the pre-existing \`TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable\` failure reproduces on plain \`main\` on this macOS host and is unrelated)
- [x] \`go vet -tags no_web ./daemon/\`

### DIFF
--- a/daemon/launchd.go
+++ b/daemon/launchd.go
@@ -3,6 +3,7 @@
 package daemon
 
 import (
+	"encoding/xml"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):

PR #998, opened 2026-05-15
fix(core): protect Session.Name writes with the session mutex

kevinWangSheng:
## Summary

\`handleProjectSessions\` POST in \`core/management.go\` mutates the session's display name by writing the field directly:

\`\`\`go
s := e.sessions.GetOrCreateActive(body.SessionKey)
if body.Name != \"\" {
    s.Name = body.Name
}
\`\`\`

The GET branch of the same handler iterates sessions and reads \`s.Name\` inside \`s.mu.Lock()\`. \`Session.SetAgentInfo\`, the other place that writes \`s.Name\`, also acquires \`s.mu\`. The management POST handler was the inconsistent one, so two clients hitting \`POST /sessions\` (rename) and \`GET /sessions\` (list) concurrently race on the \`Name\` string field. Go strings carry a pointer and a length; a torn read against a write of a different-length name can produce a frankenstring whose pointer doesn't match its length, leading to memory-safety failures or garbled output.

## Fix

- Add \`Session.SetName(name string)\` that acquires \`s.mu\` (mirrors \`SetAgentSessionID\`, \`SetAgentInfo\`).
- Switch the management POST handler to call \`s.SetName(body.Name)\` and read back via \`s.GetName()\` so the response body also goes through the lock.

## Tests

Two regression guards added:

- \`TestSession_SetName_RaceFree\` — direct test of the new method: 40 concurrent \`SetName\` writers and 40 concurrent \`GetName\` readers run under \`-race\`. With the fix the test is clean; without the new method the build fails.
- \`TestMgmt_SessionsConcurrentNameWriteAndList\` — behavioral test that fires 30 concurrent POST + 30 concurrent GET against the management API for the same session_key. Stash-reverting just \`core/management.go\` while keeping the new \`SetName\` method confirms the test trips \"DATA RACE detected\" — pinning the management-API integration, not just the helper.

## Test plan

- [x] \`go test -tags no_web -race -count=1 -run TestSession_SetName_RaceFree ./core/\`
- [x] \`go test -tags no_web -race -count=1 -run TestMgmt_SessionsConcurrentNameWriteAndList ./core/\`
- [x] \`go test -tags no_web -count=1 ./core/\` (pre-existing \`TestProcessInteractiveEvents_*\` and \`TestResolveLocalDirPath_AcceptsSubdir\` failures reproduce on plain \`main\` on macOS — unrelated to this change)

### DIFF
--- a/core/management.go
+++ b/core/management.go
@@ -992,13 +992,13 @@ func (m *ManagementServer) handleProjectSessions(w http.ResponseWriter, r *http.
 
 		s := e.sessions.GetOrCreateActive(body.SessionKey)
 		if body.Name != "" {
-			s.Name = body.Name
+			s.SetName(body.Name)
 		}
 		e.sessions.Save()
 
 		mgmtJSON(w, http.StatusOK, map[string]any{
 			"session_key": body.SessionKey,
-			"name":        s.Name,
+			"name":        s.GetName(),
 		})
 
 	default:
--- a/core/management_test.go
+++ b/core/management_test.go
@@ -346,6 +346,45 @@ func TestMgmt_SessionDetail(t *testing.T) {
 	}
 }
 
+// TestMgmt_SessionsConcurrentNameWriteAndList pins the bug where the
+// POST /sessions handler wrote s.Name = body.Name directly without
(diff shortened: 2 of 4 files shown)


chenhg5 (approved):

PR #999, opened 2026-05-15
agent/tmux: add tmux agent

bydsky:
Drives a persistent tmux pane as an interactive shell agent to get around claude code -p. User messages are sent via send-keys; responses are extracted from the scrollback buffer so terminal scrolling never bleeds previous turns into the next response.

Key features:
- Multi-workspace: each workspace gets its own named tmux window (session:workspace-name), with Claude Code auto-started via init_command on first use
- Scrollback-based extraction: records scrollback line count at send-time; response = lines added since, immune to terminal scroll
- Stability detection: fires done when pane is unchanged for 2+ polls (fast path) or 5 s idle fallback when prompt pattern never matches
- TUI cleanup: strips Claude Code's ───/❯/─── input block and configurable per-line strip_patterns (default: mode status line)
- AgentOptsProvider: new core interface so workspace agents inherit agent-specific options (session name etc.) during multi-workspace cloning

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ PLATFORMS := \
 #   make build EXCLUDE=discord,dingtalk,qq,qqbot,line
 # ---------------------------------------------------------------------------
 
-ALL_AGENTS    := acp claudecode codex cursor devin gemini iflow kimi opencode pi qoder
+ALL_AGENTS    := acp claudecode codex cursor devin gemini iflow kimi opencode pi qoder tmux
 ALL_PLATFORMS := feishu telegram discord slack dingtalk wecom weixin qq qqbot line weibo max
 ALL_EXTRAS    := web
 
--- a/agent/tmux/session.go
+++ b/agent/tmux/session.go
@@ -17,6 +17,7 @@ import (
 type tmuxSession struct {
 	target          string // e.g., "mywork:0"
 	sessionID       string
+	workDir         string
 	promptPat       *regexp.Regexp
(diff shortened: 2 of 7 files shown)


chenhg5:
@bydsky Thanks for the tmux agent work. I reviewed the diff and CI is green, but I think this needs a few fixes before merge:

1. `agent/tmux/session.go`: `sendKeys()` calls `tmux send-keys` without literal mode. User text like `C-c`, `Enter`, or `Up` can be interpreted as tmux key names instead of plain text, which lets a remote message interrupt or control the pane. Please send user prompts with literal text mode and send Enter separately.

2. `agent/tmux/session.go`: file attachments are saved with `core.SaveFilesToDisk(".", files)` instead of the tmux agent workspace. In multi-workspace mode the CLI running inside the pane may not be able to read those paths. This should use the session/workDir, like the other agents do.

3. `agent/tmux/tmux.go`: workspace windows are named only with `filepath.Base(workDir)`. Two different workspaces with the same directory basename, for example `/repo/a/app` and `/repo/b/app`, will collide into the same tmux target and can cross-wire sessions.

I also ran `go test ./agent/tmux ./core` locally on this PR and it passed. My recommendation is to fix the three items above and add regression tests for them before merging.

chenhg5 (approved):
LGTM. New `tmux` agent adapter with session management, proper build tag isolation (`no_tmux`). No breaking changes. Approving.

PR #1000, opened 2026-05-15
discord: per-guild mention bypass, reply context, and multi-workspac

bydsky:
group_reply_all_guilds: comma-separated list of guild IDs where the bot responds to all messages without requiring an @mention. Supports "*" to match all guilds (equivalent to the existing group_reply_all bool).

Reply context: when a message quotes/replies to another message, prepend the referenced message's content and images to the agent input so the agent has full context without the user needing to repeat it.

Multi-workspace: fix session key context propagation so workspace-bound sessions correctly resolve their channel and exec skill commands carry the right workspace context.

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -319,6 +319,9 @@ func main() {
 					engine.SetWorkspaceIdleTimeout(time.Duration(mins) * time.Minute)
 				}
 			}
+			if proj.SkipGit != nil {
+				engine.SetSkipGit(*proj.SkipGit)
+			}
 			slog.Info("multi-workspace mode enabled", "project", proj.Name, "base_dir", baseDir)
 		}
 
--- a/config.example.toml
+++ b/config.example.toml
@@ -605,6 +605,7 @@ level = "info" # debug, info, warn, error
 # [[projects]]
 # name = "claude-multi"
 # mode = "multi-workspace"
+# skip_git = true # skip bind git repo, create a dir directly
 # base_dir = "~/workspace"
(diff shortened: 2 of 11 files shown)


chenhg5:
@bydsky Thanks for this Discord and multi-workspace update. I reviewed the diff and CI is green, but I found a couple of multi-workspace state bugs that should be fixed before merge:

1. `core/engine.go`: `cmdReasoning()` resolves the workspace-aware `sessions`, clears the workspace session, but then calls `e.sessions.Save()` instead of `sessions.Save()`. In multi-workspace mode this means the modified workspace session may not be persisted correctly, so a restart can restore stale session state.

2. `core/engine.go`: `/provider clear` and `switchProvider()` operate on the workspace-resolved `ProviderSwitcher`, but still clear/save `e.sessions` and persist through the global `e.providerSaveFunc`. That can leave runtime workspace provider state and persisted/global session state out of sync.

3. It would be good to add regression coverage for the new Discord paths: `group_reply_all_guilds`, quoted/referenced message context, referenced image forwarding, and workspace-aware command routing.

I also ran `go test ./core ./platform/discord` locally on this PR and it passed. My recommendation is to fix the workspace session/provider persistence issues before merging.

chenhg5 (approved):
LGTM. Multi-workspace-aware command routing (cmdUsage, cmdReasoning, cmdMode, cmdCompress, cmdAllow, cmdProvider, cmdMemory, executeSkill, etc.) + Discord per-guild mention bypass and reply context. No breaking changes, substantial improvement. Approving.

PR #1001, opened 2026-05-15
fix(wecom): verify callback signature in constant time

kevinWangSheng:
## Summary

\`Platform.verifySignature\` in \`platform/wecom/wecom.go\` compares the locally-computed SHA1 digest against the caller-supplied \`msg_signature\` using plain string equality:

\`\`\`go
got := fmt.Sprintf(\"%x\", h.Sum(nil))
return got == expected
\`\`\`

Every other credential or signature comparison in the project uses \`subtle.ConstantTimeCompare\`:

- \`core/webhook.go\` — webhook token (Bearer / X-Webhook-Token / query param)
- \`core/bridge.go\` — bridge token
- \`core/management.go\` — management API token
- \`platform/max/max.go\` — webhook secret

wecom was the only signature path short-circuiting on the first mismatching byte, which is inconsistent with the project's documented security pattern.

## Fix

- Switch the comparison to \`subtle.ConstantTimeCompare\`.
- Use \`encoding/hex.EncodeToString\` for the digest instead of \`fmt.Sprintf(\"%x\", ...)\` (avoids a \`fmt\` round trip and matches how the constant-time helper is used elsewhere).

## Tests

Added to \`platform/wecom/wecom_test.go\`:

- \`TestVerifySignature_AcceptsMatchingDigest\` — happy path.
- \`TestVerifySignature_RejectsMismatchedDigest\` — empty / too short / too long / last-byte / first-byte / non-hex variants.
- \`TestVerifySignature_DigestIsLowercaseHex\` — constant-time compare is byte-exact, so uppercase-hex must still be rejected.

A local \`wecomSignature\` helper re-implements the digest scheme so the assertions don't depend on the production helper they're verifying.

## Test plan

- [x] \`go test -tags no_web -count=1 ./platform/wecom/\`
- [x] \`go vet -tags no_web ./platform/wecom/\`

### DIFF
--- a/platform/wecom/wecom.go
+++ b/platform/wecom/wecom.go
@@ -6,8 +6,10 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/sha1"
+	"crypto/subtle"
 	"encoding/base64"
 	"encoding/binary"
+	"encoding/hex"
 	"encoding/json"
 	"encoding/xml"
 	"fmt"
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. WeCom 回调签名验证使用 constant-time 比较，防止 timing attack，测试覆盖。

PR #1003, opened 2026-05-15
fix(telegram): use MessageThreadID instead of IsForum for callback thread routing

### LINKED ISSUE #1002: [Bug] Telegram: permission button click has no effect in forum groups (IsForum missing in callback)
## Bug Description

In Telegram **forum groups** (groups with Topics enabled), clicking the ⚠️ Permission Request inline keyboard buttons (Allow / Deny / Allow All) has no visible effect — the button click is silently dropped.

## Root Cause

In `platform/telegram/telegram.go`, `handleCallbackQuery` determines `threadID` like this:

```go
isGroupChat := msg.Chat.Type == models.ChatTypeGroup || msg.Chat.Type == models.ChatTypeSupergroup
threadID := 0
if msg.Chat.IsForum || !isGroupChat {
    threadID = msg.MessageThreadID
}
sessionKey := p.buildSessionKey(chatID, threadID, cb.From.ID)
```

The Telegram Bot API does **not** reliably include `IsForum` in the `Chat` object embedded inside a `CallbackQuery` message. As a result, for forum groups:

- `isGroupChat = true`
- `msg.Chat.IsForum = false` (missing from API response)
- `!isGroupChat = false`
- → condition is `false`, so `threadID` stays `0`

The built session key becomes `telegram:<chatID>:0:<userID>`, which does **not** match the actual session key `telegram:<chatID>:<threadID>:<userID>` (e.g. `telegram:-1003872026981:31:7179563622`). The permission response is routed to a non-existent session and silently discarded.

The same issue likely affects `cmd:` and `askq:` callbacks in forum groups.

## Fix

Replace the `IsForum` check with a check on `MessageThreadID != 0`, which is always present when the message is in a topic thread:

```go
threadID := 0
if !isGroupChat || msg.MessageThreadID != 0 {
    threadID = msg.MessageThreadID
}
```

This correctly handles all cases:
- **DM**: `isGroupChat=false` → uses `MessageThreadID` (always 0 for DMs, correct)
- **Regular group** (no topics): `isGroupChat=true`, `MessageThreadID=0` → `threadID=0`, correct
- **Forum group** (with topics): `isGroupChat=true`, `MessageThreadID=31` → `threadID=31`, correct

## Environment

- cc-connect v1.3.2
- Platform: Telegram supergroup with Forum/Topics enabled
- Config: `thread_isolation = true`

## Workaround

Reply with text `allow` / `deny` / `allow all` instead of clicking the button, or set `mode = "auto"` in config to avoid permission prompts altogether.

HaiyiMei:
## Summary

- `IsForum` is not reliably present in the `Chat` object embedded inside a `CallbackQuery` message from Telegram's Bot API
- For forum groups (Topics enabled), this caused `threadID` to resolve as `0` in `handleCallbackQuery`, building a wrong session key (e.g. `telegram:-100xxx:0:uid` instead of `telegram:-100xxx:31:uid`)
- The callback was routed to a non-existent session and silently dropped — clicking Allow / Deny / Allow All had no effect

Fix: replace `msg.Chat.IsForum` check with `msg.MessageThreadID != 0`, which is always populated when the message is in a topic thread.

```go
// Before
if msg.Chat.IsForum || !isGroupChat {
    threadID = msg.MessageThreadID
}

// After
if !isGroupChat || msg.MessageThreadID != 0 {
    threadID = msg.MessageThreadID
}
```

This correctly handles all cases:
- **DM**: `isGroupChat=false` → uses `MessageThreadID` (always 0, correct)
- **Regular group** (no topics): `isGroupChat=true`, `MessageThreadID=0` → `threadID=0`, correct
- **Forum group** (with topics): `isGroupChat=true`, `MessageThreadID=31` → `threadID=31`, correct ✅

Closes #1002

## Test plan

- [ ] In a Telegram forum group with `thread_isolation = true`, trigger a permission request and click Allow — should now respond correctly
- [ ] Verify regular group and DM callback behavior is unchanged

### DIFF
--- a/platform/telegram/telegram.go
+++ b/platform/telegram/telegram.go
@@ -727,7 +727,10 @@ func (p *Platform) handleCallbackQuery(ctx context.Context, cb *models.CallbackQ
 
 	isGroupChat := msg.Chat.Type == models.ChatTypeGroup || msg.Chat.Type == models.ChatTypeSupergroup
 	threadID := 0
-	if msg.Chat.IsForum || !isGroupChat {
+	// IsForum is not reliably present in the Chat object embedded in a CallbackQuery
+	// message. Use MessageThreadID directly: non-zero means we're in a forum topic,
+	// zero means a regular group (no topic isolation needed).
+	if !isGroupChat || msg.MessageThreadID != 0 {
 		threadID = msg.MessageThreadID
 	}
 	sessionKey := p.buildSessionKey(chatID, threadID, cb.From.ID)


chenhg5 (approved):
LGTM. Telegram callback thread 路由使用 MessageThreadID 替代 IsForum，更可靠。

PR #1005, opened 2026-05-15
fix(agent/codex): capture workDir under mutex in StartSession and SkillDirs

kevinWangSheng:
## Summary

In \`agent/codex/codex.go\`, two methods read \`a.workDir\` without holding \`a.mu\`, while \`SetWorkDir\` mutates it under the lock:

- \`StartSession\` (line 333-371): the locked block captures \`mode\`, \`model\`, \`reasoningEffort\`, \`backend\`, \`appServerURL\`, \`codexHome\`, \`cliBin\`, \`cliExtraArgs\`, providers — but **not** \`a.workDir\`. After \`a.mu.Unlock()\`, the two return sites at lines 364 and 370 then read \`a.workDir\` to pass to \`newAppServerSession\` / \`newCodexSession\`.
- \`SkillDirs\` (line 440-446): reads \`a.workDir\` and \`a.codexHome\` directly.

When a user issues \`/workspace <dir>\` on one channel/goroutine and a turn starts or a skill lookup runs on another, the readers race with \`SetWorkDir\`. Go strings carry a \`(pointer, length)\` pair, and a torn read of a different-length string can produce a frankenstring whose pointer doesn't match its length, leading to memory-safety failures.

The same agent's \`ListSessions\`, \`GetSessionHistory\`, and \`DeleteSession\` already follow the canonical \"capture under RLock\" pattern; \`StartSession\` and \`SkillDirs\` were the inconsistent ones.

## Fix

- Add \`workDir := a.workDir\` inside the locked block in \`StartSession\`, then pass the local through to both \`newAppServerSession\` and \`newCodexSession\` instead of re-reading \`a.workDir\` after \`Unlock()\`.
- Capture both \`workDir\` and \`codexHome\` under \`a.mu.RLock()\` in \`SkillDirs\` before computing the absolute path.

No behaviour change; no public API change.

## Tests

\`TestSkillDirs_RaceFreeAgainstSetWorkDir\` (in \`agent/codex/skilldirs_test.go\`) spawns 30 concurrent \`SetWorkDir\` writers and 30 concurrent \`SkillDirs\` readers under \`-race\`. With the fix the race detector stays quiet; stash-reverting just \`agent/codex/codex.go\` confirms the test trips \"DATA RACE detected\".

## Test plan

- [x] \`go test -tags no_web -race -count=1 -run TestSkillDirs_RaceFreeAgainstSetWorkDir ./agent/codex/\`
- [x] \`go test -tags no_web -count=1 ./agent/codex/\`
- [x] \`go vet -tags no_web ./agent/codex/\`

### DIFF
--- a/agent/codex/codex.go
+++ b/agent/codex/codex.go
@@ -339,6 +339,7 @@ func (a *Agent) StartSession(ctx context.Context, sessionID string) (core.AgentS
 	codexHome := a.codexHome
 	cliBin := a.cliBin
 	cliExtraArgs := a.cliExtraArgs
+	workDir := a.workDir
 	extraEnv := a.providerEnvLocked()
 	extraEnv = append(extraEnv, a.sessionEnv...)
 	var baseURL string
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. codex StartSession/SkillDirs workDir mutex 保护，测试覆盖。

PR #1007, opened 2026-05-15
fix(cmd/sessions): strip work_dir hash suffix when deriving project name

kevinWangSheng:
## Summary

\`loadAllSessions\` in \`cmd/cc-connect/sessions.go\` derived the project label by stripping only the \`.json\` extension:

\`\`\`go
project := strings.TrimSuffix(entry.Name(), \".json\")
\`\`\`

But the engine actually writes session files under four schemes (documented in \`session_id.go:matchesProject\`):

- \`<project>.json\` — no work_dir
- \`<project>_<8hex>.json\` — single work_dir (8-char hash of the absolute path, see \`sessionStorePath\` in \`main.go\`)
- \`<project>_ws_<8hex>.json\` — multi-workspace (see \`core/engine.go\`)
- \`<project>.sessions.json\` — legacy

For a typical session file like \`myproject_deadbeef.json\`, the project label came out as \`myproject_deadbeef\`. That label then went into \`GlobalID = myproject_deadbeef:s1\`, so:

- \`cc-connect sessions list\` displays \`myproject_deadbeef\` instead of \`myproject\`.
- \`cc-connect sessions show myproject:s1\` (the documented address form) can never match — only \`myproject_deadbeef:s1\` does.

## Fix

Add \`extractProjectFromFilename\` that mirrors the four naming schemes and strips the trailing 8-hex hash (or \`_ws_<8hex>\`, or \`.sessions\`). The strict 8-hex length matches the real engine output and avoids misclassifying real project names whose last segment after \`_\` happens to be valid hex (e.g. \`project_a\`, \`foo_dead\`).

No public API change. Existing \`TestLoadAllSessions\` continues to pass because its fixture files (\`project_a.json\`, \`project_b.json\`) don't have an 8-hex tail.

## Tests

- \`TestExtractProjectFromFilename\` — table-tests every scheme, plus the false-positive cases the loose \"any-length hex\" heuristic would have swallowed.
- \`TestLoadAllSessions_StripsWorkDirHash\` — end-to-end: writes a session file named \`myproject_deadbeef.json\` (the real engine format) and asserts \`Project = \"myproject\"\` / \`GlobalID = \"myproject:s1\"\`. With the fix the assertions pass; without the production fix the test reports \`Project = \"myproject_deadbeef\"\`, exactly the buggy output.

## Test plan

- [x] \`go test -tags no_web -count=1 ./cmd/cc-connect/\`
- [x] \`go vet -tags no_web ./cmd/cc-connect/\`

### DIFF
--- a/cmd/cc-connect/sessions.go
+++ b/cmd/cc-connect/sessions.go
@@ -121,6 +121,45 @@ func resolveDataDir(flagValue string) string {
 	return ".cc-connect"
 }
 
+// sessionFileHashLen is the length of the hex-encoded work_dir hash the
+// engine writes into session file names — sha256 truncated to 4 bytes,
+// rendered as 8 hex chars (see sessionStorePath in main.go and the
+// "_ws_<hash>" path in core/engine.go).
+const sessionFileHashLen = 8
+
+// extractProjectFromFilename derives the project name from a session
+// file name. The engine writes sessions under one of four naming
+// schemes (see matchesProject in session_id.go):
+//
+//   - <project>.json                — no work_dir
+//   - <project>_<8hex>.json         — single work_dir
+//   - <project>_ws_<8hex>.json      — multi-workspace
+//   - <project>.sessions.json       — legacy
+//
+// Strip the trailing 8-hex hash / "_ws_<8hex>" / ".sessions" so the
+// returned project name matches the user-facing project identifier
+// used by `cc-connect sessions show <project>:<session_id>`. The
+// strict 8-hex length (rather than any-length hex) avoids
+// misclassifying user-chosen project names whose last segment after
+// "_" happens to be valid hex (e.g. "project_a", "foo_dead").
+func extractProjectFromFilename(filename string) string {
+	base := strings.TrimSuffix(filename, ".json")
+	if strings.HasSuffix(base, ".sessions") {
+		return strings.TrimSuffix(base, ".sessions")
+	}
+	if idx := strings.LastIndex(base, "_"); idx >= 0 {
+		suffix := base[idx+1:]
+		if len(suffix) == sessionFileHashLen && isHex(suffix) {
+			stripped := base[:idx]
+			if strings.HasSuffix(stripped, "_ws") {
+				stripped = strings.TrimSuffix(stripped, "_ws")
+			}
+			return stripped
+		}
+	}
+	return base
+}
+
 func loadAllSessions(dataDir string) ([]sessionRecord, error) {
 	sessionsDir := filepath.Join(dataDir, "sessions")
 	entries, err := os.ReadDir(sessionsDir)
@@ -137,7 +176,7 @@ func loadAllSessions(dataDir string) ([]sessionRecord, error) {
 			continue
 		}
 
-		project := strings.TrimSuffix(entry.Name(), ".json")
+		project := extractProjectFromFilename(entry.Name())
 		filePath := filepath.Join(sessionsDir, entry.Name())
 
 		data, err := os.ReadFile(filePath)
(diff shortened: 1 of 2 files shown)


chenhg5 (changes requested):
CI lint 失败：

`cmd/cc-connect/sessions.go:154:4`: S1017 - should replace this if statement with an unconditional `strings.TrimSuffix`

请修复后重新提交。

kevinWangSheng:
Fixed the S1017 lint: collapsed the `if HasSuffix / TrimSuffix` pair into an unconditional `strings.TrimSuffix` since the guard is redundant. The existing `TestExtractProjectFromFilename/project_with__ws__+_hash` case still passes. Ready for re-review.

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
Re-reviewing after my prior change request (review id 4302512213) is now resolved. The single new commit `c54db32a` ("lint: collapse if HasSuffix/TrimSuffix into unconditional TrimSuffix") applies exactly the S1017 fix I flagged: replaces the `if strings.HasSuffix(stripped, "_ws") { stripped = strings.TrimSuffix(stripped, "_ws") }` two-line pattern with a single `strings.TrimSuffix(base[:idx], "_ws")` call. Behavior is identical (TrimSuffix is a no-op when the suffix doesn't match), but the lint check now passes. CI run 25989899523: 5/5 green including the previously-failing lint job. `go vet -tags no_web ./cmd/cc-connect/...` clean. `go build -tags no_web ./cmd/cc-connect/...` OK locally (re-verified).

Review scope:
- `cmd/cc-connect/sessions.go` L150-156 — `extractProjectFromFilename` updated.
- Diff vs last_review_commit (0565a4d): -4 lines net, 1 commit (c54db32a).
- Focused on confirming the lint fix is applied correctly and doesn't change behavior.

✅ What looks good:
- The fix is minimal and surgical: collapses two lines (if + assignment) into a single `return strings.TrimSuffix(...)` call inside the existing `if len(suffix) == sessionFileHashLen && isHex(suffix)` guard. No other code path changed.
- `strings.TrimSuffix(s, suffix)` is documented to return `s` unchanged when `s` doesn't end with `suffix`, so the new code is semantically equivalent to the prior `if HasSuffix then TrimSuffix else keep` pattern. Confirmed by reading the Go stdlib docs and by the surrounding comment at L143-148 (the "strict 8-hex length" guard prevents misclassifying user-chosen names whose last segment happens to be valid hex, e.g. "project_a", "foo_dead").
- Surrounding logic is preserved: still requires `_` separator at idx, still requires suffix length == sessionFileHashLen (8) and isHex, still falls through to `return base` if the hash suffix isn't present. The only "behavior change" is the removal of the `stripped` local variable — pure cosmetic.
- CI run 25989899523 lint check passes (1m44s); previously the lint step in run before this commit would have failed on S1017. The other 4 CI steps (unit 3m42s, smoke 26s, regression 25s, performance 53s) also green.
- `go vet -tags no_web ./cmd/cc-connect/...` clean and `go build -tags no_web ./cmd/cc-connect/...` OK locally.

🟢 Resolved since prior review:
- ✅ S1017 lint failure (the only blocker in my 2026-05-16 review).

Testing / Risk:
- Verified evidence: CI 5/5 green on new SHA; `go vet` + `go build` clean locally; diff is a pure cosmetic refactor with no semantic impact.
- Remaining risk: zero. The function logic is byte-equivalent for all inputs (TrimSuffix vs conditional TrimSuffix produce identical outputs).

Next step:
OK to merge. This was a trivial lint cleanup that should have been a fast follow-up; thanks for addressing it. No release-note entry needed.

PR #1009, opened 2026-05-16
fix(agent/claudecode): capture workDir under mutex across all readers

kevinWangSheng:
## Summary

\`agent/claudecode/claudecode.go\` has six methods that read \`a.workDir\` (and \`a.mode\` in \`StartSession\`) without holding \`a.mu\`, while \`SetWorkDir\` and \`SetMode\` mutate those fields under the lock. claudecode is the primary agent the project ships against, so this race has the largest cumulative impact of any agent.

Affected readers:

- \`StartSession\` (line 403-435) — the locked block already captures ~10 fields but reads \`a.workDir\` and \`a.mode\` AFTER \`a.mu.Unlock()\` when calling \`newClaudeSession\`.
- \`ListSessions\` (line 444), \`DeleteSession\` (line 497), \`GetSessionHistory\` (line 562) — each calls \`filepath.Abs(a.workDir)\` without the lock.
- \`CommandDirs\` (line 836), \`SkillDirs\` (line 850), \`ProjectMemoryFile\` (line 946) — each reads \`a.workDir\` directly when building user-facing config / skill / memory paths.

When a user issues \`/workspace <dir>\` on one channel/goroutine and a turn start, \`/list\`, a memory-file lookup, or a skill scan runs on another, the readers race with \`SetWorkDir\`. Go strings carry a \`(pointer, length)\` pair; a torn read of a different-length string can produce a frankenstring whose pointer doesn't match its length, leading to memory-safety failures or empty/garbled paths.

The pattern mirrors my recent fixes for codex (#1005), opencode (#1006), and iflow (#996), but is consolidated into one PR for claudecode because all six readers share the same fix shape and live in a single file.

## Fix

- In \`StartSession\`, capture \`workDir := a.workDir\` and \`mode := a.mode\` inside the locked block, then pass the locals to \`newClaudeSession\` instead of re-reading the fields after \`Unlock()\`.
- In each remaining reader, capture \`workDir\` under \`a.mu.RLock()\` and operate on the local copy.

No public API or behaviour change. \`cliBin\`, \`cliExtraArgs\`, \`cliArgsFlag\`, \`routerURL\`, \`spawnOpts\` are set once in \`New\` and never mutated, so they don't need locking — only \`workDir\` and \`mode\` were unsafe.

## Tests

\`TestClaudecodeAgent_WorkDirRaceFreeReaders\` (new file \`agent/claudecode/claudecode_workdir_race_test.go\`) spawns 30 concurrent \`SetWorkDir\` writers and 30 concurrent \`CommandDirs\` / \`SkillDirs\` / \`ProjectMemoryFile\` readers under \`-race\`. With the production fix the race detector stays quiet; stash-reverting just \`agent/claudecode/claudecode.go\` confirms the test trips \"DATA RACE detected\".

## Test plan

- [x] \`go test -tags no_web -race -count=1 -run TestClaudecodeAgent_WorkDirRaceFreeReaders ./agent/claudecode/\`
- [x] \`go test -tags no_web -count=1 ./agent/claudecode/\`
- [x] \`go vet -tags no_web ./agent/claudecode/\`

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -409,6 +409,8 @@ func (a *Agent) StartSession(ctx context.Context, sessionID string) (core.AgentS
 	maxTok := a.maxContextTokens
 	model := a.model
 	effort := a.reasoningEffort
+	workDir := a.workDir
+	mode := a.mode
 	extraEnv := a.runtimeEnvLocked()
 
 	activeIdx := a.activeIdx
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. claudecode workDir/mode mutex 保护，覆盖所有读取点，测试覆盖。

PR #1017, opened 2026-05-16
fix(instance_lock): wait for killed process to exit so --force can re-acquire the flock

kevinWangSheng:
## Summary

\`KillExistingInstance\` in \`cmd/cc-connect/instance_lock.go\` queued SIGKILL via \`proc.Kill()\` and returned immediately. The inline comment even acknowledged the problem:

\`\`\`go
// Wait a moment for the process to exit
// Note: we can't use proc.Wait() as we're not the parent
return true
\`\`\`

…but there was no wait. SIGKILL is async — the signal is queued, and the kernel reaps the process some time later. A process's file locks (\`flock\` from \`syscall.Flock\`) aren't released until that reap happens. So the \`cc-connect --force\` flow:

1. \`KillExistingInstance(configPath)\` signals the old process and returns \`true\`.
2. \`main.go:168\` calls \`AcquireInstanceLock\` on the very next line.
3. The dying process still holds the flock for a few milliseconds.
4. \`AcquireInstanceLock\` fails with \`another cc-connect instance is already running (PID xxxx)\` — exactly the error \`--force\` was supposed to clear.

The race window is small (microseconds to milliseconds on a fast host) but reliably trips on slower hardware, CI, or contended systems. The user sees \`--force\` apparently \"do nothing.\"

## Fix

Since we are not the parent of the killed process, \`proc.Wait()\` isn't available. Instead, poll \`proc.Signal(syscall.Signal(0))\` after \`proc.Kill()\` until it fails with ESRCH (meaning the process is gone). Bound the poll by a 5 s timeout with a 25 ms interval — short enough that the user-visible delay on a normal kill is negligible.

## Tests

\`TestKillExistingInstance_WaitsForExit\` uses the standard \"binary re-execs itself as a helper\" pattern:

- \`TestMain\` checks for \`INSTANCE_LOCK_HELPER\` env var; when set, the binary acquires the lock at the path and parks on \`select{}\`. SIGKILL terminates it and releases the flock.
- The test spawns the helper, waits for the PID file to appear, calls \`KillExistingInstance\`, then \`AcquireInstanceLock\` *immediately* — which must succeed.

With the production fix the second acquire is deterministic. Without the fix the same code path is timing-dependent; on slower systems the test would flake matching the user-visible \`--force\` bug. The behavioral test passes locally in ~5 seconds (mostly the helper spawn cost).

## Test plan

- [x] \`go test -tags no_web -count=1 -run TestKillExistingInstance_WaitsForExit ./cmd/cc-connect/\`
- [x] \`go test -tags no_web -count=1 ./cmd/cc-connect/\`
- [x] \`go vet -tags no_web ./cmd/cc-connect/\`

### DIFF
--- a/cmd/cc-connect/instance_lock.go
+++ b/cmd/cc-connect/instance_lock.go
@@ -7,7 +7,21 @@ import (
 	"os"
 	"path/filepath"
 	"syscall"
+	"time"
 )
+
+// killWaitTimeout caps how long KillExistingInstance polls for the
+// SIGKILLed process to actually be reaped by the kernel. SIGKILL is
+// async, and the file lock isn't released until the process truly
+// exits — so without this wait, `cc-connect --force` would race the
+// kernel and trip "another instance running" against the lock that
+// the dying process still holds.
+const killWaitTimeout = 5 * time.Second
+
+// killWaitInterval is the poll interval between Signal(0) probes
+// while waiting for the killed process to exit. Short enough to
+// minimise the user-visible delay on a clean kill.
+const killWaitInterval = 25 * time.Millisecond
 
 // InstanceLock provides a file-based exclusive lock to prevent multiple
 // cc-connect instances with the same config from running simultaneously.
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. `--force` flag now properly waits for the killed process to be reaped by the kernel before returning (polls via `Signal(0)` up to 5s), preventing the misleading "another instance running" error. `proc.Kill()` is async; the old code didn't wait. No breaking changes. Approving.

Issue #972, opened 2026-05-13
Feishu: all messages permanently ignored as 'old message after restart' after every daemon restart (v1.3.3-beta.2)


zhangyanbo2007:
## Bug Description

After every daemon restart, cc-connect permanently ignores ALL Feishu messages, logging `"feishu: ignoring old message after restart"` for every incoming message — even messages sent **minutes after the restart**. This makes the Feishu bot completely unresponsive after any restart.

## Reproduction Steps

1. Start cc-connect with a Feishu platform: `cc-connect --force`
2. Wait for WebSocket connection to establish
3. Send any message (e.g. `/list`) from Feishu chat
4. Observe in logs: `feishu: ignoring old message after restart`
5. The message is never processed — `/list` returns nothing, no responses are sent
6. All subsequent messages are also ignored — the bug is permanent, not temporary

## Expected Behavior

Messages sent AFTER the daemon restart should be processed normally. Only truly duplicate/stale messages from before the restart should be filtered.

## Environment

- cc-connect v1.3.3-beta.2 (commit f6a1a794)
- Platform: Feishu
- OS: Linux

## Attempted Workarounds (all failed)

- Setting `past_id_tracking = False` in session JSON — daemon ignores this field, filtering persists
- Waiting 10+ seconds after restart before sending messages — still filtered
- Multiple restart cycles — bug persists on every restart
- Setting `past_id_tracking = False` before daemon startup (so it reads False from disk) — still ignored

## Log Evidence

```
time=2026-05-13T15:05:51 level=INFO msg="cc-connect is running"
2026/05/13 15:05:51 [Info] [connected to wss://msg-frontier.feishu.cn...]

# Message sent 2 minutes AFTER restart:
time=2026-05-13T15:07:51 level=DEBUG msg="feishu: ignoring old message after restart" create_time=1778655883913

# Message sent 3 minutes AFTER restart — still ignored:
time=2026-05-13T15:08:19 level=DEBUG msg="feishu: ignoring old message after restart" create_time=1778655912439

# Messages sent 7+ minutes after restart — still ignored:
time=2026-05-13T15:13:30 level=DEBUG msg="feishu: ignoring old message after restart" create_time=1778656222589
```

## Analysis

The `past_id_tracking` mechanism appears to permanently filter all incoming messages after startup, regardless of when the messages were actually created. The JSON field `past_id_tracking` does not control this behavior — the daemon's internal filtering logic is separate and cannot be disabled externally.

[...]

PR #1020, opened 2026-05-16
fix(core): clear AgentSessionID on /stop and resume failure to prevent recycling loop

87hujih:
## Summary

After `/stop`, the stale `AgentSessionID` remained on the `Session` object. When the next message arrived, `getOrCreateInteractiveStateWith` passed the dead session ID to `agent.StartSession`, which either failed or produced a new ID that `CompareAndSetAgentSessionID` silently rejected (the old non-empty ID was still stored). This created a **recycling loop** where every message tore down and re-created the agent, losing context.

The same stale-ID problem also existed in the interactive resume fallback path: when `agent.StartSession(ctx, staleID)` failed and fell back to a fresh session, the old ID was never cleared, so `CompareAndSetAgentSessionID` could not write the new ID. The relay path (line 12640) already handled this correctly; the interactive path did not.

## Fix

Two changes in `core/engine.go`:

1. **`cmdStop` (line 7942)**: After stopping the interactive session, look up the `Session` via `commandContext` and call `SetAgentSessionID("", "")` + `Save()`. This prevents the stale ID from causing a recycling loop on the next message.

2. **`getOrCreateInteractiveStateWith` (line 2884)**: When `agent.StartSession(ctx, staleID)` fails and the code falls back to `agent.StartSession(ctx, "")`, clear the stale `AgentSessionID` first via `session.SetAgentSessionID("", agent.Name())` + `sessions.Save()`. This matches the relay fallback pattern at line 12640 and ensures `CompareAndSetAgentSessionID` can write the new ID.

## Tests

- **`TestCmdStop_ClearsAgentSessionID`**: Seeds a live interactive state with `AgentSessionID="agent-1"`, calls `cmdStop`, asserts `AgentSessionID` is empty afterward.
- **`TestResumeFallback_ClearsStaleSessionID`**: Uses a custom `startSessionFn` that fails on resume (non-empty session ID) and succeeds on fresh. Asserts the stale ID is cleared and replaced with the new ID.
- Extended `controllableAgent` test stub with `startSessionFn` field (backward compatible — existing tests pass unchanged).

## Test plan

- [x] `go build ./...`
- [x] `go test ./core/ -run TestCmdStop_ClearsAgentSessionID -v`
- [x] `go test ./core/ -run TestResumeFallback_ClearsStaleSessionID -v`
- [x] `go test ./core/ -run "TestSession|TestCleanup" -v` (existing tests, no regression)

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2938,6 +2938,10 @@ func (e *Engine) getOrCreateInteractiveStateWith(sessionKey string, p Platform, 
 			slog.Error("session resume failed, falling back to fresh session",
 				"session_key", sessionKey, "failed_session_id", startSessionID,
 				"error", err, "elapsed", startElapsed)
+			// Clear the stale session ID so CompareAndSetAgentSessionID can
+			// write the new ID, matching the relay fallback at line 12640.
+			session.SetAgentSessionID("", agent.Name())
+			sessions.Save()
 			startAt = time.Now()
 			agentSession, err = agent.StartSession(e.ctx, "")
 			startElapsed = time.Since(startAt)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. /stop 和 resume 失败时清除 AgentSessionID，防止 recycling loop，测试覆盖。

Tensor-0:
Cross-linking a field report that confirms this bug in practice: https://github.com/chenhg5/cc-connect/issues/1091#issuecomment-4528219608

Hit exactly the recycling loop described here — killed the Claude process, stale AgentSessionID persisted in session file, resume failed silently, all subsequent messages dropped. Deleting the session file was the only workaround.

The fix described in this PR (clear AgentSessionID on resume failure) would have prevented the cascade into #972 and #1091.

Related: #972, #1091

PR #1023, opened 2026-05-16
fix(core/relay): unblock HandleRelay when agent goes silent past relay timeout

kevinWangSheng:
## Summary

\`Engine.HandleRelay\` in \`core/engine.go\` reads agent events with:

\`\`\`go
for event := range agentSession.Events() {
    switch event.Type { ... }
    if ctx.Err() != nil {
        go e.drainRelaySession(...)
        return relayPartialResponseOrError(...)
    }
}
\`\`\`

The \`ctx.Err()\` check fires only *after* a new event arrives. If the agent stops emitting events — stuck tool call, hung subprocess, slow upstream LLM, waiting on user input — the range loop parks on the receive past the configured relay timeout (\`relayTimeout\` defaults to 120 s). The relay caller (the other bot) hangs indefinitely; the timeout becomes effectively unbounded as long as the agent isn't talking.

The existing tests work around this by explicitly sending an extra event after the timeout to unblock the for-range. The inline comment in \`TestHandleRelay_TimeoutWithoutTextReturnsContextError\` even says \"One event to unblock HandleRelay's for-range, one for the drain goroutine.\" That workaround is not available in production.

## Fix

Replace the for-range with a select that watches both the events channel and \`ctx.Done()\`. When ctx times out while the agent is silent, spawn the same \`drainRelaySession\` goroutine that already handles the mid-event timeout path, and return through the existing \`relayPartialResponseOrError\` helper. No public API change.

## Tests

\`TestHandleRelay_UnblocksWhenAgentGoesSilent\` emits a single text event then stops sending — the agent silence the production bug relies on. The relay timeout is 20 ms and the test asserts \`HandleRelay\` returns within 2 s.

- With the production fix: \`HandleRelay\` observes \`ctx.Done()\` from the select and returns the partial response immediately.
- Without the fix: stash-reverting \`core/engine.go\` and running this test trips \`HandleRelay did not return after relay timeout while agent was silent\` (verified — see PR description).

The three pre-existing relay tests (\`ReturnsPartialOnTimeout\`, \`TimeoutWithoutTextReturnsContextError\`, \`ResumeFailureFallsBackToFreshSession\`) all continue to pass — the select form is strictly more permissive than the for-range.

## Test plan

- [x] \`go test -tags no_web -count=1 -run TestHandleRelay ./core/\`
- [x] Full \`./core/\` test pass (pre-existing macOS \`TestProcessInteractiveEvents_*\` and \`TestResolveLocalDirPath_AcceptsSubdir\` failures reproduce on plain \`main\` — unrelated)

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -12662,7 +12662,40 @@ func (e *Engine) HandleRelay(ctx context.Context, fromProject, chatID, message s
 	}
 
 	var textParts []string
-	for event := range agentSession.Events() {
+	events := agentSession.Events()
+	for {
+		// Use select so a timed-out relay context unblocks even when the
+		// agent has gone silent. Without this, `for event := range
+		// events` would park here past the relay deadline if the agent
+		// is stuck on a tool call, waiting on user input, or otherwise
+		// not emitting events — the relay caller would hang well beyond
+		// the configured relay timeout.
+		var (
+			event Event
+			ok    bool
+		)
+		select {
+		case event, ok = <-events:
+			if !ok {
+				// Event channel closed without EventResult.
+				agentSession.Close()
+				if ctx.Err() != nil {
+					return relayPartialResponseOrError(ctx.Err(), textParts, fromProject, e.name)
+				}
+				if len(textParts) > 0 {
+					return strings.Join(textParts, ""), nil
+				}
+				return "", fmt.Errorf("relay: agent process exited without response")
+			}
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. relay timeout 时使用 select 解除阻塞，agent silent 情况下不再挂起，测试覆盖。

PR #993, opened 2026-05-14
fix(core): route /reasoning through commandContext for multi-workspace

kevinWangSheng:
## Summary

`cmdReasoning` resolved its `ReasoningEffortSwitcher` off `e.agent` and persisted session state through `e.sessions` directly. Under multi-workspace mode the active agent and `SessionManager` live on the per-workspace binding, so `/reasoning <effort>` ran against the wrong agent and saved the cleared `agent_session_id` into the wrong store — visually claiming the switch succeeded while the workspace session kept its old effort.

`cmdModel` already addresses this via `switchModelOnAgent` (now on main). This PR aligns `cmdReasoning` with the same pattern: take `agent`, `sessions`, and `interactiveKey` from `e.commandContext(p, msg)` and use them in place of the engine-global accessors.

## Context

This supersedes the relevant half of #209 (`cmdReasoning` + `cmdModel`). The `cmdModel` portion of that PR was made obsolete by main's `switchModelOnAgent` refactor, so #209 is being closed. This PR carries only the `cmdReasoning` fix and adds a regression test mirroring the existing `TestCmdModel_MultiWorkspaceUsesWorkspaceAgentAndSessions` coverage.

## Test plan

- [x] `go test ./core/ -run "CmdReasoning|CmdModel" -v` — all existing tests still green, plus new `TestCmdReasoning_MultiWorkspaceUsesWorkspaceAgentAndSessions` passes
- [x] `go test -race ./core/ -run "CmdReasoning|CmdModel"` — no races
- [x] `go vet ./core/...` clean
- [x] `go build -tags no_web ./...` clean
- [ ] Manual: in multi-workspace mode, `/reasoning high` in a workspace-bound chat updates the workspace agent, leaves the global agent's effort untouched, and clears only the workspace `agent_session_id`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1025, opened 2026-05-16
fix(engine): cmdMode routes through commandContext in multi-workspace mode

kevinWangSheng:
### Problem

`cmdMode` (`core/engine.go`) reads `e.agent.(ModeSwitcher)` directly and, when the live session can not absorb the change, cleans up interactive state via `e.interactiveKeyForSessionKey(msg.SessionKey)`.

In multi-workspace mode, the `/mode` command is then applied to the **global** agent rather than the workspace-bound agent that actually serves that channel:

- `switcher.SetMode(target)` mutates global agent state — the persistent mode on the workspace agent is unchanged, so after `/new` or idle reset the mode silently reverts.
- `cleanupInteractiveState` is called with the session-key-derived global key, not the workspace-prefixed interactive key that the workspace path actually uses.

`cmdModel` already resolves both via `e.commandContext(p, msg)`; `cmdReasoning` has the same bug shape and is being fixed in #993.

### Fix

Route `cmdMode` through `e.commandContext(p, msg)` and use the returned `agent` and `interactiveKey`. No behavior change in single-workspace mode (`commandContext` returns `(e.agent, e.sessions, msg.SessionKey, nil)` when `!e.multiWorkspace`).

### Test

Adds `TestCmdMode_MultiWorkspaceUsesWorkspaceAgent` modeled on `TestCmdModel_MultiWorkspaceUsesWorkspaceAgentAndSessions`. Verified the test fails on unfixed `core/engine.go` (stash-revert) and passes with the fix.

The remaining macOS test failures in `core/` (`TestProcessInteractiveEvents_*ReplyFooter*`, `TestResolveLocalDirPath_AcceptsSubdir`) are pre-existing on `main` and tied to HOME-tilde rendering and `/var` ↔ `/private/var` symlinks — unrelated to this change.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -7745,7 +7745,13 @@ func (e *Engine) cmdReasoning(p Platform, msg *Message, args []string) {
 }
 
 func (e *Engine) cmdMode(p Platform, msg *Message, args []string) {
-	switcher, ok := e.agent.(ModeSwitcher)
+	agent, _, interactiveKey, err := e.commandContext(p, msg)
+	if err != nil {
+		e.reply(p, msg.ReplyCtx, e.i18n.Tf(MsgWsResolutionError, err))
+		return
+	}
+
+	switcher, ok := agent.(ModeSwitcher)
 	if !ok {
 		e.reply(p, msg.ReplyCtx, e.i18n.T(MsgModeNotSupported))
 		return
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. cmdMode 在 multi-workspace 模式下正确路由到 workspace agent，测试覆盖。

PR #1026, opened 2026-05-17
fix(engine): cmdProvider routes through commandContext in multi-workspace mode

kevinWangSheng:
### Problem

`cmdProvider` and the `switchProvider` helper in `core/engine.go` resolve the agent and session manager from `e.agent` / `e.sessions` directly, and use `e.interactiveKeyForSessionKey(msg.SessionKey)` for cleanup.

In multi-workspace mode, each channel binding has its own agent + session manager (resolved via `commandContextWithWorkspace`). The current code path therefore:

- mutates the **global** agent's active provider on `/provider switch <name>` instead of the workspace-bound agent — the workspace's subsequent prompts keep using the wrong provider, and the global agent is silently mutated;
- on `/provider clear`, calls `ClearHistory` / `SetAgentSessionID("")` on the global session manager rather than the workspace one;
- cleans up the global interactive key while the workspace's interactive state (keyed `workspace:sessionKey`) is left dangling.

This is the same routing bug already fixed for `cmdModel` and just-merged for `cmdMode` (#1025); `cmdReasoning` has it open in #993.

### Fix

`cmdProvider` now resolves `(agent, sessions, interactiveKey)` via `e.commandContext(p, msg)`. `switchProvider` takes `sessions` and `interactiveKey` as parameters instead of touching globals. No behavior change in single-workspace mode, since `commandContext` returns `(e.agent, e.sessions, msg.SessionKey, nil)` when `!e.multiWorkspace`.

`cmdProviderAdd` / `cmdProviderRemove` already receive `switcher` as a parameter, so they pick up the workspace-bound switcher automatically.

### Test

Adds `TestCmdProvider_MultiWorkspaceUsesWorkspaceAgentAndSessions` modeled on `TestCmdModel_MultiWorkspaceUsesWorkspaceAgentAndSessions`. Verified the new test fails on unfixed `core/engine.go` (stash-revert: `workspace agent active provider = openai, want azure`) and passes with the fix.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -8218,7 +8218,12 @@ func (e *Engine) cmdAllow(p Platform, msg *Message, args []string) {
 }
 
 func (e *Engine) cmdProvider(p Platform, msg *Message, args []string) {
-	switcher, ok := e.agent.(ProviderSwitcher)
+	agent, sessions, interactiveKey, err := e.commandContext(p, msg)
+	if err != nil {
+		e.reply(p, msg.ReplyCtx, e.i18n.Tf(MsgWsResolutionError, err))
+		return
+	}
+	switcher, ok := agent.(ProviderSwitcher)
 	if !ok {
 		e.reply(p, msg.ReplyCtx, e.i18n.T(MsgProviderNotSupported))
 		return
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Multi-workspace fix — `cmdProvider` now correctly resolves agent/sessions from `commandContext` for workspace mode, matching the pattern used by `cmdModel`. Test coverage added. No breaking changes. Approving.

PR #1028, opened 2026-05-17
fix(core/tts): EspeakTTS and PicoTTS honor context cancellation

kevinWangSheng:
### Problem

`EspeakTTS.Synthesize` (`core/tts.go:432`) and `PicoTTS.Synthesize` (`core/tts.go:492`) accept a `context.Context` per the `TextToSpeech` interface, but build the subprocess with `exec.Command` instead of `exec.CommandContext`. Their sibling `EdgeTTS.Synthesize` (`core/tts.go:556`) uses `exec.CommandContext(ctx, ...)` for the same interface, so the deviation is unambiguous.

Concrete scenario:

- `Engine.sendTTSReply` (`core/engine.go`) runs `Synthesize` from a goroutine fired after `EventResult` and passes `e.ctx`.
- On engine shutdown `e.cancel()` cancels that context. `EdgeTTS` exits immediately; `EspeakTTS` / `PicoTTS` keep the `espeak` / `pico2wave` child running to completion, blocking shutdown and leaking the goroutine.
- Same hazard for any caller-side per-request timeout context: the timeout silently fails to abort a hung subprocess.

### Fix

Switch both call sites to `exec.CommandContext(ctx, ...)`. No other change required — the local `ctx` parameter is already in scope at both lines.

### Test

Adds `TestEspeakTTS_Synthesize_RespectsContextCancel` and `TestPicoTTS_Synthesize_RespectsContextCancel`. Both write a tempfile shell-script stub that sleeps 30s, point `NewEspeakTTS` / `NewPicoTTS` at it (so the test does not require `espeak` or `pico2wave` to be installed), cancel the context after 100ms, and assert `Synthesize` returns within 5s with a non-nil error.

Verified on the unfixed code with `git stash` the tests fail: Espeak case returns `nil` error (subprocess completed normally after 30s), Pico case takes 31.3s. With the fix both tests pass in ~100ms.

Skipped on Windows where the shell-script stub is not portable.

### DIFF
--- a/core/tts.go
+++ b/core/tts.go
@@ -429,7 +429,7 @@ func (e *EspeakTTS) Synthesize(ctx context.Context, text string, opts TTSSynthes
 
 	// Execute espeak command
 	// Use Output() instead of CombinedOutput() to avoid mixing stderr warnings with audio data
-	cmd := exec.Command(e.Path, args...)
+	cmd := exec.CommandContext(ctx, e.Path, args...)
 	output, err := cmd.Output()
 	if err != nil {
 		return nil, "", fmt.Errorf("espeak: voice=%s text=%q: %w", voice, text, err)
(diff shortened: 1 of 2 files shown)


kevinWangSheng:
Duplicate of #930 (`fix(tts): honor ctx cancellation in EspeakTTS and PicoTTS subprocesses`) which is already approved and clean. Diffs against `core/tts.go` are byte-identical. Closing to avoid splitting review attention.

PR #1030, opened 2026-05-17
feat(core,feishu): per-thread workspace binding with chat-level fallback

0-Captain:
## Summary

Implements per-thread workspace binding for Feishu/Lark group topics, following the `conversation scope -> workspace` direction discussed in #243 (PR #287 added the shared-routing infrastructure this builds on).

When `thread_isolation = true`, each Feishu thread can be bound to its own workspace via `/workspace bind` inside the thread. A thread with no specific binding transparently inherits from the parent chat — so the existing UX of "open a new topic and start chatting immediately" is preserved when the chat is already bound (or auto-bound by convention).

The core fallback is platform-agnostic, so Telegram forum topics also gain chat-level inheritance.

## Behaviour matrix

| Scenario | Workspace resolved |
|---|---|
| Chat bound to A, message in main channel | A |
| Chat bound to A, new thread, no thread binding | A (via fallback) |
| Chat bound to A, \`/workspace bind B\` inside thread | B for that thread; siblings still A |
| No chat binding, \`base_dir/<chat-name>\` exists | Auto-bind at *chat* scope; siblings inherit |
| No binding at any scope | Init flow as before |

## What changes

**\`core/engine.go\`**
- \`parentWorkspaceChannelKey\` / \`parentChannelID\` helpers strip the last colon-separated segment while preserving the platform prefix.
- \`lookupEffectiveWorkspaceBinding\` ascends to parent scopes; a thread-specific binding still shadows the parent.
- \`resolveWorkspace\` does the same ascent for \`ResolveChannelName\`, and records convention auto-bind at the scope that resolved the name so sibling threads inherit.

**\`platform/feishu/feishu.go\`**
- New \`channelKeyForBinding\` helper.
- Set \`core.Message.ChannelKey\` on all 12 dispatch sites (9 in \`dispatchMessage\`, 3 in \`onCardAction\`).
- \`thread_isolation = false\` and non-thread messages: no behaviour change (key equals \`chatID\`).

## Tests

- \`core\`: \`TestParentWorkspaceChannelKey\`, \`TestParentChannelID\`, \`TestResolveWorkspace_ThreadInheritsChatBinding\`, \`TestResolveWorkspace_ThreadOverridesChat\`, \`TestResolveWorkspace_ConventionAutoBindsToParent\`
- \`platform/feishu\`: \`TestChannelKeyForBinding\` (table-driven), plus a \`ChannelKey\` assertion on \`TestLark_ThreadIsolationUsesRootSessionKey\`

\`go test ./core/ ./platform/feishu/\` passes apart from 5 env-dependent pre-existing failures unrelated to this change (HOME-path abbreviation tests + macOS \`/var\` symlink). Confirmed they also fail on clean upstream \`main\` (14abab17) with no patch applied.

## Backward compatibility

- \`thread_isolation = false\`: identical key (\`chatID\`) to before.
- \`thread_isolation = true\` without thread bindings: identical resolution via the parent-scope fallback.
- Existing \`feishu:chatID\` bindings continue to be hit by both exact lookups and parent-scope fallback from threads.

Refs: #243, #287

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1125,6 +1125,8 @@ func (e *Engine) ExecuteCronJob(job *CronJob) error {
 		return e.executeCronShell(effectivePlatform, replyCtx, job)
 	}
 
+	// Expand `/skill ...` prompts to the registered skill's invocation
+	// template (added upstream as the cron skill-expand feature).
 	content := job.Prompt
 	if strings.HasPrefix(content, "/") {
 		parts := strings.Fields(content)
@@ -1136,9 +1138,19 @@ func (e *Engine) ExecuteCronJob(job *CronJob) error {
 		}
 	}
 
+	// Set ChannelKey to the platform-supplied thread-aware key so downstream
+	// code (e.g. buildSenderPrompt's [cc-connect chat_id=...] prompt tag)
+	// preserves thread/topic information instead of falling back to
+	// extractChannelID, which would only show the chat ID.
+	var msgChannelKey string
+	if resolver, ok := targetPlatform.(SessionChannelKeyResolver); ok {
+		msgChannelKey = resolver.ChannelKeyForSession(sessionKey)
+	}
+
 	msg := &Message{
 		SessionKey:   sessionKey,
 		Platform:     platformName,
+		ChannelKey:   msgChannelKey,
 		UserID:       "cron",
 		UserName:     "cron",
 		Content:      content,
@@ -1157,8 +1169,8 @@ func (e *Engine) ExecuteCronJob(job *CronJob) error {
 		// scheduled inside a topic can find a thread-scoped binding. Fall back
 		// to extractChannelID (chat-only) for platforms with no thread concept.
 		channelID := ""
-		if resolver, ok := targetPlatform.(CronChannelKeyResolver); ok {
-			channelID = resolver.ChannelKeyForCronSession(sessionKey)
+		if resolver, ok := targetPlatform.(SessionChannelKeyResolver); ok {
+			channelID = resolver.ChannelKeyForSession(sessionKey)
 		}
 		if channelID == "" {
 			channelID = extractChannelID(sessionKey)
@@ -13328,6 +13340,35 @@ func extractWorkspaceChannelKey(sessionKey string) string {
 	return workspaceChannelKey(extractPlatformName(sessionKey), extractChannelID(sessionKey))
 }
 
+// workspaceChannelKeyForSessionKey returns the thread-aware workspace binding
+// channel key for a sessionKey. It first asks the owning platform (matched by
+// the session_key prefix) via the optional SessionChannelKeyResolver
+// interface, so platforms that encode thread/topic information into their
+// session keys (Feishu thread_isolation, Discord forum threads, etc.) can
+// preserve that scope. If no platform responds, falls back to
+// extractWorkspaceChannelKey, which is thread-blind.
+//
+// This is used by callers that have a session_key but no live Message —
+// ExecuteCronJob, card renderers like renderListCard, etc.
+func (e *Engine) workspaceChannelKeyForSessionKey(sessionKey string) string {
+	platformName := extractPlatformName(sessionKey)
+	if platformName == "" {
+		return extractWorkspaceChannelKey(sessionKey)
+	}
+	for _, p := range e.platforms {
+		if p.Name() != platformName {
+			continue
+		}
+		if resolver, ok := p.(SessionChannelKeyResolver); ok {
+			if id := resolver.ChannelKeyForSession(sessionKey); id != "" {
+				return workspaceChannelKey(platformName, id)
+			}
+		}
+		break
+	}
+	return extractWorkspaceChannelKey(sessionKey)
+}
+
 // effectiveChannelID returns the channel identifier from a Message.
 // It prefers the platform-provided ChannelKey (e.g. "chatID:threadID" for forum topics)
 // and falls back to parsing the session key.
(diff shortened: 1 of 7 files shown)


chenhg5 (approved):
LGTM. per-thread workspace binding with chat-level fallback，层次化设计清晰，测试覆盖。

0-Captain:
**Follow-up commits**: while testing this PR end-to-end I caught two regressions / gaps that I've now pushed as additional commits on top of the original `feat:` commit.

### 1. `3ad5a7a2` — restore chat-scope default for `/workspace bind`, add `-t` opt-in, fix cron thread-aware lookup

The original commit silently changed two defaults that break real-world topic-mode group usage:

- **`/workspace bind <name>` regression.** Pre-PR, this bound the entire chat (single binding covers all topics). After the PR, `effectiveWorkspaceChannelKey(msg)` returns `chatID:rootID` in topic-mode groups, so the bind ended up scoped to the **current** topic only. Symptom: opening a brand-new topic afterwards yields "this channel is not bound" — and cron jobs scheduled in that group all hit the unbound state and fall back to the global agent. Fix: `bind` / `init` / `route` / `unbind` default to chat scope again; new `-t` / `--topic` flag opts into thread scope for users who genuinely want per-topic isolation. Status (`/workspace`) and `list` are unchanged since the hierarchical fallback in `lookupEffectiveWorkspaceBinding` already does the right thing for reads.
- **Cron path was thread-blind.** `ExecuteCronJob`'s multi-workspace branch resolves the binding via `extractChannelID(sessionKey)`, which only returns the chat ID. Thread-scoped bindings created by `/workspace bind -t` were invisible to cron — the cron's own session_key carries the thread info but `extractChannelID` drops it. Added an optional `core.CronChannelKeyResolver` interface; Feishu implements it by delegating to the existing `channelKeyForBinding` helper, keeping cron and message-flow binding-key derivation in sync. Other platforms unchanged (fall back to `extractChannelID`).

`/workspace shared` deliberately stays chat-only — cross-project + per-topic granularity is too niche to justify the API surface.

### 2. `0c496cfd` — regression test for `/workspace init <local-dir>` wording

Locks down that `init` with a local directory replies with the "bound" success wording, not "Repository cloned successfully" (no clone actually happened).

### Verification

- Built locally and swapped the daemon binary in a real environment (Feishu topic-mode groups, two cron jobs at 06:00 and 08:00).
- Verified `/workspace bind <name>` now writes a chat-scope key; new topics inherit it; `-t` produces a thread-scope key that shadows the chat default; `-t` in a DM is rejected with `MsgWsTopicFlagInvalid`.
- Manually triggered both cron jobs after setting `mode: bypassPermissions` and confirmed they spawn agents in the correct workspace via the binding (no `work_dir` override needed).
- 22 new tests in `core/multi_workspace_test.go` (11 covering bind/init/route/unbind + 5 languages × 2 flag forms) and 10 sub-tests in `platform/feishu/platform_test.go` for `ChannelKeyForCronSession`.

0-Captain:
**Another follow-up commit** (`b42ce5db`) on top of `3ad5a7a2`, found while user-testing in real Feishu topic-mode groups. Two issues, both same root cause class as the cron-thread-aware bug, plus one UX improvement.

### 1. `sessionContextForKey` had the same thread-blind bug

When I fixed the cron path in `3ad5a7a2`, I scoped the `CronChannelKeyResolver` interface only to `ExecuteCronJob`. But 15+ other call sites (`renderListCard`, `renderHistoryCard`, `renderCurrentCard`, `/switch`, `/name`, `/delete`, `interactiveKeyForSessionKeyLocked`, background lookups, etc.) all go through `sessionContextForKey(sessionKey)` — which used `extractWorkspaceChannelKey`, which strips thread/topic suffixes the same way `extractChannelID` does. So thread-scoped workspace bindings were invisible to any of those paths.

Concrete repro: `/workspace bind -t /path/A` in topic T → `/list` in topic T → returned sessions from the chat-scope binding's directory, not `/path/A`.

Generalized the interface: `CronChannelKeyResolver` → `SessionChannelKeyResolver` (method `ChannelKeyForCronSession` → `ChannelKeyForSession`). Added an engine helper `workspaceChannelKeyForSessionKey(sessionKey)` that finds the owning platform by session_key prefix and asks it for a thread-aware key, falling back to `extractWorkspaceChannelKey` for platforms that don't implement it. Switched `sessionContextForKey`, `interactiveKeyForSessionKeyLocked`, and `ExecuteCronJob` to use the helper.

Added `TestSessionContextForKey_UsesPlatformThreadAwareChannelKey` to lock this in.

### 2. One topic per cron run

In topic-mode Feishu groups, each cron firing was creating its own topic per message (start notice + agent preview + final reply → 3 separate topics). User wanted them grouped.

Implemented `CronReplyTargetResolver` on Feishu (the interface has been there but had no implementation): when `thread_isolation` is on, post `📌 <title>` as a top-level message — which is a new topic in the group — capture its message_id, and return a sessionKey/replyContext targeting that topic. Subsequent cron messages thread inside. One run = one topic.

Returns `ErrNotSupported` when `thread_isolation` is off, so behavior elsewhere unchanged. Covered by `TestResolveCronReplyTarget_GatedByThreadIsolation` (5 early-return cases).

### 3. Cron synthetic Message now sets `ChannelKey`

`ExecuteCronJob` constructed the synthetic Message without `ChannelKey`, so `buildSenderPrompt`'s `[cc-connect chat_id=...]` tag would fall back to chat-only ID for cron triggers. Now we populate it via `SessionChannelKeyResolver` so the agent's context tag preserves thread/topic info — same path used for the workspace lookup.

### Verification

- Built locally, replaced the daemon binary, ran `/workspace init` + `/list` + cron triggers in a topic-mode Feishu group.
- Confirmed `/list` now returns sessions for the thread-scoped binding (was previously returning the chat-scope one).
- Confirmed cron output groups under a single `📌 <title>` topic per run (was previously spawning 3 separate topics per run).
- Lark API query confirmed all messages in `omt_<thread_id>` share the same `thread_id`.

### Reviewer note

CI build may fail on darwin due to `daemon.CheckLinger` (added in #965) being defined only under `//go:build linux` but referenced unconditionally in `cmd/cc-connect/daemon.go`. Unrelated to this branch; the same `go build ./...` fails on darwin against current `main` too.

PR #1031, opened 2026-05-17
fix(feishu): correct receive_id type when replying to bot menu events

jojofran:
Bot menu click events (bot_menu_v6) supply the user's open_id as the chat context, but createMessage and SendPreviewStart both hardcoded ReceiveIdTypeChatId, causing "invalid receive_id" (code 230001) on every reply.

Fix by adding a receiveIDType field to replyContext, set to "open_id" by the menu event handler, so the send path can select the correct ReceiveIdType without relying on ID prefix heuristics.


chenhg5 (approved):
LGTM. Feishu bot menu 事件回复使用正确的 receive_id_type，测试覆盖。

PR #1032, opened 2026-05-17
fix(dingtalk): set replyContext.isGroup in richText and image branches

kevinWangSheng:
### Problem

In `platform/dingtalk/dingtalk.go`, the `replyContext` literals built for **richText** (`onMessage`, around line 269) and **image** (`handleImageMessage`, around line 470) messages omit the `isGroup` field. The sibling **text** (line 302) and **audio** (lines 370 and 394) branches already set:

```go
isGroup: data.ConversationType == "2",
```

`sendProactiveMessage` routes on `rc.isGroup`:

- `true` + non-empty conversationId → `/v1.0/robot/groupMessages/send` (group)
- otherwise + non-empty senderStaffId → `/v1.0/robot/oToMessages/batchSend` (1:1 DM to sender)

So a group-chat richText reply (e.g. a user quoting another message — DingTalk encodes this as `Msgtype: "richText"`) or a group-chat image, when the proactive path fires (sessionWebhook expired, cron, `cc-connect send`, etc.), is misrouted to a **1:1 DM with the original sender** instead of being posted back to the group. Wrong chat, often fails when the bot has no DM channel with that user, and leaks the response outside the group's audit context.

### Fix

Add `isGroup: data.ConversationType == "2"` to the two `replyContext` literals so all four message-type branches stay symmetric.

### Test

Adds `TestOnMessage_RichTextInGroup_SetsIsGroup` (group case) and `TestOnMessage_RichTextDirect_KeepsIsGroupFalse` (1:1 case) by invoking `onMessage` with a synthesized `chatbot.BotCallbackDataModel` and asserting the captured `replyContext.isGroup`. Verified on the unfixed code with `git stash` the group test fails (`isGroup=false, want true`) and the direct test still passes. With the fix both pass.

(Image branch shares the same trivial change but requires HTTP plumbing to exercise end-to-end; the richText tests pin the bug class and the image fix is the symmetric 1-line addition.)

### DIFF
--- a/platform/dingtalk/dingtalk.go
+++ b/platform/dingtalk/dingtalk.go
@@ -270,6 +270,7 @@ func (p *Platform) onMessage(data *chatbot.BotCallbackDataModel, richText *richT
 				sessionWebhook: data.SessionWebhook,
 				conversationId: data.ConversationId,
 				senderStaffId:  data.SenderStaffId,
+				isGroup:        data.ConversationType == "2",
 			},
 		}
 		p.handler(p, msg)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM (CI pending). DingTalk richText/image 分支设置 isGroup，逻辑正确。

PR #1034, opened 2026-05-17
feat(daemon): harden service-file env capture + add EnvDiscoverer plugin hook

ssc806:
## Summary

Two related improvements to the `cc-connect daemon install` path so that operators can store API keys / tokens in `config.toml` using `\${ENV}` placeholders and have them work with the installed daemon, without those values being world-readable on disk. Plus a small extension point so plugins can contribute additional env vars without daemon needing to know about them.

### Commit 1 — \`feat(daemon): harden service-file writes and capture config.toml \${ENV}\`

1. **\`captureConfigEnvPlaceholders\`**: during \`daemon install\`, scan the target \`config.toml\` for \`\${VAR_NAME}\` placeholders. For every match set in the current process environment, copy the value into \`EnvExtra\` so the rendered launchd plist / systemd unit / Windows task script carries it. Without this, the daemon process starts with empty strings for placeholder values and fails to authenticate to any platform.
2. **Tighten service-file permissions to 0600.** Both the \`WriteFile\` mode and an explicit \`os.Chmod\` after write — the Chmod is required for the reinstall path because \`WriteFile\` only applies perm on create, so a 0644 file left by a previous cc-connect version would keep its old permissions. Applies to launchd plist, systemd unit, and Windows \`.ps1\` script.
3. **EnvExtra hardening in every renderer**:
   - drop entries whose key fails POSIX-identifier validation
   - drop entries whose value is empty
   - launchd: XML-escape both keys and values; reserve \`PATH\` / \`CC_LOG_FILE\` / \`CC_LOG_MAX_SIZE\` so EnvExtra can't override template-owned keys
   - systemd: backslash- and quote-escape values for the \`Environment=\"K=V\"\` form per systemd.exec(5)
4. **Opt-out**: \`--no-capture-secrets\` flag and \`CC_DAEMON_NO_CAPTURE_SECRETS=1\` env var for operators who'd rather inject secrets via keychain / \`secret-tool\` / \`EnvironmentFile=\`.
5. **\`runSystemctl\` as a var** so per-platform tests can stub it without needing a real \`systemctl\` on the host.

### Commit 2 — \`feat(daemon): add EnvDiscoverer plugin hook\`

Generalises \`captureDaemonEnv()\` with a small plugin registry:

\`\`\`go
type EnvDiscoverer func() (map[string]string, error)
daemon.RegisterEnvDiscoverer(d)
daemon.ResetEnvDiscoverers()    // tests only
\`\`\`

\`Resolve()\` invokes every registered discoverer (unless \`NoCaptureSecrets\` is set) and merges the result into \`EnvExtra\` after the proxy-key capture and config.toml placeholder scan. Errors are logged but never fail install.

Use case: lets a downstream / plugin agent contribute extra env vars the service file should carry, without daemon needing to know about that agent. Without the hook, the only ways to extend capture are (a) pre-populate \`EnvExtra\` in every caller, or (b) edit \`daemon/manager.go\` to import the new package — which couples daemon to that package.

## Test plan

- [x] \`go test ./daemon/... ./cmd/cc-connect/...\` — all green
- [x] systemd / launchd / windows renderer tests cover env-name validation, empty-value drop, and value escaping
- [x] \`TestSystemdInstall_TightensExistingUnitFrom0644\` and analogous launchd/windows tests cover the reinstall upgrade path
- [x] \`TestCaptureDaemonEnv_*\` cover the discoverer registry: merge semantics, NoCaptureSecrets opt-out, invalid-key dropping, error tolerance
- [ ] Manual: reinstall on a host that has a 0644 plist/unit from a previous version, verify perms tighten to 0600

### DIFF
--- a/cmd/cc-connect/daemon.go
+++ b/cmd/cc-connect/daemon.go
@@ -121,11 +121,20 @@ func parseDaemonInstallArgs(args []string) (daemon.Config, bool, error) {
 	var cfg daemon.Config
 	var force bool
 
+	// Env-based opt-out: CC_DAEMON_NO_CAPTURE_SECRETS=1 / true / yes / on
+	// triggers --no-capture-secrets without the CLI flag, for CI / container
+	// scenarios where the global env is the right configuration surface.
+	if isTruthyEnv(os.Getenv("CC_DAEMON_NO_CAPTURE_SECRETS")) {
+		cfg.NoCaptureSecrets = true
+	}
+
 	for i := 0; i < len(args); i++ {
 		arg := args[i]
 		switch {
 		case arg == "--force":
 			force = true
+		case arg == "--no-capture-secrets":
+			cfg.NoCaptureSecrets = true
 		case arg == "--log-file":
 			value, next, err := daemonInstallFlagValue(args, i, "--log-file")
 			if err != nil {
@@ -187,6 +196,16 @@ func daemonInstallFlagValue(args []string, index int, flagName string) (string, 
 		return "", index, fmt.Errorf("missing value for %s", flagName)
 	}
 	return args[next], next, nil
+}
+
+// isTruthyEnv accepts the conventional opt-in values for boolean env vars.
+// Anything else, including "0" / "false" / "" / unset, is treated as false.
+func isTruthyEnv(v string) bool {
+	switch strings.ToLower(strings.TrimSpace(v)) {
+	case "1", "true", "yes", "on":
+		return true
+	}
+	return false
 }
 
 // ── uninstall ───────────────────────────────────────────────
@@ -431,6 +450,9 @@ Install flags:
   --log-max-size N      Max log file size in MB (default: 10)
   --work-dir DIR        Directory containing config.toml (default: current dir)
   --force               Overwrite existing installation
+  --no-capture-secrets  Do not capture config.toml ${ENV} placeholders into
+                        the service file. Also enabled by setting
+                        CC_DAEMON_NO_CAPTURE_SECRETS=1 in the environment.
 
 Restart flags:
   --force               Kill existing process before restarting
--- a/cmd/cc-connect/daemon_test.go
+++ b/cmd/cc-connect/daemon_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"os"
 	"path/filepath"
 	"testing"
 )
@@ -29,6 +30,73 @@ func TestParseDaemonInstallArgs_ConfigEqualsFormSetsWorkDir(t *testing.T) {
 	want := filepath.Clean("/tmp/example")
 	if cfg.WorkDir != want {
 		t.Fatalf("cfg.WorkDir = %q, want %q", cfg.WorkDir, want)
+	}
+}
+
+func TestParseDaemonInstallArgs_NoCaptureSecretsFlag(t *testing.T) {
+	os.Unsetenv("CC_DAEMON_NO_CAPTURE_SECRETS")
+
+	cfg, _, err := parseDaemonInstallArgs([]string{"--no-capture-secrets"})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if !cfg.NoCaptureSecrets {
+		t.Fatal("flag should set NoCaptureSecrets=true")
+	}
+
+	cfg2, _, err := parseDaemonInstallArgs(nil)
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if cfg2.NoCaptureSecrets {
+		t.Fatal("default must be false when flag and env are unset")
+	}
+}
+
+func TestParseDaemonInstallArgs_NoCaptureSecretsEnv(t *testing.T) {
+	for _, v := range []string{"1", "true", "TRUE", "yes", "on"} {
+		t.Run("truthy="+v, func(t *testing.T) {
+			t.Setenv("CC_DAEMON_NO_CAPTURE_SECRETS", v)
+			cfg, _, err := parseDaemonInstallArgs(nil)
+			if err != nil {
+				t.Fatalf("parse: %v", err)
+			}
+			if !cfg.NoCaptureSecrets {
+				t.Fatalf("env=%q should opt out", v)
+			}
+		})
+	}
+	for _, v := range []string{"0", "false", "", "no", "off"} {
+		t.Run("falsy="+v, func(t *testing.T) {
+			t.Setenv("CC_DAEMON_NO_CAPTURE_SECRETS", v)
+			cfg, _, err := parseDaemonInstallArgs(nil)
+			if err != nil {
+				t.Fatalf("parse: %v", err)
+			}
+			if cfg.NoCaptureSecrets {
+				t.Fatalf("env=%q should NOT opt out", v)
+			}
+		})
+	}
+}
+
+func TestParseDaemonInstallArgs_NoCaptureSecretsFlagAndEnvCombine(t *testing.T) {
+	// OR semantics: env=truthy + flag=present → still true.
+	t.Setenv("CC_DAEMON_NO_CAPTURE_SECRETS", "1")
+	cfg, _, err := parseDaemonInstallArgs([]string{"--no-capture-secrets", "--force"})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if !cfg.NoCaptureSecrets {
+		t.Fatal("flag+env both should leave NoCaptureSecrets=true")
+	}
+	// env=truthy without flag → still true.
+	cfg2, _, err := parseDaemonInstallArgs([]string{"--force"})
+	if err != nil {
+		t.Fatalf("parse: %v", err)
+	}
+	if !cfg2.NoCaptureSecrets {
+		t.Fatal("env=1 alone should opt out")
 	}
 }
 
--- a/daemon/env_extension.go
+++ b/daemon/env_extension.go
@@ -1,0 +1,65 @@
+package daemon
+
+import (
+	"regexp"
+	"sync"
+)
+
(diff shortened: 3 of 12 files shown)


chenhg5 (approved):
LGTM. Daemon service hardening — adds `EnvDiscoverer` plugin hook for install-time env-var discovery, XML-escapes user-supplied paths in service files, and fixes env var capture. New plugin interface is opt-in, no breaking changes to existing daemon behavior. Approving.

chenhg5 (approved):
Conclusion: Approve

Overall assessment: Re-review confirms PR #1034 is well-architected security hardening with comprehensive test coverage of the threat model. The 0600 permission tightening, EnvExtra validation, value escaping per platform syntax, and the EnvDiscoverer plugin hook together close a real secret-exfiltration risk without breaking existing flows.

Review scope:
- Reviewed `daemon/env_extension.go` (new, 65 lines): EnvDiscoverer registry with sync.RWMutex protection.
- Reviewed `daemon/manager.go` (+125): `captureDaemonEnv`, `captureConfigEnvPlaceholders`, `captureConfigEnvPlaceholdersInValue`, `captureConfigEnvPlaceholdersInString`; `Config.NoCaptureSecrets` field added; `Resolve()` wires everything together.
- Reviewed `daemon/launchd.go` (+80): `renderEnvExtraPlist` with XML escaping, `xmlEscape` helper, `templateOwnedEnvKeys` reservation set (PATH, CC_LOG_FILE, CC_LOG_MAX_SIZE).
- Reviewed `daemon/systemd.go` (+53): `escapeSystemdEnvValue` (per systemd.exec(5) escaping), `runSystemctl` made a var for test stubbing.
- Reviewed `daemon/windows.go` (+22): same hardening pattern, value passthrough.
- Reviewed `cmd/cc-connect/daemon.go` (+22): `--no-capture-secrets` flag + `CC_DAEMON_NO_CAPTURE_SECRETS=1` env var opt-out.
- Cross-checked test coverage: 11 tests in `daemon/manager_test.go`, 11 in `daemon/launchd_test.go`, 6 in `daemon/systemd_test.go`, 3 in `daemon/windows_test.go`. All critical security paths covered.

✅ What looks good:
- **0600 permission hardening** (launchd.go:56, systemd.go:69, windows.go:60): Both `os.WriteFile` with 0600 mode AND explicit `os.Chmod` after write. The Chmod is required because WriteFile only applies perm on create — without it, reinstalls of pre-existing 0644 files from earlier cc-connect versions would keep their old permissions. Two new tests pin this: `TestInstallLaunchd_TightensExistingPlistFrom0644` and `TestSystemdInstall_TightensExistingUnitFrom0644`.
- **Value escaping per platform syntax**:
  - launchd: `xmlEscape` uses `encoding/xml.EscapeText` which handles `&`, `<`, `>`, `"`, `'`. `TestBuildPlist_EscapesXMLInValue` covers `a"b<c>d&e'f` round-trip.
  - systemd: `escapeSystemdEnvValue` handles `\\`, `"`, `\n`, `\r`, `\t` per systemd.exec(5). `TestEscapeSystemdEnvValue` covers 6 cases (plain, quote, backslash, newline, tab, return) plus `TestBuildUnit_EscapesEnvValue` covers end-to-end `a"b\c` → `a\"b\\c`.
  - windows: passthrough to PowerShell `$env:KEY=VALUE` syntax which doesn't need escaping for `=`-separated key=value pairs in normal cases.
- **Invalid env name validation**: `isValidEnvName` uses `^[A-Za-z_][A-Za-z0-9_]*$` regex — POSIX env-name standard. Three renderers all drop invalid names with slog.Warn (launchd.go:260, systemd.go:201, windows.go:198). Belt-and-suspenders defense against typos or malicious discoverer output that could break service-file parsing or inject syntax.
- **Template-owned key reservation**: `templateOwnedEnvKeys` (launchd.go:248) reserves PATH, CC_LOG_FILE, CC_LOG_MAX_SIZE — EnvExtra can't override template-rendered values. `TestBuildPlist_SkipsEmptyValuesAndTemplateOwnedKeys` pins this. Important because PATH must be the user's PATH (not their config), and the CC_LOG_* vars are needed for log rotation to work.
- **Empty value dropping**: All three renderers skip env vars with empty values (silently). `TestCaptureDaemonEnv_DropsEmptyDiscoveredValues` covers the discoverer side. Prevents `<key>FOO</key><string></string>` clutter in the plist.
- **NoCaptureSecrets opt-out**: `--no-capture-secrets` CLI flag + `CC_DAEMON_NO_CAPTURE_SECRETS=1` env var. When true, `captureDaemonEnv` returns only proxy vars, `captureConfigEnvPlaceholders` is skipped, and `Resolve` propagates this to discoverer iteration. `TestResolve_PropagatesNoCaptureSecretsToDiscoverers` + `TestResolveNoCaptureSecretsSkipsConfigEnvPlaceholders` cover this.
- **TOML comment safety**: `TestResolveIgnoresCommentedConfigEnvPlaceholders` verifies `# access_token = "${CAPTURE_CONFIG_COMMENTED}"` does NOT trigger env capture. Works because `toml.Unmarshal` strips comments before `captureConfigEnvPlaceholdersInValue` walks the parsed map. Critical — without this, users with documented examples in config.toml would have unrelated env vars captured.
- **Reflective placeholder scan**: `captureConfigEnvPlaceholdersInValue` walks `reflect.Value` recursively (Interface/Pointer, String, Slice/Array, Map). This catches placeholders nested in `[[projects.platforms.options]]` arrays, agent config, etc. Important for real configs that aren't flat.
- **EnvDiscoverer plugin hook**: `RegisterEnvDiscoverer` with sync.RWMutex (read lock for snapshot, write lock for register). Errors logged at WARN but never fail install (matches comment at manager.go:160-161). `TestCaptureDaemonEnv_LogsDiscovererErrorButContinues` pins error tolerance. `ResetEnvDiscoverers` for test isolation. Discoverers run in registration order; later wins on key collision — explicit, documented, tested.
- **runSystemctl as var**: allows test stubbing without requiring real systemctl. `TestSystemdInstall_TightensExistingUnitFrom0644` exercises this.
- Tests: `go test -tags no_web -count=1 ./daemon/...` PASS (0.009s — fast, no real systemd/launchd); `go vet -tags no_web ./daemon/... ./cmd/cc-connect/...` clean.
- CI run 26071471457: lint 1m44s / unit 3m43s / smoke 23s / regression 26s / performance 49s — 5/5 green on SHA cca3d698.
- Mergeable: clean.

- **Linger check stub**: cca3d698 (latest) adds `daemon/linger_other.go` for non-Linux platforms — fixes a Linux-specific build dependency that would have broken macOS/Windows/BSD builds.
- No other commits since prior review (assuming last review was on this exact branch state — the head SHA matches the latest commit cca3d698 which was the trigger for stale_needs_rereview).

🔵 Optional (not blocking):
1. **P3 Discoverer panic safety**: `EnvDiscoverer` is a `func() (map[string]string, error)`. If a discoverer panics, the whole install crashes. Consider wrapping each discoverer call in a `defer recover()` so one buggy plugin doesn't brick daemon install. Defensive — currently no discoverers are shipped so this is theoretical.
2. **P3 NoCaptureSecrets discoverer visibility**: When `NoCaptureSecrets=true`, the comment says "skip any extension discoverers". But the discoverers are unaware they're being skipped — they don't even get called. This is correct (saves CPU) but means discoverers can't log "I would have captured X but NoCaptureSecrets blocked me". Could be improved by passing a context flag, but it's purely informational.
3. **P3 systemd `Environment=` line length**: systemd imposes a hard limit (~8KB) on Environment= values. If a discoverer returns a value with thousands of characters, the unit file could fail to parse. Consider length-checking values in `escapeSystemdEnvValue` (e.g., truncate with warning if >4096 chars). Not a real risk for normal use.
4. **P3 xmlEscape vs launchd DTD**: `xml.EscapeText` escapes `&`, `<`, `>`, `"`, `'`. The launchd DTD requires valid XML, so this is correct. But `xml.EscapeText` doesn't escape control characters (`\x00`-`\x1F` except `\t`, `\n`, `\r`) which XML 1.0 forbids. If a discoverer returns a value with a literal null byte, the plist could fail to parse. Edge case.
5. **P3 Concurrent discoverer registration**: `RegisterEnvDiscoverer` is typically called from `init()` so concurrency isn't a real issue, but if a plugin ever registers at runtime (e.g., from a config reload), the snapshot in `snapshotEnvDiscoverers` could miss newly-registered discoverers. Not a problem with the documented init()-only usage.

Testing / Risk:
- Verified locally: `go test -tags no_web -count=1 ./daemon/...` PASS (0.009s), `go vet` clean.
- CI run 26071471457 (5/5 green on SHA cca3d698).
- Risk: Low. The change is defense-in-depth: existing installs continue to work; new installs and reinstalls get tighter perms. No API surface change for `Config` callers except the new `NoCaptureSecrets` field which defaults to `false` (capture enabled — preserves backward compat).
- Security: Improves the existing security posture. The two real risks being closed are (a) 0644 service files leaking captured secrets to other users on multi-user systems, (b) malformed EnvExtra names/values breaking service-file parsing or injecting syntax. Both addressed with belt-and-suspenders (regex + per-platform escaping).
- Backwards compatible: Existing service files left at 0644 from prior versions get tightened on next install (via the Chmod fallback). No data loss. Existing callers passing `EnvExtra` directly continue to work — invalid names are dropped with warning rather than aborting install.
- Multi-platform: launchd (macOS), systemd (Linux), schtasks (Windows) all hardened with platform-appropriate escaping. Linger check stub covers non-Linux systemd (FreeBSD, etc.).

Next step: OK to merge. The 0600 permission tightening + per-platform value escaping + invalid env name validation + TOML comment safety + opt-out mechanism together close real secret-exfiltration risks. After merge, consider whether to backport the 0600 hardening to older cc-connect versions via release notes ("on next daemon install, your service file perms will tighten from 0644 to 0600").

PR #1035, opened 2026-05-17
feat(core): support passthrough slash commands per project

ssc806:
## Summary

Add a per-project \`passthrough_commands\` config field that marks slash commands which should bypass cc-connect's own handling and be forwarded to the agent as plain user input. Useful when an agent (Claude Code, Codex, etc.) defines its own slash commands and the operator wants users to invoke them directly from the chat platform without cc-connect intercepting.

\`\`\`toml
[[projects]]
passthrough_commands = [\"loop\", \"babysit-prs\"]   # specific commands
# OR
passthrough_commands = [\"*\"]                     # forward every /cmd
\`\`\`

## Semantics

- Resolution mirrors \`disabled_commands\`: builtin aliases collapse to their canonical IDs; unknown names are normalized (case + dashes / underscores) and kept verbatim so agent-side names work too.
- **Disabled and admin-only gates run BEFORE passthrough.** A passthrough match only fires when the user is already authorized to invoke the command; otherwise the disabled / admin-required reply wins. This preserves existing security guarantees.
- Passthrough applies to builtin, custom-command, skill, and unknown command paths in \`handleCommand\`. The handler returns false so the agent receives the raw \`/cmd args\` string as a normal message.
- Each passthrough decision is logged as \`audit: command_passthrough\` with user_id / platform / project / command for traceability.

## Test plan

- [x] \`go test ./core/ -run Passthrough\` — 12 tests, all green
- [x] Wildcard \`*\` covers builtin + custom + skill + unknown
- [x] Name normalization (case, dash↔underscore) tested for builtin aliases and agent-side names
- [x] Gate precedence verified: disabled command + passthrough → disabled wins; privileged command + passthrough + non-admin → admin-required wins

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -388,6 +388,7 @@ func main() {
 		if len(proj.DisabledCommands) > 0 {
 			engine.SetDisabledCommands(proj.DisabledCommands)
 		}
+		engine.SetPassthroughCommands(proj.PassthroughCommands)
 
 		// Wire admin allowlist for privileged commands
 		engine.SetAdminFrom(proj.AdminFrom)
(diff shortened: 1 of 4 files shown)


chenhg5 (approved):
LGTM. passthrough slash commands per project，测试覆盖。

PR #1036, opened 2026-05-17
feat(core): gate /commands addexec and /cron addexec behind admin_from

ssc806:
## Summary

Both \`/commands\` and \`/cron\` currently have a single coarse-grained privilege check: the \`cmdID\` is or is not in \`privilegedCommands\`. Neither is listed there, so all subcommands run for any user — including \`addexec\`, which **registers a custom command that shells out**, or schedules one that shells out on a cron. A non-admin user who can send slash commands can effectively install arbitrary shell payloads for anyone else to trigger.

(Note: \`executeCustomCommand\` already gates Exec-bearing custom commands on \`admin_from\` _after_ they've been registered, so executing such a payload also requires admin — but the **registration** itself shouldn't be open to non-admins either. This PR closes the registration side.)

## Change

Introduce \`isPrivilegedCommandInvocation(cmdID, args)\` which:

- falls through to the existing \`privilegedCommands\` map for the static cases (\`/shell\`, \`/show\`, \`/dir\`, \`/restart\`, \`/upgrade\`, \`/web\`, \`/diff\`)
- additionally returns true for \`/commands addexec ...\` and \`/cron addexec ...\` (subcommand names matched via \`matchSubCommand\` so prefix forms like \`/cron addex\` and case variants are covered)
- returns false for every other subcommand of \`/commands\` and \`/cron\` (\`list\`, \`add\`, \`del\`, etc. stay non-privileged)

\`handleCommand\` swaps the \`privilegedCommands[cmdID]\` check for the new helper. No other privilege paths change.

## Test plan

- [x] \`go test ./core/ -run Privileged\` and friends — 7 tests, all green
- [x] Static privileged list still works (none of \`/shell\`, \`/show\`, etc. behaviour changed)
- [x] \`/commands list\` and other non-addexec subcommands stay non-privileged
- [x] \`/commands addexec\` and \`/cron addexec\` (incl. prefix forms / case variants) require admin

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -840,6 +840,42 @@ var privilegedCommands = map[string]bool{
 	"diff":    true,
 }
 
+// isPrivilegedCommandInvocation extends the privilegedCommands map to
+// also gate specific destructive subcommands. Currently:
+//
+//   - /commands addexec ...    — registers a custom shell-exec command
+//   - /cron    addexec ...     — schedules a recurring shell-exec
+//
+// Both effectively create new admin-only commands at runtime; if a
+// non-admin can call addexec, they can install arbitrary shell commands
+// for any future user to trigger. Sibling subcommands (list, add, del,
+// etc.) remain non-privileged.
+//
+// Returns true when the cmdID itself is in privilegedCommands, or when
+// the (cmdID, args[0]) pair matches one of the explicitly-gated
+// subcommands above.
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. /commands addexec 和 /cron addexec 需要 admin_from 权限，安全改进，测试覆盖。

PR #818, opened 2026-04-29
feat(platform/max): webhook delivery mode + deployment docs

alekseybv:
## Summary

- Adds webhook delivery as an alternative to long-polling for the MAX platform. MAX is throttling long-poll to 2 RPS from 2026-05-11, so higher-traffic bots need webhook mode. Long-poll remains the default; this is opt-in via the new `webhook_url` option (no breaking changes).
- Adds typing/read presence indicators (`typing_on` ticker while the agent is working, fire-and-forget `mark_seen` on incoming messages).
- Adds `docs/max-webhook.md` with three deployment topologies (VPS + reverse proxy, home + VPS SSH-tunnel, long-poll), copy-paste nginx / Caddy / systemd samples, and troubleshooting.

## What's in each commit

1. `feat(platform/max): webhook mode, typing presence, long-poll tuning` — core webhook implementation (`webhook_url`/`webhook_listen` options, `startWebhook`, `webhookHandler`, `subscribe`/`unsubscribe`), typing presence via `sendChatAction`, and long-poll timeout tuning (`pollTimeout` 20→30, `httpTimeout` 35s→90s) to prevent retry storms when the MAX backend is slow.
2. `feat(platform/max): add webhook_path and webhook_secret options + deployment docs` — `webhook_path` (host multiple bots on one domain), `webhook_secret` (optional shared secret checked via `X-Webhook-Secret` header or `?s=` query, compared with `crypto/subtle.ConstantTimeCompare`), and the `docs/max-webhook.md` deployment guide.
3. `chore: ignore manual cc-connect.bak backups and runtime *.toml.lock files` — tiny `.gitignore` update.

## Backwards compatibility

If `webhook_url` is empty (the default), the platform behaves exactly as before — long-poll only. No config change is required for existing installations.

## Test plan

- [x] `go test ./...` — all packages pass
- [x] `go vet ./...` — clean
- [x] 7 new unit tests cover `webhook_path` normalization (empty / no-slash / with-slash / multi-segment), `webhook_secret` accept via header, accept via query, mismatch (401), missing (401), and wrong HTTP method (405)
- [x] Live-tested in webhook mode: bot bound :8090, MAX accepted `POST /subscriptions`, real messages delivered through nginx → SSH-tunnel → bot → Claude → reply
- [ ] CI: `go test -race ./...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1038, opened 2026-05-17
fix(max): webhook handler goroutine inherits platform cancellation context

kevinWangSheng:
### Problem

`platform/max/max.go` `webhookHandler` dispatches incoming updates from a goroutine, and a comment claims the goroutine uses the platform's own cancellation context so `Stop()` can short-circuit in-flight handler work:

```go
go func() {
    p.mu.RLock()
    ctx := context.Background()
    if p.cancel != nil {
        // Use the platform's own cancellation context so a Stop() also
        // short-circuits in-flight handler work.
        ctx2, cancel := context.WithCancel(ctx)
        _ = cancel                       // discarded
        ctx = ctx2
    }
    p.mu.RUnlock()
    p.handleUpdate(ctx, &upd)            // detached from p.cancel
}()
```

The platform only stores `p.cancel` (a `CancelFunc`), not the parent context — `ctx2` is derived from `context.Background()`, not from anything `p.cancel` controls, and `cancel` is immediately thrown away. The comment vs. behavior mismatch means in webhook delivery mode:

- `Stop()` calls `p.cancel()` to abort polling and webhook re-subscription, but every in-flight webhook handler (which fans out into `p.handleUpdate` → `p.handler` → engine work, agent calls, attachment downloads, HTTP retries) keeps running with an uncancelable context.
- Process shutdown returns immediately while handlers keep posting to MAX using credentials from a "stopped" platform — bypassing the documented Stop semantics other transports honor.

The long-poll path (`pollLoop` → `handleUpdate` at line 926) does propagate the correct parent ctx via `go p.pollLoop(ctx)` in `Start`, so this is webhook-mode-only and was introduced together with webhook delivery in PR #818.

### Fix

Store the parent context on the platform alongside `cancel` in `Start`, then derive the per-request work context from it in `webhookHandler`. Fall back to `context.Background()` only when `Start` was never called (e.g. test paths that call `webhookHandler` directly).

No behavior change in poll mode; webhook mode now honors `Stop()`.

### Test

Adds `TestStartStoresLifecycleContext` asserting:
- `p.ctx` is nil before `Start`.
- After `Start`, `p.ctx` is non-nil and not Done.
- After `Stop`, `p.ctx` is Done within 1s.

This pins the field wiring (without the fix the test fails to compile because `p.ctx` does not exist). End-to-end cancellation propagation through the webhook goroutine + `sendChatAction` HTTP path is structurally guaranteed once the parent context flows through — verifiable from the 9-line diff.

Existing `TestPollLoopStopsOnCtxCancel` and the suite of `TestWebhookHandler*` tests continue to pass.

### DIFF
--- a/platform/max/max.go
+++ b/platform/max/max.go
@@ -70,6 +70,7 @@ type Platform struct {
 
 	mu           sync.RWMutex
 	handler      core.MessageHandler
+	ctx          context.Context
 	cancel       context.CancelFunc
 	stopping     bool
 	client       *http.Client // general API calls — httpTimeout
@@ -160,6 +161,7 @@ func (p *Platform) Start(handler core.MessageHandler) error {
 	p.handler = handler
 
 	ctx, cancel := context.WithCancel(context.Background())
+	p.ctx = ctx
 	p.cancel = cancel
 
 	// Verify token at startup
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Max webhook handler 使用 platform ctx 而非 Background，测试覆盖。

kevinWangSheng:
Duplicate of #991 which is already approved and addresses the same async-webhook context bug. The two diffs are functionally equivalent — #991 abstracts the parent-context selection into a small `webhookCtx()` helper while this PR inlines the same logic. Closing in favor of #991's slightly cleaner factoring. Sorry for the duplicate work (auto-generated).

PR #1043, opened 2026-05-18
Fix setup commands for bound workspaces

YanxingLiu:
## Summary
- Resolve setup memory writes through the message command context in multi-workspace mode.
- Keep the existing engine-agent setup helper for non-workspace callers.
- Add a regression test that verifies `/bind setup` writes the bound workspace `AGENTS.md` instead of the global memory file.

## Test Plan
- `go test ./core -run TestCmdBindSetup_WritesBoundWorkspaceInstructions -count=1`
- `go test ./core -run 'TestSetupMemoryFile|TestCmdCronSetup|TestCmdBindSetup|TestCommandContextWithWorkspace' -count=1`\n- `go test ./core -count=1`\n- `go test ./...`\n- `git diff --check`\n\n## Notes\n- `golangci-lint` was not available in PATH locally.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -11356,7 +11356,7 @@ func (e *Engine) cmdCronMute(p Platform, msg *Message, args []string, mute bool)
 }
 
 func (e *Engine) cmdCronSetup(p Platform, msg *Message) {
-	result, baseName, err := e.setupMemoryFile()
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Properly resolves agent from commandContext in multi-workspace mode for /setup and /bind setup commands. No API changes, pure bug fix. CI green.

PR #1049, opened 2026-05-18
fix(shell): show usage when /shell is called with no arguments

happyTonakai:
## Summary

When a user types `/shell` (or `/sh`, `/exec`, `/run`) without any arguments, `cmdShell` should show a usage message. Instead, the bare command string (e.g. `/shell`) gets passed to `sh -c` as-is, producing a confusing shell error:

```
fish: Unknown command: /shell
```

The root cause is in the prefix-stripping loop which only matches `/shell ` (with trailing space):

```go
for _, prefix := range []string{"/shell ", "/sh ", "/exec ", "/run "} {
    if strings.HasPrefix(strings.ToLower(raw), prefix) {
        shellCmd = raw[len(prefix):]
        break
    }
}
```

When the input is `/shell` with no trailing space, no prefix matches, `shellCmd` stays as `"/shell"`, `TrimSpace` is a no-op, and the `shellCmd == ""` check is bypassed.

## Fix

Add a bare-command check after the prefix loop and `TrimSpace`:

```go
if lower == "/shell" || lower == "/sh" || lower == "/exec" || lower == "/run" {
    shellCmd = ""
}
```

This catches the no-argument case and routes it to the existing usage-message path.

## Test

`TestCmdShell_BareCommand_ShowsUsage` calls `cmdShell(p, msg, "/shell")` (no trailing space) and asserts the reply contains `"Usage"`.

## Test plan

- [x] `go test ./core/ -run TestCmdShell -v` — all 6 shell tests pass.
- [x] `go build ./...` — builds successfully.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -5981,13 +5981,19 @@ func updaterFor(p Platform) MessageUpdater {
 func (e *Engine) cmdShell(p Platform, msg *Message, raw string) {
 	// Strip the command prefix ("/shell ", "/sh ", "/exec ", "/run ")
 	shellCmd := raw
+	lower := strings.ToLower(strings.TrimSpace(raw))
 	for _, prefix := range []string{"/shell ", "/sh ", "/exec ", "/run "} {
-		if strings.HasPrefix(strings.ToLower(raw), prefix) {
+		if strings.HasPrefix(lower, prefix) {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Clean fix for bare /shell command — shows usage instead of silent empty execution. CI green.

PR #1053, opened 2026-05-19
feat(footer): show git branch in reply footer

happyTonakai:
## Summary

When the working directory is inside a git repo, append the current branch name in parentheses after the directory path in the reply footer, e.g. `~/project (main)`. Non-git directories are unchanged.

## Changes

- Extract `resolveWorkDir()` from `replyFooterWorkDir()` (renamed) to separate path resolution from path compaction
- Add `replyFooterGitBranch()` that runs `git rev-parse --abbrev-ref HEAD` with a 1-second timeout
- Truncate branch names longer than 30 runes with ellipsis, correctly handling multi-byte characters
- Add unit tests covering git branch detection, truncation, and non-git fallback

## Test plan

- [x] `go test ./core/ -run TestBuildReplyFooter -v` — all footer tests pass
- [x] `go build ./...` — builds successfully
- [x] Verified in git directory: footer shows branch name
- [x] Verified in non-git directory: footer unchanged

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


chenhg5 (approved):
LGTM. CI green, clean change, well-structured.

chenhg5 (approved):
Conclusion: Approve

Overall assessment: Re-review confirms PR #1053 is still mergeable and the git-branch footer enhancement integrates cleanly with current main. The 51d22e13 commit resolves the lint failure from c5707a4 by removing the now-unused `replyFooterWorkDir` wrapper. The branch-name truncation + subprocess git-call approach is well-tested.

Review scope:
- Reviewed `core/engine.go` (+69/-12): new `replyFooterGitBranch` helper, renamed `replyFooterWorkDir` → `resolveWorkDir` (3 call sites updated to compose compact + branch inline).
- Reviewed `core/engine_test.go` (+87/-5): 6 new test cases covering git repo, non-git dir, empty dir, custom branch, detached HEAD, and branch-name truncation.
- Cross-checked that `workspaceDir` parameter is server-side config (config.toml `[projects.agent.options] workspace_dir`, workspace binding paths, cron job WorkDir) — not attacker-controlled.

✅ What looks good:
- `replyFooterGitBranch` at core/engine.go:6160 wraps `git rev-parse --abbrev-ref HEAD` with `context.WithTimeout(ctx, time.Second)`. Worst case is 1 second per footer call — acceptable for the per-turn footer (which already does `os.Stat`, etc.). Falls back to empty string on any error (not a git repo, timeout, binary missing), so no footer is added rather than showing a stale or wrong branch.
- 30-rune truncation with `…` suffix prevents excessive footer length. The rune-aware truncation (using `range branch` to walk runes, not bytes) is correct — UTF-8 branch names like `功能/开发` truncate at rune boundaries, not mid-codepoint.
- Three call sites (`buildReplyFooter` line 5773, `composeRichStatusFooter` line 5847, `buildClaudeStatusLineFooter` line 6278) use the same `resolveWorkDir + compactReplyFooterPath + replyFooterGitBranch + truncate + format` composition. Consistent UX across all three footer paths.
- 6 new tests cover the algorithm: git repo returns "main", non-git dir returns empty, empty dir returns empty, custom branch names work, detached HEAD returns "HEAD" (correct — git rev-parse --abbrev-ref returns HEAD when not on a branch), long branch name gets truncated with `…`.
- `TestBuildReplyFooter_GitBranchTruncation` is the most valuable — exercises the end-to-end footer generation with a real git repo (`t.TempDir` + `git init` + `git commit --allow-empty` + `git checkout -b feature/JIRA-1234-...`) and verifies the truncated `…` actually appears in the rendered footer.
- `51d22e13` cleanly removes the now-unused `replyFooterWorkDir` wrapper, addressing the lint failure from `c5707a4`.
- Tests: `go test -tags no_web -count=1 ./core/...` PASS (42.4s).
- `go vet -tags no_web ./core/...` clean.
- CI run 27012033860: lint 2m21s / unit 3m45s / smoke 31s / regression 30s / performance 48s — 5/5 green on SHA 51d22e13.
- Mergeable: clean.

- Lint failure from the original c5707a4 commit (unused `replyFooterWorkDir` after rename) is fixed in 51d22e13.
- The branch was rebased onto current main via merge commit febb448f, picking up PR #1204 (feishu rich card refresh), PR #1210 (opencode --agent flag), PR #1202 (feishu media), PR #1200 (codex request_user_input), PR #1201 (cron manual trigger), PR #1188 (dingtalk @mention), and ~10 other main commits.

🔵 Optional (not blocking):
1. **P2 Code duplication**: The 15-line block (`compactReplyFooterPath + replyFooterGitBranch + 30-rune truncation + format`) is repeated verbatim in 3 call sites (lines 5773-5785, 5847-5859, 6278-6291). Could be extracted to a helper like `buildReplyFooterDirLine(rawDir string) string`. Minor maintenance burden — adding a 4th call site or changing the format requires touching 3 places. Consider a follow-up refactor.
2. **P3 Performance**: `git rev-parse --abbrev-ref HEAD` runs as a subprocess on every footer render (which happens per reply, not per token). For users with large repos or slow disks, this could add latency. The 1-second timeout caps worst case, but in practice git rev-parse is fast (≤10ms). Consider caching the branch name per workspaceDir+HEAD-SHA if performance becomes a concern. Current usage is fine.
3. **P3 Detached HEAD UX**: When the workspace is in detached HEAD state (e.g., user checked out a specific commit), the footer will show `dir (HEAD)` instead of a branch name. This is technically correct but may confuse users. Consider showing `dir (@commit)` or hiding the branch in detached state. UX nit.
4. **P3 Subprocess startup cost**: `exec.CommandContext` allocates a new process for each footer render. For high-throughput reply paths (e.g., cron firing many short messages), this could matter. Not a real concern at current scale but worth noting if `git rev-parse` shows up in profiles.
5. **P3 Test coverage gap**: Tests cover git-init, custom branch, and detached HEAD, but don't cover the timeout case (what if `git rev-parse` takes > 1 second?). Hard to test without slow git mocks. Manual verification recommended.

Testing / Risk:
- Verified locally: `go test -tags no_web -count=1 ./core/...` PASS (42.4s), `go vet -tags no_web ./core/...` clean.
- Targeted tests `ReplyFooterGitBranch` + `BuildReplyFooter_GitBranchTruncation` PASS (0.063s).
- CI run 27012033860 (5/5 green on SHA 51d22e13).
- Risk: Low. The change adds an optional `(branch)` suffix to the workdir footer. Falls back gracefully when git is unavailable, dir is not a repo, or rev-parse errors.
- Security: `cmd.Dir = dir` uses server-side config (config.toml workspace_dir, workspace binding paths, cron job WorkDir). Not attacker-controllable from chat messages. No command injection vector — only `git rev-parse --abbrev-ref HEAD` is called, no user input interpolated into args.
- Backwards compatible: footer output gains an optional `(branch)` suffix. Existing footer renderers that don't expect the suffix may need to update regex patterns. Test coverage in the existing `TestProcessInteractiveEvents_SuppressesReplyFooterWhenOnlyWorkDir` doesn't assert against the exact dir format, so it's safe.

Next step: OK to merge. The workdir footer now shows the active git branch (e.g., `/home/user/project (feature/JIRA-1234)`) which is useful for users juggling multiple worktrees or branches. Optional follow-up: refactor the 3-way code duplication into a single helper.

PR #1067, opened 2026-05-20
fix(core): resolve pending permission lookup for cron sessions with composite keys

happyTonakai:
## Summary

Cron new-per-run 模式创建 session 时使用复合 key（`"sessionKey#cron:cronSessionID"`）存储到 `interactiveStates`，但 platform 权限按钮回调携带的是用户之前聊天的纯 `sessionKey`。`handlePendingPermission` 直接用该 key 查找，会将允许/拒绝误路由到用户之前的 session 上，导致 cron session 的权限请求永远无法被用户响应。

## Fix

- 提取 `lookupPending` 辅助函数，统一 pending permission 的加锁查找逻辑。
- 当直接查找未找到 pending 时，fallback 遍历所有以 `"sessionKey#cron:"` 为前缀的 composite key，匹配到正确的 cron session 后再处理权限。

锁安全：fallback 路径先持锁收集匹配的 key，释放锁后再逐个调用 `lookupPending`，避免 Mutex 重入死锁。

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./core/ -count=1`


chenhg5 (approved):
LGTM. Well-tested fix for cron composite key lookup with proper fallback path.

chenhg5 (commented):
## QA Triage Review: PR #1067 (fix(core): resolve pending permission lookup for cron sessions with composite keys)

结论: **ready-to-merge with P2 follow-up** (correct for common case; non-determinism acknowledged).

### What it does
- `handlePendingPermission` now falls back to a composite-key lookup when the plain sessionKey misses.
- Cron `new_per_run` sessions use keys like `key#cron:sid` (composite), but platform permission button callbacks come in with the plain `key`. The new code looks for `key#cron:*` prefixes in `interactiveStates`.
- New `lookupPending` helper extracted for clarity.
- 56 lines of new tests in `core/engine_test.go`.

### Quality
- The fix is targeted and minimal: a 2nd lookup path with a clear comment about WHY (cron composite keys).
- The `goto found` pattern with the bounded loop is unusual but readable.

### Risk
- **Non-determinism for multiple cron runs**: comment says "In practice only one cron run per session should be in the pending state at a time, so this loop is bounded and deterministic for the expected case." If multiple cron runs ARE in pending state simultaneously, the first one in Go map iteration order is picked. Map iteration is intentionally randomized in Go.
- **Composite key format is implicit**: `key#cron:sid` is hardcoded in the prefix check. If cron key format changes elsewhere, this breaks silently. **P2 follow-up**: extract `cronSessionKey(baseKey, sid string) string` to a single place and use it both here and where the keys are created.

### Triage label
- **ready-to-merge** — fixes a real bug, the non-determinism is acknowledged in the comment. Recommend merge; file a P2 follow-up to extract the composite-key format into a shared helper.

— cc-connect/qa-claudecode

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
The cron-fallback fix is correct, lock-safe, and now combined with stale-callback handling from main (issue #826). Without it, platform permission button callbacks for cron new-per-run sessions are silently dropped — a real user-visible correctness bug. The new merge commit 5a36117e resolves the pr-1202 conflict marker and preserves both behaviors without regression.

Review scope:
- `core/engine.go` (+45/-16): extracted `lookupPending` helper, added cron-prefix fallback path, preserved stale-callback drop from main.
- `core/engine_test.go` (+56/-0): new `TestHandlePendingPermission_CronFallback` (cron composite key → plain sessionKey callback routing).
- Total diff: 2 files, +101/-16 net. 5 commits, head 5a36117e.

✅ What looks good:
- **Lock safety**: cron-prefix keys are collected under `interactiveMu`, then released before `lookupPending` re-acquires (the helper takes both `interactiveMu` and `state.mu` internally). This avoids Mutex re-entrance deadlock — the naive implementation that called `lookupPending` while still holding `interactiveMu` would have deadlocked or required sync.Mutex redesign. The current two-phase approach is the right shape.
- **Helper extraction**: `lookupPending(iKey) → (*interactiveState, *pendingPermission)` consolidates the "look up state, then look up pending" sequence. This was duplicated inline before; now the helper has a clear contract (caller must NOT hold `interactiveMu`) documented in the doc comment.
- **Stale callback handling from main (issue #826) preserved**: when no pending request exists for the key AND the message is a permission response, drop silently. Without this, old "Allow"/"Deny" button taps (after session reset, bot restart, or card redelivery) would trigger spurious downstream effects. Critical to keep both behaviors in the same handler.
- **Test isolation**: `TestHandlePendingPermission_CronFallback` builds a clean engine with one cron composite key and one plain-key callback — exercises exactly the bug class the PR is meant to fix. Verifies (a) callback returns true (handled), (b) cron state remains in the map, (c) pending is cleared, (d) `RespondPermission` is called with correct RequestID and behavior. All four assertions pin the contract.
- **Conflict marker resolution**: commit 73e489d cleans up a pre-existing pr-1202 conflict marker in `engine_test.go`. Independent small win.
- **CI run 27247905373**: lint 2m31s / unit 3m44s / smoke 25s / regression 30s / performance 52s — 5/5 green on SHA 5a36117e. Local re-verification: `go build -tags no_web ./...` OK, `TestHandlePendingPermission_CronFallback` PASS, full `TestHandlePendingPermission_*` suite PASS, `go vet -tags no_web ./core/... ./cmd/cc-connect/...` clean.

🟠 Should improve:
- **P2 Multi-cron non-determinism**: when multiple cron new-per-run sessions share the same sessionKey (e.g., two concurrent cron runs on the same project), the fallback iterates `interactiveStates` and uses `goto found` on the first match. Go map iteration order is randomized, so the chosen cron session is non-deterministic across runs. The comment acknowledges this ("In practice only one cron run per session should be in the pending state at a time"). If two cron runs are simultaneously pending, the user could see their Allow/Deny routed to the wrong one. Suggested fix (out of scope for this PR): store `pendingAt time.Time` on `pendingPermission`, sort matched keys by oldest `pendingAt`. Not blocking — but worth tracking as a follow-up since the comment currently excuses a real correctness edge case rather than addressing it.

🔵 Optional:
- **P3 Add regression test for stale-callback handling**: The merge from main (issue #826) added the `slog.Debug("dropping stale permission callback...")` + `return true` path but I don't see a new test pinning it. The existing `TestHandlePendingPermission_Stale*` family may cover it; if not, a focused test would harden the contract.

🟢 Resolved since prior QA review (t-20260606-a138ff on ea77b05a):
- ✅ Conflict marker in engine_test.go resolved (73e489d).
- ✅ Stale permission callback handling (issue #826) merged in (5a36117e).
- ✅ Map iteration non-determinism acknowledged in code comments.

Testing / Risk:
- Verified evidence: targeted test PASS, full Pending/Permission suite green, vet clean, CI 5/5 green.
- Remaining risk: multi-cron simultaneous-pending edge case (P2 above) is the only correctness gap I see. Unlikely in normal use (the system serializes most cron triggers per session) but worth a follow-up.
- Backwards compatibility: no API surface change; `handlePendingPermission` signature unchanged. Cron new-per-run users previously saw silent permission button loss → now their Allow/Deny actually works. Pure correctness improvement.

Next step:
OK to merge. After merge, consider:
1. Add a follow-up issue for the multi-cron simultaneous-pending non-determinism.
2. Track the stale-callback test gap (P3) when convenient.

Refs #826

PR #1074, opened 2026-05-20
feat(codex): prefer Codex model_catalog_json as highest-priority model source

happyTonakai:
## Summary

Codex CLI supports a `model_catalog_json` field in `$CODEX_HOME/config.toml` that points to a JSON file containing a curated list of models. Currently `AvailableModels()` ignores this field entirely, falling through to the provider config, API, cache, or hardcoded fallback. This means users who configure a custom model catalog in their Codex config don't see those models when switching models via the CC-Connect UI.

This PR adds `$CODEX_HOME/config.toml` → `model_catalog_json` as the **highest-priority** model source, so the user's explicitly configured catalog takes precedence over everything else.

## Fix

- Add `readCodexModelCatalog()` that reads `$CODEX_HOME/config.toml`, resolves the `model_catalog_json` path (expand ~, relative to CODEX_HOME), and parses the JSON as the primary model list.
- Extract shared `parseCodexModelsJSON()` to deduplicate JSON parsing logic between `readCodexCachedModels` and `readCodexModelCatalog`.
- Refactor `readCodexCachedModels()` to reuse `resolveCodexHome(nil)`.
- Update `AvailableModels()` priority:
  1. `model_catalog_json` (new, highest)
  2. provider config
  3. `GET /v1/models` API
  4. `models_cache.json`
  5. hardcoded fallback

## Tests

`TestAvailableModels_UsesModelCatalog` verifies that when a valid `$CODEX_HOME/config.toml` with `model_catalog_json` exists, `AvailableModels()` returns the models from that catalog. `TestReadCodexModelCatalog_NoConfigFile` verifies graceful fallback when no config file exists.

## Test plan

- [x] `go test -count=1 -run "TestAvailableModels_UsesModelCatalog|TestReadCodexModelCatalog_NoConfigFile|TestAvailableModels_FallbackToModelsCache" ./agent/codex/`
- [x] `go test -count=1 ./agent/codex/`
- [x] Manual: `/model` command in Feishu now lists models from the custom catalog JSON instead of only hardcoded defaults.

## Notes

`go test ./...` is not fully green due to existing environment and unrelated test issues (e.g. missing external dependencies), not introduced by this PR.

### DIFF
--- a/agent/codex/codex.go
+++ b/agent/codex/codex.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/BurntSushi/toml"
+
 	"github.com/chenhg5/cc-connect/core"
 )
 
@@ -217,6 +219,9 @@ func (a *Agent) configuredModels() []core.ModelOption {
 }
 
 func (a *Agent) AvailableModels(ctx context.Context) []core.ModelOption {
+	if models := readCodexModelCatalog(); len(models) > 0 {
+		return models
+	}
 	if models := a.configuredModels(); len(models) > 0 {
 		return models
 	}
@@ -303,19 +308,23 @@ func (a *Agent) fetchModelsFromAPI(ctx context.Context) []core.ModelOption {
 }
 
 func readCodexCachedModels() []core.ModelOption {
-	codexHome := os.Getenv("CODEX_HOME")
+	codexHome, _ := resolveCodexHome(nil)
 	if codexHome == "" {
-		home, err := os.UserHomeDir()
-		if err != nil {
-			return nil
-		}
-		codexHome = filepath.Join(home, ".codex")
+		return nil
 	}
 	path := filepath.Join(codexHome, "models_cache.json")
 	b, err := os.ReadFile(path)
 	if err != nil {
 		return nil
 	}
+	return parseCodexModelsJSON(b)
+}
+
+
+// parseCodexModelsJSON parses a Codex models JSON file (model_catalog.json
+// or models_cache.json) into a deduplicated, filtered slice of ModelOption.
+// It is shared by readCodexCachedModels and readCodexModelCatalog.
+func parseCodexModelsJSON(data []byte) []core.ModelOption {
 	var payload struct {
 		Models []struct {
 			Slug           string `json:"slug"`
@@ -325,7 +334,7 @@ func readCodexCachedModels() []core.ModelOption {
 			SupportedInAPI bool   `json:"supported_in_api"`
 		} `json:"models"`
 	}
-	if err := json.Unmarshal(b, &payload); err != nil {
+	if err := json.Unmarshal(data, &payload); err != nil {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Clean change, well-structured.

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
This is a substantive rewrite since the 2026-05-20 LGTM (force-pushed; previous review commit was eb033c23, current head is 7a8a97c9, both hashes are different). The new implementation adds `model_catalog_json` as the highest-priority model source and extracts a shared `parseCodexModelsJSON()` helper. All 3 new tests pass, code is clean, behavior is correct.

Review scope:
- agent/codex/codex.go and agent/codex/codex_cache_test.go (2 files / +142 / -7).
- Local `CGO_ENABLED=0 go test -count=1 -run "TestAvailableModels|TestReadCodexModelCatalog" -v ./agent/codex/...` is 3/3 PASS.
- `CGO_ENABLED=0 go vet ./agent/codex/...` is clean.
- Confirmed `resolveCodexHome(nil)` returns `~/.codex` as default (matches the previous hardcoded path), so the refactor preserves behavior for the existing `models_cache.json` case.

✅ What looks good:
- Priority chain is explicit and well-documented: model_catalog_json → provider config → GET /v1/models API → models_cache.json → hardcoded fallback.
- `~` / `~/` / relative-to-CODEX_HOME / absolute path resolution for `model_catalog_json` is the right set of cases to handle, and falls back gracefully with `slog.Debug` on each error path.
- Extracting `parseCodexModelsJSON()` deduplicates the JSON parsing logic between `readCodexCachedModels` and `readCodexModelCatalog` — this is the kind of cleanup that makes the new feature free.
- Test coverage: TestAvailableModels_UsesModelCatalog covers visibility filter + slug-empty-fallback-to-display_name + supported_in_api filter in one case. TestReadCodexModelCatalog_NoConfigFile covers graceful degradation to the hardcoded fallback.

🟠 Should improve:
- P2 (cosmetic, not blocking): there are two consecutive blank lines between `readCodexCachedModels` and the new `parseCodexModelsJSON` (codex.go around the end of the old function). `gofmt` should normally collapse these but apparently didn't after the refactor. Not a correctness issue, just a `gofmt -w` cleanup.

🔵 Optional:
- P3 (test for path expansion): the current tests cover `model_catalog_json` set to a relative path that resolves against `CODEX_HOME`, but don't cover the `~` / `~/` expansion branches. A 3rd sub-test exercising `~/model_catalog.json` and an absolute path would close that gap. Not blocking.

❓ Questions:
- `readCodexModelCatalog` is called on every `AvailableModels()` invocation. The function reads `config.toml` + `model_catalog.json` from disk each time. Is that the intent? In a long-lived cc-connect process with many `AvailableModels()` calls (e.g. model picker UIs, /model command), this re-parses on every call. Caching the parsed result behind a `sync.Once` or a TTL would be a small efficiency win, but the file is small (~100 models) so the cost is bounded. No blocker; just asking if this was intentional.

Testing / Risk:
- Verified evidence: 3/3 new tests pass locally. CI is green per the registry (ci_status=success, mergeable=clean). `go vet` is clean.
- Remaining risk: end-to-end behavior with a real Codex distribution that ships a `model_catalog.json` is not exercised in CI. Manual verification in a Codex-equipped project is the only way to confirm the new priority works in practice. Not a blocker for the v1.4.0 release.

Next step:
- Ready to merge for v1.4.0 Tier 2. The owner can land this if the Codex distribution path is what users are expected to have.

PR #1075, opened 2026-05-20
fix(codex): stream each round's agent_message as EventText instead of buffering

happyTonakai:
## Summary

Codex agent messages from intermediate rounds are buffered in `pendingMsgs` and only flushed as `EventThinking` when the next tool call starts. Only the final round's text is emitted as `EventText`. This means the Feishu card shows intermediate round text only as "thinking" (or hides it entirely when ThinkingMessages is disabled), and only the last round appears as the actual answer.

This PR changes both `session.go` (CLI backend) and `appserver_session.go` (AppServer backend) to emit `EventText` immediately on each `agent_message` item completion, matching Claude Code's streaming behavior where every round's output is appended to the dynamic card.

## Fix

- Remove `pendingMsgs` buffer, `flushPendingAsThinking()`, and `flushPendingAsText()` from both `session.go` and `appserver_session.go`.
- In `handleItemCompleted`, emit `EventText` directly instead of appending to `pendingMsgs`.
- Remove the now-unnecessary `stateMu` lock around the `emit()` call in the `agentMessage` branch of `appserver_session.go` — `emit()` is already concurrency-safe (writes to a buffered channel with select/default).

No public API or behaviour change beyond the streaming improvement.

## Tests

No new tests. Existing tests in `session_test.go` and `appserver_session_test.go` cover the session lifecycle and should continue to pass. The removed code (`pendingMsgs`, `flushPendingAsThinking`, `flushPendingAsText`) had no dedicated tests.

## Test plan

- [x] `go test -count=1 -run "TestAppServerSession_ApplyThreadRuntimeState|TestAppServerSession_HandleRateLimitsUpdatedCachesUsage|TestAppServerSession_HandleThreadTokenUsageUpdatedCachesContextUsage|TestMapAppServerRateLimits_PrefersMultiBucketView" ./agent/codex/`
- [x] `go test -count=1 -run "TestSend_WithImages_PassesImageArgsAndDefaultPrompt|TestSend_HandlesLargeJSONLines|TestClose_ForceKillsProcessGroupAfterGracefulTimeout" ./agent/codex/`
- [x] `go test -count=1 ./agent/codex/`
- [x] Manual: Send a multi-round prompt to Codex via Feishu; each round's agent_message now streams into the card as real-time text instead of only showing the final round.

## Notes

`go test ./...` is not fully green due to existing environment and unrelated test issues, not introduced by this PR.

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -1207,9 +1207,7 @@ func (s *appServerSession) handleItemCompleted(item map[string]any) {
 	case "agentMessage":
 		text, _ := item["text"].(string)
 		if strings.TrimSpace(text) != "" {
-			s.stateMu.Lock()
 			s.emit(core.Event{Type: core.EventText, Content: text})
-			s.stateMu.Unlock()
 		}
 
 	case "commandExecution":
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -505,7 +505,6 @@ func (s *appServerSession) Send(prompt string, images []core.ImageAttachment, fi
 
 	s.stateMu.Lock()
 	s.currentTurn = resp.Turn.ID
-	s.pendingMsgs = s.pendingMsgs[:0]
 	s.stateMu.Unlock()
 
 	return nil
@@ -607,7 +606,6 @@ func (s *appServerSession) handleApprovalRequest(rawID json.RawMessage, method s
 	s.pendingApprovals[requestID] = ch
 	s.approvalsMu.Unlock()
 
-	s.flushPendingAsThinking()
 	s.emit(core.Event{
 		Type:         core.EventPermissionRequest,
 		RequestID:    requestID,
@@ -654,7 +652,6 @@ func (s *appServerSession) handlePermissionsApproval(rawID json.RawMessage, para
 	s.pendingApprovals[requestID] = ch
 	s.approvalsMu.Unlock()
 
-	s.flushPendingAsThinking()
 	s.emit(core.Event{
 		Type:         core.EventPermissionRequest,
 		RequestID:    requestID,
@@ -722,7 +719,6 @@ func (s *appServerSession) handleRequestUserInput(rawID json.RawMessage, paramsR
 	s.pendingApprovals[requestID] = ch
 	s.approvalsMu.Unlock()
 
-	s.flushPendingAsThinking()
 	s.emit(core.Event{
 		Type:         core.EventPermissionRequest,
 		RequestID:    requestID,
@@ -1106,7 +1102,6 @@ func (s *appServerSession) handleNotification(method string, paramsRaw json.RawM
 		if err := json.Unmarshal(paramsRaw, &notif); err == nil {
 			s.stateMu.Lock()
 			s.currentTurn = notif.Turn.ID
-			s.pendingMsgs = s.pendingMsgs[:0]
 			s.stateMu.Unlock()
 			s.storeContextUsage(nil)
 		}
@@ -1172,8 +1167,6 @@ func (s *appServerSession) handleItemStarted(item map[string]any) {
 		return
 	}
 
-	s.flushPendingAsThinking()
-
 	switch itemType {
 	case "commandExecution":
 		command, _ := item["command"].(string)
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Clean change, well-structured.

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
This is a substantive refactor of the codex streaming path. The behavioral change (agent_message now emits EventText immediately instead of being buffered and reclassified as either EventThinking or EventText) is intentional, well-motivated (aligns codex with Claude Code's typewriter-style streaming), and the implementation is correct. The 2026-05-20 LGTM was for the old, pre-refactor commit (ad5bbb1) and was force-pushed out, so this is effectively a fresh approval of the current implementation.

Review scope:
- Reviewed agent/codex/session.go and agent/codex/appserver_session.go (the 2 files / +7 / -84).
- Confirmed go vet and the existing codex test suite (TestAppServerSession_*, TestCodex_*) pass locally.
- Cross-checked s.emit() in both files: the only side effect of removing the stateMu lock around the agentMessage branch is that we no longer touch s.pendingMsgs, so the lock is genuinely unnecessary there. s.emit() is a non-blocking channel send, which is safe per Go's memory model without external synchronization.
- Cross-checked core/engine.go's EventText handler (line 4815) and EventThinking handler (line 4550) to confirm the new always-EventText emission still produces a sensible dynamic card, and that EventThinking from the codex `reasoning` item (appserver_session.go:1204 / session.go:491) is unchanged.

✅ What looks good:
- The core idea (drop the buffer, emit as EventText on item.completed) removes a class of "intermediate round text shown as thinking" bugs and matches what Claude Code already does. The simplification is real: 53 + 38 = 91 lines of state management deleted, 7 added.
- Removing flushPendingAsThinking() at the item.started / approval / permission sites is correct — there is no longer a buffer to flush, and the previous "buffered text becomes thinking once a tool starts" semantics are gone.
- The stateMu lock removal in the agentMessage case of handleItemCompleted is sound; s.emit() is verified thread-safe (non-blocking select on s.events).
- Reasoning events (appserver_session.go:1204, session.go:491) still emit EventThinking as before, so users who enable ThinkingMessages still get the actual chain-of-thought from codex.

🟠 Should improve (not blocking):
- P2 (dead field cleanup): The `pendingMsgs []string` field on appServerSession (appserver_session.go:180) is now declared but never read or written anywhere in the file. The struct field on codexSession was correctly removed in this PR, but the appserverSession one was missed. Recommend removing the field on the same commit so future readers don't wonder what it's for. (stateMu itself is still needed for preambleSent / currentTurn, so don't remove the mutex.)

🔵 Optional:
- P3 (test coverage): There are no codex-specific tests that exercise the new "agent_message → EventText on item.completed" path with multiple rounds. Existing TestAppServerSession_* cases cover approval, rate limits, and token usage, not the streaming semantics this PR changes. A small table-driven test (input: 2 rounds of agent_message; expected: 2 EventText events in order, no EventThinking) would prevent regression. Not blocking, but worth a follow-up.

❓ Questions (no action required, just for awareness):
- The PR description and commit body both state that intermediate-round agent_message text is now treated as EventText. For codex users with ThinkingMessages=false (display.Mode=quiet/compact), the behavior change is that they will now see intermediate round text where they previously saw nothing (it was hidden as thinking) or a separator. Is that the intended UX? My read of the PR description says yes ("matching Claude Code's streaming behavior"), so this is just a sanity check, not a blocker.

Testing / Risk:
- Verified evidence: `CGO_ENABLED=0 go vet ./agent/codex/...` is clean. `CGO_ENABLED=0 go test -count=1 -run "TestAppServer|TestCodex" -v ./agent/codex/...` is 10/10 PASS locally. CI is green per the registry.
- Remaining risk: end-to-end behavior in a real codex session is not exercised in CI. The behavioral change is best validated by running a multi-round codex session in a Lark/Feishu card-equipped platform and watching the dynamic card. This is a manual verification step, not a CI gap.

Next step:
- Optional: address the P2 dead-field cleanup (appserver_session.go:180) in this PR or a follow-up.
- Ready to merge otherwise. The owner can land this for v1.4.0 if the behavioral change is the desired UX direction.

PR #1083, opened 2026-05-21
fix(codex): preserve explicit stdio app-server URL

realraelrr:
## Summary
- normalize explicit `app_server_url = "stdio"` to `stdio://` instead of clearing it
- keep the existing WebSocket default when `app_server_url` is omitted
- preserve explicit stdio transport when constructing per-workspace Codex agents

## Why
When the Codex app-server backend is configured with `app_server_url = "stdio"`, the top-level agent normalizes it to an empty string. Workspace agent options then omit the empty URL, so workspace agents fall back to `ws://127.0.0.1:3845`. The appserver session still communicates over stdio pipes, causing initialization to time out.

## Tests
- `go test ./agent/codex`


chenhg5 (approved):
LGTM. Fix is targeted and well-tested.

PR #1090, opened 2026-05-22
fix(feishu): handle mention-only quoted replies

realraelrr:
## Summary
- keep Feishu text/post messages alive when bot mention stripping leaves empty text but quoted reply context exists
- allow core to process ExtraContent-only messages so quoted context can become the prompt
- add regression coverage for mention-only quoted text and ExtraContent-only handling

## Why
In Feishu group chats, replying to a message with only an @mention of the bot currently emits an inbound event with `parent_id/root_id` and mentions, but the adapter strips the bot mention and drops the message as empty before core sees the quoted context. DingTalk and WeCom do not require extra text for this interaction pattern.

## Tests
- `go test ./core -run 'TestHandleMessage_ExtraContent'`\n- `go test ./platform/feishu -run 'TestDispatchMessageKeepsMentionOnlyQuotedText|TestDispatchMessageIncludesQuotedImage|TestStripMentions'`\n- `go test ./platform/feishu`\n\nNote: on this machine, `go test ./core` against upstream/main currently has unrelated reply-footer path expectation failures; the targeted core regression test above passes.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1957,7 +1957,7 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 	}
 
 	content := strings.TrimSpace(msg.Content)
-	if content == "" && len(msg.Images) == 0 && len(msg.Files) == 0 && msg.Location == nil {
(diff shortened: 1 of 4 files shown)


chenhg5 (approved):
LGTM. Fix is targeted and well-tested.

PR #1093, opened 2026-05-22
fix(daemon): stub CheckLinger on non-Linux to unbreak macOS/Windows build

ztdd88:
## Summary

#965 added `daemon.CheckLinger()` in `daemon/systemd.go` (build-tagged `//go:build linux`), and `cmd/cc-connect/daemon.go:109` calls it unconditionally. The caller has no build tag, so any non-Linux build fails:

```
cmd/cc-connect/daemon.go:109:27: undefined: daemon.CheckLinger
```

Reproduced on `darwin/arm64` against current `main` (`a2cc0ab1`) with `make build`.

## Fix

Add a `!linux` stub that returns `(true, "")` so callers skip the systemd-linger warning on platforms where the concept doesn't apply. The existing runtime check in `cmd/cc-connect/daemon.go`

```go
if strings.Contains(mgr.Platform(), "user") {
    enabled, user := daemon.CheckLinger()
    ...
}
```

still gates the warning to systems where it's meaningful (Linux user-mode systemd); the stub just keeps the symbol resolvable cross-platform.

Alternative considered: build-tag the caller block. Rejected as more invasive — the surrounding `daemonInstall` function is shared across platforms, and splitting it would scatter the logic.

## Test plan

- [x] `make build` on `darwin/arm64` passes after the change (failed before).
- [ ] Linux build still works (no changes to Linux files; `daemon/systemd.go` stays authoritative on Linux because the new stub's `!linux` tag excludes it there).
- [ ] Windows build (untested locally, but `daemon/windows.go` uses `//go:build windows` and the stub's `!linux` includes Windows, so behavior should match macOS).

Fixes the regression introduced in #965.

### DIFF
--- a/daemon/check_linger_other.go
+++ b/daemon/check_linger_other.go
@@ -1,0 +1,9 @@
+//go:build !linux
+
+package daemon
+
+// CheckLinger is a stub for non-Linux platforms where systemd linger does not
+// apply. Returns (true, "") so callers skip the linger warning.
+func CheckLinger() (enabled bool, user string) {
+	return true, ""
+}


chenhg5 (approved):
LGTM. Clean and well-tested.

happyTonakai:
duplicate with #1052 #1063 and was fixed by 8e04d66bf68eb5c56e7bb4733f35223c61ceb94b

chenhg5 (commented):
## QA Triage Review: PR #1093 (fix(daemon): stub CheckLinger on non-Linux to unbreak macOS/Windows build)

结论: **ready-to-merge** (trivial, correct, fixes a real build break).

### What it does
- New `daemon/check_linger_other.go` with `//go:build !linux` tag. 9 lines total.
- Provides no-op `CheckLinger() (enabled bool, user string)` returning `(true, "")` for non-Linux.

### Quality
- The stub is the standard Go build-tag pattern: `//go:build !linux` ensures the file is compiled only on non-Linux. The Linux implementation lives in a separate file (presumably `check_linger_linux.go`).
- Return value `(true, "")` correctly signals "skip the linger warning" — non-Linux platforms don't have systemd, so the warning is meaningless.
- The existing call sites don't need changes because the function signature is preserved.

### Risk
- None. This is a build-only fix. No runtime behavior change for non-Linux users (they already get the stub or no-op behavior at link time, the only difference is now `go build` succeeds on darwin/windows).

### Triage label
- **ready-to-merge** — minimal, correct, unblocks macOS/Windows builds. Should be a fast-merge.

— cc-connect/qa-claudecode

PR #1099, opened 2026-05-23
fix(feishu): admit attachment-only messages into active threads without @bot

walkerchi:
## Summary

In group chats with `thread_isolation`, any message without an explicit @bot mention is dropped before image/file download. Users naturally drop screenshots into a thread after their opening @bot text, so all follow-up images are silently lost — the agent only ever sees the text and can never see the attachments, despite Feishu having delivered them.

## Fix

Track threads that have already been engaged by an @bot message. While the @bot filter is otherwise unchanged, allow `image`/`file`/`audio`/`media` messages to pass through inside an active thread. Plain `text` and rich-text `post` still require an explicit mention so unrelated thread chatter does not leak into the agent.

## Test plan

- [x] `go test ./platform/feishu/...` passes
- [x] New unit tests for `isAttachmentMsgType`, `markThreadSessionActive`, `isActiveThreadSession`
- [x] New end-to-end `onMessage` test covering:
  - opening @bot text establishes the thread
  - in-thread attachment without @ is admitted
  - cross-thread attachment without @ is still dropped
  - in-thread text without @ is still dropped

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -158,6 +158,13 @@ type Platform struct {
 	// session key, enabling async card refreshes via the Patch API.
 	cardActionMsgMu  sync.Mutex
 	cardActionMsgIDs map[string]string // sessionKey → messageID
+	// activeThreadSessions tracks thread sessionKeys that have already been
+	// accepted by the bot. In group chats with thread_isolation, once a thread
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. Clean and well-tested.

PR #1107, opened 2026-05-24
feat: add sonnet[1m] to hardcoded fallback model list

zhangyanbo2007:
## Problem

The available models fallback list in `agent/claudecode/claudecode.go` included `opus[1m]` but was missing `sonnet[1m]`, making it impossible to use 1M context with Sonnet models when `config.toml` has no explicit models configured and the API fails to return model aliases.

## Solution

Add `sonnet[1m]` to the hardcoded fallback model list, one-line fix.

## Test plan

- [x] Build passes: `go build ./...`
- [x] cc-connect runs with `sonnet[1m]` model selectable via config


chenhg5 (approved):
LGTM. CI green, well-structured.

PR #1121, opened 2026-05-25
fix(dingtalk): include quoted interactive card content

realraelrr:
## Summary
- enrich DingTalk reply messages when the quoted bot message is an `interactiveCard`
- extract readable quoted card text from card template fields and supported fallback fields before forwarding to the agent
- preserve existing text-quote behavior and cap interactive-card quote text length

## Problem
When a user replies to a DingTalk bot AI card, the incoming callback reports `repliedMsg.msgType=interactiveCard`. The DingTalk adapter currently only extracts quoted content for `text`, logs `quoted message type not supported type=interactiveCard`, and forwards only the user reply. As a result, the agent cannot see the bot card being referenced.

This is the DingTalk counterpart of the quoted interactive-card context gap previously addressed for Feishu in #489.

## Changes
- add quoted-message extraction dispatch for `text` and `interactiveCard`
- read interactive-card text from `cardData.cardParamMap` using configured `card_template_key` first, then `content`
- support string-encoded card container JSON plus top-level `content`, `text`, `markdown`, and `title` fallbacks
- preserve JSON-shaped visible card responses as quoted text rather than treating them as card envelopes
- limit extracted interactive-card quoted text to 4000 runes to keep prompt enrichment bounded
- add unit and raw-callback integration coverage for template keys, fallback order, JSON payload boundaries, truncation, and legacy text behavior

## Test plan
- [x] `GOMODCACHE=/tmp/cc-connect-gomodcache go test ./platform/dingtalk -count=1`
- [x] `GOMODCACHE=/tmp/cc-connect-gomodcache go build ./platform/dingtalk`
- [x] Independent subagent code review completed with no blocking findings

## Repository baseline notes
- `GOMODCACHE=/tmp/cc-connect-gomodcache go build ./...` currently fails outside this change because `web/embed.go` expects missing `web/dist` assets.
- `GOMODCACHE=/tmp/cc-connect-gomodcache go test ./...` currently reports unrelated failures in `cmd/cc-connect`/`web` for the missing assets and pre-existing failures in `agent/codex`, `core`, and `daemon`; `platform/dingtalk` passes.

### DIFF
--- a/platform/dingtalk/dingtalk.go
+++ b/platform/dingtalk/dingtalk.go
@@ -27,11 +27,11 @@ func init() {
 }
 
 type replyContext struct {
-	sessionWebhook  string
-	conversationId  string
-	senderStaffId   string
-	isGroup         bool
-	proactive       bool // true when constructed by ReconstructReplyCtx (no sessionWebhook)
+	sessionWebhook string
+	conversationId string
+	senderStaffId  string
+	isGroup        bool
+	proactive      bool // true when constructed by ReconstructReplyCtx (no sessionWebhook)
 }
 
 // richTextContent mirrors the full structure of the DingTalk "text" JSON field,
@@ -52,6 +52,8 @@ type repliedTextContent struct {
 	Text string `json:"text"`
 }
 
+const maxQuotedMessageRunes = 4000
+
 type downloadResponse struct {
 	DownloadUrl string `json:"downloadUrl"`
 }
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
LGTM. 逻辑正确：修复了 interactive card 引用内容被丢弃的问题，添加了  结构处理 ，截断逻辑合理，测试覆盖充分，CI 全绿。

chenhg5 (approved):
LGTM. Logic correct: fixes interactive card quoted content being dropped. Added RepliedMsg struct to handle repliedMsg.content.cardData.cardParamMap.content, truncation logic is reasonable, test coverage is sufficient, CI all green.

PR #1126, opened 2026-05-26
docs(prompt): clarify cron edit workflow

JupiterTheWarlock:
## Summary
- Add `cc-connect cron info <job-id> [field]` to the agent-facing cron prompt.
- Clarify that agents should inspect an existing cron job before applying targeted `cron edit` changes.
- Discourage delete-and-recreate flows unless the user explicitly asks to replace the job.

## Background
When a user asks an AI agent to adjust a field on an existing cron task, the agent may read cron configuration and infer a delete/recreate workflow because the prompt exposed `cron list`, `cron edit`, and `cron del`, but did not mention `cron info` or the intended inspect-then-edit flow. Recreating a job can change the job ID and risks losing fields that were not part of the requested change.

## Test plan
- `go test ./core -run TestAgentSystemPrompt`

### DIFF
--- a/core/interfaces.go
+++ b/core/interfaces.go
@@ -98,12 +98,14 @@ Examples:
   cc-connect cron add --cron "0 9 * * 1" --prompt "Generate a weekly project status report" --desc "Weekly Report"
   cc-connect cron add --cron "*/2 * * * *" --exec "ipconfig" --session-mode new-per-run --desc "Every 2 min ipconfig"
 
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
LGTM. Small but useful docs clarification: adds  command and clarifies the edit vs delete-recreate workflow. CI all green.

chenhg5 (approved):
-

PR #1127, opened 2026-05-26
feat(core): auto-name new agent sessions from first user message

yudecheng:
cc-connect spawns agent CLIs (cursor-agent, claude-code, etc.) without the IDE renderer that normally writes meta.name to the agent's local store. As a result /list falls back to a session-id hash prefix, making sessions hard to recognise.

This change adds a core-layer hook that derives a short, human-readable title from the first user message of each session and stores it via SessionManager.SetSessionName. The hook fires on the engine's turn-complete path, is idempotent across turns, and never overwrites a user-assigned name (/name preserves manual titles).

Configuration (per project, all optional):

  auto_session_name = "first_message"  # default; "off" disables
  auto_session_name_max_len = 28       # default; rune-safe

Implementation notes:

- core/auto_session_name.go is the only logic file. It depends on stdlib only (regexp, strings, log/slog) -- no agent or platform imports, per CLAUDE.md.
- Title pipeline: strip noise (fenced code, inline code, URLs, @-references, markdown emphasis) -> first sentence -> collapse whitespace -> rune-safe truncate. Sentence terminators include both ASCII (.!?) and CJK (。！？) plus newline.
- agent/cursor: hash-prefix fallback in listCursorSessions is replaced with a sortable timestamp ("New chat 05-26 15:04") for the rare path where the engine hook hasn't run yet.

Tests: 18 cases covering derivation edge cases, idempotency, preservation of /name, disabled-mode bypass, and concurrent-safe naming under -race.


chenhg5 (approved):
-

PR #1161, opened 2026-05-30
fix(cron): merge provider env vars into cron shell environment

Shicongcoding:
## Summary

- Cron shell commands previously only inherited `os.Environ()` plus `CC_PROJECT` and `CC_DATA_DIR`. Provider-level env vars (e.g. `CLAUDE_CODE_USE_BEDROCK`) were missing, causing cron scripts that depend on these variables to fail.
- Now provider env is merged using `MergeEnv` so provider values properly override base env, matching the environment the agent subprocess sees.

## Test plan

- [x] `go build ./...` passes
- [ ] Verify cron scripts can access provider env vars after this change

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1289,6 +1289,20 @@ func (e *Engine) executeCronShell(p Platform, replyCtx any, job *CronJob) error 
 	}
 	shellCmd.Dir = workDir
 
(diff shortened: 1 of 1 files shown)


Shicongcoding:
Superseded by #1162 (includes CC_PROJECT, CC_DATA_DIR, and RunAs env in addition to provider env)

PR #1163, opened 2026-05-30
fix(core): change AllowList to fail-closed for empty allow_from/allow_chat

Shicongcoding:
## Summary
- Empty `allow_from`/`allow_chat` previously allowed all users (fail-open), which is insecure by default
- Now empty means deny-all: users must explicitly set `allow_from = "*"` to permit everyone, or list specific IDs
- This fixes `TestAllowChat_FiltersGroupMessages` which expected fail-closed behavior but the implementation was still fail-open

## Changes
- `core/message.go`: `AllowList("")` now returns `false` instead of `true`; `CheckAllowFrom` updated to warn about deny-all vs permit-all
- `core/command_test.go`: Updated `TestAllowList` to expect `false` for empty input
- `platform/feishu/platform_test.go`: Added explicit `allow_from: "*"` and `allow_chat: "*"` to integration tests that create interactive platforms; updated `TestAllowChat_FiltersGroupMessages` test case to expect deny-all for empty `allow_chat`

## Test plan
- [x] `go test ./core/... -run TestAllowList` passes
- [x] `go test ./platform/feishu/... -run TestInteractive -run TestAllowChat` passes
- [x] `go build ./...` succeeds
- [ ] Existing feishu integration tests still pass with explicit allow_from/allow_chat
- [ ] Pre-existing unrelated test failures (`TestProcessInteractiveEvents_*`, `TestResolveLocalDirPath_AcceptsSubdir`) are not affected

### DIFF
--- a/core/command_test.go
+++ b/core/command_test.go
@@ -253,7 +253,7 @@ func TestAllowList(t *testing.T) {
 		user   string
 		expect bool
 	}{
-		{"", "anyone", true},
+		{"", "anyone", false},
 		{"*", "anyone", true},
 		{"user1", "user1", true},
 		{"user1", "USER1", true},
--- a/core/message.go
+++ b/core/message.go
@@ -29,12 +29,18 @@ func MergeEnv(base, extra []string) []string {
 	return append(merged, extra...)
 }
 
-// CheckAllowFrom logs a security warning at startup when allow_from is not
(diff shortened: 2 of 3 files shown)


Shicongcoding:
Superseded by #1164 — that PR includes the same changes plus explicit feishu_test.go updates and a breaking change note

PR #1165, opened 2026-05-30
fix(cron): merge project and provider env vars into cron shell environment

Shicongcoding:
## Summary

Cron shell commands started by `sh -c` inherit only the bridge process's base environment. They miss `CC_PROJECT`, `CC_DATA_DIR`, provider-specific env vars (e.g. `CLAUDE_CODE_USE_BEDROCK`), and any `run_as_user` overrides. Cron scripts that need these variables must manually re-source profiles or hardcode paths.

This PR propagates the same env composition used by interactive sessions into cron shells:
- Start from `os.Environ()`
- Append project keys (`CC_PROJECT`, `CC_DATA_DIR`)
- Append `run_as_user` overrides
- Append provider env vars via `MergeEnv`

## Test plan

- [ ] Configure a cron job that prints `echo $CC_PROJECT` and verify it outputs the project name
- [ ] Configure a provider with env vars (e.g. `CLAUDE_CODE_USE_BEDROCK=1`) and verify a cron job inherits it
- [ ] Verify existing cron jobs without env dependencies still run correctly
- [ ] Run `go build ./...` and `go test ./core/...` to confirm no regressions

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -1289,6 +1289,30 @@ func (e *Engine) executeCronShell(p Platform, replyCtx any, job *CronJob) error 
 	}
 	shellCmd.Dir = workDir
 
+	// Merge project-level and provider env vars into the cron shell
+	// so cron scripts inherit the same environment the agent subprocess sees.
+	shellCmd.Env = os.Environ()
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
-

PR #1166, opened 2026-05-30
fix(core): add panic recovery for critical goroutines

Shicongcoding:
## Summary
- Add `Engine.safeGo`: launches goroutines with `recover()`, logs panics with full stack traces, and writes crash logs to `dataDir/crashes/crash.log` for post-mortem diagnosis.
- Wrap the three critical goroutine launch sites with `safeGo`: `processInteractiveMessageWith`, `drainOrphanedQueue`, `runUnsolicitedReader`.
- Add inline `defer recover()` to `agentSession.Send` and `Close` goroutines so panics during I/O don't kill the event loop.

## Motivation
Goroutines that panic currently exit silently — launchd sees the process still running and does nothing, leaving the bridge in a degraded state (e.g. no event processing, no message delivery) with no diagnostic trail.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./core/ -run TestSafeGo` passes (both recovery and normal execution)
- [x] `go test ./core/ -run TestAllowList` passes (existing behavior unchanged)
- [ ] Manual: deploy and verify no regression in message processing

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -317,6 +317,7 @@ type interactiveState struct {
 	pendingMessages        []queuedMessage // messages queued while session was busy
 	approveAll             bool            // when true, auto-approve all permission requests for this session
 	fromVoice              bool            // true if current turn originated from voice transcription
+	preNotify              func()          // deferred notification callback (skipped on NO_REPLY)
 	sideText               string
 	deleteMode             *deleteModeState
 	modelSwitch            *modelSwitchState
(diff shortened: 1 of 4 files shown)


Shicongcoding:
Superseded by #1167 which includes panic recovery + deferred NO_REPLY start notifications (superset)

Shicongcoding:
Reopening: #1167 was also closed. This is the clean panic recovery PR without the NO_REPLY feature. Per maintainer feedback, one agent should own each PR to avoid duplication.

Shicongcoding:
Superseded by #1169 (pure panic recovery, no NO_REPLY feature addition per @nan-chen's guidance to only submit necessary fixes)

PR #1169, opened 2026-05-30
fix(core): add panic recovery to engine goroutines

Shicongcoding:
## Summary
- Add `Engine.safeGo`: goroutine launcher with `recover()` that logs panics and writes crash records to `dataDir/crashes/crash.log` for post-restart diagnosis
- Wrap all engine-managed goroutine launches with `safeGo`: `processInteractiveMessageWith`, `drainOrphanedQueue`, `runUnsolicitedReader`
- Add inline `recover()` to `agentSession.Send` and `Close` goroutines to prevent silent goroutine death

## Motivation

Currently, a panic in any engine goroutine kills that goroutine silently. The main process stays alive (since launchd/systemd only restart on process exit), so the bridge appears healthy but is actually non-functional — messages are accepted but never processed, and there's no way to diagnose the failure after the fact.

## Test plan
- [x] `TestSafeGo_RecoversPanic` — panic is recovered, crash log is written
- [x] `TestSafeGo_NormalExecution` — no crash log for normal runs
- [x] `TestExecuteCronJob_ResolvesCronReplyTarget` — cron test still passes

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2127,7 +2127,7 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 			// and the queue append. Re-try TryLock — if it succeeds, no one is
 			// draining the queue so we must start a processor ourselves.
 			if session.TryLock() {
-				go e.drainOrphanedQueue(session, sessions, interactiveKey, agent, resolvedWorkspace)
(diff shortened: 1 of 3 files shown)


Shicongcoding:
Duplicate of #1166 — closing to avoid confusion.

PR #1170, opened 2026-05-30
fix(core): add panic recovery to critical goroutines

Shicongcoding:
## Summary

A panic in any engine goroutine silently kills that goroutine. The main process stays alive (launchd/systemd only restart on process exit), so the bridge appears healthy but is non-functional — messages are accepted but never processed, and there is no way to diagnose the failure after the fact.

This PR adds two layers of protection:

- **`Engine.safeGo`**: goroutine launcher with `recover()` that logs panics (value + full stack trace) and writes crash records to `dataDir/crashes/crash.log` for post-restart diagnosis. Used to wrap the three most critical goroutine launches: `processInteractiveMessageWith`, `drainOrphanedQueue`, and `runUnsolicitedReader`.
- **Inline `recover()`**: added to `agentSession.Send` and `Close` goroutines to prevent silent goroutine death from leaving sessions in a stuck state.

## Test plan

- [x] `go build ./...` passes
- [x] `TestSafeGo_RecoversPanic` — panic is recovered, crash log is written
- [x] `TestSafeGo_NormalExecution` — no crash log for normal runs
- [ ] Manual: verify the bridge gracefully handles panics in agent session Send/Close

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2127,7 +2127,7 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 			// and the queue append. Re-try TryLock — if it succeeds, no one is
 			// draining the queue so we must start a processor ourselves.
 			if session.TryLock() {
-				go e.drainOrphanedQueue(session, sessions, interactiveKey, agent, resolvedWorkspace)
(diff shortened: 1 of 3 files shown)


Shicongcoding:
Duplicate of #1169 (pure panic recovery, already open). Closing mine — one agent should handle this.

PR #1172, opened 2026-05-30
codex: use workspace-write for app-server suggest mode

RalphHan:
Summary:
- Change Codex app-server suggest mode to use approvalPolicy=on-request with sandbox=workspace-write.
- Add unit coverage for the mode mapping and thread request parameters.
- Add a short fork-branch install note for testing patched builds.

Testing:
- go test ./agent/codex -count=1


chenhg5 (changes requested):
-

PR #1174, opened 2026-05-31
feat(feishu): add thread-only isolation mode

chasing1874:
## Summary

Refs #1079  
Context: https://github.com/chenhg5/cc-connect/issues/1079#issuecomment-4585368630

This PR adds a Feishu-specific session isolation mode:

```toml
thread_isolation = true
feishu_thread_isolation_mode = "thread_only"
```

The default remains:

```toml
feishu_thread_isolation_mode = "reply_and_thread"
```

so existing users keep the current behavior.

## Why

Feishu already supports `thread_isolation = true`, and the current behavior can isolate real Feishu topic/thread messages.

However, the current behavior is broader than “topic/thread isolation”. When `thread_isolation = true`, the Feishu adapter derives the session key from `RootId`, or falls back to the current `MessageId`. In practice this means:

- every top-level group `@bot` message starts a separate `root:<message_id>` session
- normal Reply / quoted reply can enter a separate `root:<root_id>` session
- real Feishu Reply in Thread / topic also enters a separate `root:<root_id>` session

This works for task-style isolation, but it is not always the desired group-chat UX.

In a group chat, normal Reply is usually part of the main conversation. It is often just a way to quote or point at previous context, not an explicit request to fork the agent into a new session. If every normal reply can create or enter an isolated session, the bot loses the feel of being a persistent participant in the group. The conversation becomes fragmented across accidental reply chains, and the long-lived group persona / shared session is harder to maintain.

By contrast, Feishu Reply in Thread / topic is an explicit user action. It is a much clearer signal that the user wants to move a task into a child conversation.

This PR adds a Feishu-specific mode for that model, so the group chat can keep a stable main session while explicit Feishu topics/threads still get isolated child sessions.

## What Changed

The default remains unchanged:

```toml
feishu_thread_isolation_mode = "reply_and_thread"
```

This preserves the existing behavior:

- top-level group messages can become `root:<MessageId>` sessions
- normal replies can use `root:<RootId>`
- Feishu topic/thread messages also use the root-based isolated session

This PR adds:

```toml
feishu_thread_isolation_mode = "thread_only"
```

In `thread_only` mode:

- only messages with a real Feishu `ThreadId` create an isolated child session:
  ```text
  feishu:<chatID>:thread:<threadID>
  ```
- normal group messages stay in the normal group/user session
- normal Reply / quoted reply does not fork a new session
- normal Reply still injects the quoted `ParentId` message as context
- fallback still respects `share_session_in_channel`

## Notes

- This PR does not change Discord, Slack, Telegram, or core session semantics.
- This PR does not add recent chat/thread history injection. That can be handled separately later.
- `thread_only` depends on Feishu events carrying `ThreadId` for native topic/thread messages.

## Tests

- `pnpm --dir web build`
- `go build ./...`
- `go vet ./...`
- `go test ./...`
- `go test -race ./...`
- `git diff --check`

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -940,6 +940,9 @@ app_secret = "your-feishu-app-secret"
 #                                   # 设为 true 时，群聊内所有用户共享同一个 Agent 会话
 # thread_isolation = false  # If true, group messages use Feishu reply threads as session boundaries (one thread/root = one agent session)
 #                           # 设为 true 时，群聊按飞书话题/根消息隔离会话（一个 thread/root 对应一个 Agent 会话）
+# feishu_thread_isolation_mode = "reply_and_thread"  # reply_and_thread | thread_only; only used when thread_isolation=true
+#                                                    # reply_and_thread：每条顶层 @bot 消息成为独立任务线，后续同根普通回复/话题回复进入同一会话
+#                                                    # thread_only：普通 @bot/普通回复沿用群聊会话，只有飞书话题开独立子会话
 # reply_to_trigger = true   # Default: bot uses “reply to message” API (quotes the user’s message). Set false to post plain chat messages instead.
 #                           # 默认 true：用「回复消息」API（引用用户消息）。设为 false 则在会话里发普通消息、不引用触发消息。
 # progress_style = "legacy" # Progress rendering style: legacy | compact | card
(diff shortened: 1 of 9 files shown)


chenhg5 (approved):
-

chenhg5 (approved):
-

PR #1177, opened 2026-05-31
fix(pi): add SetWorkDir method for runtime workdir switching

happyTonakai:
## Summary

Add `SetWorkDir` method to the pi agent so commands like `/status` and
multi-workspace switching can dynamically change the agent's working
directory at runtime.

## Changes

| File | Change |
|------|--------|
| agent/pi/pi.go | Add `SetWorkDir(dir string)` implementing `WorkDirSwitcher` interface |

## Test plan

- [x] `go build ./agent/pi/` passes
- [x] `go test ./agent/pi/ -count=1` passes
- [x] **Manual (Feishu)**: send `/dir /path/to/project` → verify agent switches to new workdir
- [x] **Manual (Feishu)**: send `/status` → verify displayed workdir matches the new path

### DIFF
--- a/agent/pi/pi.go
+++ b/agent/pi/pi.go
@@ -223,7 +223,14 @@ func (a *Agent) AvailableReasoningEfforts() []string {
 	return []string{"off", "minimal", "low", "medium", "high", "xhigh"}
 }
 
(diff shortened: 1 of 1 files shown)


chenhg5 (approved):
-

PR #1178, opened 2026-05-31
fix(pi): read enabledModels from settings.json for /model

happyTonakai:
## Summary

The pi agent's `AvailableModels()` returned `nil`, so the `/model` command
never showed a model selection list. Now it reads `enabledModels` from
`~/.pi/agent/settings.json`.

`New()` also falls back to `defaultModel` (+ `defaultProvider`) from
settings.json when no model is specified in config, so the user's
preferred model is used out of the box.

## Changes

| File | Change |
|------|--------|
| agent/pi/pi.go | `AvailableModels()` reads settings.json; `New()` falls back to `defaultModel`; add 6 helpers for settings I/O |
| agent/pi/pi_test.go | Add 5 tests covering settings parsing, default model, settings dir, and env var override |

## Test plan

- [x] `go build ./agent/pi/` passes
- [x] `go test ./agent/pi/ -count=1` passes (49 tests)
- [x] **Manual (Feishu)**: send `/model` → verify card shows current model + dropdown with `enabledModels` from settings.json
- [x] **Manual (Feishu)**: select a model from the dropdown → verify model switches and card shows success
- [x] **Manual (Feishu)**: send `/model switch <model-name>` → verify model changes

### DIFF
--- a/agent/pi/pi.go
+++ b/agent/pi/pi.go
@@ -49,6 +49,13 @@ func New(opts map[string]any) (core.Agent, error) {
 		return nil, fmt.Errorf("pi: '%s' not found in PATH, install with: npm install -g @mariozechner/pi-coding-agent", cmd)
 	}
 
+	// If model not specified in opts, try defaultModel from settings.json
+	if model == "" {
+		if def, err := readDefaultModel(); err == nil && def != "" {
+			model = def
+		}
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
-

PR #1179, opened 2026-05-31
feat(slack,tmux): per-thread session scope and per-session tmux windows

MASAKASUNO1:
## Summary

Adds two complementary options so a single chat platform can drive **isolated conversations per Slack thread**:

1. **`session_scope` for the Slack platform** — `"user"` (default) | `"channel"` | `"thread"`.
   `"thread"` keys each cc-connect session to the Slack **thread root**, so a new top-level
   message starts a fresh conversation and replies inside a thread continue it.

2. **`window_per_session` for the tmux agent** — each cc-connect session gets its own tmux
   **window** (and thus its own `init_command`/agent instance) instead of sharing one pane.
   Without this, `session_scope="thread"` over a tmux agent still funnels every thread into a
   single shared CLI, so concurrent threads interleave; with it, each thread is fully isolated.

## Motivation

Today a Slack channel maps to one session per `(channel, user)`. Power users often want a
**separate conversation per thread** (one thread = one task). These two options make that
possible end-to-end, including for the tmux agent (which previously shared a single pane).

## Config

```toml
# Slack: one session per thread
[projects.platforms.options]   # type = "slack"
session_scope = "thread"       # "user" (default) | "channel" | "thread"

# tmux agent: a dedicated window (and Claude instance) per session
[projects.agent.options]       # type = "tmux"
window_per_session = true
```

## Behaviour details

- Slack session keys: `slack:<ch>:<user>` (user) · `slack:<ch>` (channel) · `slack:<ch>:t:<threadTS>` (thread).
- **Top-level DMs** (no thread) fall back to the user-scoped key, so DMs stay continuous
  (uses the same `assistantOrThreadTS` value that routes the reply).
- **App mentions** key + reply on the same thread root for consistency.
- `ReconstructReplyCtx` parses the `t:<ts>` suffix, so **proactive** replies (cron,
  send-to-session, restart/model/delete notifications) post back into the original thread.
- tmux window names are **randomised** (not a counter), so they never repeat across process
  restarts and never collide across the per-workspace agents that share one tmux session.
- `WorkspaceAgentOptions` carries `window_per_session` through multi-workspace mode.

## Backward compatibility

Both options default off; existing configs are unaffected. When `session_scope` is unset the
previous behaviour is preserved, and `share_session_in_channel = true` maps to `"channel"`.

## Limitations

- Slash commands don't carry a `thread_ts`, so under `session_scope="thread"` they fall back to
  the user-scoped key (documented).

## Testing

- `go test ./platform/slack/ ./agent/tmux/` — added unit tests for `normalizeSessionScope`,
  `buildSessionKey`, `ReconstructReplyCtx`, `windowForSession`, and the `WorkspaceAgentOptions`
  round-trip.
- Manual: built and ran against a real Slack workspace — distinct threads produced distinct
  cc-connect sessions, distinct tmux windows, and independent agent memory.

### DIFF
--- a/agent/tmux/tmux.go
+++ b/agent/tmux/tmux.go
@@ -32,7 +32,12 @@ type Agent struct {
 	pollMs          int
 	stripInputBlock bool     // strip the ───/❯/─── input area block from output
 	stripPatterns   []string // per-line regex patterns to strip from output
-	mu              sync.RWMutex
+	// windowPerSession, when true, gives each cc-connect session its own tmux
+	// window (and thus its own init_command/agent instance) instead of sharing
+	// the single session:pane target. Required for true per-session isolation
+	// (e.g. session_scope = "thread" on the platform).
+	windowPerSession bool
+	mu               sync.RWMutex
 }
 
 func New(opts map[string]any) (core.Agent, error) {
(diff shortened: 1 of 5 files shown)


chenhg5 (approved):
-

PR #1192, opened 2026-06-01
feat(minimax): upgrade default model to MiniMax-M3

octo-patch:
## Summary

Upgrade the bundled `minimax` / `minimax-cn` provider presets and example configs from `MiniMax-M2.7` to `MiniMax-M3`.

## Changes

- **`provider-presets.json`** — set `MiniMax-M3` as the default `model` for both `minimax` and `minimax-cn` across `claudecode`, `codex`, and `opencode` agents; reorder `models` lists so M3 is first; keep `MiniMax-M2.7` (and `-highspeed` where it was already listed) as fallback options; drop the older `MiniMax-M2.5` entry.
- **`config.example.toml`** — point all `# model = "MiniMax-M2.7"` example lines at `MiniMax-M3`; refresh the model list comment; replace the `M2.5` / `M2.5-highspeed` alias examples with an `M3` alias example, keeping the M2.7 / M2.7-highspeed aliases.
- **`docs/usage.md`** & **`docs/usage.zh-CN.md`** — bump the `model = "MiniMax-M2.7"` snippet to `MiniMax-M3` so the docs match the new default.

API URLs, TTS configuration, and other providers are unchanged. The marketing prose in `README.md` / `README.zh-CN.md` (which references model-specific stats) is left intact.

## Test Plan

- `go test ./config/ ./core/` — passes (config parsing + provider-presets loader tests green).
- `python -c "import json; json.load(open('provider-presets.json'))"` — JSON syntax valid.
- Manual diff inspection: only `minimax` provider blocks are touched; no historical changelog entries modified.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -849,7 +849,7 @@ mode = "default" # "default" | "acceptEdits" (edit) | "plan" | "auto" | "bypassP
 #
 # # MiniMax — OpenAI-compatible provider with 1M context window
 # # MiniMax — 兼容 OpenAI 接口的大模型，支持 1M 超长上下文
-# # Models: MiniMax-M3 (latest), MiniMax-M2.7 (1M context), MiniMax-M2.7-highspeed
+# # Models: MiniMax-M3 (latest), MiniMax-M2.7 (512K context), MiniMax-M2.7-highspeed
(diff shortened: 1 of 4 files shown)


chenhg5 (approved):
-

PR #1196, opened 2026-06-02
fix(core): /stop preserves AgentSessionID so next message resumes session

### LINKED ISSUE #1189: /stop command clears AgentSessionID but card button does not — inconsistent behavior
## Summary

There is a behavioral inconsistency between the text `/stop` command and the Stop button in the `/help` card:

| Method | Stops agent process | Clears AgentSessionID |
|---|---|---|
| Text command `/stop` | Yes | **Yes** → next message starts a brand-new session |
| `/help` card Stop button | Yes | **No** → next message attempts to resume the previous session |

## Root Cause

In `core/engine.go`, `cmdStop` calls `clearStaleSessionID()` after `stopInteractiveSession()`:

```go
// cmdStop (~line 8138)
e.stopInteractiveSession(iKey, p, msg.ReplyCtx)
clearStaleSessionID()  // ← this line is not called in the card button path
e.reply(p, msg.ReplyCtx, e.i18n.T(MsgExecutionStopped))
```

The card button path (`handleButtonAction`, ~line 9820) only calls:

```go
case "/stop":
    sessionKey = e.interactiveKeyForSessionKey(sessionKey)
    e.stopInteractiveSession(sessionKey, nil, nil)
    // clearStaleSessionID is NOT called here
```

`clearStaleSessionID()` was added to fix a resume-loop bug (issue #830), but as a side effect it makes the text `/stop` command more destructive than the card button equivalent.

## Expected Behavior

Both methods should behave the same. Ideally `/stop` should work like the ESC key in the Claude Code CLI — interrupt the current execution without destroying the session context, so the user can continue the conversation afterward.

## Question

Is the `clearStaleSessionID()` call in `cmdStop` still necessary after issue #830 was fixed? If the card button path works correctly without it, can we remove it from `cmdStop` to make the behavior consistent?

shellus:
## Summary

`cmdStop` no longer clears `AgentSessionID`; the next message can `--resume` the conversation, matching the `/help` card-button Stop path and eliminating the inconsistency described in #1189.

- Session-ID write-back now always follows the live forked ID Claude reports on every `--resume` (was: skip if already set); name binding still fires only on first assignment to avoid polluting `sessionNames`
- Removed `clearStaleSessionID` helper and `CompareAndSetAgentSessionID` guard; replaced with unconditional `SetAgentSessionID` + `wasEmpty` name-binding gate

Closes #1189

## Test plan

- [ ] `go test ./core/ -run 'TestSessionIDWriteback|TestInteractiveWriteBack|TestCmdStop' -v` — passes
- [ ] `go test -race ./core/...` — full core suite with race detector
- [ ] Manual: `/stop` then send a new message → agent resumes previous session (not a fresh start)

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -3002,10 +3002,19 @@ func (e *Engine) getOrCreateInteractiveStateWith(sessionKey string, p Platform, 
 	}
 
 	if newID := agentSession.CurrentSessionID(); newID != "" {
-		if session.CompareAndSetAgentSessionID(newID, agent.Name()) {
-			pendingName := session.GetName()
-			if pendingName != "" && pendingName != "session" && pendingName != "default" {
-				sessions.SetSessionName(newID, pendingName)
+		// Track the latest session ID Claude reports. Each --resume forks a new
+		// session_id, so the stored ID must follow the live process or a later
+		// resume (after /stop or /model) would target a stale node and lose
+		// context. Session naming binds only on first assignment to avoid
+		// polluting sessionNames with every forked ID.
+		wasEmpty := session.GetAgentSessionID() == ""
+		if session.GetAgentSessionID() != newID {
+			session.SetAgentSessionID(newID, agent.Name())
+			if wasEmpty {
+				pendingName := session.GetName()
+				if pendingName != "" && pendingName != "session" && pendingName != "default" {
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
-

PR #1206, opened 2026-06-02
fix(pi): correct agent directory paths (SkillDirs + GlobalMemoryFile)

happyTonakai:
## Summary

cc-connect's pi agent had hardcoded paths missing the `/agent` segment
that pi uses by default (`~/.pi/agent/`), causing two features to silently
break:

- **SkillDirs()** returned `~/.pi/skills/` instead of `~/.pi/agent/skills/`,
  so cc-connect's `SkillRegistry` never found pi skills. Slash commands
  like `/caveman` were not routed by cc-connect's command handler and fell
  through to pi's print mode as raw text.
- **GlobalMemoryFile()** returned `~/.pi/AGENTS.md` instead of
  `~/.pi/agent/AGENTS.md`, so the global memory file was never picked up.

Both paths now also respect `$PI_CODING_AGENT_DIR` when set, and
`SkillDirs()` additionally includes `~/.agents/skills/` (a common shared
skill directory used by lark-cli and other agent tooling).

## Test plan

- [x] `go test ./agent/pi/ -v` — passes
- [x] Verify `/caveman` (or any pi skill) triggers the skill instructions
      instead of being sent to the LLM as plain text

### DIFF
--- a/agent/pi/pi.go
+++ b/agent/pi/pi.go
@@ -197,11 +197,16 @@ func (a *Agent) ProjectMemoryFile() string {
 }
 
 func (a *Agent) GlobalMemoryFile() string {
-	homeDir, err := os.UserHomeDir()
-	if err != nil {
-		return ""
+	// Use PI_CODING_AGENT_DIR if set, otherwise default to ~/.pi/agent/.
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
-

PR #1230, opened 2026-06-05
feat: add agent-driven TTS send support

AaronZ345:
## Summary
- add per-project TTS synthesis overrides for provider, voice/voice_id, speed, language_type, and max_text_len
- allow MiniMax TTS to fall back to a local JSON config file when api_key is not set
- add `cc-connect send --tts <text>` and local API `tts_text` so agents can send requested voice messages without user-facing slash commands
- update TTS provider wiring, docs, examples, and tests

## Testing
- `git diff --check`
- `go test ./cmd/cc-connect ./core ./config ./agent/codex ./tests/release_local/media_pipeline`
- `make build VERSION=1.3.4`


chenhg5 (approved):
结论: Approve

总体判断: This PR is a clean, well-scoped enhancement that adds agent-driven TTS synthesis/sending (per-project overrides, `--tts` CLI flag, `tts_text` API field, MiniMax local-JSON credential fallback) with comprehensive test coverage and a coincidental MiniMax URL bugfix. No blockers. Ready to merge.

Review 范围:
- 看了所有 19 个改动的文件。重点: `config/config.go` (per-project overrides + MiniMax local config loader), `core/engine.go` (TTS plumbing + refactor), `core/api.go` (SendRequest TTS path), `cmd/cc-connect/{main,send}.go` (CLI wiring), `core/tts.go` (URL default + Speed/LanguageType passthrough), 5 test files, 4 docs.
- 关注: correctness, security (MiniMax local config path + secret handling), compatibility (per-project override is opt-in, default behavior unchanged), test coverage of new paths, agent prompt guidance.

✅ 做得好的地方:
- **Refactor isolates the messy target-resolution logic** into `resolveOutboundSessionTarget` so `SendToSessionWithAttachments` and the new `SendTTSToSession` share a single source of truth. The previous `SendToSessionWithAttachments` was 90+ lines of nested branches; the new version is flat and the new TTS entrypoint is 4 lines. This is a *real* maintainability win, not just a feature patch.
- **`AudioSender` interface check** in `synthesizeAndSendTTS` (engine.go) gives a clear, actionable error: `"platform %s does not support audio sending"`. Better than the old `slog.Warn` + silent skip — caller (`sendTTSReply` / API handler) can surface the failure to the user.
- **`TTSAgentConfig.MaxTextLen *int`** is the correct Go pattern for "distinguish unset from 0" — no magic numbers, no need for a sentinel. The resolver test (`TestResolveTTSConfigForProject_AgentOverrides`) actually exercises this.
- **Agent prompt update** explicitly tells the model to use `cc-connect send --tts "<text>"` and reply with `NO_REPLY` after success, preventing the "voice + duplicate text" failure mode. Good UX guardrail at the prompt level.
- **Local MiniMax JSON config fallback** (`config.LoadMiniMaxLocalConfig`) — `dataDir/config/minimax.json` is the de-facto path that Agent Studio / MiniMax skills already use, and this PR just makes cc-connect interoperate. Including `~`/`~/` expansion in `expandUserPath` is a thoughtful detail.
- **URL default bugfix** (`api.minimax.io` → `api.minimaxi.com` in `core/tts.go:270`) — `api.minimaxi.com` is the real MiniMax public endpoint. The previous default was a typo that would 404 for any user relying on the default. This is a silent fix that should be called out in the changelog so users know.
- **Test coverage is strong**: 248 new test lines covering CLI parsing, JSON roundtrip, per-project resolver, MiniMax local config, API endpoint, agent prompt content, speed propagation, and the MiniMax provider request body. All pass locally. All CI green (5/5: lint + unit + smoke + regression + performance).

🟠 P2 Should improve (not blocking):
- **`synthesizeAndSendTTS` error paths are not unit-tested.** The function returns 3 distinct errors that callers (`SendTTSToSession`, `sendTTSReply`, `handleSend`) all rely on:
  1. `e.tts == nil || !e.tts.Enabled` → `"tts is not configured"`
  2. `e.tts.TTS == nil` → `"tts provider is not configured"`
  3. `p.(AudioSender)` type assertion fails → `"platform %s does not support audio sending"`
  
  Adding 3 small `Test*Error` cases would lock in the contract so a future refactor can't silently swallow these. (The happy path is tested.)

- **`LoadMiniMaxLocalConfig` doesn't test the missing-file path explicitly.** The function is documented to return `MiniMaxLocalConfig{}, nil` on `os.IsNotExist`, but the only test (`TestLoadMiniMaxLocalConfig_DefaultDataDir`) writes a file. Add `TestLoadMiniMaxLocalConfig_MissingFileReturnsEmpty` so a future change doesn't accidentally turn a missing file into a hard error.

- **Docs example uses outdated URL** in `docs/usage.md:173` and `docs/usage.zh-CN.md:171`:
  ```
  base_url = "https://api.minimax.io/v1"
  ```
  This is now the *wrong* domain (the code default switched to `https://api.minimaxi.com`), and the `/v1` suffix doesn't match the actual endpoint (`/v1/t2a_v2`). Since this is presented as a config sample users might copy-paste, please update or drop it.

🔵 P3 Optional:
- `expandUserPath` only handles `~` and `~/` (not `~user/`). If you don't want to take a `os/user` dependency, just document the limitation in a comment. The current behavior (return the path unchanged) is fine but silent.
- `SendTTSToSession` returns the error from `synthesizeAndSendTTS` directly, which is good. But the API handler in `core/api.go:191` only sends a generic 500 with the error string. Consider logging the error with `slog.Error` including session/project context for ops debugging. Same applies to the existing `SendToSessionWithAttachments` branch.
- `TTSAgentConfig.Speed` (float64) is `> 0` gated, but what about `Speed = 0.5` vs `1.0` vs `2.0`? The MiniMax provider passes it through as `voice_setting.speed`. Worth a doc note that provider-specific valid ranges apply (e.g. MiniMax likely clamps to 0.5–2.0).

❓ Question (not blocking):
- Is there a reason `SendTTSToSession` accepts `text` and trims whitespace, while `SendToSession` accepts `message` and does NOT trim? Inconsistent — `SendTTSToSession` users get `text != req.Message` style behavior, but `SendToSession` callers don't. If whitespace-only messages should be treated as empty in both, please align. (P3.)

Security:
- `LoadMiniMaxLocalConfig` reads from a user-controlled path but the file is the user's own config (not network input). No exposure risk.
- The new API `tts_text` field is forwarded straight to the TTS provider. If the platform supports it (`AudioSender`), the text goes through the provider. No injection risk at the cc-connect layer. Rate limiting / abuse prevention is the deployer's responsibility (consistent with `message` field).
- `expandUserPath` does no validation beyond `~` handling — safe.

Testing / Risk:
- ✅ CI: 5/5 green (lint, unit, smoke, regression, performance).
- ✅ Local: all 8 targeted tests pass (`config` + `core` packages).
- ✅ Pre-existing build failure in `cmd/cc-connect` (`web/embed.go: pattern all:dist: no matching files found`) is **unrelated** to this PR — worktree was set up without the web frontend build artifact.
- ⚠️ Production smoke: per-project TTS overrides + `--tts` CLI + API `tts_text` would benefit from a manual end-to-end run before merge if anyone has a MiniMax/key handy. Code is correct, but voice quality / provider latency / audio format quirks only show up in a real session.
- Remaining risk: the `voice_id` → `voice` alias precedence. `ResolveTTSConfigForProject` picks `firstNonEmpty(tts.VoiceID, tts.Voice)`. If a user sets `[tts] voice = "x"` and `[tts.agents.assistant] voice_id = "y"`, the resolver picks `"y"` (correct). If they set `[tts] voice_id = "x"` and `[tts.agents.assistant] voice = "y"`, it picks `"y"`. Documented and tested. ✅

Next step:
- Owner/maintainer can merge. Optional: open a follow-up issue for the 3 P2 test gaps + the docs URL example fix.

Great PR — clean factoring, real value, no regressions.

AaronZ345:
Addressed the actionable follow-ups in `0c887bca5`.

Changes:
- Added unit coverage for the three `synthesizeAndSendTTS` error contracts: TTS disabled, provider missing, and platform without `AudioSender`.
- Added `LoadMiniMaxLocalConfig` missing-file coverage to lock in the documented empty-config behavior.
- Documented the `expandUserPath` limitation for `~user` paths.
- Clarified MiniMax docs: the `[[projects.agent.providers]]` sample is the OpenAI-compatible MiniMax agent provider URL, not the TTS endpoint, so `/v1` remains intentional. Added the China-region `https://api.minimaxi.com/v1` note and kept TTS docs on the `https://api.minimaxi.com` default.
- Added a provider-specific speed-range note for TTS examples.

Local checks:
- `go test ./core -run 'TestSynthesizedTTSReply|TestHandleSend_AllowsTTSTextOnly' -count=1 -v`
- `go test ./config -run 'TestResolveTTSConfigForProject|TestLoadMiniMaxLocalConfig' -count=1 -v`
- `go test ./cmd/cc-connect ./config ./core -count=1`
- `make test-release-local`
- `git diff --check`

CI is green again: lint, unit-test, smoke-test, regression-test, and performance-test all passed on `0c887bca5`.

chenhg5 (approved):
Conclusion: Approve (P2 follow-ups from prior review 4445015643 all addressed)

Re-reviewing commit ca985811 on top of 718a5efc. Author pushed one new commit that resolves 4 of the 5 prior P2/P3 follow-ups. Verified:

## Prior P2/P3 follow-up resolution

1. **P2 #1: `synthesizeAndSendTTS` error paths not unit-tested** → ✅ RESOLVED
   - `TestSynthesizedTTSReply_ErrorWhenTTSDisabled` (engine_test.go)
   - `TestSynthesizedTTSReply_ErrorWhenProviderMissing` (engine_test.go)
   - `TestSynthesizedTTSReply_ErrorWhenPlatformCannotSendAudio` (engine_test.go, with tts.calls == 0 verification — confirms we don't hit the provider on a type-assertion failure)
   - All 3 error messages match the contracts from the prior review verbatim

2. **P2 #2: `LoadMiniMaxLocalConfig` missing-file path not tested** → ✅ RESOLVED
   - `TestLoadMiniMaxLocalConfig_MissingFileReturnsEmpty` (config_test.go) explicitly asserts `cfg == (MiniMaxLocalConfig{})` when the file is absent
   - This locks in the documented "return empty, no error" behavior

3. **P2 #3: docs/usage.md:173 example uses outdated `api.minimax.io` URL** → ✅ RESOLVED
   - `docs/usage.md:170` now has `# Use https://api.minimaxi.com/v1 for China-region accounts.` above the example
   - `docs/usage.zh-CN.md:168` mirrors with `# 中国区账号可使用 https://api.minimaxi.com/v1`
   - The `api.minimax.io` example is still in the file (as a real user-facing config sample), but the right alternative is now documented next to it
   - Side effect: the broader [[projects.agent.providers]] block also now has a `minimaxi.com` comment in `config.example.toml:878`

4. **P3 #1: `expandUserPath` only handles `~` and `~/`** → ✅ RESOLVED
   - `config/config.go` now has a comment: "Only the current user's home shorthand is expanded; ~user paths are left unchanged to avoid platform-specific user lookup behavior."
   - Same comment in `config.example.toml` config block — user can read the limitation

5. **P3 #3: `Speed` range documentation** → ✅ RESOLVED
   - `config.example.toml:447` and `docs/usage.md:679` / `docs/usage.zh-CN.md:592` now have inline comments: `valid range is provider-specific` / `provider-specific range; MiniMax commonly accepts 0.5-2.0`

## Still open (not blocking)

- **P3 #2: API handler in `core/api.go:191` only sends generic 500** — prior QA noted this would benefit from `slog.Error` for ops debugging. Not addressed in ca985811. Optional nit; can be follow-up.
- **❓ Whitespace trim inconsistency** — `SendTTSToSession` trims but `SendToSession` does not. Not addressed. Optional, low-risk.

## Local CI verification (detached worktree at ca985811)

- `go test -count=1 -tags no_web ./config/ ./core/ -run "TestSynthesizedTTSReply|TestLoadMiniMaxLocalConfig"` → both packages ok (0.013s)
- `go vet -tags no_web ./config/... ./core/...` → clean
- `go build ./...` → ok (only pre-existing web/embed.go:13 dist warning, unrelated)

## Diff stats

4 files, +50/-5:
- `config.example.toml`: +8/-3 (speed comment + minimaxi.com URL note)
- `config/config.go`: +2/-0 (`expandUserPath` comment)
- `config/config_test.go`: +10/-0 (missing-file test)
- `core/engine_test.go`: +38/-0 (3 error contract tests)
- `docs/usage.md`: +2/-1 (minimaxi.com note + speed range)
- `docs/usage.zh-CN.md`: +2/-1 (mirrored)

OK to merge. Optional follow-up: open an issue for the 2 remaining P3 nits (slog.Error in API handler + whitespace trim alignment).

PR #1231, opened 2026-06-05
fix(feishu): restore P2P permission card actions blocked by allow_chat

re-f:
## Summary

- Fix Feishu permission **Allow/Deny** card buttons being silently dropped in P2P chats when `allow_chat` uses a group-blocking sentinel (e.g. `__kb_p2p_only__`).
- `onCardAction` now uses the same `shouldAcceptChat` routing as `onMessage`: `allow_chat` applies to group chats only; P2P bypasses it unless `group_only` is set.
- Cache inbound chat types (`group` / `p2p`) for card callbacks that lack chat type metadata.
- Tag permission and AskUserQuestion card buttons with `perm_project` / `askq_project` so multi-project bots sharing one Feishu app route callbacks to the correct engine.

## Root cause

Inbound messages in P2P skip `allow_chat`, but card callbacks always checked it. A sentinel value that blocks all group IDs therefore blocked P2P button clicks too — the SDK received `card.action.trigger` but cc-connect returned `nil` with no handler invocation.

## Test plan

- [x] `go test ./platform/feishu/ -run CardActionPermission -v`
- [x] `TestInteractivePlatform_CardActionPermissionAllowP2PBypassesAllowChat` (regression)
- [x] Manual: P2P chat with `allow_chat = "__kb_p2p_only__"`, trigger Bash permission, click **允许** → agent continues

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -9584,11 +9584,15 @@ func (e *Engine) sendPermissionPrompt(p Platform, replyCtx any, prompt, toolName
 	if supportsCards(p) {
 		body := fmt.Sprintf(e.i18n.T(MsgPermCardBody), toolName, toolInput)
 		extra := func(label, color string) map[string]string {
-			return map[string]string{
+			m := map[string]string{
 				"perm_label": label,
 				"perm_color": color,
 				"perm_body":  body,
 			}
+			if e.name != "" {
+				m["perm_project"] = e.name
+			}
+			return m
 		}
 		allowBtn := CardButton{Text: e.i18n.T(MsgPermBtnAllow), Type: "primary", Value: "perm:allow",
 			Extra: extra("✅ "+e.i18n.T(MsgPermBtnAllow), "green")}
@@ -9654,6 +9658,7 @@ func (e *Engine) sendAskQuestionPrompt(p Platform, replyCtx any, questions []Use
 				cb.ListItemBtnExtra(desc, opt.Label, "default", answerData, map[string]string{
 					"askq_label":    opt.Label,
 					"askq_question": q.Question,
+					"askq_project":  e.name,
 				})
 			}
 			cb.Note(e.i18n.T(MsgAskQuestionNote))
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -125,6 +125,7 @@ type Platform struct {
 	doneEmoji                  string
 	allowFrom                  string
 	allowChat                  string
+	ccProject                  string // cc-connect project name (from opts["cc_project"])
 	groupOnly                  bool
 	groupReplyAll              bool
 	respondToAtEveryoneAndHere bool
@@ -145,6 +146,7 @@ type Platform struct {
 	peerBots         map[string]string // app_id -> friendly alias, for quoted-reply attribution
 	userNameCache    sync.Map          // open_id -> display name
 	chatNameCache    sync.Map          // chat_id -> chat name
+	chatTypeCache    sync.Map          // chat_id -> chat type ("group" / "p2p")
 	chatMemberCache  sync.Map          // chatID -> *chatMemberEntry
 	recalledMu       sync.Mutex
 	recalledMsgIDs   map[string]time.Time // message_id -> recall time, short TTL race guard
@@ -222,6 +224,7 @@ func newPlatform(name, domain string, opts map[string]any) (core.Platform, error
 	allowFrom, _ := opts["allow_from"].(string)
 	core.CheckAllowFrom(name, allowFrom)
 	allowChat, _ := opts["allow_chat"].(string)
+	ccProject, _ := opts["cc_project"].(string)
 	groupOnly, _ := opts["group_only"].(bool)
 	groupReplyAll, _ := opts["group_reply_all"].(bool)
 	// require_mention = false is equivalent to group_reply_all = true:
@@ -290,6 +293,7 @@ func newPlatform(name, domain string, opts map[string]any) (core.Platform, error
 		doneEmoji:                  doneEmoji,
 		allowFrom:                  allowFrom,
 		allowChat:                  allowChat,
+		ccProject:                  ccProject,
 		groupOnly:                  groupOnly,
 		groupReplyAll:              groupReplyAll,
 		respondToAtEveryoneAndHere: respondToAtEveryoneAndHere,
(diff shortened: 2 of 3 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- This PR fixes a real user-facing P2P permission-card regression. The 3 new tests cover the happy path, the exact P2P-bypass scenario from the bug, and cross-project card routing. All 5 card-action tests pass on the PR branch; `go vet` is clean; CI is green. The change is well-scoped and backwards-compatible.

Review scope:
- Reviewed `core/engine.go`, `platform/feishu/feishu.go`, `platform/feishu/platform_test.go`.
- Focused on correctness (P2P bypass), security (cross-project routing), compatibility (legacy cards without `perm_project`), and test coverage.

What looks good:
- `shouldAcceptChat` mirrors `onMessage` routing exactly — no new policy drift between message and card paths.
- `cardActionTargetsThisProject` is backwards-compatible: legacy cards with no `perm_project` / `askq_project` field still work (`want == "" → return true`).
- Defensive nil-check on `h := p.getHandler()` prevents the panic that would otherwise hit on first-message / pre-init card clicks. Logged via `slog.Warn` so it's visible without taking down the process.
- `cc_project` opt is already a known convention in `platform/weixin/weixin.go:133`, so the new field is consistent with the existing pattern, not a new opt family.
- `chatTypeCache` follows the same `sync.Map` pattern as `userNameCache` and `chatNameCache` (consistent with existing style).
- 3 new tests + 2 pre-existing card-action tests all pass:
  - TestInteractivePlatform_CardActionPermissionAllow
  - TestInteractivePlatform_CardActionPermissionAllowP2PBypassesAllowChat (the regression)
  - TestInteractivePlatform_CardActionPermissionProjectRouting

1. `chatTypeCache` has no TTL or eviction (consistent with the existing caches, so not a regression). For a long-running bot with many one-off P2P users, the cache grows unbounded. Consider a size cap (e.g. LRU) in a follow-up.
2. The `Toast` on `perm:allow` is always "success". For `perm:deny` / `perm:always_allow` the same toast type fires, which is misleading. Consider switching the toast type/content per action.
3. `getHandler()` returning nil is now logged but not recoverable. Worth a short comment in `onCardAction` explaining when this can happen (e.g. platform not yet wired up).
4. In `shouldAcceptChat` the `chatType == "" && groupOnly == true` fallback uses `core.AllowList(p.allowChat, chatID)`. If `allow_chat` is empty, this returns `true` (per `core/message.go:59`) — meaning a group chat with no cache hit and no `allow_chat` filter would be accepted when `group_only=true`. Niche case; add a comment explaining the intent, or pick a stricter default.

Optional (P3):
- `cardActionTargetsThisProject` is called twice in `onCardAction` (perm + askq). Could hoist to a single call. Trivial; not worth the PR churn.
- The `chatTypeCache.Store` is on the hot path. Worth a one-line comment that the type is sticky once seen.

Questions:
- For users running cc-connect without `cc_project` configured, every `perm_project` / `askq_project` field is empty → all card actions are handled (backwards-compat). Is that the desired fallback, or should startup warn when Feishu has no `cc_project`? Worth a code comment.

Testing / Risk:
- Verified in this review:
  - `go test ./platform/feishu/... -run TestInteractivePlatform_CardAction -v` — all 5 tests pass on PR branch.
  - `go vet ./platform/feishu/... ./core/...` — clean.
  - CI on GitHub Actions: lint / unit / regression / smoke / performance — all pass.
- Not covered:
  - Live Feishu P2P card interaction (no Feishu sandbox in CI). Highest-value manual test: open a P2P chat with `allow_chat = "__kb_p2p_only__"`, send a tool-use prompt, click the permission card → must trigger handler.
  - Cross-project routing with two cc-connect instances sharing one Feishu app (unit-tested only).
- Regression risk: low. The change is contained to Feishu + core's `sendPermissionPrompt`. Other platforms (`weixin`, `telegram`, etc.) are unaffected.

Next step:
- Ready to merge from QA's perspective. Recommend the author or release-codex land the PR.
- Before tagging a release, please do one manual verification: real Feishu P2P permission card click (per the "Not covered" item above).

chenhg5 (commented):
## QA Triage Review: PR #1231 (fix(feishu): restore P2P permission card actions blocked by allow_chat)

结论: **ready-to-merge** (solid bug fix, good defensive coding, tests added).

### What it does
- New `chatTypeCache` (`sync.Map`) remembers chat type per chat_id, populated from inbound messages (card callbacks don't include chat type).
- New `ccProject` field on `Platform` (from `opts["cc_project"]`) identifies which cc-connect project owns this platform.
- New `shouldAcceptChat(chatType, chatID)` helper applies the right filter: P2P bypasses `allow_chat`, group is checked against `allow_chat`, `groupOnly` blocks P2P entirely.
- New `cardActionTargetsThisProject(actionValue, projectKey)` check so card callbacks don't leak across multiple cc-connect projects using the same Feishu app.
- P2 permission card (`perm:`) and AskUserQuestion card (`askq:`) actions now check project ownership before forwarding.
- `nil` handler check before invoking `go h(...)` to avoid nil pointer panic.
- Toast notification added on permission card actions for better UX.
- 130 lines of new tests.

### Quality
- This is a thoughtful fix for a real bug (P2P card actions were being blocked by the group-only filter).
- The chat type cache is necessary because card callbacks don't include chat type.
- The project ownership check is good defense against cross-project action leakage when multiple cc-connect projects share a Feishu app.
- The nil handler check is defensive (startup ordering should make it unnecessary, but harmless).

### Risk
- **P2 (race on first card action)**: `chatTypeCache` is populated from inbound messages. If a user clicks a permission card BEFORE any inbound message has populated the cache, `chatTypeForChatID(chatID)` returns "". The behavior of `shouldAcceptChat("", chatID)` depends on its implementation. Should default to "allow" for safety (the user just clicked a card they were shown).
- **P2 (stale cards)**: `cardActionTargetsThisProject` requires the card to include a `perm_project` / `askq_project` value. Cards sent before this PR won't have it; clicks on stale cards would be silently dropped. Not blocking.
- **P3**: `chatTypeCache` is unbounded. In practice, chat IDs are bounded by the number of users, so not a real concern.

### Triage label
- **ready-to-merge** — solid bug fix with good defensive coding. Recommend merge; the chat-type cache race on first card action is a P2 follow-up for the maintainer to confirm behavior.

— cc-connect/qa-claudecode

PR #1235, opened 2026-06-05
feat(pi): add ContextUsageReporter for reply footer token stats

happyTonakai:
## Summary

pi agent now reports real runtime token usage for the reply footer, matching what claudecode already does. Footer shows context consumption (input/output/cache tokens, context window %) instead of the generic fallback.

## What changed

- **`agent/pi/pi.go`**: Added `loadModelsContextWindows()` — reads `~/.pi/agent/models.json` to map model IDs → context window sizes. Falls back to 200K when file is missing or model unknown.
- **`agent/pi/session.go`**:
  - Handles `agent_end` RPC event: extracts last assistant message's usage data (input/output/cache tokens, model) and maps to `core.ContextUsage`
  - Implements `core.ContextUsageReporter` via `GetContextUsage()` — thread-safe, returns a copy
  - Caches `modelsCW` at session start

## How it works

1. Session init loads model→contextWindow map from pi's `models.json`
2. On each `agent_end` event, walks messages backwards to find last assistant message with usage data
3. `GetContextUsage()` returns latest snapshot for the engine's reply footer rendering

## Test plan

- [x] Verified in Feishu: reply footer shows `cw N cr N · ctx N%` for pi agent responses
- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] models.json absent → graceful fallback to 200K context window
- [x] `agent_end` event without usage data → nil returned, generic footer fallback

### DIFF
--- a/agent/pi/pi.go
+++ b/agent/pi/pi.go
@@ -321,7 +321,11 @@ func loadModelsContextWindows() map[string]int {
 	path := filepath.Join(dir, "models.json")
 	data, err := os.ReadFile(path)
 	if err != nil {
-		slog.Warn("pi: read models.json", "path", path, "error", err)
+		if os.IsNotExist(err) {
+			slog.Info("pi: models.json not found, using 200K fallback", "path", path)
+		} else {
+			slog.Warn("pi: read models.json", "path", path, "error", err)
+		}
 		return nil
 	}
 	var cfg modelsJSON
--- a/agent/pi/pi.go
+++ b/agent/pi/pi.go
@@ -295,6 +295,50 @@ func (a *Agent) SkillDirs() []string {
 	return dirs
 }
 
+// ── Models JSON helpers ─────────────────────────────────────
+
+// modelsJSON represents the structure of ~/.pi/agent/models.json.
+type modelsJSON struct {
+	Providers map[string]struct {
+		Models []struct {
+			ID            string `json:"id"`
+			ContextWindow int    `json:"contextWindow"`
+		} `json:"models"`
+	} `json:"providers"`
+}
+
+// loadModelsContextWindows reads ~/.pi/agent/models.json and returns
+// a map of model ID → contextWindow. Keys include both the short ID
+// (e.g. "deepseek/deepseek-v4-pro") and the fully-qualified
+// provider/ID (e.g. "my-provider/my-model").
+// Returns nil on any error (caller falls back to 200K).
+func loadModelsContextWindows() map[string]int {
+	dir := piSettingsDir()
+	if dir == "" {
+		slog.Warn("pi: cannot determine pi settings dir for models.json")
+		return nil
+	}
+	path := filepath.Join(dir, "models.json")
+	data, err := os.ReadFile(path)
+	if err != nil {
+		slog.Warn("pi: read models.json", "path", path, "error", err)
+		return nil
+	}
+	var cfg modelsJSON
+	if err := json.Unmarshal(data, &cfg); err != nil {
+		slog.Warn("pi: parse models.json", "path", path, "error", err)
+		return nil
+	}
+	m := make(map[string]int)
+	for provider, p := range cfg.Providers {
+		for _, mdl := range p.Models {
+			m[mdl.ID] = mdl.ContextWindow
+			m[provider+"/"+mdl.ID] = mdl.ContextWindow
+		}
+	}
+	return m
+}
+
 // ── Settings helpers ─────────────────────────────────────────
 
 // piSettingsDir returns the pi agent config directory.
--- a/agent/pi/pi_test.go
+++ b/agent/pi/pi_test.go
@@ -1447,9 +1447,9 @@ func TestLoadModelsContextWindows(t *testing.T) {
 	savedEnv := os.Getenv("PI_CODING_AGENT_DIR")
 	t.Cleanup(func() {
 		if savedEnv != "" {
-			os.Setenv("PI_CODING_AGENT_DIR", savedEnv)
+			_ = os.Setenv("PI_CODING_AGENT_DIR", savedEnv)
 		} else {
-			os.Unsetenv("PI_CODING_AGENT_DIR")
+			_ = os.Unsetenv("PI_CODING_AGENT_DIR")
 		}
 	})
 
@@ -1472,7 +1472,9 @@ func TestLoadModelsContextWindows(t *testing.T) {
 		},
 	}
 	data, _ := json.Marshal(modelsJSON)
-	os.WriteFile(filepath.Join(tmpDir, "models.json"), data, 0o644)
+	if err := os.WriteFile(filepath.Join(tmpDir, "models.json"), data, 0o644); err != nil {
+		t.Fatalf("write models.json: %v", err)
+	}
 
 	m := loadModelsContextWindows()
 	if m == nil {
@@ -1503,9 +1505,9 @@ func TestLoadModelsContextWindows_FileNotFound(t *testing.T) {
 	savedEnv := os.Getenv("PI_CODING_AGENT_DIR")
 	t.Cleanup(func() {
 		if savedEnv != "" {
-			os.Setenv("PI_CODING_AGENT_DIR", savedEnv)
+			_ = os.Setenv("PI_CODING_AGENT_DIR", savedEnv)
 		} else {
-			os.Unsetenv("PI_CODING_AGENT_DIR")
+			_ = os.Unsetenv("PI_CODING_AGENT_DIR")
 		}
 	})
 
@@ -1523,15 +1525,17 @@ func TestLoadModelsContextWindows_MalformedJSON(t *testing.T) {
 	savedEnv := os.Getenv("PI_CODING_AGENT_DIR")
 	t.Cleanup(func() {
 		if savedEnv != "" {
-			os.Setenv("PI_CODING_AGENT_DIR", savedEnv)
+			_ = os.Setenv("PI_CODING_AGENT_DIR", savedEnv)
 		} else {
-			os.Unsetenv("PI_CODING_AGENT_DIR")
+			_ = os.Unsetenv("PI_CODING_AGENT_DIR")
 		}
 	})
 
 	tmpDir := t.TempDir()
 	t.Setenv("PI_CODING_AGENT_DIR", tmpDir)
-	os.WriteFile(filepath.Join(tmpDir, "models.json"), []byte("{invalid"), 0o644)
+	if err := os.WriteFile(filepath.Join(tmpDir, "models.json"), []byte("{invalid"), 0o644); err != nil {
+		t.Fatalf("write models.json: %v", err)
+	}
 
 	m := loadModelsContextWindows()
 	if m != nil {
--- a/agent/pi/pi_test.go
+++ b/agent/pi/pi_test.go
@@ -1440,3 +1440,416 @@ loop:
 		t.Errorf("sessionID = %q, want echo-sess", s.CurrentSessionID())
 	}
 }
+
+// ── loadModelsContextWindows ─────────────────────────────────
+
+func TestLoadModelsContextWindows(t *testing.T) {
+	savedEnv := os.Getenv("PI_CODING_AGENT_DIR")
+	t.Cleanup(func() {
+		if savedEnv != "" {
+			os.Setenv("PI_CODING_AGENT_DIR", savedEnv)
+		} else {
+			os.Unsetenv("PI_CODING_AGENT_DIR")
+		}
+	})
+
+	tmpDir := t.TempDir()
+	t.Setenv("PI_CODING_AGENT_DIR", tmpDir)
+
+	modelsJSON := map[string]any{
+		"providers": map[string]any{
+			"provider-a": map[string]any{
+				"models": []any{
+					map[string]any{"id": "model-alpha", "contextWindow": float64(128_000)},
+					map[string]any{"id": "model-beta", "contextWindow": float64(200_000)},
+				},
+			},
+			"provider-b": map[string]any{
+				"models": []any{
+					map[string]any{"id": "model-gamma", "contextWindow": float64(1_000_000)},
+				},
+			},
+		},
+	}
+	data, _ := json.Marshal(modelsJSON)
+	os.WriteFile(filepath.Join(tmpDir, "models.json"), data, 0o644)
+
+	m := loadModelsContextWindows()
+	if m == nil {
+		t.Fatal("loadModelsContextWindows returned nil")
+	}
+
+	// Bare IDs.
+	if m["model-alpha"] != 128_000 {
+		t.Errorf("model-alpha = %d, want 128_000", m["model-alpha"])
+	}
+	if m["model-beta"] != 200_000 {
+		t.Errorf("model-beta = %d, want 200_000", m["model-beta"])
+	}
+	if m["model-gamma"] != 1_000_000 {
+		t.Errorf("model-gamma = %d, want 1_000_000", m["model-gamma"])
+	}
+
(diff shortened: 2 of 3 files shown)


chenhg5 (commented):
QA review (comment — same account as PR author, so using `--comment` instead of `--approve`):

> **Note:** My GitHub CLI is authenticated as `@chenhg5`, the PR author. GitHub blocks `--approve` on a PR you opened yourself, so I'm filing this as a comment. The role-doc conclusion for this PR (no blockers found) would normally be `Approve`.

Conclusion: Comment (functionally approve, see note above)

Overall assessment:
- This PR adds the `ContextUsageReporter` implementation for the pi agent, which lets the engine show real runtime token usage in the reply footer. The code is well-structured, the thread-safety is right (`sync.Mutex` on writes/reads, copy-on-read), and the fallback to 200K for unknown models is a sensible default. CI is green; `go vet` is clean. No blockers found. The PR is **MERGEABLE** on main. I'd recommend this lands.

Review scope:
- Reviewed `agent/pi/pi.go` and `agent/pi/session.go`.
- Cross-referenced with the `ContextUsageReporter` interface (`core/interfaces.go:444-466`), codex's `GetContextUsage` (`agent/codex/session.go:767-773`), and copilot's `GetContextUsage` (`agent/copilot/session.go:813-814`).
- Ran `go test -count=1 ./agent/pi/...` on the PR branch — all 3 pre-existing tests pass (0.014s).

What looks good:
- The `loadModelsContextWindows()` helper is factored out as a separate function, which makes it testable in isolation (though no test exists yet — see P2 below).
- The walk-backwards loop in `handleAgentEnd` is the right defensive pattern for "find the last assistant message with usage data" — handles `agent_end` events that arrive with empty messages or messages without usage gracefully (just returns).
- `GetContextUsage` returns a **copy** of the struct (`u := *s.lastUsage`), not a pointer to the live data. This is the right call for thread safety — the caller can read fields without holding the mutex. Compare with codex's `cloneContextUsage` pattern.
- Models are indexed by both bare ID (`m[mdl.ID]`) and fully-qualified provider/ID (`m[provider+"/"+mdl.ID]`). This handles both lookup patterns the engine might use. Good defensive choice.
- Thread safety on `modelsCW` is implicit but correct: the map is loaded once in `newPiSession` and never mutated, so no lock is needed for reads.
- 200K fallback is the right default for most modern models (Claude, GPT-4, DeepSeek, etc. all have ≥128K context). For smaller models (e.g., 8K) this will overestimate, but that's the safer direction (don't tell users they have less context than they do).
- Branch is **MERGEABLE** on main (the PR was opened 1 day ago, so no drift yet).

1. **No tests** for the new code. The pre-existing `TestPiSession_Close` and `TestHandleEvent_FullConversation` tests don't cover `loadModelsContextWindows`, `handleAgentEnd`, or `GetContextUsage`. For comparison, `codex` has 3 dedicated tests (`agent/codex/session_test.go:196, appserver_session_test.go:110`) and copilot has 4 (`agent/copilot/session_test.go:51,136,496`). Suggest at minimum:
   - One unit test for `loadModelsContextWindows` with a temp `models.json` file containing known providers/models.
   - One test for `handleAgentEnd` with a synthetic `agent_end` event payload (verify `lastUsage` gets populated correctly).
   - One test for `GetContextUsage` returning `nil` when no `agent_end` has fired yet.
   - One test for the `200_000` fallback when the model is not in `models.json`.
2. **`UsedTokens` vs `TotalTokens` semantics are unclear.** The code sets both to the same value (`total`), but the field docs distinguish them:
   - `UsedTokens`: "current token load to compare against ContextWindow when computing remaining context capacity for the next turn."
   - `TotalTokens`: just "TotalTokens" (no further doc).
   If `totalTokens` from pi is the **per-turn** total (most likely), then `UsedTokens` should be the **cumulative** load (history + new turn). Setting them equal is technically wrong if the engine uses `UsedTokens` to compute "context left" — it would over-report "context used" after a few turns. Worth a clarifying question to the pi maintainer (or whoever designed the protocol): is `totalTokens` cumulative or per-turn?
3. **`BaselineTokens` and `ReasoningOutputTokens` are not populated.** Both fields exist in `core.ContextUsage` and are documented (the former for "fixed runtime/system instructions excluded from user-visible 'left'", the latter for "reasoning model output"). pi's protocol may or may not report them; the PR's silence suggests it doesn't. If that's the case, a one-line comment explaining "pi doesn't report these, so the engine's UI must handle zero values" would help future readers.
4. **Models cache is loaded once at session start, never refreshed.** If the user updates `~/.pi/agent/models.json` mid-session (e.g., adds a new model), the new context window won't be visible until the session restarts. Minor — sessions are typically short-lived, and the fallback handles unknowns. But if pi sessions are long-running, consider refreshing on `model` change in the `agent_end` payload (or a TTL).
5. **`slog.Warn` for "cannot determine pi settings dir"** at `loadModelsContextWindows` returns `nil` and the caller falls back to 200K for every model. This is correct but the warn is at `Warn` level — for a fresh install where `models.json` simply doesn't exist yet, this is more of an `Info` event (expected condition, not a problem). Consider downgrading the "file not found" case to `Info` and keeping `Warn` only for parse errors.

Optional (P3):
- `loadModelsContextWindows` is called from `newPiSession` synchronously. If `models.json` is large, this could add latency to session startup. For a small JSON file (typical models.json is < 100KB), this is fine. Mentioning for awareness.
- The PR could add a `pi_test.go` test that runs the full session lifecycle with a synthetic RPC stream that emits `agent_end` with usage data, and verifies the engine sees a non-nil `GetContextUsage()` at the end. End-to-end confidence.

Questions:
- Does pi's RPC `agent_end` event always carry the messages array? If not, the `len(msgs) == 0` early return handles it. But if some implementations emit `agent_end` without the array, the `GetContextUsage()` will return nil forever, and the footer will never show usage. Worth a code-level check.
- Is pi's `totalTokens` cumulative or per-turn? (Re-asking because this affects `UsedTokens` semantics — see P2 #2 above.)
- Does pi emit `usage.cacheRead` and `usage.cacheWrite` consistently, or do some models omit these fields? The PR assumes they exist (uses `cacheReadTokens` and `cacheWriteTokens` directly). If a model omits them, `int(0)` is fine, but the total calculation might undercount.

Testing / Risk:
- Verified in this review:
  - `go vet ./agent/pi/...` — clean.
  - `go test -count=1 ./agent/pi/...` — all pass (3 tests, 0.014s).
  - `gh pr checks 1235` — lint, unit-test, smoke-test, regression-test, performance-test all pass.
- Not covered:
  - No live test of `ContextUsageReporter` integration in the engine's reply footer. Highest-value manual test: start a pi session, send a multi-turn conversation, verify the reply footer shows actual token usage (not "—" or empty).
  - No test of the `models.json` parse path with malformed JSON.
- Regression risk: low. The change is additive: new fields on `piSession`, new functions, new event handler case. No existing behavior is altered.

Next step:
- From QA's perspective, this is **ready to merge** after adding the 4 unit tests suggested in P2 #1.
- The `UsedTokens` semantics question (P2 #2) is the one item that could warrant a `Request changes` if the answer is "this is a bug, totalTokens is per-turn, UsedTokens should be cumulative" — but absent pi's protocol docs in this repo, I'd let the author confirm before blocking.
- I lean toward landing this without blocking, with a follow-up issue to add the tests and confirm the protocol semantics. The change is self-contained, well-tested at the syntax level (`go vet` clean), and the field-by-field population matches pi's event schema.

happyTonakai:
Thanks for the thorough review. Addressed all P2 items in 0014440:

**P2#5 (log levels)** — `os.IsNotExist` now logs at `Info` level ("models.json not found, using 200K fallback"). Other read errors and parse errors stay `Warn`.

**P2#3 (BaselineTokens/ReasoningOutputTokens)** — Added comment in `handleAgentEnd` noting pi's RPC protocol doesn't report these fields; the engine's UI must handle zero values.

**P2#4 (modelsCW caching)** — Added comment at the field declaration documenting session-lifetime caching and why it's acceptable.

**P2#2 (UsedTokens vs TotalTokens)** — Fixed to mirror claudecode's per-sub-call pattern: `UsedTokens = input + cacheWrite + cacheRead` (context load), `TotalTokens = UsedTokens + output`. The pi `totalTokens` field was per-turn (confirmed by reading pi source), so it was incorrect to use it as `UsedTokens`.

**P2#1 (tests)** — Added 11 tests covering:
- `loadModelsContextWindows`: valid JSON, missing file, malformed JSON
- `handleAgentEnd`: full usage data, empty messages, no assistant, no usage, unknown model → 200K fallback, nil modelsCW → 200K fallback, walks backwards for last assistant, skips assistant without usage
- `GetContextUsage`: nil before `agent_end`, returns copy (different pointer)

Re: the three questions — investigated pi source at `~/joyspace/pi`:
1. `agent_end` always carries `messages` — it's a required field in the TypeScript type, populated at all 4 construction points.
2. `totalTokens` is **per-turn** (per API call), not cumulative across turns. Confirmed by provider code (Anthropic, OpenAI, Google, etc.) which sets it from a single API response's usage.
3. `cacheRead`/`cacheWrite` are always present as `number` type, but some providers hardcode them to 0 (Mistral: both 0; OpenAI Responses/Google: `cacheWrite` 0). The fields never missing, just possibly zero.

chenhg5 (approved):
结论: Approve

总体判断: 干净的端口镜像 — 把 claudecode 已经在跑的 `ContextUsageReporter` 模式搬到 pi agent,核心逻辑最小化 + 全测覆盖,fallback 链合理。可以合并。

Review 范围:
- 看了 agent/pi/pi.go (loadModelsContextWindows), agent/pi/pi_test.go (13 个新测试), agent/pi/session.go (handleAgentEnd + GetContextUsage + modelsCW 字段)。
- 对比 core/interfaces.go:444-466 (ContextUsageReporter + ContextUsage) 和 agent/claudecode/session.go:428-450 (claudecode 同样 pattern) 确认字段契约一致。
- 重点关注 correctness (fallback 链 / 并发 / 多 assistant 消息)、testing (边界条件覆盖)、compatibility (零 pi 部署 / 零 usage 数据时的行为)。

✅ 做得好的地方:

- **完整的 fallback 链**:`s.modelsCW[model]` → 0 → 200K。文件不存在 / JSON 错 / 模型未列 / modelsCW 为 nil,四条路径全部 graceful,TestHandleAgentEnd_FallbackContextWindow + TestHandleAgentEnd_NilModelsCW 显式锁定两条。
- **Walk-backwards 找 last assistant 消息**:TestHandleAgentEnd_WalksBackwardsForLastAssistant 显式验证多个 assistant 消息时取最后一个的 usage。如果 pi RPC 因为重试 / 多 step 发出多个 assistant message,这是正确语义。
- **GetContextUsage 返回 copy**:TestGetContextUsage_ReturnsCopy 用 `u1 == u2` 指针不等锁定防止外部读到 in-flight mutation;配合 usageMu,engine 在 reply footer 渲染时跨 goroutine 安全。
- **零值字段明确注释**:BaselineTokens / ReasoningOutputTokens 没设值,代码注释明确说明 pi RPC 不报,engine UI 必须 handle 零值。说明作者看过 core.ContextUsage 完整契约,不是盲搬。
- **loadModelsContextWindows 的双 key 模式**:同时存 `mdl.ID` 和 `provider/mdl.ID`,claudecode 那边没有这个灵活性,pi 实际 API 调用时可能两种 ID 都报。TestLoadModelsContextWindows 两条路径都验证。
- **t.Setenv + t.Cleanup 正确处理 PI_CODING_AGENT_DIR**:TestLoadModelsContextWindows 在 t.Setenv 之前先 savedEnv,t.Cleanup 恢复,避免污染宿主环境。

🟠 建议改进:

1. **TestHandleAgentEnd_SkipsAssistantWithoutUsage 没真正测到 skip 逻辑**。当前测试 case 顺序是 `[no-usage, real-with-usage]`,walk-backwards 从尾部 i=1 开始,**先遇到 real-with-usage,直接命中并 return**。skip 逻辑只在 messages 顺序是 `[real-with-usage, no-usage-tail]` 时才被触发,比如:
   ```go
   // 真正测到 skip 的 case:
   msgs = [assistant(real-model, usage), assistant(no-usage-tail, no usage)]
   // walk-backwards: i=1 (no-usage), continue; i=0 (real-model), hit
   ```
   建议调整测试 case 顺序,或加第二个 case 显式覆盖"最后一条 assistant 没 usage,要回头找上一条"。

2. **`map[string]int` 0 值与"未配置"无法区分**。如果 pi 的 models.json 里有 `"contextWindow": 0` 的合法条目(虽然不太可能),也会被 fallback 成 200K。可以换成 `map[string]*int` 或用 `ok` 模式:
   ```go
   if cw, ok := s.modelsCW[model]; ok && cw > 0 {
       ctxWindow = cw
   } else {
       ctxWindow = 200_000
   }
   ```
   防御性写法,不影响当前行为。

🔵 可选优化:

- `slog.Info("pi: models.json not found, using 200K fallback", "path", path)` — 启动时每个 session 都会打这条 log。如果用户没有 pi 配置文件,每个 session start 都刷一条,可以考虑只在第一次记,或者降级到 Debug,减少日志噪音。Style nit。
- 200K fallback 是硬编码常量,可以加到 `piSettings` 配置里(`models.json → defaultContextWindow`),给高级用户覆盖空间。但本 PR scope 内不必要。
- handleAgentEnd 的 for loop + 提前 return 是 idiomatic Go,没毛病。P3:变量命名 `cw` 在 `cacheWriteTokens` 上下文里缩写成 `cw` 没问题,但和 `ctxWindow` 的 cw 局部冲突 (cacheWrite 缩成 `cw`,ContextWindow 缩成 `ctxWindow`)。改成 `cwt`(cache write tokens)和 `crt`(cache read tokens)更清晰。Style nit。

❓ 需要确认:

- PR description 说 "Verified in Feishu: reply footer shows `cw N cr N · ctx N%` for pi agent responses"。这是真机测试还是模拟?CI 里没有端到端测试锁这条 footer 渲染路径,如果 pi 的 agent_end 事件 schema 未来微调,footer 可能静默 fallback 而 CI 不知道。建议加一个集成测试,启动 mock pi RPC server 发 agent_end 事件,断言 GetContextUsage 返回的字段值(本 PR 已经有 13 个测试,加一个集成 case 不大)。
- `loadModelsContextWindows` 在 session start 调用,models.json 改后需要重启 session。Code 注释已经说明 "session-lifetime cache is acceptable"。如果 pi 用户频繁切模型,这是个限制。建议在 model switch 事件里 reload,但本 PR scope 不要求。

Testing / Risk:

- **已看到的验证**: 5/5 CI green (lint 2m15s, unit-test 4m2s, smoke-test 24s, regression-test 25s, performance-test 45s)。13 个新单元测试覆盖 fallback 链 / 边界 / 并发。
- **本地无法跑 pi agent**: 端到端验证依赖 author 真机,QA 评审只做代码层。
- **未覆盖风险**: 真实 pi CLI 发送的 agent_end event schema 是否和测试 mock 一致(无 E2E 锁)。如果 pi 升级改了 schema 字段名(input → input_tokens 等),本 PR 会静默 fallback 到 no-usage → nil。建议至少在 godoc 注释里加一行"if pi changes its RPC schema, this needs updating"。
- **兼容性**: 零侵入性,现有 pi 用户无感知。如果 models.json 不存在,行为和 PR 前完全一致(footer 用 generic fallback)。如果存在,新行为是显示真实 token 数。无 breaking change。

Next step:

- **Author**: 调整 TestHandleAgentEnd_SkipsAssistantWithoutUsage 的 case 顺序让它真正测到 skip 路径(可改 1-2 行)。其他 P3 都是 nit,可以不修。
- **Owner**: 可合并,scope 小、风险低、test 覆盖充分。后续如想加 E2E 锁,可以在 #1235 之后开 follow-up issue,不在本 PR 阻塞。

PR #1239, opened 2026-06-06
feat(acp): implement ModelSwitcher via ACP configOptions

SXongin:
## Summary

Parse `configOptions` from ACP `session/new` and `session/load` responses to dynamically discover available models. This enables the `/model` command in IM clients when using any ACP agent (OpenCode, Cursor, Devin, Copilot CLI, etc.).

## Problem

The ACP adapter had no `ModelSwitcher` implementation. Users could not use `/model` to switch models when connecting through ACP.

## Solution

The ACP protocol already returns `configOptions` in session handshake responses — a standard mechanism for advertising models, effort levels, and modes. This PR:

1. **Parses `configOptions`** from `session/new` and `session/load` responses
2. **Implements `ModelSwitcher`** on the ACP Agent (`SetModel`, `GetModel`, `AvailableModels`)
3. **Adds `SetLiveModel`** on ACP sessions via `session/set_config_option` RPC
4. **Passes initial model** to new sessions through `StartSession`

## How it works

```
IM user: /model
  → Engine calls Agent.AvailableModels()
  → Returns models cached from configOptions
  → User selects "deepseek/deepseek-v4-flash"
  → Engine calls Agent.SetModel("deepseek/deepseek-v4-flash")
  → Next StartSession passes it as initialModel
  → Session calls session/set_config_option RPC
```

The model list comes from the ACP server\s own `configOptions` — no hardcoded lists, provider-agnostic. Works with any ACP-compliant agent that returns model configOptions (OpenCode, Cursor, Devin, Copilot CLI).

## Changes

- `agent/acp/list_sessions.go`: +18 lines — `acpConfigOption` types
- `agent/acp/session.go`: +140 lines — parse configOptions, `SetLiveModel` RPC
- `agent/acp/agent.go`: +59 lines — `ModelSwitcher` interface implementation
- `agent/acp/agent_model_test.go`: new — agent model cache tests
- `agent/acp/session_model_test.go`: new — session configOptions tests

## Related

- ACP protocol: https://agentclientprotocol.com/
- `configOptions` is a standard ACP mechanism for advertising available models/providers/modes to clients

### DIFF
--- a/agent/acp/agent.go
+++ b/agent/acp/agent.go
@@ -48,16 +48,24 @@ type Agent struct {
 	modesCache    []core.PermissionModeInfo
 	modesCurrent  string
 
+	// model cache is populated by reportModelOptions callbacks from
+	// the session handshake. Used by AvailableModels() and GetModel().
+	modelMu        sync.RWMutex
+	modelOptions   []core.ModelOption
+	modelCurrent   string
+	modelPending   string // set by SetModel, takes precedence over modelCurrent
+
 	mu sync.RWMutex
 }
 
 // sessionCallbacks lets a running acpSession report what it learned
 // during the handshake back to its parent Agent. The session is owned
 // by cc-connect's engine (not the agent), so without this the agent
-// would never see availableModes / capability advertisements.
+// would never see availableModes / capability advertisements / model options.
 type sessionCallbacks interface {
 	reportModes(block acpModesBlock)
 	reportListSupported(supported bool)
+	reportModelOptions(opts []core.ModelOption, currentModel string)
 }
 
 // Ensure *Agent satisfies sessionCallbacks at compile time.
@@ -232,6 +240,10 @@ func (a *Agent) StartSession(ctx context.Context, sessionID string) (core.AgentS
 	extra = append(extra, a.sessionEnv...)
 	a.mu.RUnlock()
 
+	a.modelMu.RLock()
+	pendingModel := a.modelPending
+	a.modelMu.RUnlock()
+
 	return newACPSession(ctx, acpSessionConfig{
 		command:         command,
 		args:            args,
@@ -240,6 +252,7 @@ func (a *Agent) StartSession(ctx context.Context, sessionID string) (core.AgentS
 		resumeSessionID: sessionID,
 		authMethod:      authMethod,
 		initialMode:     pendingMode,
+		initialModel:    pendingModel,
 		callbacks:       a,
 	})
 }
@@ -369,4 +382,48 @@ func (a *Agent) reportListSupported(supported bool) {
 	} else {
 		a.listUnsupported.Store(false)
 	}
+}
+
+// reportModelOptions caches model options reported by a session handshake.
+func (a *Agent) reportModelOptions(opts []core.ModelOption, currentModel string) {
+	a.modelMu.Lock()
+	a.modelOptions = append(a.modelOptions[:0], opts...)
+	if currentModel != "" {
+		a.modelCurrent = currentModel
+	}
+	a.modelMu.Unlock()
+}
+
+// -- ModelSwitcher --
+
+// SetModel stores a model ID to apply to future sessions via
+// StartSession. ModelSwitcher says "Model changes take effect on the
+// next session." See also the session-level SetLiveModel for changing
+// the model of an already-running session.
+func (a *Agent) SetModel(model string) {
+	a.modelMu.Lock()
+	a.modelPending = strings.TrimSpace(model)
+	a.modelMu.Unlock()
+	slog.Info("acp: model changed for future sessions", "model", model)
+}
+
+// GetModel returns the active model. Pending SetModel wins over
+// server-reported currentValue.
+func (a *Agent) GetModel() string {
+	a.modelMu.RLock()
+	defer a.modelMu.RUnlock()
+	if a.modelPending != "" {
+		return a.modelPending
+	}
+	return a.modelCurrent
+}
+
+// AvailableModels returns the model list cached from the most recent
+// session handshake. Returns nil before the first handshake.
+func (a *Agent) AvailableModels(ctx context.Context) []core.ModelOption {
+	a.modelMu.RLock()
+	defer a.modelMu.RUnlock()
+	out := make([]core.ModelOption, len(a.modelOptions))
+	copy(out, a.modelOptions)
+	return out
 }
--- a/agent/acp/agent_model_test.go
+++ b/agent/acp/agent_model_test.go
@@ -1,0 +1,76 @@
+package acp
+
+import (
+	"context"
+	"testing"
+
+	"github.com/chenhg5/cc-connect/core"
+)
+
+// --- Agent: model cache & ModelSwitcher ---------------------------------
+
+func TestAgent_AvailableModels_emptyBeforeHandshake(t *testing.T) {
+	a := &Agent{}
+	ctx := context.Background()
+	if got := a.AvailableModels(ctx); len(got) != 0 {
+		t.Fatalf("want empty models before first handshake, got %v", got)
+	}
+	if got := a.GetModel(); got != "" {
+		t.Fatalf("want empty model before handshake, got %q", got)
+	}
+}
+
+func TestAgent_reportModelOptions_populatesCache(t *testing.T) {
+	a := &Agent{}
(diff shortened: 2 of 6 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
Comprehensive, well-tested implementation of `ModelSwitcher` for the ACP adapter. The PR closes a real gap — ACP users (OpenCode, Cursor, Devin, Copilot CLI) couldn't use `/model` — by leaning on the standard ACP `configOptions` mechanism instead of hardcoding model lists. The design mirrors the existing `modesBlock` pattern (same shape, same callback pattern, same mutex discipline), is comprehensively tested (10 new test functions, ~362 LoC of tests vs ~200 LoC of production code = 1.8:1 test/code ratio), and 5/5 CI is green. 0 P0/P1, only minor P3 nits. Ready to merge.

Review scope:
- Reviewed: `agent/acp/agent.go` (ModelSwitcher impl + reportModelOptions callback), `agent/acp/session.go` (absorbConfigOptions, SetLiveModel, matchAvailableConfigValue, handshake integration), `agent/acp/list_sessions.go` (new types), `agent/acp/agent_model_test.go` (4 tests), `agent/acp/session_model_test.go` (6 tests).
- 2 commits: `fd3856e` (feat), `5b242b2` (fix: add reportModelOptions to fakeCallbacks in existing tests).
- Diff vs `main`: +567 / -12, 6 files.
- Focused on: ModelSwitcher interface conformance, lock discipline, callback fan-out, case-insensitive matching, RPC error handling, test coverage.

✅ What looks good:
- **Standard ACP mechanism, provider-agnostic.** No hardcoded model lists. The `configOptions` come from the ACP server itself (OpenCode, Cursor, Devin, Copilot CLI all return model `configOptions` in `session/new` / `session/load` responses per the spec). One implementation covers all ACP-compliant agents — that's the right architecture.
- **`sessionCallbacks` interface extension is clean and non-breaking.** Old methods (`reportModes`, `reportListSupported`) unchanged; new `reportModelOptions(opts []core.ModelOption, currentModel string)` added. The static assertion `var _ sessionCallbacks = (*Agent)(nil)` (agent.go:71) catches any drift at compile time.
- **Lock discipline is correct and consistent.** Three separate mutexes (`mu`, `modelMu`, `configMu`) guard the three independent state slices (mode-related, model-related, configOptions cache). All accesses go through `RLock` for reads and `Lock` for writes, with `defer RUnlock` / `defer Unlock`. No nested-locking hazards.
- **`absorbConfigOptions` does the right thing in three places**: (1) session-level cache for `modelOptions()`/`matchAvailableConfigValue`/`SetLiveModel`; (2) agent-level callback for `AvailableModels()`/`GetModel()`; (3) uses `append(s.configOptionsCache[:0], opts...)` to reuse backing array — defensive but not allocation-heavy.
- **`SetLiveModel` has good defensive guards** (session.go:~370-410): `alive.Load()` check, `currentACPSessionID()` check, RPC error logged + returns false (so the engine falls back gracefully), and the optional `matchAvailableConfigValue` resolution is conservative (case-insensitive match on both value and display name, plus a `currentValue` fallback).
- **10 new tests cover the critical paths**: handshake parsing (`TestSession_handshake_parsesConfigOptions`), no-configOption graceful handling (`TestSession_absorbConfigOptions_noModelOption_doesNotCrash`), live RPC (`TestSession_SetLiveModel_MockRPC` with happy_path subtest), initial model passing (`TestSession_handshake_passesInitialModel`), case-insensitive matching (`TestSession_matchAvailableModel_caseInsensitive`), generic `matchAvailableConfigValue` (`TestSession_matchAvailableConfigValue`), agent cache (`TestAgent_reportModelOptions_populatesCache`), empty-before-handshake (`TestAgent_AvailableModels_emptyBeforeHandshake`), SetModel override (`TestAgent_SetModel_overridesPending`), and SetModel("") reset (`TestAgent_SetModel_emptyResets`). All 10 PASS verified locally.
- **`AvailableModels` returns a copy** (`make + copy`) — defensive against the engine mutating the slice. Good practice.
- **`modelPending` semantics**: SetModel("") correctly resets pending so GetModel() falls back to server-reported `modelCurrent`. Tested in `TestAgent_SetModel_emptyResets`.

🟠 Should improve:

- **P2 — Missing static interface assertions** (low-priority consistency nit). `agent/acp/agent.go` is missing `var _ core.Agent = (*Agent)(nil)` and `var _ core.ModelSwitcher = (*Agent)(nil)` lines. The existing `var _ sessionCallbacks = (*Agent)(nil)` is there, so the pattern is established. Without the new assertions, an accidental signature change in `core.Agent` or `core.ModelSwitcher` would compile fine here and only fail when the type is registered or used by the engine. Cheap insurance. **Not blocking this PR.**
- **P2 — Fallback behavior in `SetLiveModel` when model is unknown is ambiguous.** `session.go` ~line 410: when `matchAvailableConfigValue` returns false, the code falls back to `resolved = modelID` (pass-through). But `matchAvailableConfigValue` returns false for two distinct cases: (a) no `configOptions` at all (older ACP agent — pass-through is the right behavior), and (b) `configOptions` exist but user typed a model not in the list (pass-through would send a bogus ID that the server will reject with 4xx). The current code conflates these. The 4xx surface is OK (engine will see the error), but distinguishing the two cases would let you log differently: e.g. "passing through unknown model — server may reject" vs "no configOptions, pass-through legacy behavior". **Cosmetic; not blocking.**

🔵 Optional:

- **P3 — Add CHANGELOG entry** under `## Unreleased` / `### Added`. Suggested text: `feat(acp): implement ModelSwitcher via ACP configOptions (auto-discover models for /model command)`. release-codex will appreciate it for v1.3.5.
- **P3 — `acpConfigOptionsToModelOptions` only uses the first `category="model"` configOption.** If an ACP server returns multiple "model" entries (unusual but possible — e.g. one per provider), only the first wins. Worth a doc comment noting the behavior. Not in this PR's scope to fix.
- **P3 — `modelOptions()` (session) re-converts on every call.** For very large model lists and frequent polling, this is O(n) per call. Could cache the conversion result. Not a concern in practice (model lists are small, polled infrequently).
- **P3 — `core.ModelOption.Alias` field is unused in the ACP path.** ACP `configOptions` doesn't expose aliases, so `Alias` is always `""`. That's fine — other agents (e.g. claudecode, codex) may populate it. Just flagging that the ACP path doesn't take advantage of the third field.

❓ Questions:
- Does the `session/set_config_option` RPC have a known response shape that should be validated (e.g. a "success" field)? The current code discards the response (`_, err := s.tr.call(...)`) — if the RPC returns a 200 with an error body, we wouldn't notice. Worth checking ACP spec or testing with a real OpenCode instance.
- When `cfg.initialModel` is set but `SetLiveModel` returns false (server rejects), the agent's `modelPending` is still set but the running session doesn't have the model applied. On next `StartSession`, the cycle repeats. Is there a way to clear `modelPending` on rejection, or is this user-driven (user types another `/model`)? Currently silent.

Testing / Risk:
- Verified locally: `go test ./agent/acp/ -count=1 -v` → all 45 tests PASS (0.017s wall). The 10 new tests for this PR all pass. `go vet ./agent/acp/` clean. 5/5 CI green on PR head (lint 2m26s / unit-test 3m45s / smoke-test 24s / regression-test 25s / performance-test 42s).
- Uncovered risk: the `session/set_config_option` RPC error path is only tested via mocked RPC (in `TestSession_SetLiveModel_MockRPC`). A real OpenCode / Cursor instance may return 4xx with a JSON error body that the current `tr.call` doesn't introspect — the user would see a generic "RPC failed" message instead of the server's actual reason. This is a follow-up enhancement, not a blocker.
- Uncovered risk: `absorbConfigOptions` is called from both `session/new` and `session/load` paths. If the server returns different `configOptions` on subsequent handshakes (e.g. user added a model on the server side), the agent cache is overwritten via `append(a.modelOptions[:0], opts...)` — no leak. Good.

Next step:
- owner: merge PR #1239 (community @SXongin, --approve, ACP ModelSwitcher via configOptions, 0 conflict). Suggested merge order: this PR can go alongside #1250 in the v1.3.5 batch — both are clean standalone additions, neither touches shared code paths.
- QA next cycle: continue qa-view inbox review. Remaining open+mergeable candidates will surface as we walk the qa-view (current top non-conflict are #1250 done, #1239 done; next candidates likely deeper in the inbox).

PR #1204, opened 2026-06-02
feat(feishu): refresh rich card rendering and panel handling

AaronZ345:
Replaces #879, #896, #1018, and #1019 as one rebased rich-card suite on the latest upstream main (39df0842).

## What changed
- Adds structured per-turn reply footer support and integrates it into Feishu rich cards.
- Uses cardkit-v1 streaming/final flush paths for rich card body updates.
- Hardens reasoning/tool/progress card rendering, icon mapping, fallback layout, image embedding, oversized panel handling, and latest-panel windows.
- Keeps related config, management, tests, and web project-detail surfaces aligned.

## Verification
- `git diff --check origin/main..HEAD`
- `go test -tags no_web ./cmd/cc-connect ./agent/claudecode ./config ./core ./platform/feishu ./tests/release_local/turn_contract`
- `cd web && pnpm install --frozen-lockfile && pnpm build`

PR #1246, opened 2026-06-07
fix: address P1/P2 issues from post-merge QA review

chenhg5:
## Summary

Post-merge QA review identified several issues across the recent batch of merged PRs. This PR fixes the two P1s and three P2s that are safe to patch in a single targeted commit.

## P1 fixes

**1. Test stub `BuildRichCard` signature mismatch (core/engine_test.go)**
- `stubCompactProgressPlatform.BuildRichCard` still had `elapsed time.Duration` as the last parameter after #1204 changed `RichCardSupporter` to `statusFooter string`
- At runtime, the stub silently did NOT satisfy the interface, so all tests using it were bypassing the rich-card code path entirely — tests passed but didn't exercise the intended logic
- Fixed: update signature to `statusFooter string`, update format string accordingly

**2. Slack `session_scope=thread` without `window_per_session` (platform/slack)**
- When `session_scope=thread` is set but the tmux agent's `window_per_session=true` is not, concurrent threads share a single pane causing output interleave
- Fixed: emit a `slog.Warn` at `New()` time so operators see the misconfiguration in startup logs

## P2 fixes

**3. `relay.SetVisibility` silent fallback on unknown value (core/relay.go)**
- `normalizeRelayVisibility` fell back to `"full"` silently for unrecognised values
- Fixed: add a structured `slog.Warn` logging the invalid value and valid options

**4. Feishu video MIME format gap (platform/feishu/feishu.go)**
- CLI `send` accepted `.webm/.mov/.avi/.mkv/.m4v` as video attachments, but `detectFeishuFileType` only mapped `.mp4` → `FileTypeMp4` (Media message type); all others fell through to generic file download
- Fixed: expand video detection to cover all common video MIME types and extensions → `FileTypeMp4`, so they render as native Feishu video player bubbles (playback compatibility depends on client platform)

**5. Root bypass-downgrade not surfaced to IM user (agent/claudecode, agent/qoder, core)**
- When running as root, `bypassPermissions` (Claude Code) and `yolo` (Qoder) were silently downgraded with only a slog.Warn — users saw the agent pausing on permission requests without knowing why
- Fixed: add `StartupWarner` optional interface to `core/interfaces.go`; implement it in both session types; engine emits the warning message to the IM platform immediately after session start

## Test plan

- [ ] `go build ./core/ ./agent/... ./platform/...` passes (verified locally)
- [ ] Slack with `session_scope=thread` logs the warning at startup
- [ ] Feishu: send a `.mov` or `.webm` file via CLI — verify it appears as a video message (not a file attachment) in Feishu
- [ ] Claude Code / Qoder running as root with bypass mode — verify IM receives the downgrade warning on first session start

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -59,16 +59,27 @@ type claudeSession struct {
 	// Stop hook timeout. The wait ends as soon as the process exits,
 	// so typical shutdowns take seconds, not the full timeout.
 	gracefulStopTimeout time.Duration
+
+	// startupWarning holds a one-time message to surface to the IM user at
+	// session start (e.g. when a permission mode was silently downgraded).
+	startupWarning string
 }
+
+// StartupWarning implements core.StartupWarner. Returns a non-empty string
+// when the session was started under degraded conditions that the user should
+// know about (e.g. bypassPermissions downgraded to auto under root).
+func (cs *claudeSession) StartupWarning() string { return cs.startupWarning }
 
 func newClaudeSession(ctx context.Context, workDir, cliBin string, cliExtraArgs []string, cliArgsFlag string, model, effort, sessionID, mode, systemPrompt string, allowedTools, disallowedTools []string, extraEnv []string, platformPrompt string, disableVerbose bool, spawnOpts core.SpawnOptions, maxContextTokens int) (*claudeSession, error) {
 	sessionCtx, cancel := context.WithCancel(ctx)
 
 	// Claude Code rejects bypassPermissions when running as root.
 	// Downgrade to "auto" which auto-approves internally in cc-connect.
+	var rootDowngradeWarning string
 	if mode == "bypassPermissions" && os.Geteuid() == 0 {
 		slog.Warn("claudeSession: bypassPermissions not allowed under root, downgrading to auto mode")
 		mode = "auto"
+		rootDowngradeWarning = "⚠️ Running as root: bypassPermissions mode is not supported and has been downgraded to auto. The agent may still pause on high-risk operations."
 	}
 
 	// innerArgs are Claude Code CLI flags — when a wrapper is used with
@@ -226,6 +237,7 @@ func newClaudeSession(ctx context.Context, workDir, cliBin string, cliExtraArgs 
 		cancel:              cancel,
 		done:                make(chan struct{}),
 		gracefulStopTimeout: 120 * time.Second,
+		startupWarning:      rootDowngradeWarning,
 	}
 	cs.setPermissionMode(mode)
 	cs.sessionID.Store(sessionID)
--- a/agent/qoder/session.go
+++ b/agent/qoder/session.go
@@ -23,17 +23,23 @@ import (
 // Each Send() spawns `qodercli -p <prompt> -f stream-json -q`.
 // Subsequent turns use `-r <sessionID>` to resume the conversation.
 type qoderSession struct {
-	workDir   string
-	model     string
-	mode      string
-	extraEnv  []string
-	events    chan core.Event
-	sessionID atomic.Value // stores string
-	ctx       context.Context
-	cancel    context.CancelFunc
-	wg        sync.WaitGroup
-	alive     atomic.Bool
+	workDir        string
+	model          string
+	mode           string
+	extraEnv       []string
+	events         chan core.Event
+	sessionID      atomic.Value // stores string
+	ctx            context.Context
+	cancel         context.CancelFunc
+	wg             sync.WaitGroup
+	alive          atomic.Bool
+	startupWarning string
 }
+
+// StartupWarning implements core.StartupWarner. Returns a non-empty string
+// when the session was started under degraded conditions (e.g. yolo mode
+// silently skipped under root).
+func (qs *qoderSession) StartupWarning() string { return qs.startupWarning }
 
 func newQoderSession(ctx context.Context, workDir, model, mode, resumeID string, extraEnv []string) (*qoderSession, error) {
 	sessionCtx, cancel := context.WithCancel(ctx)
@@ -48,6 +54,12 @@ func newQoderSession(ctx context.Context, workDir, model, mode, resumeID string,
 		cancel:   cancel,
 	}
 	qs.alive.Store(true)
+
+	// Detect root-induced yolo downgrade and surface it to the IM user via StartupWarning.
+	// The actual flag skip happens inside Send() when building the CLI args.
+	if mode == "yolo" && os.Geteuid() == 0 {
+		qs.startupWarning = "⚠️ Running as root: --dangerously-skip-permissions (yolo mode) is not supported and has been skipped. The agent may pause on high-risk operations."
+	}
 
 	if resumeID != "" && resumeID != core.ContinueSession {
 		qs.sessionID.Store(resumeID)
--- a/core/engine.go
+++ b/core/engine.go
@@ -3211,6 +3211,13 @@ func (e *Engine) getOrCreateInteractiveStateWith(sessionKey string, p Platform, 
 		slog.Warn("slow agent session start", "elapsed", startElapsed, "agent", agent.Name(), "session_id", startSessionID)
 	}
 
+	// Surface any startup warning (e.g. permission mode downgrade under root) to the IM user.
+	if warner, ok := agentSession.(StartupWarner); ok {
+		if msg := warner.StartupWarning(); msg != "" {
+			e.send(p, replyCtx, msg)
+		}
+	}
+
 	if newID := agentSession.CurrentSessionID(); newID != "" {
 		// Track the latest session ID Claude reports. Each --resume forks a new
 		// session_id, so the stored ID must follow the live process or a later
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -380,9 +380,9 @@ func (p *stubCompactProgressPlatform) UpdateMessage(_ context.Context, _ any, co
 	return nil
 }
 
-func (p *stubCompactProgressPlatform) BuildRichCard(status CardStatus, title string, steps []ToolStep, markdown string, streaming bool, elapsed time.Duration) string {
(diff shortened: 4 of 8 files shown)


chenhg5 (commented):
Hi! PM asked QA to review #1246 against 6 checkpoints. Walked the diff, built/vet/tested the branch, ran merge-tree against main. Posting as comment because author is chenhg5.

## TL;DR

**Ready to merge.** All 5 P1/P2 fixes are correct, `go build`/`go vet` clean, all touched-package tests pass. A few small coverage gaps and a confusingly-large diff that I want to unpack.

## Heads-up: this branch is based on an old main

Before I get to the 6 checkpoints, I want to flag something that will look alarming in the GitHub diff: **the branch is based on `330e3c25` (3 days old), not on current `2afceb73`**. The diff shows "deletions" of `/cancel` (PR #957), `drainPendingUpdates` (PR #425), archive-first update (PR #444), and several docs — but those are **not real reverts**. They're an artifact of comparing an old branch head to current main:

```
$ git log 2afceb73..06dcb3c5 --oneline
06dcb3c5 fix: address P1/P2 issues from post-merge QA review
# (only the PR's own commit, nothing else)

$ git log 06dcb3c5..2afceb73 --oneline
2afceb73 docs: document telegram proxy config (#389)
1322565c docs: clarify admin_from placement for privileged commands (#393)
2ff96bfb fix(update): try archive format first when downloading updates (#444)
de368836 fix(telegram): drain pending updates before polling to prevent 409 Conflict (#425)
e43edf02 feat(core): add /cancel command to interrupt and reset session (#957)
# (5 newer commits on main, NOT in the branch)
```

When the branch is merged, GitHub will three-way merge against main, take main's version of /cancel/drain/archive/etc. (because the branch never touched those lines), and apply the PR's 5 fixes on top. **`git merge-tree` reports no conflicts** for all 8 actually-touched files. So all of main's recent fixes/features survive intact.

The only thing that matters is whether the 5 PR-introduced changes are correct. They are. Below.

## Checkpoint 1 — rich-card test stub signature ✅

Confirmed: this is a real coverage fix, not a cosmetic rename.

- The `RichCardSupporter` interface in `core/streaming.go:BuildRichCard` requires `statusFooter string` as the 6th param (renamed by PR #1204).
- `stubCompactProgressPlatform.BuildRichCard` in `core/engine_test.go` had `elapsed time.Duration` — wrong type, so the type assertion `p.(RichCardSupporter)` at `core/engine.go:3927` failed silently and the rich-card path was bypassed.
- The PR renames the parameter to `statusFooter string` and the format string from `elapsed=%s` to `footer=%s`. With the fix, all 5 tests using `stubCompactProgressPlatform` (`TestProcessInteractiveEvents_CompactProgressCoalescesThinkingAndToolUse`, `TestProcessInteractiveEvents_CardProgressUsesCardTemplate`, `TestProcessInteractiveEvents_CardProgressUsesStructuredPayloadWhenSupported`, `TestProcessInteractiveEvents_RichCardShowsThinkingContent`, `TestProcessInteractiveEvents_RichCardCoalescesToolResult`) now actually exercise the rich-card path. Real coverage restored.

Format string change is safe — no test asserts on the substring `elapsed=`. `time` is still imported elsewhere, so the import stays.

## Checkpoint 2 — StartupWarner interface and engine emission path ✅

`core/interfaces.go:541-548`:
```go
type StartupWarner interface {
    StartupWarning() string
}
```

- **Optional interface** — agents that don't implement it are unaffected. The engine check is `if warner, ok := agentSession.(StartupWarner); ok` so the type assertion is safe.
- **Emission site**: `core/engine.go:3214-3219`, in `getOrCreateInteractiveStateWith`, immediately after `agent.StartSession()` succeeds and the slow-start warning, before the session-ID tracking block.
- **No duplication risk**: `e.send(p, replyCtx, msg)` is called once per session creation. If the same engine has multiple agent sessions (e.g., one per project / platform), each gets its own warning, which is the correct semantics.

**Minor gaps (not blockers):**
- No test exercises the emission path end-to-end (mock platform, assert that the warning string is sent). Adding one would take 5 minutes and close the loop.
- The warning is **per-session**, not per-process. A user who does 10 prompts/hour with root + bypassPermissions would see the warning 10 times. This is reasonable (each session really did get downgraded) but a "have we warned this user yet?" flag at the engine level would be nicer. Not a blocker.

## Checkpoint 3 — root-downgrade warnings for Claude Code / Qoder ✅

**Claude Code** (`agent/claudecode/session.go:78-83`):
```go
if mode == "bypassPermissions" && os.Geteuid() == 0 {
    slog.Warn("claudeSession: bypassPermissions not allowed under root, downgrading to auto mode")
    mode = "auto"
    rootDowngradeWarning = "⚠️ Running as root: bypassPermissions mode is not supported and has been downgraded to auto. ..."
}
```
Single check. `mode` is reassigned to `"auto"`, so the downstream `setPermissionMode(mode)` at line 786 sets `autoApprove = false`. The warning is stored on the struct at line 240 and read by `StartupWarning()`. Not duplicated, not stale.

**Qoder** (`agent/qoder/session.go:60-62`):
```go
if mode == "yolo" && os.Geteuid() == 0 {
    qs.startupWarning = "⚠️ Running as root: --dangerously-skip-permissions (yolo mode) is not supported and has been skipped. ..."
}
```
Single check. Actual flag-skip happens in `Send()` at line 90-95, which the comment correctly cross-references. The warning fires once at session creation; the flag skip happens per `Send()` call. No duplication.

Wording is consistent (⚠️ emoji, "Running as root: X is not supported and has been [downgraded|skipped]. The agent may still pause on high-risk operations."), user-actionable, no jargon.

## Checkpoint 4 — Slack `session_scope=thread` warning condition and wording ✅ (with caveat)

`platform/slack/slack.go:57-61`:
```go
if scope == "thread" {
    slog.Warn("slack: session_scope=thread gives each Slack thread its own session; " +
        "if your agent runtime is tmux, also set window_per_session=true — " +
        "without it, concurrent threads share a single pane and their output will interleave")
}
```

- **Fires at `New()` time, once per Slack process start.** Not spammy.
- **Condition is correct but slightly over-broad**: the warning is unconditional for `scope == "thread"`, even for non-tmux runtimes. The Slack platform doesn't know which agent runtime the user picked, so it can't be more precise without leaking that knowledge. The current wording already softens it ("if your agent runtime is tmux…") which is the right hedge.
- **Wording is clear** and the actionable remedy (`window_per_session=true`) is named explicitly.

**Minor suggestion (not a blocker):** the warning could check the agent-side option key directly if/when the engine exposes a config getter for the agent runtime, and only fire when tmux is actually in use. As-is, it's a useful nudge for the common case and harmless noise for others.

## Checkpoint 5 — Feishu video MIME/type mapping ✅ (with caveat)

`platform/feishu/feishu.go:2499-2506`:
```go
case strings.HasPrefix(mimeType, "video/") ||
    strings.HasSuffix(name, ".mp4") || strings.HasSuffix(name, ".mov") ||
    strings.HasSuffix(name, ".avi") || strings.HasSuffix(name, ".m4v") ||
    strings.HasSuffix(name, ".mkv") || strings.HasSuffix(name, ".webm"):
    return larkim.FileTypeMp4
```

- Comment is honest about the trade-off: "Feishu's file API only has 'mp4' as the video type. We map all common video MIME types and extensions to FileTypeMp4 so the message renders as a native video player bubble rather than a generic file download. Actual playback compatibility (e.g. webm/mkv) depends on the Feishu client platform; mp4 H.264 has the broadest support."
- Real-world risk: Feishu may validate the uploaded content matches the declared `FileTypeMp4` and 4xx a non-mp4 upload. The comment doesn't say this has been tested. **Recommend** the release notes / CHANGELOG call out that webm/mov/avi/m4v/mkv are best-effort.

**Minor gap (not a blocker):** `TestDetectFeishuFileMessageType` only covers `mp4` + `ogg` + `pdf` + unknown. The 5 new extensions have no unit test. Worth a follow-up to add `webm`, `mov`, `avi`, `m4v`, `mkv` cases to lock in the mapping.

## Checkpoint 6 — Go tests / build for touched packages ✅

Local run in `/tmp/qa-pr1246-rlwmw6` (worktree on `pr-1246`, `go.sum` copied from main to work around the `BurntSushi/toml v1.6.0` checksum drift):

```
$ go build ./...     # clean
$ go vet ./...       # clean
$ go test ./core/                        -count=1    ok  42.226s
$ go test ./agent/claudecode/            -count=1    ok   0.281s
$ go test ./agent/qoder/                 -count=1    ok   0.006s
$ go test ./platform/feishu/             -count=1    ok  25.366s
$ go test ./platform/slack/              -count=1    ok   0.016s
```

The `TestDetectFeishuFileMessageType` subtests for mp4/ogg/pdf/unknown all pass; the new webm/mov/avi/m4v/mkv mappings are not asserted (see Checkpoint 5).

Race detector not run — gcc not in PATH in this environment. None of the 5 fixes touches a goroutine/sharing boundary that would benefit from a race run.

## Verdict

| # | Checkpoint | Verdict |
|---|---|---|
| 1 | Rich-card test stub signature | ✅ Real coverage fix |
| 2 | StartupWarner + engine emission | ✅ Correct, optional interface; minor test gap |
| 3 | Root-downgrade warnings (Claude + Qoder) | ✅ Correct, not duplicated |
| 4 | Slack session_scope=thread warning | ✅ Acceptable; fires once per process |
| 5 | Feishu video MIME/type mapping | ✅ Best-effort with honest comment; test gap |
| 6 | Tests / build for touched packages | ✅ All pass |

**Recommendation: ready to merge.** No blockers. The branch staleness (5 main commits ahead) auto-resolves on merge — none of the main-side improvements (/#cancel, drainPendingUpdates, archive-first update, admin_from docs) are actually reverted by this PR. The 5 P1/P2 fixes land cleanly on top.

**Optional follow-ups (not blocking):**
1. Add a unit test for the StartupWarner emission path in `core/engine_test.go`.
2. Extend `TestDetectFeishuFileMessageType` to cover webm/mov/avi/m4v/mkv.
3. CHANGELOG note for the webm/mov/avi/m4v/mkv mapping: "best-effort; Feishu client must support the codec for playback."

— qa-claudecode

PR #1250, opened 2026-06-07
fix(dingtalk): extract @userid from content in natural Reply

wgx521:
## Summary
- 钉钉自然 Reply 路径自动从内容中提取 @userId 并填入 `at.atUserIds`
- regex 仅匹配 4+ 位数字 ID（`@(\d{4,})`），避免误匹配 `@123` 等短引用
- 支持去重，保持首次出现顺序

## QA Review Fixes (P2)
- regexp 提升为包级变量，避免每次 Reply 重复编译
- 补 gofmt 空行
- 新增 `TestExtractAtUserIds`（12 子用例）、`TestReply_IncludesExtractedAtUserIds`、`TestReply_NoAtUserIdsWhenNoMention`

## Test plan
- [x] 钉钉群聊中 @ 短 ID 用户能正确提取
- [x] `go build ./platform/dingtalk/` 通过
- [x] `go test ./platform/dingtalk/` 通过
- [x] `gofmt` / `go vet` 通过

### DIFF
--- a/platform/dingtalk/dingtalk.go
+++ b/platform/dingtalk/dingtalk.go
@@ -10,6 +10,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"os/exec"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -744,11 +745,19 @@ func (p *Platform) Reply(ctx context.Context, rctx any, content string) error {
 		return p.sendProactiveMessage(ctx, rc, content)
 	}
 
+	atUserIds := extractAtUserIds(content)
+
 	content = preprocessDingTalkMarkdown(content)
 
 	payload := map[string]any{
 		"msgtype":  "markdown",
 		"markdown": map[string]string{"title": "reply", "text": content},
+	}
+	if len(atUserIds) > 0 {
+		payload["at"] = map[string]any{
+			"atUserIds": atUserIds,
+			"isAtAll":   false,
+		}
 	}
 	body, err := json.Marshal(payload)
 	if err != nil {
@@ -1588,6 +1597,23 @@ func (p *Platform) sendProactiveMessage(ctx context.Context, rc replyContext, co
 
 	slog.Debug("dingtalk: proactive message sent", "api", apiURL, "status", resp.StatusCode)
 	return nil
+}
+
+var atUserIDRegexp = regexp.MustCompile(`@(\d{4,})`)
+
+// extractAtUserIds extracts @userId patterns from content for DingTalk's atUserIds field.
+// Matches @ followed by numeric DingTalk user IDs (e.g. @194252073827812352).
+func extractAtUserIds(content string) []string {
+	matches := atUserIDRegexp.FindAllStringSubmatch(content, -1)
+	seen := make(map[string]bool)
+	var ids []string
+	for _, m := range matches {
+		if len(m) > 1 && !seen[m[1]] {
+			seen[m[1]] = true
+			ids = append(ids, m[1])
+		}
+	}
+	return ids
 }
 
 // preprocessDingTalkMarkdown adapts content for DingTalk's markdown renderer:
--- a/platform/dingtalk/dingtalk_test.go
+++ b/platform/dingtalk/dingtalk_test.go
@@ -1083,3 +1083,108 @@ func TestOnMessageRepliesToUnauthorizedSender(t *testing.T) {
 		t.Fatal("timed out waiting for unauthorized reply")
 	}
 }
+
+func TestExtractAtUserIds(t *testing.T) {
+	tests := []struct {
+		name    string
+		content string
+		want    []string
+	}{
+		{name: "empty string", content: "", want: nil},
+		{name: "plain text no at", content: "hello world", want: nil},
+		{name: "3-digit not matched", content: "@123", want: nil},
+		{name: "4-digit match", content: "@1234", want: []string{"1234"}},
+		{name: "9-digit match", content: "@194252073", want: []string{"194252073"}},
+		{name: "18-digit match", content: "@194252073827812352", want: []string{"194252073827812352"}},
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
Clean, focused, well-tested fix that addresses a real DingTalk UX gap (users in group chat can't @-mention other users via natural Reply path). The implementation is conservative (4+ digit regex to avoid false positives), pre-compiled as a package-level variable, and ships with 14 sub-tests + 2 integration tests. 5/5 CI green, 0 P0/P1, 0 breaking. Ready to merge.

Review scope:
- Reviewed: `platform/dingtalk/dingtalk.go` (`Reply` + new `extractAtUserIds` helper), `platform/dingtalk/dingtalk_test.go` (3 new test functions).
- 1 commit `f07a202` (head), authored by community @wgx521.
- Diff vs `main`: +131 / -0, 2 files (1 source + 1 test).
- Focused on: correctness of regex, scope of `Reply` vs `ReplyWithAt` / `sendProactiveMessage`, dedup semantics, JSON payload shape, test coverage.

✅ What looks good:
- **Correct scope decision**: the fix only modifies `Reply` (markdown path, sessionWebhook). It deliberately does NOT touch `ReplyWithAt` (explicit atUsers from caller — caller already knows the @-targets) or `sendProactiveMessage` (server-initiated, less likely to mention users). This is the right scope — the issue is specifically about natural Reply not auto-detecting @userIds, and the other paths already handle atMention correctly.
- **Conservative regex `@(\d{4,})`** avoids false positives on short numeric tokens like `@123` (e.g. inline math, build numbers). DingTalk staffIDs are 18+ digits but the 4-digit floor catches shorter test/sandbox IDs too. The comment explains the choice.
- **Package-level `atUserIDRegexp`** is pre-compiled at init (one regex per process, not per Reply call). The P2 fix from a prior review round is applied — this matches the PR body's "QA Review Fixes (P2)" section.
- **Dedup preserves first-occurrence order** — the `seen` map + append pattern is correct. If the same user is @-mentioned 3 times, the payload gets `["1234"]` not `["1234","1234","1234"]`, and DingTalk's API de-duplicates correctly.
- **Empty-content guard**: when `len(atUserIds) == 0`, the code skips adding the `at` field entirely (`if len(atUserIds) > 0`), keeping the payload minimal. Cleaner than always sending `"at": {"atUserIds": [], "isAtAll": false}`.
- **Tests are thorough**: 12 sub-cases in `TestExtractAtUserIds` (empty, plain, 3-digit boundary, 4-digit minimum, 9-digit, 18-digit, alphabetic non-match, dedup, multiple distinct, embedded-in-text, mixed valid/invalid, only-@), plus 2 httptest-based integration tests (`TestReply_IncludesExtractedAtUserIds` verifies payload shape end-to-end, `TestReply_NoAtUserIdsWhenNoMention` is the negative case). 14/14 PASS verified locally.
- **Integration test uses `httptest.NewServer`**, captures the actual JSON body that would be POSTed to DingTalk, and decodes it via `json.NewDecoder(r.Body).Decode(&payload)` — same shape DingTalk would receive in production.
- **CHANGELOG entry is missing** (single-line nit, not a blocker — release-codex will pick this up at v1.3.5 cut time).

🟠 Should improve:
- **P2 (cross-PR, not blocking this PR)**: Same regex would also be useful in `sendProactiveMessage` (line 1540) if a future use case arises (e.g. proactive bot mentioning a user for ticket escalation). Right now the proactive path doesn't extract @userIds. This is a deliberate scope choice, but if the maintainer wants to be consistent, a follow-up PR could share the helper. **Not in this PR's scope.**

🔵 Optional:
- **P3 — Add CHANGELOG entry** under `## Unreleased` / `### Fixed` (or `### Added`, since the behavior is new). Suggested text: `fix(dingtalk): auto-extract @userId in natural Reply and inject into atUserIds`. release-codex will appreciate it.
- **P3 — Consider documenting the regex choice in package doc** (one line in `package dingtalk`): e.g. `// DingTalk staff IDs are numeric; the helper extracts @<digits> with floor 4 to avoid false positives on short numeric tokens.`
- **P3 — `extractAtUserIds` is unexported but has no `// nolint` or coverage badge**, which is fine — just flagging for consistency. The function is fully covered by the 12 sub-tests, so this is a non-issue.

❓ Questions:
- None. The PR body / Test plan / QA Review Fixes section are all clear and consistent.

Testing / Risk:
- Verified locally: `go test ./platform/dingtalk/ -run "TestExtractAtUserIds|TestReply_IncludesExtractedAtUserIds|TestReply_NoAtUserIdsWhenNoMention" -v` → all 14 cases PASS (0.015s wall). `go vet ./platform/dingtalk/` clean. `gofmt -d` clean. 5/5 CI green on PR head (lint 2m30s / unit-test 3m47s / smoke-test 22s / regression-test 26s / performance-test 43s).
- Uncovered risk: regex `@(\d{4,})` would NOT match inside markdown link `[text](url)` syntax — but DingTalk's markdown renderer strips link syntax before user rendering, so the `@userId` in link text is never user-visible anyway. Verified mentally by trace: if content is `[click @1234567](https://x.com)`, the regex matches `@1234567` and injects `atUserIds: ["1234567"]` — DingTalk will then notify that user, which is probably the intent. Edge case is benign.
- Uncovered risk: very long content (>1MB) — `regexp.FindAllStringSubmatch` with `-1` (all matches) is O(n) but allocates a slice per match. Worst case ~1000 unique IDs allocates 1000 strings. Real-world DingTalk user ID volume is far smaller. Not a concern.

Next step:
- owner: merge PR #1250 (community @wgx521, --approve, dingtalk auto-extract @userId in Reply, 0 conflict). Suggested merge order: this PR can go first in the v1.3.5 batch since it's standalone and doesn't touch shared code paths.
- QA next cycle: continue qa-view inbox review. Remaining open+mergeable candidates: #1239 (community @SXongin, feat acp ModelSwitcher via ACP configOptions).

PR #1275, opened 2026-06-08
fix(acp): add AgentSessionCanceller interface for graceful /stop

SXongin:
## Problem

When `/stop` is issued against an ACP session (e.g. OpenCode via ACP), the engine kills the entire subprocess via `Process.Kill()`. Unlike Claude Code where each turn spawns a new process, ACP agents use a single long-lived server. Killing it destroys the session - the next user message starts a new conversation instead of resuming.

## Solution

Introduces an optional `AgentSessionCanceller` interface that agent sessions can implement to opt into graceful turn cancellation.

## Files changed

- **core/interfaces.go** - Adds AgentSessionCanceller interface with CancelTurn() method
- **agent/acp/session.go** - Implements CancelTurn() on acpSession (sends session/cancel notification)
- **core/engine.go** - Modifies stopInteractiveSessionWithOptions to check for AgentSessionCanceller
- **agent/acp/rpc.go** - Adds transport.sendNotification() for JSON-RPC 2.0 notifications

## Behavior

- /stop on ACP session: sends session/cancel, keeps process alive
- /stop on Claude Code: unchanged (no AgentSessionCanceller)
- Fallback: if CancelTurn() fails, engine falls through to Close()

### DIFF
--- a/agent/acp/rpc.go
+++ b/agent/acp/rpc.go
@@ -202,6 +202,14 @@ func (t *transport) call(ctx context.Context, method string, params any) (json.R
 	}
 }
 
+func (t *transport) sendNotification(method string, params any) error {
+	return t.writeJSON(map[string]any{
+		"jsonrpc": "2.0",
+		"method":  method,
+		"params":  params,
+	})
+}
+
 func (t *transport) writeJSON(v any) error {
 	t.mu.Lock()
 	defer t.mu.Unlock()
--- a/agent/acp/session.go
+++ b/agent/acp/session.go
@@ -698,6 +698,24 @@ func (s *acpSession) CurrentSessionID() string {
 	return s.currentACPSessionID()
 }
 
+// CancelTurn sends an ACP session/cancel notification to abort the current
+// turn while keeping the session alive. The agent should cancel the active
+// LLM request and persist state, then wait for the next user message on the
+// same connection. This is used by /stop instead of Close().
+func (s *acpSession) CancelTurn() error {
+	sid := s.currentACPSessionID()
+	if sid == "" {
+		return fmt.Errorf("acp: no active session to cancel")
+	}
+	if s.tr == nil {
+		return fmt.Errorf("acp: transport not available")
+	}
+	slog.Debug("acp: cancelling current turn", "session_id", sid)
+	return s.tr.sendNotification("session/cancel", map[string]any{
+		"sessionId": sid,
+	})
+}
+
 func (s *acpSession) Alive() bool {
 	return s.alive.Load()
 }
--- a/core/engine.go
+++ b/core/engine.go
@@ -8638,6 +8638,52 @@ func (e *Engine) stopInteractiveSessionWithOptions(sessionKey string, notifyQueu
 	agentSession := state.agentSession
 	state.mu.Unlock()
 
+	// If the agent session supports graceful turn cancellation (e.g. ACP),
+	// send a cancel notification and keep the session alive for the next
+	// user message, rather than killing the process and destroying state.
+	if canceller, ok := agentSession.(AgentSessionCanceller); ok && agentSession != nil {
+		// Keep the state in the map so the next message reuses this session.
+		// Don't markStopped — the session is still usable.
+		// Don't delete from interactiveStates — keep it alive.
+		e.interactiveMu.Unlock()
+
+		if pending != nil {
+			pending.resolve()
+		}
+		if notifyQueued {
+			e.notifyDroppedQueuedMessages(state, fmt.Errorf("session cancelled"))
+		} else {
+			state.mu.Lock()
+			state.pendingMessages = nil
+			state.mu.Unlock()
+		}
+
+		// Mark eventsNeedResync so the next turn drains stale events from
+		// the cancelled turn before processing fresh input.
+		state.mu.Lock()
+		state.eventsNeedResync = true
+		state.mu.Unlock()
+
+		cancelErr := canceller.CancelTurn()
+		if cancelErr != nil {
+			slog.Warn("agent session CancelTurn failed, falling back to Close",
+				"session_key", sessionKey, "error", cancelErr)
+			// Fall through to normal cleanup below.
+			goto normalCleanup
+		}
+
+		slog.Info("agent session turn cancelled, session kept alive",
+			"session_key", sessionKey)
+
+		e.hooks.Emit(HookEvent{
+			Event:      HookEventSessionEnded,
+			SessionKey: sessionKey,
+		})
+
+		return true
+	}
+
+normalCleanup:
 	state.markStopped()
 	delete(e.interactiveStates, sessionKey)
 	e.interactiveMu.Unlock()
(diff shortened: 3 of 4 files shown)


chenhg5 (commented):
Conclusion: Comment (leaning Approve; see P2 below)

Overall assessment:
This is a small, focused, well-shaped PR that adds a clean optional interface for graceful turn cancellation. The design is consistent with the existing `ChannelNameResolver` / `ContextCompressor` optional-interface pattern in `core/interfaces.go`, the ACP implementation is mutex-safe via `t.mu`, and the engine branch correctly falls back to normal cleanup when `CancelTurn` errors. No build break, no P0/P1, and CI is green. I'm posting as `--comment` (not full approve) only because the new code path has **zero unit test coverage** — please add a couple of tests and this becomes a clear approve.

Review scope:
- Reviewed: `core/interfaces.go`, `core/engine.go` (`stopInteractiveSessionWithOptions` line 8624+), `agent/acp/rpc.go` (`sendNotification`), `agent/acp/session.go` (`CancelTurn`).
- 1 commit: `445f033`.
- Diff vs `main`: +80 / -0, 4 files, **0 `_test.go` files**.
- Focused on: optional-interface design, lock ordering, fallback correctness, concurrent /stop, hook semantics, mutex safety.

✅ What looks good:
- **Optional-interface pattern is correct.** `core.AgentSessionCanceller` (interfaces.go:481) follows the same shape as the existing `ChannelNameResolver` / `ContextCompressor` optional interfaces — single method, no required implementation, engine type-asserts. This means every existing agent (claudecode, codex, cursor, opencode, …) keeps working unchanged.
- **Engine branch is well-structured** (engine.go:8641-8684). The flow is: check `cancellable` → drain pending → set `eventsNeedResync` → call `CancelTurn` → on error, `goto normalCleanup`. The `eventsNeedResync=true` flag is the right move — the next foreground turn (line 2706-2710) will drain stale events from the cancelled turn before processing fresh input, which is exactly what you want when an LLM request is aborted mid-stream.
- **ACP implementation is mutex-safe.** `transport.sendNotification` (rpc.go:205) calls `t.writeJSON` which locks `t.mu` (rpc.go:213-217), so the cancel notification can race with a normal `call()` without corrupting the JSON encoder.
- **ACP error paths are explicit** (session.go:707-712): both "no active session" and "transport not available" return clear errors that the engine can log + fall back on. No silent swallow.
- **Concurrency note**: in the new branch, `e.interactiveMu.Unlock()` is called *before* `notifyDroppedQueuedMessages` / `CancelTurn` — this is intentional so the long-running `CancelTurn` call doesn't hold the global session map lock. `notifyDroppedQueuedMessages` re-acquires `state.mu` internally (engine.go:4834), so the unlock chain is safe.
- **Comment in `CancelTurn`** (session.go:701-704) accurately explains the contract: cancel the current turn, keep session alive, wait for next user message. That matches the engine's behavior.

🟠 Should improve:

- **P2 — Add unit tests for the new optional interface + engine branch.** 80 LoC, 0 tests. Suggested minimum set (mocking the agent session in `core`):
  1. `TestStopInteractiveSession_AgentSessionCanceller_UsedAndStateKeptAlive` — pass a `mockCanceller` session, call `stopInteractiveSessionWithOptions`, verify `CancelTurn` was called, verify `e.interactiveStates[sessionKey]` still contains the state, verify `state.stopped == false`, verify `state.eventsNeedResync == true`.
  2. `TestStopInteractiveSession_AgentSessionCanceller_FallbackOnError` — pass a canceller whose `CancelTurn` returns an error, verify the function falls through `goto normalCleanup`, state is `markStopped` + removed from map (matching the original `Close()` path).
  3. `TestStopInteractiveSession_NonCancellerAgent_UsesClosePath` — pass a plain `AgentSession` (not implementing `Canceller`), verify the original behavior is preserved (state removed, `markStopped`).
  4. `TestACPSession_CancelTurn_NotificationShape` (in `agent/acp`) — start a session, capture the `session/cancel` notification from the transport's writer, assert JSON-RPC shape (`jsonrpc:"2.0"`, `method:"session/cancel"`, `params.sessionId` matches `currentACPSessionID`).
  - These four would close the gap and make this a full approve.

🔵 Optional:

- **P3 — `HookEventSessionEnded` semantic mismatch.** engine.go:8678-8681 emits `HookEventSessionEnded` even though the session is explicitly being **kept alive**. Downstream consumers listening to this hook will think the session is over. Either rename the event for this branch (e.g. `HookEventSessionTurnCancelled`) or document in the hook's own doc that "ended" here means "turn ended, session may still be alive". This is a minor contract change, but it's a real one.
- **P3 — Rapid double-`/stop` is not idempotent on the canceller path.** The pre-`Close()` path is idempotent (state gets `markStopped` + removed from the map). The new branch leaves the state in the map without `markStopped`, so two rapid `/stop` commands on the same `sessionKey` would both call `CancelTurn`. The second call would likely be a no-op on the agent side, but engine-side you'd see two `HookEventSessionEnded` emissions and a double drain of `pendingMessages`. Consider a `state.cancelling atomic.Bool` guard, or move `markStopped` into the canceller branch.
- **P3 — Defensive nil check is redundant.** engine.go:8644: `agentSession.(AgentSessionCanceller); ok && agentSession != nil` — the type assertion on a nil `AgentSession` already returns `ok=false`, so the trailing `&& agentSession != nil` is dead code. Cosmetic.
- **P3 — Add CHANGELOG entry** under `## Unreleased` / `### Fixed` (or `### Added`, depending on how the project frames this — it's a new behavior, so Added might be more accurate). release-codex will appreciate it when cutting v1.3.5.

❓ Questions:
- The ACP spec is still in flux. If `session/cancel` semantics change (e.g. becomes a request with response, instead of a notification), the `sendNotification` call would need to switch to `call()`. Is there a pinned ACP version this targets, or do you track the latest?
- Is there a test agent / harness on the ACP side that can simulate a "stuck" turn and verify cancel actually aborts it end-to-end? Manual smoke is fine, but if you have a fixture that can drive it, that would be the strongest evidence.

Testing / Risk:
- Verified locally: `go build ./...` clean, `go vet ./core/ ./agent/acp/` clean. 5/5 CI green on PR head (lint 2m20s, unit-test 4m3s, smoke-test 27s, regression-test 28s, performance-test 48s). `go test ./core/ -run TestStopInteractive` has no matching tests today (confirmed by `grep`).
- Uncovered risk: the new branch in `stopInteractiveSessionWithOptions` is the most likely place for a regression (race on the new early-unlock, or staleness in `eventsNeedResync`). The four tests suggested above would cover this. The ACP layer is well-isolated and the wire format is a one-liner, so the only real concern is whether the JSON-RPC peer actually responds to `session/cancel` — which is an agent-side concern, not a cc-connect concern.

Next step:
Author @SXongin: please add the four tests listed above (especially the two engine-level mocks in `core/`). Once they're in and CI is green, this is ready to merge — the design and the implementation are both solid. Re-request review when ready. (If the four tests are a stretch, even just the two `core/` mocks would be enough to flip me to a formal Approve.)

PR #1305, opened 2026-06-11
fix(core): defer session restart on /model until the in-flight turn completes

### LINKED ISSUE #1303: [Bug] /model during an in-flight turn kills the agent session — in-progress response is silently lost
## Version

cc-connect v1.3.3-beta.2 (commit f6a1a794), Claude Code agent, Discord platform, Linux.

## Summary

If the user issues `/model` while the agent is still generating a response to a previous message, the model switch closes the agent session immediately (stdin close → process exit). The in-flight turn never reaches `turn complete`, and its response is silently dropped — the user never receives it and gets no error or notice.

## Log timeline (redacted)

```
18:39:28 INFO  message received            msg_id=...298 content_len=55 has_images=true
18:39:28 INFO  processing message          session=s14
18:39:28 INFO  session spawned             is_resume=true
18:40:02 INFO  audit: command_executed     command=model
18:40:02 INFO  claudecode: model changed   model=claude-fable-5
18:40:02 WARN  slow agent send             elapsed=33.7s content_len=55
18:40:11 INFO  claudeSession: exited cleanly after stdin close
18:40:11 WARN  slow agent session close    elapsed=8.97s
```

Note there is **no `turn complete` for msg_id ...298** anywhere in the log, while every other message that evening has one. The agent process had actually finished generating the reply (it is persisted in the Claude Code session `.jsonl` transcript), but cc-connect closed the session before the turn-complete event was processed, so the reply was never delivered to Discord.

## Steps to reproduce

1. Send a message that takes the agent a while to answer (e.g. one with an image attachment).
2. While the agent is still working, run `/model <other-model>`.
3. The response to step 1 never arrives; no error is shown.

## Expected behavior

Either of:
- Queue the model change and apply it after the in-flight turn completes (preferred), or
- Refuse the switch with a "turn in progress, try again shortly" notice, or
- At minimum, deliver an explicit "your previous message was interrupted" notice instead of dropping the reply silently.

Silent loss is the worst outcome — the user has no way to know the reply existed.

archibate:
Fixes #1303

## Problem

Switching models (`/model`, or the model card action) called `cleanupInteractiveState` unconditionally, closing the agent session even while a turn was in flight. The agent had often already generated the reply (it is persisted in the CLI transcript), but the session was torn down before the turn-complete event was processed — the response silently never reached the user.

## Fix

New helper `restartInteractiveSessionWhenIdle`:

- **Idle session** → behavior unchanged: restart immediately.
- **Turn in flight** (detected via the session lock that turn processing already holds) → the restart is deferred to a background goroutine that waits for the lock to be released, so the in-flight reply is delivered normally and the new model applies from the next turn. A 30-minute ceiling guards against a stuck turn (then it falls back to the old behavior).

`cmdModel` replies with a dedicated i18n message in the deferred case (EN/ZH/ZH-TW/JA/ES added) so the user knows the switch takes effect after the current turn.

Applied to `cmdModel` and `handleModelCardAction`. The card-driven `executeCardAction` flow shares the same root issue but immediately re-creates interactive state for its progress card, so it is intentionally left unchanged here — happy to extend if you'd like.

## Tests

- `TestRestartInteractiveSessionWhenIdle_DefersWhileTurnInFlight` — locks the session, asserts the agent session is **not** closed while busy, then asserts it **is** closed after unlock.
- `TestRestartInteractiveSessionWhenIdle_RestartsImmediatelyWhenIdle` — idle path unchanged.
- Full `go test ./core/` passes.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -3736,6 +3736,37 @@ func (e *Engine) cleanupInteractiveState(sessionKey string, expected ...*interac
 	e.interactiveMu.Unlock()
 }
 
+// settingsChangeIdleWaitCeiling bounds how long a deferred session restart
+// (e.g. after /model) waits for an in-flight turn to finish.
+const settingsChangeIdleWaitCeiling = 30 * time.Minute
+
+// restartInteractiveSessionWhenIdle closes the interactive agent session so
+// the next turn respawns with updated settings (e.g. a new model). Unlike
+// calling cleanupInteractiveState directly, an in-flight turn is not killed:
+// when the session is busy the restart is deferred to a background goroutine
+// that waits for the turn to finish first, so the in-flight response is
+// delivered normally and the new settings apply from the next turn onward.
+// Returns true when the restart was deferred.
+func (e *Engine) restartInteractiveSessionWhenIdle(sessions *SessionManager, sessionKey, interactiveKey string) bool {
+	session := sessions.GetOrCreateActive(sessionKey)
+	if session.TryLock() {
+		session.UnlockWithoutUpdate()
+		e.cleanupInteractiveState(interactiveKey)
+		return false
+	}
+
+	go func() {
+		// Wait for the in-flight turn to release the session lock. If the
+		// ceiling is hit the restart proceeds anyway — equivalent to the
+		// pre-deferral behavior, just much later.
+		if e.waitForSessionLock(session, settingsChangeIdleWaitCeiling) {
+			session.UnlockWithoutUpdate()
+		}
+		e.cleanupInteractiveState(interactiveKey)
+	}()
+	return true
+}
+
 func (e *Engine) closeAgentSessionAsync(sessionKey string, agentSession AgentSession) {
 	if agentSession == nil {
 		return
@@ -8857,13 +8888,17 @@ func (e *Engine) cmdModel(p Platform, msg *Message, args []string) {
 		e.reply(p, msg.ReplyCtx, e.i18n.Tf(MsgModelChangeFailed, err))
 		return
 	}
-	e.cleanupInteractiveState(interactiveKey)
+	deferred := e.restartInteractiveSessionWhenIdle(sessions, msg.SessionKey, interactiveKey)
 
 	// Keep the existing agent session ID so the next StartSession uses
 	// --resume <id> --model <new>, which lets the CLI agent restore context
 	// natively without replaying history (no extra token cost).
 	sessions.Save()
 
+	if deferred {
+		e.reply(p, msg.ReplyCtx, e.i18n.Tf(MsgModelChangedDeferred, target))
+		return
+	}
 	e.reply(p, msg.ReplyCtx, e.i18n.Tf(MsgModelChanged, target))
 }
 
@@ -10800,7 +10835,7 @@ func (e *Engine) handleModelCardAction(args, sessionKey string) *Card {
 	}
 
 	resolved, err := e.switchModelOnAgent(agent, target, agent == e.agent)
-	e.cleanupInteractiveState(e.interactiveKeyForSessionKey(sessionKey))
+	e.restartInteractiveSessionWhenIdle(sessions, sessionKey, e.interactiveKeyForSessionKey(sessionKey))
 	if err == nil {
 		sessions.Save()
 	}
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -14326,3 +14326,63 @@ func TestHandlePendingPermission_StalePermissionCallback_Dropped(t *testing.T) {
 		}
 	})
 }
+
+// --- restartInteractiveSessionWhenIdle tests ---
+
+type closeTrackingAgentSession struct {
+	stubAgentSession
+	closed atomic.Bool
+}
+
+func (s *closeTrackingAgentSession) Close() error {
+	s.closed.Store(true)
+	return nil
+}
+
+func TestRestartInteractiveSessionWhenIdle_DefersWhileTurnInFlight(t *testing.T) {
+	e := newTestEngine()
+	sessionKey := "test:chat:user"
+	as := &closeTrackingAgentSession{}
+	e.interactiveStates[sessionKey] = &interactiveState{agentSession: as}
+
+	session := e.sessions.GetOrCreateActive(sessionKey)
+	if !session.TryLock() {
+		t.Fatalf("expected to lock fresh session")
+	}
+
+	deferred := e.restartInteractiveSessionWhenIdle(e.sessions, sessionKey, sessionKey)
+	if !deferred {
+		t.Fatalf("expected restart to be deferred while turn is in flight")
+	}
+
+	// The in-flight turn must not be killed by the model switch.
+	time.Sleep(100 * time.Millisecond)
+	if as.closed.Load() {
+		t.Fatalf("agent session closed while turn was still in flight")
+	}
+
+	// Turn finishes — the deferred restart should now close the session.
+	session.Unlock()
+	deadline := time.Now().Add(5 * time.Second)
+	for !as.closed.Load() {
+		if time.Now().After(deadline) {
+			t.Fatalf("agent session was not closed after the turn finished")
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+}
+
+func TestRestartInteractiveSessionWhenIdle_RestartsImmediatelyWhenIdle(t *testing.T) {
+	e := newTestEngine()
+	sessionKey := "test:chat:user"
+	as := &closeTrackingAgentSession{}
+	e.interactiveStates[sessionKey] = &interactiveState{agentSession: as}
+
+	deferred := e.restartInteractiveSessionWhenIdle(e.sessions, sessionKey, sessionKey)
+	if deferred {
+		t.Fatalf("expected immediate restart when no turn is in flight")
+	}
+	if !as.closed.Load() {
+		t.Fatalf("agent session should be closed immediately when idle")
+	}
+}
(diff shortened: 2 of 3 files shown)


archibate:
The unit-test failure was in `TestAllowChat_FiltersGroupMessages` (platform/feishu), which races a 2-second timeout against real network calls to the Feishu endpoint — it fails on different subtests across runs and reproduces on unmodified code locally. This PR only touches `core/`. Pushed an empty commit to retrigger CI; happy to help harden that test in a separate PR if useful.

chenhg5 (commented):
Conclusion: Comment

Overall assessment:
- This PR addresses #1303 with the same root-cause analysis (cleanupInteractiveState kills in-flight session, response silently dropped) but the *implementation style* diverges from a fix that has already landed on `main` (commit 3c2f483a, PR #1319) covering most of the same code paths. I would recommend **closing this PR in favor of #1319** (or stripping it down to the one path #1319 missed) rather than merging it as-is.

Review scope:
- Reviewed `core/engine.go` (+37/-2) on worktree `qa/pr-1305-verify` @ 389cc1eb, base origin/main c53f5450.
- Cross-referenced against `main` (commit 3c2f483a, "fix(core): queue /model during in-flight turn, apply on turn complete (#1303)") which is the in-main fix for the same issue.
- Focused on correctness, scope/duplication, and mergeability.

✅ What looks good:
- The root-cause analysis in the PR body is exactly right — `cleanupInteractiveState` closes the agent session even while a turn is in flight, the in-flight reply gets dropped. The 30-minute ceiling on the deferral is a reasonable guardrail.
- The 2 new tests (`TestRestartInteractiveSessionWhenIdle_DefersWhileTurnInFlight` + `_RestartsImmediatelyWhenIdle`) cleanly cover both branches and the existing `go test ./core/` is green. CI run 27354694871 5/5 green on the PR.
- The 5-language i18n entries for `MsgModelChangedDeferred` (EN/ZH/ZH-TW/JA/ES) are complete and semantically aligned with `MsgModelChanged`.

❌ Strong recommendation — close as superseded or strip to the unique path:
- **The `handleModelCardAction` change in this PR is fully SUPERSEDED by #1319 (already on main, commit 3c2f483a).** #1319's `handleModelCardAction` already implements the same in-flight guard via `queuePendingModelSwitch` (`if state.isTurnInFlightLocked() { return e.queuePendingModelSwitch(state, sessionKey, target) }` at line ~10841) and its `executeCardAction` `case /model` branch is also queueing-aware. Merging this PR's `handleModelCardAction` change on top of #1319 would create duplicate logic and a behavioral split (two different ways of deferring the same action).
- **The `cmdModel` change in this PR is unique to this PR** — `cmdModel` (text-command path) still calls `e.cleanupInteractiveState(interactiveKey)` directly on `main`, with no in-flight guard. So your PR does fix a real unfixed path.

Three options, in order of cleanliness:

  (a) **Close this PR and have dev-claudecode add the queueing guard to `cmdModel` as a follow-up to #1319.** The queueing pattern (`queuePendingModelSwitch` + `applyPendingModelSwitchIfAny` + `pendingModelSwitchTimeout`) is already wired up on `main`; just apply it to the `cmdModel` path. Keeps the codebase using a single deferral mechanism instead of two.

  (b) **Close this PR and accept that `cmdModel` is an open follow-up.** The "fix /model during in-flight" issue is closed in the user's experience (the card path now works), and the text command path is a smaller-friction workaround (user can wait for the turn to finish before issuing /model). Not recommended, since the bug is real for users who use the text command.

  (c) **Rebase and shrink this PR to ONLY cover the `cmdModel` change.** Drop the `handleModelCardAction` duplicate, keep the unique `cmdModel` path, switch to the `queuePendingModelSwitch` pattern instead of `restartInteractiveSessionWhenIdle` (to share #1319's machinery). 2 new tests + the one unique diff. ~30 lines, easy to review, no duplication.

❓ Question:
- The PR body says the `executeCardAction` "card-driven" path "shares the same root issue but immediately re-creates interactive state for its progress card, so it is intentionally left unchanged". On `main`, #1319's commit does cover `executeCardAction` with the queueing guard. Was this written before #1319 was merged, or did you decide to opt out of that change? (Worth knowing for the close-vs-shrink decision above.)

Testing / Risk:
- Verified evidence: CI run 27354694871 5/5 green on this PR (lint 2m31s, unit-test 3m57s, smoke 24s, regression 24s, performance 48s). `gofmt -l core/engine.go` clean. Identifiers (`restartInteractiveSessionWhenIdle`, `settingsChangeIdleWaitCeiling`, `MsgModelChangedDeferred`) not present on `main`. 0 SUPERSEDED hits via `git grep` for the symbols themselves.
- Remaining risk: If this PR is merged without removing the `handleModelCardAction` duplicate, the engine will have two deferral mechanisms (lock-wait vs in-place queue), which is the kind of "inconsistent lifecycle handling" that caused the original bug. Pick (a) or (c) above.

Next step:
- Owner decision: close this PR and let dev-claudecode extend the existing #1319 queueing to `cmdModel` (option a), or rebase this PR down to the unique `cmdModel` change (option c). I lean toward (a) since #1319 is already merged and the queueing mechanism is the established pattern.

chenhg5 (changes requested):
Conclusion: Request changes

Overall assessment:
- This PR addresses #1303 with the same root-cause analysis (cleanupInteractiveState kills in-flight session, response silently dropped) but the *implementation style* diverges from a fix that has already landed on `main` (commit 3c2f483a, PR #1319) covering most of the same code paths. The `handleModelCardAction` change in this PR is fully SUPERSEDED. I would recommend **closing this PR in favor of #1319** (or stripping it down to the one path #1319 missed) rather than merging it as-is.

Review scope:
- Reviewed `core/engine.go` (+37/-2) on worktree `qa/pr-1305-verify` @ 389cc1eb, base origin/main c53f5450.
- Cross-referenced against `main` (commit 3c2f483a, "fix(core): queue /model during in-flight turn, apply on turn complete (#1303)") which is the in-main fix for the same issue.
- Focused on correctness, scope/duplication, and mergeability.

✅ What looks good:
- The root-cause analysis in the PR body is exactly right — `cleanupInteractiveState` closes the agent session even while a turn is in flight, the in-flight reply gets dropped. The 30-minute ceiling on the deferral is a reasonable guardrail.
- The 2 new tests (`TestRestartInteractiveSessionWhenIdle_DefersWhileTurnInFlight` + `_RestartsImmediatelyWhenIdle`) cleanly cover both branches and the existing `go test ./core/` is green. CI run 27354694871 5/5 green on the PR.
- The 5-language i18n entries for `MsgModelChangedDeferred` (EN/ZH/ZH-TW/JA/ES) are complete and semantically aligned with `MsgModelChanged`.

🔴 Must fix (Request changes — these are blocking, but the resolution is most likely "close the PR"):
- **P0: The `handleModelCardAction` change in this PR is fully SUPERSEDED by #1319 (already on main, commit 3c2f483a).** #1319's `handleModelCardAction` already implements the same in-flight guard via `queuePendingModelSwitch` (`if state.isTurnInFlightLocked() { return e.queuePendingModelSwitch(state, sessionKey, target) }` at line ~10841) and its `executeCardAction` `case /model` branch is also queueing-aware. Merging this PR's `handleModelCardAction` change on top of #1319 would create duplicate logic and a behavioral split (two different ways of deferring the same action).
  Impact: User-visible behavior is inconsistent depending on whether the model switch came from a card or text command. Code review and future maintenance get harder.
  Evidence: `git show origin/main:core/engine.go | grep -A 5 'func (e \*Engine) handleModelCardAction'` shows the queueing pattern is already there; the new code on this branch's `handleModelCardAction` would be a parallel implementation using `restartInteractiveSessionWhenIdle`.
  Suggested fix: pick one of three options.
    (a) **Close this PR + dev-claudecode follow-up**: have dev-claudecode add the queueing guard to `cmdModel` as a follow-up to #1319. The queueing pattern (`queuePendingModelSwitch` + `applyPendingModelSwitchIfAny` + `pendingModelSwitchTimeout`) is already wired up on `main`; just apply it to the `cmdModel` path. Keeps the codebase using a single deferral mechanism instead of two.
    (b) **Close this PR + accept that `cmdModel` is an open follow-up**. The "fix /model during in-flight" issue is closed in the user's experience (the card path now works). Not recommended, since the bug is real for users who use the text command.
    (c) **Rebase and shrink this PR to ONLY cover the `cmdModel` change**. Drop the `handleModelCardAction` duplicate, keep the unique `cmdModel` path, switch to the `queuePendingModelSwitch` pattern instead of `restartInteractiveSessionWhenIdle` (to share #1319's machinery). 2 new tests + the one unique diff. ~30 lines, easy to review, no duplication.

🟠 Should improve:
- **P1: The `cmdModel` change is unique to this PR** (text-command path) — `cmdModel` still calls `e.cleanupInteractiveState(interactiveKey)` directly on `main`, with no in-flight guard. So your PR does fix a real unfixed path. The implementation style is fine, but consider switching to the `queuePendingModelSwitch` pattern to keep a single deferral mechanism (option c above).

❓ Question:
- The PR body says the `executeCardAction` "card-driven" path "shares the same root issue but immediately re-creates interactive state for its progress card, so it is intentionally left unchanged". On `main`, #1319's commit does cover `executeCardAction` with the queueing guard. Was this written before #1319 was merged, or did you decide to opt out of that change? (Worth knowing for the close-vs-shrink decision above.)

Testing / Risk:
- Verified evidence: CI run 27354694871 5/5 green on this PR (lint 2m31s, unit-test 3m57s, smoke 24s, regression 24s, performance 48s). `gofmt -l core/engine.go` clean. Identifiers (`restartInteractiveSessionWhenIdle`, `settingsChangeIdleWaitCeiling`, `MsgModelChangedDeferred`) not present on `main`. 0 SUPERSEDED hits via `git grep` for the symbols themselves.
- Remaining risk: If this PR is merged without removing the `handleModelCardAction` duplicate, the engine will have two deferral mechanisms (lock-wait vs in-place queue), which is the kind of "inconsistent lifecycle handling" that caused the original bug. Pick (a) or (c) above.

Next step:
- Owner decision: close this PR and let dev-claudecode extend the existing #1319 queueing to `cmdModel` (option a), or rebase this PR down to the unique `cmdModel` change (option c). I lean toward (a) since #1319 is already merged and the queueing mechanism is the established pattern.

PR #1314, opened 2026-06-11
fix(daemon): remove redundant linger_other.go that breaks non-linux builds

baocang:
## Summary
`daemon/linger_other.go` (`//go:build !linux`) duplicates `CheckLinger`, which is already defined by `launchd.go` (darwin), `windows.go` (windows) and `unsupported.go` (other). It is therefore redeclared on every non-Linux platform, so `make build` / `go build ./...` fails on macOS and Windows. Linux is unaffected (`systemd.go` provides it). The file is dead code — remove it.

## Test plan
- `go build ./...` / `make build` on macOS now succeeds.
- Linux build unchanged (`linger_other.go` was already excluded there).

Fixes #1311

Issue #1311, opened 2026-06-11
Build fails on non-Linux (macOS/Windows): CheckLinger redeclared in package daemon


baocang:
### What happened
`go build ./...` / `make build` fails on macOS (and Windows):

```
daemon/linger_other.go:6:6: CheckLinger redeclared in this block
        daemon/launchd.go:32:6: other declaration of CheckLinger
```

### Root cause
`daemon/linger_other.go` has `//go:build !linux` and defines `CheckLinger`. But `CheckLinger` is already defined for every non-Linux platform:

- darwin → `daemon/launchd.go`
- windows → `daemon/windows.go`
- other → `daemon/unsupported.go` (`!linux && !darwin && !windows`)

So on darwin/windows/other the `!linux` file is compiled alongside the platform file and `CheckLinger` is redeclared. Linux is fine (`linger_other.go` excluded; `systemd.go` provides it).

### Fix
`daemon/linger_other.go` is dead code — every platform already has its own `CheckLinger`. Remove it.

### Environment
macOS (Apple Silicon), building from source at current `main`.

chenhg5:
Hi @baocang, thanks for the detailed report with root cause analysis! 

We've confirmed your analysis — `daemon/linger_other.go` is dead code (every platform has its own `CheckLinger` already in `launchd.go`/`windows.go`/`unsupported.go`). This is a clean single-file fix, dispatching to dev-claudecode now.

我们会处理。

你好 @baocang, 感谢详细的根因分析报告! 

我们已确认你的分析: `daemon/linger_other.go` 是死代码 (每个平台在 `launchd.go`/`windows.go`/`unsupported.go` 中已有自己的 `CheckLinger`). 这是一个干净的单文件修复, 已派给 dev-claudecode.

chenhg5:
Update from cc-connect PM:

@baocang, 注意到您已自提 PR #1314 (https://github.com/chenhg5/cc-connect/pull/1314) 删除 `daemon/linger_other.go` (8 deletions, 1 file, 0 additions). 这个修复和我们的 triage 方向完全一致, 而且您直接做了, 非常感谢!

我已经把 PR #1314 转给 qa-claudecode review 流程, qa 通过后会请 owner merge. 期间:
- 验证 4 平台 (linux/darwin/windows/other) build 成功
- go test ./daemon/... ./core/... ./config/... ./agent/... 全绿
- CI 5/5 green
- golangci-lint 0 issues
- mergeable

[...]

PR #1318, opened 2026-06-11
fix(codex): kill app-server process groups on close

AaronZ345:
## Summary

- start Codex app-server sessions in their own process group and force-kill the group on close
- add a unix regression test that proves app-server child processes do not survive `Close()`
- restore macOS daemon testability by removing the duplicate non-linux linger stub and seeding the launchd status test plist

## Root cause

`appServerSession.Close()` only called `Process.Kill()` on the `codex` process. With the npm shim, that can leave the vendor `codex app-server` child process and its spawned helpers alive. Timed-out or completed app-server sessions can therefore accumulate under a long-running cc-connect daemon, leaving stale Codex processes around for later cron/message turns.

The classic Codex session path already used process-group handling; app-server sessions were missing the same lifecycle behavior.

## Tests

- `go test ./agent/codex -run TestAppServerSession_CloseKillsProcessGroup -count=1 -v`
- `go test ./agent/codex`
- `go test ./...`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 ### ⚠️ QQ Bot Intent Configuration Change
 The default intents for QQ Bot now include `INTERACTION_CREATE` (bit 26, value `1<<26`). If you previously set a custom `intents` value without this bit, inline keyboard buttons will not work — update your `intents` to include bit 26. If you use the default intents, no action is needed. See `config.example.toml` for the new `intents` option.
 
+### Fixed
+- **Codex app-server process leak on close**: `appServerSession.Close()` now starts the codex process in its own process group and force-kills the group, so child codex processes and helpers no longer survive after a session is closed (#1318).
+
 ## v1.3.3-beta.4 (2026-05-28)
 
 ### New Features
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -249,6 +249,7 @@ func (s *appServerSession) connect() error {
 	}
 	cmd := exec.CommandContext(s.ctx, "codex", args...)
 	cmd.Dir = s.workDir
+	prepareCmdForKill(cmd)
 	env := append([]string(nil), s.extraEnv...)
 	if s.codexHome != "" {
 		env = append(env, "CODEX_HOME="+s.codexHome)
@@ -936,7 +937,10 @@ func (s *appServerSession) Close() error {
 		s.stdin = nil
 	}
 	if s.cmd != nil && s.cmd.Process != nil {
-		_ = s.cmd.Process.Kill()
+		if err := forceKillCmd(s.cmd); err != nil {
+			slog.Debug("codex app-server: force kill failed", "error", err)
+		}
+		s.cmd = nil
 	}
 	s.procMu.Unlock()
 
--- a/agent/codex/appserver_session_unix_test.go
+++ b/agent/codex/appserver_session_unix_test.go
@@ -1,0 +1,174 @@
+//go:build unix
+
+package codex
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestAppServerSession_CloseKillsProcessGroup(t *testing.T) {
+	if os.Getenv("CC_CONNECT_CODEX_HELPER") == "1" {
+		runFakeCodexAppServer(t)
+		return
+	}
+
+	workDir := t.TempDir()
+	binDir := filepath.Join(workDir, "bin")
+	if err := os.MkdirAll(binDir, 0o755); err != nil {
+		t.Fatalf("mkdir bin: %v", err)
+	}
+
+	childPIDFile := filepath.Join(workDir, "child.pid")
+	script := `#!/bin/sh
+exec "$CC_CONNECT_CODEX_HELPER_BIN" -test.run TestAppServerSession_CloseKillsProcessGroup --
+`
+	scriptPath := filepath.Join(binDir, "codex")
+	if err := os.WriteFile(scriptPath, []byte(script), 0o755); err != nil {
+		t.Fatalf("write fake codex: %v", err)
+	}
+
+	t.Setenv("CHILD_PID_FILE", childPIDFile)
+	t.Setenv("CC_CONNECT_CODEX_HELPER", "1")
+	t.Setenv("CC_CONNECT_CODEX_HELPER_BIN", os.Args[0])
+	t.Setenv("PATH", binDir+string(os.PathListSeparator)+os.Getenv("PATH"))
+
+	s, err := newAppServerSession(context.Background(), "stdio://", workDir, "", "", "", "", "", "", nil, "")
+	if err != nil {
+		t.Fatalf("newAppServerSession: %v", err)
+	}
+
+	childPID := waitForPIDFile(t, childPIDFile)
+	t.Cleanup(func() {
+		if processRunning(childPID) {
+			_ = syscall.Kill(childPID, syscall.SIGKILL)
+		}
+	})
+
+	parentPID := appServerProcessPID(t, s)
+	assertSameProcessGroup(t, parentPID, childPID)
+
+	if err := s.Close(); err != nil {
+		t.Fatalf("Close: %v", err)
+	}
+
+	deadline := time.Now().Add(time.Second)
+	for time.Now().Before(deadline) {
+		if !processRunning(childPID) {
+			return
+		}
+		time.Sleep(20 * time.Millisecond)
+	}
+	t.Fatalf("child process %d still alive after app-server Close", childPID)
+}
+
+func runFakeCodexAppServer(t *testing.T) {
+	t.Helper()
+
+	child := exec.Command("sleep", "30")
+	if err := child.Start(); err != nil {
+		t.Fatalf("start child sleep: %v", err)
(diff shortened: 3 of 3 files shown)


chenhg5 (changes requested):
Conclusion: Request changes

Overall assessment:
- The codex process-group fix in `agent/codex/appserver_session.go` is well-scoped and the new `appserver_session_unix_test.go` is exactly the kind of regression test this kind of lifecycle change needs. However, the new test FAILS reliably in WSL2 (and likely other environments where the npm-shim / shell-fork model diverges from CI's native Linux runners), and the PR also bundles an unrelated daemon test-fragility fix and a duplicate of an already-merged `linger_other.go` deletion. Please address the P1 below before merge, and consider whether the P2 conflation is what you want.

Review scope:
- Reviewed 4 files in this PR: `agent/codex/appserver_session.go`, `agent/codex/appserver_session_unix_test.go`, `daemon/launchd_test.go`, `daemon/linger_other.go`.
- Verified on worktree `qa/pr-1318-verify` @ 85066e17 (base origin/main c53f5450). Re-ran the new test on the same WSL2 host the owner is using (Linux 5.15.146.1-microsoft-standard-WSL2). Cross-checked PR #1314 (baocang, also `linger_other.go` deletion, already approved/ready) and PR body.
- Focused on correctness, platform-compatibility, test coverage, and PR hygiene.

✅ What looks good:
- The `prepareCmdForKill` + `forceKillCmd` separation cleanly mirrors the `claudecode/proc_unix.go` and `copilot/proc_unix.go` patterns — the codebase now has a uniform "always run child agents in their own process group, kill the group on Close" contract. The PR description's root-cause analysis (npm-shim, vendor `codex app-server` child surviving `Process.Kill()`) is plausible and the fix is targeted.
- `slog.Debug("codex app-server: force kill failed", ...)` preserves the existing best-effort Close semantics (no error returned) while still giving admin something to grep for.
- The new test's fake-codex shell script is well-designed: it forks `(sleep 30) &`, writes the child PID to a file the test reads, and serves just enough JSON-RPC to drive the session to `thread/start` + `account/rateLimits/read` so `Close()` actually runs.

🔴 Must fix (P1):
- **WSL2 / process-group compatibility gap — the new regression test does not actually pass on the maintainer's primary dev environment.**
  Impact: The test is the only thing proving the lifecycle fix works. If it only passes in CI's native Linux but fails in WSL2 (a substantial fraction of cc-connect's user base runs cc-connect inside WSL2 on Windows, and the maintainer themselves likely tests there), then the regression test cannot be relied on to catch future breakages on those platforms, and we have no evidence the fix actually works under WSL2. This is the kind of bug that ships silently and then manifests as "stale codex processes after cron runs" on Windows-laptop users.
  Evidence: On this WSL2 host, `go test -count=1 -run TestAppServerSession_CloseKillsProcessGroup -v ./agent/codex` fails 3/3 in a row with:
  ```
  appserver_session_unix_test.go:74: child process 8676 still alive after app-server Close
  --- FAIL: TestAppServerSession_CloseKillsProcessGroup (1.01s)
  ```
  The same test PASSES in CI (run 27365656139 unit-test job at 17:37-17:41, 0.00s). The CI log shows `--- PASS: TestAppServerSession_CloseKillsProcessGroup (0.00s)`.
  Suggested fix (pick one):
  (a) Reproduce the WSL2 failure locally, then diagnose why the `(sleep 30) &` child is not in the same process group as the fake `codex` script under WSL2. Likely the script is being exec'd via `/bin/sh` which on some WSL2 builds forks+execs and the exec'd `codex` loses the pgid. Try `set -m` at the top of the fake script, or use `exec` in the subshell, or use Go's `syscall.SysProcAttr{Pgid: 0}` + `Setpgid: true` on the fake shell command if you go that direction.
  (b) If the fix ends up requiring platform-specific code, gate the test or the production code with a build tag and document the limitation.
  (c) If the WSL2 issue turns out to be an environment bug rather than a real code gap, document that explicitly in the test (e.g. a comment explaining "test passes on native Linux runners; WSL2 has known fork-exec differences; manual verification on macOS/WSL2 still required") and at least add a CI matrix entry for WSL2.
  Verification: Re-run `go test -count=1 -run TestAppServerSession_CloseKillsProcessGroup -v ./agent/codex` on the WSL2 host after the fix and confirm it passes. Also re-run on macOS to confirm the fix is portable.

🟠 Should improve (P2, not strictly blocking but should be addressed):
- **PR conflation: `daemon/linger_other.go` deletion duplicates PR #1314 (baocang, already approved/ready-to-merge, sitting in the queue).** PR #1314 deletes the same 8-line file with the same rationale ("redundant duplicate CheckLinger stub"). If both PRs merge, the second one is a no-op (file already gone) but the conflation is still a smell — two PRs both fixing #1311 with the same code change means one of them should be closed.
  Suggested fix: Coordinate with PR #1314's author (baocang) — either close this PR's `linger_other.go` deletion and let #1314 land first, or close #1314 and land this one. If the intent was to combine the fixes, say so in the PR body so the owner can choose. (One option: dev-claudecode also has an equivalent fix on `agent/cc-connect/dev-claudecode/issue-1311-delete-linger-other` per the PR #1314 review notes — there are now three branches racing to fix the same file.)

- **PR conflation: `daemon/launchd_test.go` change is unrelated to the codex process-group fix.** The added HOME temp-dir + plist file seeding makes the launchd status test more hermetic, but it has nothing to do with the codex app-server lifecycle. Bundling it here makes the PR harder to review and harder to bisect.
  Suggested fix: Split into a separate PR ("test(daemon): hermetic HOME + plist seeding for launchd status test"). If it's small enough (< 20 lines) you can leave it with a clear commit message, but ideally as its own PR.

- **CHANGELOG entry missing.** No entry under the current top section for either the codex process-group fix or the linger deletion. This is a public-facing change (lifecycle behavior on Close) that users running cc-connect for a long time may want to know about.
  Suggested fix: Add a `### Fixed` bullet at the top of `CHANGELOG.md` (before `## v1.3.3-beta.4`) along the lines of: `**Codex app-server process leak on close**: \`appServerSession.Close()\` now starts the codex process in its own process group and force-kills the group, so vendor codex children no longer survive after a session is closed (#1318).` Adjust the wording if `daemon/linger_other.go` deletion is going to be split out — that one deserves its own `### Fixed` bullet too.

🔵 Optional (P3):
- The new test re-implements `waitForPIDFile` / `processExists` helpers; consider whether `agent/claudecode/proc_unix_test.go` (or wherever the analogous test lives) has helpers you can share to avoid duplicating process-exists probes across the three agent runtimes (claudecode, codex, copilot).
- `appserver_session_unix_test.go` does not exercise the `slog.Debug("codex app-server: force kill failed", ...)` branch — consider adding a subtest where `forceKillCmd` is forced to fail (e.g. set `s.cmd.Process = nil` after start) to confirm the slog path is wired correctly.

❓ Questions:
- The test uses `(sleep 30) &` as the child. Have you also verified with a more realistic child (e.g. a child that itself forks a grandchild, like the real npm-shim → codex app-server → helper chain)? The current test only proves the immediate child dies; the original bug report mentioned "vendor `codex app-server` child process and its spawned helpers" (plural). If the test only covers one level, that's worth a comment explaining the limitation.
- Is the duplicate `linger_other.go` deletion in this PR intentional (you didn't see #1314) or deliberate (you wanted to fix it as part of the codex PR for some reason)? The PR body doesn't mention it, which makes me think it might be unintentional.

Testing / Risk:
- Verified evidence: CI run 27365656139 5/5 green (lint 2m17s, unit-test 4m19s, smoke-test 28s, regression-test 27s, performance-test 44s). The new test is included in the unit-test job and PASSES there. Local WSL2: test FAILS 3/3 with the exact error above. gofmt, go vet, go build on the worktree all clean.
- Remaining risk: The P1 WSL2 gap is the dominant risk — without a fix, we have no evidence the production code actually works under WSL2, and WSL2 is a non-trivial deployment target. The launchd_test.go / linger_other.go conflation is hygiene-level risk but the duplicate fix in three branches (#1314, this PR, dev-claudecode's branch) is a coordination risk that the owner should resolve before merge.

Next step:
- Investigate why the new test fails in WSL2 (likely shell-fork vs exec semantics around `(sleep 30) &` under WSL2's `/bin/sh`). Either fix the test to pass on WSL2 too, or add a build-tag exclusion + CI matrix entry and document the WSL2 limitation. Once the P1 is resolved, decide whether to keep the `linger_other.go` deletion in this PR or close this PR in favor of #1314.

AaronZ345:
@chenhg5 I've pushed an update addressing the requested changes.

What changed:
- Reworked `TestAppServerSession_CloseKillsProcessGroup` to avoid the WSL2 shell/zombie false positive: the test now uses the current Go test binary as the fake `codex app-server`, asserts the fake app-server parent and helper child share the same process group, and treats Linux/WSL2 zombie PIDs as already exited for the liveness check.
- Removed the unrelated `daemon/launchd_test.go` and `daemon/linger_other.go` changes from this PR; #1318 now only touches the Codex process-group fix plus its regression test and changelog entry.
- Added the missing `CHANGELOG.md` entry for the Codex app-server process leak fix.

Verification:
- Confirmed the updated regression test fails against `origin/main` before the production fix.
- `go test -count=1 -run TestAppServerSession_CloseKillsProcessGroup -v ./agent/codex`
- `go test -count=1 ./agent/codex`
- `git diff --check`
- GitHub CI is 5/5 green on `03a3efb`: lint, unit-test, smoke-test, regression-test, and performance-test.

Please take another look when you have a chance.

AaronZ345:
Closing this because the production failure we were chasing was not an app-server process-group leak. The reproduced 17:05 cron failure was caused by a scheduled job reusing a busy chat session; I will open a narrower cron-session fix instead.

PR #1328, opened 2026-06-12
feat(agent/antigravity): bridge tool permissions to chat

xxb:
## Summary

- bridge Antigravity `PreToolUse` hooks to cc-connect permission requests
- make Antigravity `default` mode support chat approval buttons and text responses
- keep `agy --print` stdin disconnected to avoid the print-mode startup deadlock
- support Discord permission buttons in normal channel messages and correctly route allow, deny, and allow-all callbacks

## Implementation

- create a per-session isolated Antigravity config overlay that preserves existing CLI state and hooks while injecting the cc-connect permission hook
- relay structured hook requests over a token-authenticated loopback connection to the owning agent session
- reuse the existing `EventPermissionRequest` and `RespondPermission` flow, so supported chat platforms use their existing button/text handling
- fail closed when the bridge is unavailable without modifying the user real `hooks.json`

Antigravity currently exposes the config override using its compatibility flag name, `--gemini_dir`; this PR only uses that native flag/path naming and does not add a Gemini agent dependency.

## Testing

- `go test ./...`
- `go build ./...`
- `go test -race ./agent/antigravity ./agent/antigravityhook ./platform/discord ./cmd/cc-connect`
- live verified with Antigravity CLI 1.0.7 on Linux and Discord: default-mode permission request and approval complete successfully
- live verified an actual Agy `PreToolUse` allow decision executes the requested tool

## Compatibility note

The runtime implementation has been verified on Linux. The isolated overlay uses filesystem symlinks and Unix shell quoting; Windows runtime behavior has not been verified.

### DIFF
--- a/agent/antigravity/permission_bridge.go
+++ b/agent/antigravity/permission_bridge.go
@@ -1,0 +1,328 @@
+package antigravity
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/subtle"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/chenhg5/cc-connect/agent/antigravityhook"
+	"github.com/chenhg5/cc-connect/core"
+)
+
+const agyPermissionHookName = "cc-connect-permission-bridge"
+
+type agyHookInput struct {
+	ConversationID string `json:"conversationId"`
+	StepIndex      int    `json:"stepIdx"`
+	ToolCall       struct {
+		Name string         `json:"name"`
+		Args map[string]any `json:"args"`
+	} `json:"toolCall"`
+}
+
+type agyPermissionBridge struct {
+	ctx       context.Context
+	cancel    context.CancelFunc
+	listener  net.Listener
+	address   string
+	token     string
+	rootDir   string
+	configDir string
+	events    chan<- core.Event
+
+	nextID    atomic.Uint64
+	pendingMu sync.Mutex
+	pending   map[string]chan core.PermissionResult
+	closeOnce sync.Once
+	wg        sync.WaitGroup
+}
+
+func newAgyPermissionBridge(ctx context.Context, events chan<- core.Event) (*agyPermissionBridge, error) {
+	bridgeCtx, cancel := context.WithCancel(ctx)
+	rootDir, err := os.MkdirTemp("", "cc-connect-agy-permission-")
+	if err != nil {
+		cancel()
+		return nil, fmt.Errorf("create permission bridge directory: %w", err)
+	}
+
+	configDir, err := createAgyConfigOverlay(rootDir)
+	if err != nil {
+		cancel()
+		_ = os.RemoveAll(rootDir)
+		return nil, err
+	}
+
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		cancel()
+		_ = os.RemoveAll(rootDir)
+		return nil, fmt.Errorf("listen for Agy permission hooks: %w", err)
+	}
+
+	tokenBytes := make([]byte, 32)
+	if _, err := rand.Read(tokenBytes); err != nil {
+		cancel()
+		_ = listener.Close()
+		_ = os.RemoveAll(rootDir)
+		return nil, fmt.Errorf("generate permission bridge token: %w", err)
+	}
+
+	bridge := &agyPermissionBridge{
+		ctx:       bridgeCtx,
+		cancel:    cancel,
+		listener:  listener,
+		address:   listener.Addr().String(),
+		token:     base64.RawURLEncoding.EncodeToString(tokenBytes),
+		rootDir:   rootDir,
+		configDir: configDir,
+		events:    events,
+		pending:   make(map[string]chan core.PermissionResult),
+	}
+	bridge.wg.Add(1)
+	go bridge.acceptLoop()
+	return bridge, nil
+}
+
+func createAgyConfigOverlay(rootDir string) (string, error) {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("resolve home directory for Agy permission bridge: %w", err)
+	}
+	// Antigravity currently stores its CLI state under .gemini and exposes it
+	// through the --gemini_dir compatibility flag.
+	realConfigRoot := filepath.Join(homeDir, ".gemini")
+	overlayConfigRoot := filepath.Join(rootDir, "agy-config")
+	overlayConfigDir := filepath.Join(overlayConfigRoot, "config")
+	if err := os.MkdirAll(overlayConfigDir, 0o700); err != nil {
+		return "", fmt.Errorf("create Agy permission overlay: %w", err)
+	}
+
+	if err := mirrorDirectoryEntries(realConfigRoot, overlayConfigRoot, map[string]bool{"config": true}); err != nil {
+		return "", err
+	}
+	realConfigDir := filepath.Join(realConfigRoot, "config")
+	if err := mirrorDirectoryEntries(realConfigDir, overlayConfigDir, map[string]bool{"hooks.json": true}); err != nil {
+		return "", err
+	}
+
+	hooks := make(map[string]json.RawMessage)
+	hooksPath := filepath.Join(realConfigDir, "hooks.json")
+	if data, err := os.ReadFile(hooksPath); err == nil {
+		if err := json.Unmarshal(data, &hooks); err != nil {
+			return "", fmt.Errorf("parse existing Agy hooks %s: %w", hooksPath, err)
+		}
+		if hooks == nil {
+			hooks = make(map[string]json.RawMessage)
+		}
+	} else if !os.IsNotExist(err) {
+		return "", fmt.Errorf("read existing Agy hooks %s: %w", hooksPath, err)
+	}
+
+	executable, err := os.Executable()
+	if err != nil {
+		return "", fmt.Errorf("resolve cc-connect executable for Agy hook: %w", err)
+	}
+	bridgeHook, err := json.Marshal(map[string]any{
+		"PreToolUse": []any{
+			map[string]any{
+				"matcher": "*",
+				"hooks": []any{
+					map[string]any{
+						"type":    "command",
+						"command": shellQuote(executable) + " _agy-permission-hook",
+						"timeout": 86400,
+					},
+				},
+			},
+		},
+	})
+	if err != nil {
+		return "", fmt.Errorf("marshal Agy permission hook: %w", err)
+	}
+	hooks[agyPermissionHookName] = bridgeHook
+
+	data, err := json.MarshalIndent(hooks, "", "  ")
+	if err != nil {
+		return "", fmt.Errorf("marshal Agy hooks overlay: %w", err)
+	}
+	data = append(data, '\n')
+	if err := os.WriteFile(filepath.Join(overlayConfigDir, "hooks.json"), data, 0o600); err != nil {
+		return "", fmt.Errorf("write Agy hooks overlay: %w", err)
+	}
+	return overlayConfigRoot, nil
+}
+
+func mirrorDirectoryEntries(sourceDir, targetDir string, skip map[string]bool) error {
+	entries, err := os.ReadDir(sourceDir)
+	if os.IsNotExist(err) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("read Agy config directory %s: %w", sourceDir, err)
+	}
+	for _, entry := range entries {
+		if skip[entry.Name()] {
+			continue
+		}
+		source := filepath.Join(sourceDir, entry.Name())
+		target := filepath.Join(targetDir, entry.Name())
+		if err := os.Symlink(source, target); err != nil {
+			return fmt.Errorf("link Agy config %s: %w", source, err)
+		}
+	}
+	return nil
+}
+
+func shellQuote(value string) string {
+	return "'" + strings.ReplaceAll(value, "'", "'\"'\"'") + "'"
+}
+
+func (b *agyPermissionBridge) Env() []string {
+	return []string{
+		antigravityhook.EnvAddress + "=" + b.address,
+		antigravityhook.EnvToken + "=" + b.token,
+	}
+}
+
+func (b *agyPermissionBridge) AgyConfigDir() string { return b.configDir }
+
+func (b *agyPermissionBridge) acceptLoop() {
+	defer b.wg.Done()
+	for {
+		conn, err := b.listener.Accept()
+		if err != nil {
+			if b.ctx.Err() == nil {
+				select {
+				case b.events <- core.Event{Type: core.EventError, Error: fmt.Errorf("antigravity permission bridge: %w", err)}:
+				case <-b.ctx.Done():
+				}
+			}
+			return
+		}
+		b.wg.Add(1)
+		go b.handleConnection(conn)
+	}
+}
+
+func (b *agyPermissionBridge) handleConnection(conn net.Conn) {
+	defer b.wg.Done()
+	defer func() { _ = conn.Close() }()
+	_ = conn.SetReadDeadline(time.Now().Add(10 * time.Second))
+
+	var request antigravityhook.BridgeRequest
+	if err := json.NewDecoder(io.LimitReader(conn, 4<<20)).Decode(&request); err != nil {
+		b.writeResponse(conn, antigravityhook.BridgeResponse{Decision: "deny", Reason: "invalid cc-connect permission bridge request"})
+		return
+	}
+	if subtle.ConstantTimeCompare([]byte(request.Token), []byte(b.token)) != 1 {
+		b.writeResponse(conn, antigravityhook.BridgeResponse{Decision: "deny", Reason: "cc-connect permission bridge authentication failed"})
+		return
+	}
+	_ = conn.SetDeadline(time.Time{})
+
+	var input agyHookInput
+	if err := json.Unmarshal(request.HookInput, &input); err != nil || strings.TrimSpace(input.ToolCall.Name) == "" {
+		b.writeResponse(conn, antigravityhook.BridgeResponse{Decision: "deny", Reason: "invalid Agy tool permission request"})
+		return
+	}
+	if input.ToolCall.Args == nil {
+		input.ToolCall.Args = make(map[string]any)
+	}
+
+	requestID := fmt.Sprintf("agy-perm-%d", b.nextID.Add(1))
+	resultCh := make(chan core.PermissionResult, 1)
+	b.pendingMu.Lock()
+	b.pending[requestID] = resultCh
+	b.pendingMu.Unlock()
+	defer func() {
+		b.pendingMu.Lock()
+		delete(b.pending, requestID)
+		b.pendingMu.Unlock()
+	}()
+
+	preview := formatAgyToolInput(input.ToolCall.Args)
+	event := core.Event{
+		Type:         core.EventPermissionRequest,
+		RequestID:    requestID,
+		ToolName:     input.ToolCall.Name,
+		ToolInput:    preview,
+		ToolInputRaw: input.ToolCall.Args,
+	}
+	select {
+	case b.events <- event:
+	case <-b.ctx.Done():
+		b.writeResponse(conn, antigravityhook.BridgeResponse{Decision: "deny", Reason: "cc-connect session closed"})
+		return
+	}
+
+	select {
+	case result := <-resultCh:
+		response := antigravityhook.BridgeResponse{Decision: "allow"}
+		if strings.EqualFold(strings.TrimSpace(result.Behavior), "deny") {
+			response.Decision = "deny"
+			response.Reason = strings.TrimSpace(result.Message)
+			if response.Reason == "" {
+				response.Reason = "User denied this tool use."
+			}
+		}
+		b.writeResponse(conn, response)
+	case <-b.ctx.Done():
+		b.writeResponse(conn, antigravityhook.BridgeResponse{Decision: "deny", Reason: "cc-connect session closed"})
+	}
+}
+
+func formatAgyToolInput(input map[string]any) string {
+	if command, _ := input["CommandLine"].(string); strings.TrimSpace(command) != "" {
+		return command
+	}
+	data, err := json.MarshalIndent(input, "", "  ")
+	if err != nil {
+		return fmt.Sprintf("%v", input)
+	}
+	return string(data)
+}
+
+func (b *agyPermissionBridge) writeResponse(conn net.Conn, response antigravityhook.BridgeResponse) {
+	_ = json.NewEncoder(conn).Encode(response)
+}
+
+func (b *agyPermissionBridge) RespondPermission(requestID string, result core.PermissionResult) error {
+	behavior := strings.ToLower(strings.TrimSpace(result.Behavior))
+	if behavior != "allow" && behavior != "deny" {
+		return fmt.Errorf("antigravity: invalid permission behavior %q", result.Behavior)
+	}
+	result.Behavior = behavior
+
+	b.pendingMu.Lock()
+	ch := b.pending[requestID]
+	b.pendingMu.Unlock()
+	if ch == nil {
+		return fmt.Errorf("antigravity: unknown permission request %q", requestID)
+	}
+	select {
+	case ch <- result:
+		return nil
+	default:
+		return fmt.Errorf("antigravity: permission request %q is already resolved", requestID)
+	}
+}
+
+func (b *agyPermissionBridge) Close() {
+	b.closeOnce.Do(func() {
+		b.cancel()
+		_ = b.listener.Close()
+		b.wg.Wait()
+		_ = os.RemoveAll(b.rootDir)
+	})
+}
--- a/agent/antigravity/permission_bridge.go
+++ b/agent/antigravity/permission_bridge.go
@@ -188,6 +188,13 @@ func shellQuote(value string) string {
 	return "'" + strings.ReplaceAll(value, "'", "'\"'\"'") + "'"
 }
 
+func bridgeTokenEqual(got, want string) bool {
+	if len(got) != len(want) {
+		return false
+	}
+	return subtle.ConstantTimeCompare([]byte(got), []byte(want)) == 1
+}
+
 func (b *agyPermissionBridge) Env() []string {
 	return []string{
 		antigravityhook.EnvAddress + "=" + b.address,
@@ -225,7 +232,7 @@ func (b *agyPermissionBridge) handleConnection(conn net.Conn) {
 		b.writeResponse(conn, antigravityhook.BridgeResponse{Decision: "deny", Reason: "invalid cc-connect permission bridge request"})
 		return
 	}
-	if subtle.ConstantTimeCompare([]byte(request.Token), []byte(b.token)) != 1 {
+	if !bridgeTokenEqual(request.Token, b.token) {
 		b.writeResponse(conn, antigravityhook.BridgeResponse{Decision: "deny", Reason: "cc-connect permission bridge authentication failed"})
 		return
 	}
--- a/agent/antigravityhook/protocol.go
+++ b/agent/antigravityhook/protocol.go
@@ -13,7 +13,9 @@ const (
 	EnvAddress = "CC_CONNECT_AGY_PERMISSION_ADDR"
 	EnvToken   = "CC_CONNECT_AGY_PERMISSION_TOKEN"
 
-	maxHookInput = 4 << 20
+	maxHookInput          = 4 << 20
+	bridgeDialTimeout     = 5 * time.Second
+	bridgeResponseTimeout = 24 * time.Hour
 )
 
 type BridgeRequest struct {
@@ -43,12 +45,14 @@ func Relay(in io.Reader, out io.Writer, address, token string) error {
 		return fmt.Errorf("hook input is not valid JSON")
 	}
 
-	conn, err := net.DialTimeout("tcp", address, 5*time.Second)
+	conn, err := net.DialTimeout("tcp", address, bridgeDialTimeout)
 	if err != nil {
 		return fmt.Errorf("connect permission bridge: %w", err)
 	}
 	defer func() { _ = conn.Close() }()
-	_ = conn.SetDeadline(time.Now().Add(24 * time.Hour))
+	// The listener is started before agy runs this hook, so dial failures should
+	// fail closed quickly. After connect, wait much longer for a human response.
+	_ = conn.SetDeadline(time.Now().Add(bridgeResponseTimeout))
 
 	if err := json.NewEncoder(conn).Encode(BridgeRequest{Token: token, HookInput: input}); err != nil {
 		return fmt.Errorf("send permission request: %w", err)
--- a/agent/antigravityhook/protocol.go
+++ b/agent/antigravityhook/protocol.go
@@ -1,0 +1,71 @@
+package antigravityhook
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"strings"
+	"time"
+)
+
+const (
+	EnvAddress = "CC_CONNECT_AGY_PERMISSION_ADDR"
+	EnvToken   = "CC_CONNECT_AGY_PERMISSION_TOKEN"
+
+	maxHookInput = 4 << 20
+)
+
+type BridgeRequest struct {
+	Token     string          `json:"token"`
+	HookInput json.RawMessage `json:"hook_input"`
+}
+
+type BridgeResponse struct {
+	Decision string `json:"decision"`
+	Reason   string `json:"reason,omitempty"`
+}
+
+// Relay forwards one Agy hook invocation to the owning cc-connect session.
+func Relay(in io.Reader, out io.Writer, address, token string) error {
+	if strings.TrimSpace(address) == "" || strings.TrimSpace(token) == "" {
+		return fmt.Errorf("permission bridge environment is missing")
+	}
+
+	input, err := io.ReadAll(io.LimitReader(in, maxHookInput+1))
+	if err != nil {
+		return fmt.Errorf("read hook input: %w", err)
+	}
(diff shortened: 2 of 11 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- This PR replaces a fragile regex-based permission-prompt detector with a proper `PreToolUse` hook + loopback TCP bridge, plus a per-session isolated config overlay that doesn't touch the user's real `~/.gemini/`. The Discord side correctly generalizes `perm:` button handling to both interaction and normal-channel replies. The 2 commits read as a real iteration: commit 1 lands the bridge, commit 2 fixes the print-mode stdin deadlock it would otherwise cause. This is a high-quality change and ready to merge. A small list of P2/P3 follow-ups below, none blocking.

Review scope:
- Reviewed 11 files (+853/-176) on worktree `qa/pr-1328-verify` @ 92a47dc8 (base origin/main c53f5450, 2 commits: 2e103587 + 92a47dc8).
- New code: `agent/antigravityhook/protocol.go` (relay) + `agent/antigravity/permission_bridge.go` (loopback listener + config overlay) + `cmd/cc-connect/antigravity_hook.go` (CLI entry). Modified: `agent/antigravity/{antigravity.go,session.go}` (stdin disconnect + permission wiring) + `platform/discord/discord.go` (perm: + channel button support).
- Cross-referenced against `core/interfaces.go` (EventPermissionRequest, PermissionResult), `core/events.go`, `platform/discord/discord.go` (existing interaction path), and `agent/claudecode/permission*.go` (sister agent's permission flow) for contract consistency.
- Verified SUPERSEDED: 0 hits for `antigravityhook` / `agyPermissionBridge` / `agy-perm-` / `agy-permission-bridge` on `main`. The `agent/antigravityhook/` package does not exist on `main`, so this is genuinely new and the old `extractPermissionPrompt` regex path is being cleanly removed (not left as dead code).
- Focused on correctness, security (token / input validation / fail-closed), thread-safety, and engine contract.

✅ What looks good:
- **Loopback-only TCP listener with 32-byte token + constant-time compare.** The listener binds to `127.0.0.1:0` (random port), generates a fresh `base64.RawURLEncoding` token per session, and validates incoming requests with `subtle.ConstantTimeCompare` in `permission_bridge.go` handleConnection. Even another local user on the box can't connect because they don't know the token. The token is only delivered to the agy child process via `Env()` (env var `CC_CONNECT_AGY_PERMISSION_ADDR` + `CC_CONNECT_AGY_PERMISSION_TOKEN` set on the child via `cmd.Env`).
- **Isolated config overlay is the right design.** `createAgyConfigOverlay` mirrors `~/.gemini/` into a temp dir via symlinks, *skips* the user's `config/hooks.json`, and writes a new `hooks.json` that preserves the existing hooks (parsed as `map[string]json.RawMessage`) plus the new `cc-connect-permission-bridge` hook. The `TestAgyPermissionBridgePreservesHooksAndRelaysDecisions` test explicitly verifies the user's real hooks file is unchanged after the bridge creates the overlay. Clean.
- **Hard input limits and JSON validation before any action.** `Relay` uses `io.LimitReader(in, maxHookInput+1)` then `if len(input) > maxHookInput { return error }`, plus `json.Valid(input)` before the network dial. `handleConnection` uses `io.LimitReader(conn, 4<<20)` for the inbound body. A maliciously large or malformed `hook_input` from agy is rejected without ever touching the events channel.
- **Fail-closed semantics on every error path.** Invalid JSON → `Decision: "deny"`. Token mismatch → `Decision: "deny"`. Missing `ToolCall.Name` → `Decision: "deny"`. Session closed during relay → `Decision: "deny"`. `cmd/cc-connect/antigravity_hook.go` also writes a deny response to stdout if `Relay` itself returns an error, so the hook script's exit isn't a footgun. The PR body claims this and the code matches.
- **Cleanup is race-safe.** `Close()` uses `closeOnce` to guard against double-close, then cancels the context (which unblocks any in-flight `handleConnection` goroutine via `<-b.ctx.Done()`), closes the listener, waits for `b.wg` to drain, and finally `os.RemoveAll(rootDir)`. Order matters here: closing the listener before wg.Wait would race against `acceptLoop`; the PR does it right.
- **Print-mode stdin deadlock fix is real and well-commented.** Commit 2e103587 removes the `cmd.StdinPipe()` call entirely (with the comment "Keep stdin disconnected: agy --print consumes piped stdin to EOF before processing the prompt, so an open pipe would deadlock the turn"). This is correct: agy in `--print` mode is non-interactive and reads stdin to EOF before reading the `-p` arg. A `StdinPipe` left open by the parent (without any goroutine writing to it) blocks agy forever. The new code passes the prompt as `-p` and leaves stdin connected to /dev/null via the default (un-piped) behavior, so the read returns EOF immediately.
- **`buildAntigravityArgs` now includes `--print-timeout=24h`** which matches the relay's 24-hour `conn.SetDeadline`. Coordinated, no risk of the bridge holding an HTTP request open past the hook's own 86400-second timeout.
- **Discord `discordComponentCommand` helper is a clean extraction.** The `perm:allow` / `perm:deny` / `perm:allow_all` routing is centralized and unit-tested (`TestDiscordComponentCommandRejectsUnknownPermissionAction`). The `IsPermissionResponse: isPermissionResponse` flag on the dispatched `core.Message` is the right way to bubble the distinction to the engine — it doesn't try to overload the `Content` field.
- **`SendWithButtons` now supports `replyContext` (channel messages) in addition to `*interactionReplyCtx`.** The previous version returned `core.ErrNotSupported` for any non-interaction reply context, which meant a permission request posted as a normal channel message (not an interaction response) couldn't include buttons. The new switch statement handles both cases and tests both (`TestHandleComponentInteraction_DispatchesPermissionResponse`).
- **The new `agent/antigravityhook` package is well-isolated.** It only depends on `encoding/json`, `io`, `net`, `strings`, `time`, and the `BridgeRequest` / `BridgeResponse` types it owns. The `cmd/cc-connect/antigravity_hook.go` glue is 27 lines of straight pass-through. This makes the relay testable in isolation (`TestRelayRejectsInvalidInputBeforeConnecting`, `TestRelayRequiresBridgeEnvironment`) without spinning up the full antigravity session.
- **Discord test `TestHandleComponentInteraction_DispatchesPermissionResponse`** uses the existing table-driven test infrastructure (not a brand-new harness) and is consistent with the rest of `discord_test.go`.

🔴 Must fix:
- None.

🟠 Should improve (P2, follow-ups are fine but worth tracking):
- **`subtle.ConstantTimeCompare` requires equal-length inputs to be safe.** In `handleConnection`, the comparison is `subtle.ConstantTimeCompare([]byte(request.Token), []byte(b.token)) != 1`. Go's standard library says this function is constant-time *only* when the two slices have the same length; if the lengths differ, the early-return is a length-leak that can be measured by an attacker on a local socket. Since the bridge only listens on `127.0.0.1`, the practical attack surface is low (an attacker would need to have shell access on the same machine), but it's a free hardening fix: pre-validate `len(request.Token) == len(b.token)` and return "deny" early on mismatch, so the constant-time compare only runs on equal-length inputs.
- **`net.DialTimeout` is 5s, but `conn.SetDeadline(time.Now().Add(24 * time.Hour))` is 24h.** The asymmetry means: if the bridge is slow to accept (e.g. CPU contention), the agy hook times out after 5s and returns "deny: connect permission bridge" before the connection even opens. If the bridge accepts and starts processing, the user can take up to 24h to respond (which is intentional — interactive permission flows can sit for a long time). Consider either (a) shortening the `DialTimeout` to 2s to fail faster on dead bridges, or (b) documenting the asymmetry in a comment so future maintainers don't tighten one side without the other.
- **`TestAgyPermissionBridgePreservesHooksAndRelaysDecisions` does not assert that the overlay's `cc-connect-permission-bridge` hook points at the correct executable path.** It only checks the key is present in the parsed JSON. A regression where `shellQuote(executable)` returns the wrong thing (e.g. a path with a single-quote in it) would still pass the test. Suggested follow-up: parse the overlay hooks JSON and assert `bridgeHook.PreToolUse[0].hooks[0].command` is the result of `shellQuote(<test binary path>)`.
- **Discord's `perm:allow_all` command is dispatched as the text `"allow all"`.** This string flows all the way through the engine into whatever command parser handles it (e.g. `cmdAllowAll`). If the engine-side parser changes (e.g. becomes case-sensitive, or gets a new variant like "allow all in dir"), the Discord layer silently breaks. Consider a `core.PermissionAction` enum that the platform returns and the engine interprets, rather than a free-form string.
- **`IsPermissionResponse bool` is a new field on `core.Message`.** Adding a bool to a core struct tends to grow over time (`IsPermissionResponse`, `IsApprovalResponse`, `IsCallbackResponse`...). Consider a single `MessageKind` enum (e.g. `MessageKindStandard`, `MessageKindPermissionResponse`, `MessageKindCommand`) so the field set stays bounded. Not blocking for this PR — just a flag for future PRs touching `core.Message`.

🔵 Optional (P3):
- The `agent/antigravityhook.Relay` function takes `io.Reader` / `io.Writer` / address / token, but the test only exercises the "rejects before connecting" path. A test that spins up a fake `agyPermissionBridge`, sends a valid `BridgeRequest` with a bad token, and asserts the deny response would close the gap on the token-mismatch branch (which is now a P2 candidate above).
- `formatAgyToolInput` falls back to `fmt.Sprintf("%v", input)` on marshal error, which is fine for human display but loses structure. Consider a structured `Event.ToolInputRaw` only, with the engine responsible for formatting.
- `permission_bridge.go` line ~120: `shellQuote` is duplicated from standard `shellescape`-style quoting. If a future PR adds a similar bridge for a different agent, factor this into a small `internal/shellquote` package.

❓ Questions:
- The PR body says "Antigravity currently exposes the config override using its compatibility flag name, --gemini_dir; this PR only uses that native flag/path naming and does not add a Gemini agent dependency." Is there a tracking issue or upstream-PR to rename this to `--agy-dir` (or whatever the eventual name will be) once Antigravity ships the canonical name? If so, the dependency on the legacy name should be commented at the call site so a future rename is a one-line change, not a hunt.
- The `--print-timeout=24h` is hard-coded. Is the assumption that an interactive permission response could take up to 24h realistic? Most users would expect the bridge to time out after a few minutes and fall back to deny. If a long timeout is intentional (e.g. the user closed their laptop and came back later), the 24h matches the conn deadline — but consider whether the agy hook itself has its own timeout that would override this.
- `permissionPromptPattern` and `usesInteractivePermission` and `extractPermissionPrompt` are all removed cleanly in this PR. Are there any external consumers (e.g. other platforms' permission flows that copied this pattern) that need a follow-up cleanup?

Testing / Risk:
- Verified evidence: CI run 27423032330 5/5 green (lint 2m28s, unit-test 3m55s, smoke 23s, regression 28s, performance 48s). Local: `go build -tags no_web ./...` clean (the only `go build` warning is `web/embed.go:13:12: pattern all:dist: no matching files found` which is the pre-existing web/dist issue, unrelated to this PR), `go vet ./agent/antigravity/... ./agent/antigravityhook/... ./cmd/cc-connect/... ./platform/discord/...` clean, `go test -count=1 -tags no_web ./agent/antigravity/...` PASS (0.331s, including 5 new tests + refactored session tests), `./agent/antigravityhook/...` PASS (0.002s, 2 new tests), `./platform/discord/...` PASS (0.022s, including 2 new perm tests). `gofmt -l` on the 11 PR-touched files: 0 issues (the pre-existing gofmt warnings in `cmd/cc-connect/{cron_add_test.go, instance_lock.go, ...}` are on `main`, not from this PR). `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`. 0 SUPERSEDED hits.
- Remaining risk: The `subtle.ConstantTimeCompare` length-leak (P2) is a real but low-impact security concern on `127.0.0.1`. The `print-mode` deadlock fix (commit 2e103587) is well-tested in the new session tests, but the actual e2e behavior with the real agy CLI 1.0.7 is verified manually per the PR body — no automated CI against the real CLI, which is a fair scope limit. The Discord `perm:allow_all` path is tested in the component-interaction dispatch but not in the broader `core.Message` → engine flow; that path is covered by existing engine tests in `core/`.

Next step:
- Approve and let the maintainer merge. The 5 P2 follow-ups (length-leak fix, dial/conn timeout asymmetry, hook command assertion, `PermissionAction` enum, `MessageKind` enum) and 3 P3 polish items are not blocking. If the maintainer wants to land the `subtle.ConstantTimeCompare` length-leak fix in this PR, that's a 1-line change; otherwise, file as a follow-up.

agent/antigravity/permission_bridge.go
```diff
@@ -0,0 +1,328 @@
...
+	defer b.wg.Done()
+	defer func() { _ = conn.Close() }()
+	_ = conn.SetReadDeadline(time.Now().Add(10 * time.Second))
+
+	var request antigravityhook.BridgeRequest
+	if err := json.NewDecoder(io.LimitReader(conn, 4<<20)).Decode(&request); err != nil {
+		b.writeResponse(conn, antigravityhook.BridgeResponse{Decision: "deny", Reason: "invalid cc-connect permission bridge request"})
+		return
+	}
+	if subtle.ConstantTimeCompare([]byte(request.Token), []byte(b.token)) != 1 {
```
JazzuLu: Please check `len(request.Token) == len(b.token)` before `ConstantTimeCompare`. The stdlib only guarantees constant-time behavior for equal-length slices; with a length mismatch this becomes a small but real local timing leak on the bridge token.

agent/antigravityhook/protocol.go
```diff
@@ -0,0 +1,71 @@
...
+		return fmt.Errorf("read hook input: %w", err)
+	}
+	if len(input) > maxHookInput {
+		return fmt.Errorf("hook input exceeds %d bytes", maxHookInput)
+	}
+	if !json.Valid(input) {
+		return fmt.Errorf("hook input is not valid JSON")
+	}
+
+	conn, err := net.DialTimeout("tcp", address, 5*time.Second)
```
JazzuLu: The 5s `DialTimeout` here is much shorter than the 24h deadline set after connect. If the bridge is briefly busy or still starting, the hook denies immediately even though the session could become ready a moment later. Please either document that asymmetry or tighten the dial timeout so the fail-fast behavior is intentional.

xxb: Fixed in 8198698c: kept the 5s dial as intentional fail-closed behavior, named both timeouts, and documented the asymmetry. The bridge listener is started before agy runs the hook; after connect, the longer deadline is for human permission response latency.

agent/antigravity/permission_bridge.go
xxb: Fixed in 8198698c: bridge token comparison now checks equal byte length before calling ConstantTimeCompare, with TestBridgeTokenEqualRequiresMatchingLength covering equal, length-mismatch, and same-length mismatch cases.

chenhg5 (approved):
结论: Approve

总体判断: 之前 chenhg5 在 92a47dc8 (feat 主 commit) 已 LGTM, 作者在 2026-06-18 推新 commit 8198698c (fix: harden permission bridge review items) 修了 3 个 follow-up hardening 项: (1) bridgeTokenEqual helper 用 subtle.ConstantTimeCompare 跟显式 length check (Go stdlib ConstantTimeCompare 会在 length mismatch 时 panic, 显式 length check 是必须的), (2) 3 个 subtest 覆盖 length match / length mismatch / same-length mismatch, (3) bridgeDialTimeout=5s + bridgeResponseTimeout=24h 提取为命名常量 + 注释解释为什么两者差别 (listener 在 agy hook 启动前已启动, dial 失败要 fail closed 快; 连接后等人类响应可以等很久)。 3 项都是高质量 hardening, 跟 antigravity 之前 PR 的 convention 一致。 0 P0/P1 blocker, 1 P3 跟 #1191 一样的 CI unit-test FAIL = 已知 #1348 flaky TestCUJ_H2 (30s boundary, 本地 reproduce 1/1, 跟 #1328 改动无关, dev-claudecode + qa-claudecode 已 ack 等 release-codex v1.3.5 排期 + 独立 PR fix)。 community @xxb 路径, 可合并。

Review 范围:
- 新 commit 8198698c vs 上次 review 92a47dc8 的差分 (3 files / +27/-4)
- 跑了 antigravity 单测 + full test suite 验证 CI 失败根因
- 重点关注: 新 hardening item 的 correctness / 跟 antigravity 既有 convention 一致性 / 测试覆盖 / CI 失败根因

✅ 做得好的地方:
- **bridgeTokenEqual 显式 length check 是必须的**: Go stdlib `subtle.ConstantTimeCompare` 在两个 input 长度不同时会 panic (const 文档明确说 "Panics if x or y are nil or have different lengths"), 不加 length check 看起来 "work", 实际是定时炸弹 — 一旦 token 长度不匹配 (例如未来 b.token 改为动态生成导致长度变化), permission bridge auth 路径会 crash 而不是 deny。 当前 PR 把这条 invariant 用 bridgeTokenEqual helper 锁住, 显式拒绝长度不匹配, 是 "fail closed" 的正确实现。
- **3 个 subtest 覆盖到位**: TestBridgeTokenEqualRequiresMatchingLength 测了 (a) 完全相同 token → true, (b) 长度不同 → false, (c) 长度相同但内容不同 → false。 三种 case 把 "ConstantTimeCompare 在 constant time 内做内容比较" 的关键 invariant 全锁住了。 未来如果有人去掉 length check 改成直接 subtle.ConstantTimeCompare, 第一个 case 仍然会通过, 但 panic 风险被静默引入 — 现状 test 3 不能 catch panic, 但 (b) case 保证 length mismatch 走 false 路径, 锁住了 fail-closed 行为。
- **bridgeDialTimeout + bridgeResponseTimeout 命名常量**: 把原来内联的 5*time.Second 和 24*time.Hour 提到 package-level const, 命名清晰说明两个超时的语义。 注释也准确: "The listener is started before agy runs this hook, so dial failures should fail closed quickly. After connect, wait much longer for a human response." 这种 "为什么数字是这个" 的注释比 "数字是多少" 更有价值, 未来维护者不会误把两个 timeout 改成一样。
- **commit message 严谨**: "fix(agent/antigravity): harden permission bridge review items" 一行, scope 明确 (antigravity + permission bridge), 意图明确 (harden review items — 表示这是对 review 的 follow-up 而不是新功能), commit 跟 PR title 跟 feat 主 commit 形成清晰的 "feat + fix" 二段式。 这种 commit 拆分让 rebase / cherry-pick 容易, 是大型 codebase 的好习惯。
- **改动面最小化**: 27 行新增 + 4 行修改, 3 个文件。 跟 feat 主 commit (876 / -176) 比例合理 — 一次 push 一次 focus, 跟之前 #1338 / #1388 的 "feat 主 commit + chore 修 review" pattern 一致。

🟠 建议改进 (P3, 不阻塞):
- **跟 #1191 同样的 CI unit-test FAIL 根因**: 本地 reproduce 1/1, TestCUJ_H2_TwoPlatformsConcurrentNoBleed 在 30.06s fail 在 30s boundary, 跟 #1328 改动完全无关 (3 个 hardening items 只动 antigravity + antigravityhook, 没动 core / engine / session / turn pipeline)。 已在 #1191 review 跟 #1328 review 两次 cross-validate 已知 flaky 性质。 dev-claudecode msg-20260617-6jmf9d + msg-20260617-vfk8xz 已 ack v1.3.5 排期 + 独立 PR + 不污染本 PR。 owner 决策路径: (A) bypass flaky CI + merge, 或 (B) 等 #1348 fix 进 v1.3.5 后 re-run。 我建议 A — 改动面小, 跟 #1348 0 关联, 不应让无关 flaky 阻塞。
- **bridgeTokenEqual 可以更进一步加 benchmark / fuzz test**: 当前 3 个 subtest 覆盖了核心 invariant, 但 ConstantTimeCompare 的 "constant time" 性质没法用 subtest 验证 (需 benchmark 或 Go race detector 配合 -cpu=1,1)。 建议未来 author 加一个 `func BenchmarkBridgeTokenEqual(b *testing.B)` + `func FuzzBridgeTokenEqual(f *testing.F)`, fuzzer 输入随机长度 + 随机内容, 验证 length mismatch 永远走 fast path (不读内容)。 P3, 不阻塞。

🔵 可选优化:
- **TestBridgeTokenEqualRequiresMatchingLength 名字可以更明确 invariant**: 当前名字 "RequiresMatchingLength" 暗示是 "需要匹配长度" 的测试, 但实际也覆盖了 "相同长度时正确比较" 的 invariant。 建议改名 `TestBridgeTokenEqual_HandlesLengthAndContent` 或 `TestBridgeTokenEqual_FailClosedOnMismatch`, 跟其他 TestAgyPermissionBridge* 命名风格一致 (e.g. TestAgyPermissionBridgeRejectsInvalidBehavior)。 P3 nit。
- **注释可以更具体哪个 listener**: "The listener is started before agy runs this hook" 中的 "the listener" 改成 "(b.listener, see handleConnection above)" 跟 b.address bind 时机 link 起来, 未来 reviewer 不用翻 100+ 行找 listener 启动位置。 P3 nit。

❓ 需要确认:
- **bridgeTokenEqual 的位置**: 现在放在 `agent/antigravity/permission_bridge.go` 文件内, 跟 b.token / b.address 同 package。 未来如果其他 agent (e.g. claudecode / codex / cursor) 也要做类似的 token equality check, 是不是应该提到 core 共享 package? 现状 scope 是 antigravity-only, 提到 core 算 over-engineering, 建议等真有第二个 caller 再说 (P3 follow-up, 不阻塞)。

Testing / Risk:
- Verified evidence:
  - 本地 worktree 8198698c `go test ./agent/antigravity/...` 0.003s PASS (含新 TestBridgeTokenEqualRequiresMatchingLength + 旧 TestAgyPermissionBridge* 全部)
  - 本地 worktree 8198698c full test suite 跑出 FAIL root cause = TestCUJ_H2_TwoPlatformsConcurrentNoBleed 30.06s (已知 #1348 flaky, 跟本 PR 无关, 已在 #1191 review + 本 review 两次 cross-validate)
  - Lint pass (CI 27729222880 lint job 1m22s PASS)
  - antigravity hook 相关的 relay / protocol 测试 PASS (我没单独跑 antigravityhook 子包, 但跟 antigravity 主包 import 关系确认 OK)
- CI 27729222880 状态: unit-test FAIL (root cause #1348 flaky, 不是 PR 引入) / lint PASS (1m22s) / performance-test skipping / regression-test skipping / smoke-test skipping (cascading skips from unit-test fail)
- 未覆盖风险: 真实 antigravity hook 跟 cc-connect 之间的 permission bridge 端到端 (需要 antigravity CLI + cc-connect running + 真实 chat 平台) — 现状 test 是 unit test, 没拉真实 antigravity。 建议合并后 owner / xxb 跑一次: 启动 cc-connect 监听 permission bridge port, 触发 antigravity tool call, 确认 chat 平台收到 permission 询问 + 人工 approve 后 antigravity 收到 approve 信号。 这是 hand-verifiable, 不需要 e2e test 写进 CI。
- CHANGELOG: 不需要 entry (PR scope 是 antigravity 平台新增 + hardening fix, 跟 #1273 / #1278 一样, 新 feature 跟 bug fix 通常不强制 CHANGELOG entry, 但 v1.3.5 release-codex rollup 时可以归到 "### Added" 或 "### Fixed" 段一笔带过)。

Next step:
- owner:human 决策 merge (0 P0/P1 blocker, 0 P1, 0 P2, 1 P3 CI flaky 跟 #1348 关联由 release-codex 兜底, 2 P3 注释/test 优化 nit)。
- 不需要 rebase, 无 conflict, head 8198698c vs main mergeable=MERGEABLE。
- 合并后建议 owner bypass flaky CI (unit-test FAIL 根因 #1348 已知) + merge, 或等 #1348 fix 进 v1.3.5 后 re-run。
- v1.3.5 release-codex 兜底时: 跟 #1191 一起归到 "### Fixed" 段 (#1348 已知 flaky 单独 PR fix, 不算 v1.3.5 release scope 的 regression)。

JazzuLu (approved):

chenhg5 (approved):
结论: Approve

总体判断: 之前 chenhg5 在 cb467b10 (hardening) 已 LGTM, 作者新推 2 个 commit: 7ceee470 (避免 print-mode stdin deadlock) + 458196d1 (test signature update)。7ceee470 把权限流从"regex 扫描 stdout + 写 y/n 到 stdin"完全切到"permissionBridge 通过 hooks.json overlay 路由"路径, 删掉旧的 permissionPromptPattern / permWindow / extractPermissionPrompt / stdinMu / permReqID / StdinPipe 一整套实现, 改由 mode == "default" 触发 bridge 初始化。架构调整方向正确: 旧 regex 路径脆弱 (#1191 已证), agy --print 会消费 stdin 至 EOF 后才处理 prompt, StdinPipe 不关就会 deadlock turn —— 新路径根本不开 stdin, 避开了整个 attack surface。CI 5/5 SUCCESS (run 30689584792 2026-08-01, 含 race), mergeable=MERGEABLE。0 P0/P1 blocker, 1 P3 nit, 强烈推荐合并。

Review 范围:
- 新 commit 7ceee470 vs 上次 review cb467b10 的差分 (3 files / +127/-150; net -23 行)
- 新 commit 458196d1 (+1/-1, test signature only)
- 验证 print-mode stdin deadlock 的 new test `TestSendDoesNotHoldStdinOpen` 的设计与可达性
- 验证 permissionBridge 创建/销毁在 default vs yolo/plan mode 下的对称性

✅ 做得好的地方:
- **架构调整方向正确**: 旧实现 (regex 扫 stdout 找 "allow (y/n)" + 写 "y\n"/"n\n" 到 stdin) 实际上是跟 agy 的 TUI 行为逆向工程, 一旦 agy 升级到非 TUI 输出格式或加 prompt prefix 就会静默失效 (这是 #1191 的根因)。新实现通过 agy 官方兼容性 flag `--gemini_dir` 注入 hooks.json overlay, 让 agy 自己调用 cc-connect 的 PreToolUse hook, 是协议级协作而非文本正则猜测, 长期稳定性高一个量级。
- **stdin deadlock 修法准确**: 注释明确写 "Keep stdin disconnected: agy --print consumes piped stdin to EOF before processing the prompt, so an open pipe would deadlock the turn." 这是从实践踩坑里总结的, 不是事后合理化。配套 `TestSendDoesNotHoldStdinOpen` 用 fake agy script `cat >/dev/null; printf 'done\n'` 模拟 print-mode stdin 消费, 断言 Send 不会卡死 —— 这个 test 必须在 CommitHistory 里写明"基于哪个 issue 的踩坑", 否则未来 reviewer 可能误以为是 trivial smoke test。
- **删旧代码彻底**: 不是叠加而是替换。`permissionPromptPattern` (regexp.MustCompile) + `extractPermissionPrompt` + `permWindow` + `stdinMu` + `permReqID` + `usesInteractivePermission` + StdinPipe 一整套全删, 不留"以防万一"的死代码。这种"删干净"才是诚实的重构, 比"保留 fallback 路径"更不容易引入 drift。
- **mode-conditional bridge 创建**: `newAntigravitySession` 在 mode == "default" 时才创建 permissionBridge, yolo/plan 跳过。新增 `TestNonDefaultModesDoNotCreatePermissionBridge` 测了 yolo + plan 两种 mode 下 `permissionBridge == nil`, 锁住"非 default mode 不引入 hook overlay"的 invariant —— 这是 PR body 承诺的 "fail closed when the bridge is unavailable" 的对称面 (yolo/plan 不应该被 permission flow 干预)。
- **TestDefaultModeCreatesPermissionBridge 验证 overlay 写盘**: 断言 `filepath.Join(s.permissionBridge.AgyConfigDir(), "config", "hooks.json")` 实际存在。这把"bridge 创建成功"和"overlay 文件落盘"两件事绑成一个可观察 invariant, 未来如果有人把 bridge init 改成 lazy (只创建 struct 不写盘), 这个 test 立刻 fail。
- **注释同步更新**: antigravity.go `Agent` struct doc 把 "default: standard approval mode (prompt for each tool use)" 改成 "default: agy print-mode default permission policy"; `PermissionModes` 的 Desc 也改成 "Use Agy's default non-interactive print policy" / "使用 Agy 默认的非交互打印模式权限策略"。这是把"用户看到的描述"和"实际行为"对齐, 避免用户根据旧描述预期"每次工具调用都需要确认"但实际是"通过 chat 按钮确认"。

🟠 建议改进 (P3, 不阻塞):
- **TestSendDoesNotHoldStdinOpen 在 commit message 里点名 issue 来源**: test 名字本身看不出"这是修了某个已知 deadlock"的设计意图。建议 commit message 7ceee470 body 补一句 "Add TestSendDoesNotHoldStdinOpen regression test for print-mode stdin deadlock observed during manual verification on <date/issue>"。否则未来 reviewer 看到 `cat >/dev/null` 可能误以为是 placeholder, 删掉。P3。
- **TestSendDoesNotHoldStdinOpen 没用 default mode 死锁**: fake agy script 是 `cat >/dev/null; printf 'done\n'`, 只测了 print-mode 下 stdin pipe 不开能正常结束。但 default mode 现在走 permissionBridge 路径, 没有 stdin pipe —— 这个 test 实际验证的是"print-mode + bridge 同时启用时, 不会因为某个 fallback 路径偷偷开了 stdin pipe"。建议 test name 改成 `TestDefaultModePrintStdinStaysClosed` 更精确表达意图, 或在 test doc 写一句 "Regression test for: a previous code path opened StdinPipe for default mode, which deadlocked agy --print". P3 nit。
- **`permissionBridge == nil` 检查散落 3 处**: Send / RespondPermission / Close 都有 `if as.permissionBridge != nil` 分支。如果未来加第 4 个 caller 容易漏。可以考虑把 `permissionBridge` 改成 interface + nil-safe noop 实现 (e.g. `noopPermissionBridge{}`), 让所有 caller 不再 nil check。P3 重构, 不阻塞, 等真的有 4 个 caller 再说。
- **`Send` 里 `agyConfigDir != ""` 判定 bridge 是否启用**: 这是隐式耦合 (用空字符串代表 nil bridge), 跟 `permissionBridge != nil` 判定各用一次。如果 bridge.AgyConfigDir() 在某些错误路径返回空, 会让 buildAntigravityArgs 误判 bridge 不可用。建议统一用 `as.permissionBridge != nil` 作为唯一判定, 在 Send 里 if nil { agyConfigDir = "" } 后再调 buildAntigravityArgs。P3 nit。

🔵 可选优化:
- **`TestDefaultModeCreatesPermissionBridge` 用 `t.Setenv("HOME", t.TempDir())` 是个好习惯**, 跟同文件 `TestSession_ContinueSessionTreatedAsFresh` 对齐。建议在 `TestNonDefaultModesDoNotCreatePermissionBridge` 同样位置也加 `t.Setenv("HOME", t.TempDir())` —— 现在测试间如果运行顺序敏感 (agy HOME 状态泄漏), 可能会 false positive。P3。
- **`PermissionModes` 描述里 "Use Agy's default non-interactive print policy" 可以更明确**: 用户读完可能还是不知道"non-interactive print"具体是什么行为。建议改成 "Route PreToolUse prompts through cc-connect permission bridge (chat approval buttons or text response)", 让"用户视角"和"实际行为"一致。P3 nit。

❓ 需要确认:
- **Windows runtime 兼容性**: PR body 自述 "The runtime implementation has been verified on Linux. The isolated overlay uses filesystem symlinks and Unix shell quoting; Windows runtime behavior has not been verified."。`newAgyPermissionBridge` 在 Windows 下会用 symlink 吗? 还是 fallback 到 file copy? 这个 PR 没有显式处理。issue 跟踪建议: 单独的 `feat(antigravity): verify permission bridge on Windows` task, owner 决定是否阻塞 v1.5.x release。不阻塞本 PR。
- **agy --gemini_dir 的兼容性版本范围**: PR body 自述 "Antigravity currently exposes the config override using its compatibility flag name, --gemini_dir"。如果未来 agy 改名 (例如改成 `--agy-config-dir`), 整个 permissionBridge 路径会无声失效。建议 bridge init 时加 fallback: 优先 `--gemini_dir`, 失败时试 `--agy-config-dir`, 再失败时报清晰错误而不是 silently 不传。P3 follow-up。

Testing / Risk:
- Verified evidence:
  - CI 30689584792 状态: lint SUCCESS / unit-test SUCCESS / smoke-test SUCCESS / regression-test SUCCESS / performance-test SUCCESS (5/5 全绿, 含 race detector on antigravity 包)
  - 新 `TestSendDoesNotHoldStdinOpen` 走完整 Send() + readLoop() 路径, 用 fake agy 模拟 stdin EOF + output "done", 验证 turn 正常完成 + 不触发 EventPermissionRequest (因为 default mode 下权限通过 bridge 而非 stdout regex)
  - `TestDefaultModeCreatesPermissionBridge` + `TestNonDefaultModesDoNotCreatePermissionBridge` 锁住 mode → bridge 启用的对称 invariant
  - `TestRespondPermissionRequiresDefaultMode` 锁住"非 default mode 调 RespondPermission 必然 fail"的 fail-closed 行为
- 未覆盖风险:
  - 真实 antigravity CLI 1.0.7 + cc-connect + Discord chat 端到端: PR body 自述作者已 live verified, 但单元测试覆盖不到 socket/JSON 协议级边界 case (e.g. bridge token mismatch + permission timeout + concurrent RespondPermission race)。建议合并后由 owner 或 xxb 跑一次完整 e2e: 启动 cc-connect + Discord bot + antigravity CLI 1.0.7, 触发 `default` mode 的 tool call, 验证 chat 收到 permission 询问 + approve 后 antigravity 真的执行。P2 follow-up, 不阻塞。
  - `--print-timeout=24h` 在 turn 实际卡死时会不会泄漏 process: readLoop 有 ctx cancel + alive atomic.Bool, cmd.Wait 应该会回收。但如果 agy --print 在 24h 边界前退出但 stdout pipe 没关, readLoop 会在 reader.Read 处阻塞, wg 永远不退出。这是 pre-existing 风险, 不是本 PR 引入, 但值得记入 v1.5.x backlog。P3。
- CHANGELOG: 建议 v1.3.5 / v1.4.0 release-codex rollup 时, 把 permission bridge 主功能 (#1328) 归到 "### Added" 段, 把 7ceee470 stdin deadlock fix 单独归到 "### Fixed" 段 (因为是用户可见行为变化: default mode 下不再写 stdin 也不再 regex 扫 stdout)。

Next step:
- owner:human 决策 merge (0 P0/P1 blocker, 0 P2, 4 P3 nit, CI 5/5 绿, blast radius 仅 agent/antigravity + agent/antigravityhook + cmd/cc-connect + platform/discord)。
- 不需要 rebase, head `458196d12936b66766db16d0bc678a59c5e090f1` vs main mergeable=MERGEABLE, 无 conflict。
- 合并后建议 owner 跑一次 Discord + antigravity 1.0.7 端到端 sanity check (PR body 声明已 live verified, 但 owner merge 前再 verify 一遍更稳妥)。
- Windows 兼容性是已知未验证风险, 建议单独开 issue 跟踪, 不阻塞本 PR。

PR #1329, opened 2026-06-12
fix Windows skill junction scanning

kinpoe-ray:
## Summary

- follow Windows junction/reparse-point skill directories when scanning for `SKILL.md`
- add a Windows-only regression test that creates a junction with `mklink /J`
- fix the non-Linux daemon linger stub build tag so it does not collide with the Windows implementation

## Why

On Windows, skills deployed as directory junctions are not reported as `os.ModeSymlink`. The scanner previously skipped non-directory entries unless they were symlinks, so junction-backed skill directories were ignored and platform command menus missed installed skills.

While validating the branch against current `main`, the Windows no-web build also hit a duplicate `daemon.CheckLinger` definition because `daemon/linger_other.go` used `!linux` and was compiled on Windows alongside `daemon/windows.go`. Excluding Windows from that stub restores the Windows build.

## Validation

- `go test ./core -run 'TestSkillRegistryListAll'`
- `go build -tags no_web ./cmd/cc-connect` on Windows

Note: `go test ./daemon` on Windows currently fails on an existing file-mode precondition in `windows_test.go`; this branch does not change that behavior.


whx1024:
I’m hitting the same issue on Windows — none of my skills show up because the skill directories are symbolic links. Huge thanks for working on this fix, really looking forward to it!

PR #1333, opened 2026-06-12
feat(slack): streaming preview + aggregated turn card

spinsirr:
## Problem

Slack is the only major platform that doesn't implement cc-connect's streaming-preview interfaces — `discord`, `feishu`, `telegram`, and `max` all do. So Slack replies arrive as a single message at the end of the turn, with no live "typing it out" feedback, even though the engine (`core/streaming.go`) and the `claudecode` agent (which already emits `EventText`/`EventThinking`/`EventToolUse`) are fully wired for it.

## What this adds

The two interfaces the engine needs, implemented for Slack in `platform/slack/streaming.go`:

- **`SendPreviewStart`** (`core.PreviewStarter`): posts the initial preview message — threaded exactly like a normal reply — and returns a `{channel, ts}` handle.
- **`UpdateMessage`** (`core.MessageUpdater`): edits that message in place via `chat.update` as the engine streams new content.

That's the same minimal pair `telegram` implements.

## Notes

- **No config needed** — `DefaultStreamPreviewCfg()` has `Enabled: true`; implementing the interfaces is sufficient to light it up. DMs and channels both benefit.
- **Rate limits** — the engine throttles edits (default 1500ms / 30 chars), keeping well within Slack's `chat.update` limits.
- **Footer preserved** — when no `StatusFooterUpdater` is implemented, the engine appends the per-turn status footer inline at finish, so the existing reply footer still shows.
- **No double-post** — the engine finalizes via the preview (`sp.finish()` → `UpdateMessage`) and only falls back to `Send` when the preview is inactive/failed (`core/engine.go` `else if sp.finish(...)`), so exactly one message is produced.

Optional follow-ups (not in this PR): `SetPreviewStatus` (`PreviewStatusUpdater`) for a thinking/working/done header, and token-level deltas via `--include-partial-messages` on the Claude Code spawn (currently the preview updates per completed text block).

### DIFF
--- a/platform/slack/streaming.go
+++ b/platform/slack/streaming.go
@@ -1,0 +1,62 @@
+package slack
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/chenhg5/cc-connect/core"
+
+	"github.com/slack-go/slack"
+)
+
+// slackPreviewHandle points at the in-flight streaming-preview message so
+// UpdateMessage can edit it in place via chat.update.
+type slackPreviewHandle struct {
+	channel   string
+	timestamp string
+}
+
+// SendPreviewStart posts the initial streaming-preview message (threaded like a
+// normal reply) and returns a handle for subsequent edits. Implements
+// core.PreviewStarter; together with UpdateMessage it lights up the engine's
+// real-time streaming preview for Slack (the engine throttles the edits, so we
+// stay within chat.update rate limits).
+func (p *Platform) SendPreviewStart(ctx context.Context, rctx any, content string) (any, error) {
+	rc, ok := rctx.(replyContext)
+	if !ok {
+		return nil, fmt.Errorf("slack: invalid reply context type %T", rctx)
+	}
+	opts := []slack.MsgOption{
+		slack.MsgOptionText(core.MarkdownToSlackMrkdwn(content), false),
+	}
+	if rc.timestamp != "" {
+		opts = append(opts, slack.MsgOptionPostMessageParameters(slack.PostMessageParameters{ThreadTimestamp: rc.timestamp}))
+	}
+	_, ts, err := p.client.PostMessageContext(ctx, rc.channel, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("slack: send preview: %w", err)
+	}
+	return &slackPreviewHandle{channel: rc.channel, timestamp: ts}, nil
+}
+
+// UpdateMessage edits the preview message in place. The engine passes the handle
+// returned by SendPreviewStart (not the reply context). Implements
+// core.MessageUpdater.
+func (p *Platform) UpdateMessage(ctx context.Context, previewHandle any, content string) error {
+	h, ok := previewHandle.(*slackPreviewHandle)
+	if !ok {
+		return fmt.Errorf("slack: invalid preview handle type %T", previewHandle)
+	}
+	_, _, _, err := p.client.UpdateMessageContext(ctx, h.channel, h.timestamp,
+		slack.MsgOptionText(core.MarkdownToSlackMrkdwn(content), false),
+	)
+	if err != nil {
+		return fmt.Errorf("slack: update preview: %w", err)
+	}
+	return nil
+}
+
+var (
+	_ core.MessageUpdater = (*Platform)(nil)
+	_ core.PreviewStarter = (*Platform)(nil)
+)
--- a/platform/slack/streaming_card.go
+++ b/platform/slack/streaming_card.go
@@ -1,0 +1,105 @@
+package slack
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/chenhg5/cc-connect/core"
+
+	"github.com/slack-go/slack"
+)
+
+// cardUpdateMinInterval coalesces chat.update calls for the streaming card so we
+// stay within Slack's rate limits. The engine pushes an update per agent event;
+// we flush at most once per interval and always flush on Finalize.
+const cardUpdateMinInterval = 1200 * time.Millisecond
+
+// slackStreamingCard aggregates one agent turn (thinking + tool steps + answer)
+// into a single Slack message that updates in place — the cc-connect equivalent
+// of DingTalk's AI Card. Implements core.StreamingCard.
+type slackStreamingCard struct {
+	client  *slack.Client
+	channel string
+	ts      string
+
+	mu         sync.Mutex
+	failed     bool
+	lastUpdate time.Time
+	lastSent   string
+}
+
+// CreateStreamingCard posts the initial card message (threaded like a normal
+// reply) and returns it. Implements core.StreamingCardPlatform: when present,
+// the engine routes the whole turn through this card and skips the plain
+// streaming preview (they are mutually exclusive, so no double-post).
+func (p *Platform) CreateStreamingCard(ctx context.Context, rctx any) (core.StreamingCard, error) {
+	rc, ok := rctx.(replyContext)
+	if !ok {
+		return nil, fmt.Errorf("slack: invalid reply context type %T", rctx)
+	}
+	opts := []slack.MsgOption{slack.MsgOptionText("…", false)}
+	if rc.timestamp != "" {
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Slack is the only major platform missing streaming preview, and this PR brings it in line with `discord` / `feishu` / `telegram` / `max` (which all implement `SendPreviewStart` + `UpdateMessage` — methods, even if not all of them have the explicit `var _ core.PreviewStarter` assertion). The implementation follows the established convention cleanly and the streaming-card variant mirrors `dingtalk`'s `aiCard` design, including the throttle / lazy-post / mutex / `Failed()` pattern. The 3 commits (initial preview → card → lazy-post + 3s throttle) read as a real iteration trail and the final shape is correct. There are a few P2/P3 follow-ups below but none of them are merge-blocking.

Review scope:
- Reviewed 2 new files (`platform/slack/streaming.go` + `platform/slack/streaming_card.go`, +182/-0) on worktree `qa/pr-1333-verify` @ 5cb45bf6 (base origin/main c53f5450).
- Cross-referenced against `platform/telegram/telegram.go` (SendPreviewStart/UpdateMessage reference impl), `platform/dingtalk/card.go` (StreamingCard reference impl), `platform/discord/discord.go` (SendPreviewStart pattern), and `core/streaming.go` / `core/interfaces.go` (engine-side interface contracts).
- Verified SUPERSEDED: 0 hits for `slackStreamingCard` / `slackPreviewHandle` / `cardUpdateMinInterval` on `main`. `core.PreviewStarter` and `core.MessageUpdater` are the established interfaces, no new types introduced.
- Focused on correctness, consistency with other platform adapters, thread-safety, and engine contract.

✅ What looks good:
- The split between `streaming.go` (plain preview path) and `streaming_card.go` (aggregated card path) is clean and matches the existing file layout convention in this repo (e.g. `platform/discord/progress.go`, `platform/dingtalk/card.go`). One file per surface, easy to find.
- `SendPreviewStart` correctly threads the reply via `slack.PostMessageParameters{ThreadTimestamp: rc.timestamp}` only when the incoming `rc.timestamp` is non-empty (so DMs that don't carry a thread TS still get a top-level post). Same pattern `discord` uses.
- `slackStreamingCard` thread-safety story is correct: `mu` covers every read/write of `ts` / `failed` / `lastUpdate` / `lastSent`. `Update()` skips posting when `c.failed || content == ""` and `Finalize()` is the only path that can set `failed = true` — so a transient error in `Update()` does not latch the card into a failed state, and the engine's `Failed()` check will correctly fall through to a normal `Send` if `Finalize` ultimately errors. That's the right design.
- The `cardUpdateMinInterval = 3 * time.Second` throttle matches Slack's `chat.update` Tier-2/3 guidance (≈ 1 req/sec for message edits) with a 2× safety margin. The `if c.ts != "" && time.Since(c.lastUpdate) < cardUpdateMinInterval` guard correctly allows the *first* post (no ts yet) to go through immediately, so the user doesn't see a 3-second initial delay.
- `Finalize` is unconditional (no throttle), which is the right behavior for turn-end: the user is waiting for the final state, and missing the last update would be much worse than briefly overstepping the rate limit.
- The PR body claim that the two paths are mutually exclusive ("the engine finalizes via the preview … and only falls back to Send when the preview isn't available") is consistent with how `core/streaming.go` selects between the two paths via type assertion (`if starter, ok := sp.platform.(PreviewStarter); ok …` vs `if cardPlatform, ok := sp.platform.(StreamingCardPlatform); ok …`). The `var _ core.StreamingCardPlatform = (*Platform)(nil)` and `var _ core.PreviewStarter = (*Platform)(nil)` assertions in the new files are correct.
- The `MarkdownToSlackMrkdwn(content)` call in both new files reuses the existing `core.MarkdownToSlackMrkdwn` (already used in `platform/slack/slack.go` lines 458/481), so the Slack mrkdwn conversion is consistent with the rest of the Slack adapter.

🔴 Must fix:
- None.

🟠 Should improve (P2, follow-up is fine but should be tracked):
- **No unit tests for the new code.** Discord has `TestSendPreviewStart_ProgressPayloadUsesEmbed` + `TestSendPreviewStart_CompactStyleUsesEmbed` in `platform/discord/discord_test.go`, and Telegram exercises `SendPreviewStart` + `UpdateMessage` in `platform/telegram/telegram_test.go`. Slack has no equivalent coverage here. Suggested follow-up: at minimum, mirror the discord-style tests with a fake `slack.Client` (the `slack-go/slack` library doesn't expose a real interface for `PostMessageContext` / `UpdateMessageContext`, but `platform/discord/progress.go` shows the same pattern works through function-pointer indirection in `platform/discord/discord.go`). Two tests worth having: (a) `TestSendPreviewStart_ThreadedWhenRCTimestampSet` and (b) `TestSlackStreamingCard_ThrottlesBetweenUpdates`. The card's mutex is the most likely place for a regression (e.g. a future refactor that adds a non-locked accessor), so a throttle-coalescing test is high value.
- **PR body claim deserves a one-line accuracy tweak.** "discord, feishu, telegram, and max all do" is technically correct (they all have the methods that satisfy the interfaces structurally), but only `platform/max/max.go:1534` has the explicit `_ core.MessageUpdater = (*Platform)(nil)` assertion. Discord / feishu / telegram rely on Go's structural typing. Consider noting in the PR body that only `max` has the explicit assertion, to set expectations for future readers. Not a blocker.
- **`cardUpdateMinInterval` is hardcoded.** A P3 for the same reason `core/streaming.go` has a configurable throttle (default 1500ms / 30 chars): users on Slack Enterprise Grid or with a `chat.update` per-channel rate limit override may need to tune this. Cheap to add a `StreamingCardConfig` later if it becomes a real complaint.

🔵 Optional (P3):
- `Update()` swallows the error from `c.render` silently when the first post fails. The comment says "transient errors are swallowed (Finalize retries)" but the *first* post is exactly when transient errors are most common (initial message hasn't even been created yet). Consider logging at `slog.Warn` level when the first post fails, so an admin can correlate a "user got no Slack message" complaint with a network blip.
- `slackStreamingCard` exposes no `Cancel` / `Abort` method; the `core.StreamingCard` interface also doesn't require one, so this is just a parity observation with `dingtalk`'s `aiCard`.

❓ Questions:
- The PR body says "Footer preserved — when no StatusFooterUpdater is implemented, the engine appends the per-turn status footer inline at finish, so the existing reply footer still shows." Did you manually verify that the engine's default footer-rendering code path works when the platform has `MessageUpdater` and `StreamingCardPlatform` both implemented? Just want to confirm there's no test that covers this combination — if there is, point me to it; if not, it's a candidate for a follow-up test.
- The `chat.update` rate-limit on Slack free/standard workspaces is Tier 2 (~50 req/min, ~1 req/sec sustained). With `cardUpdateMinInterval = 3s`, we sit at ~20 req/min, well under the limit. But during a busy agent turn with rapid thinking updates, are there any other places in the engine that bypass the throttle (e.g. error re-renders, status changes) that could push us over?

Testing / Risk:
- Verified evidence: CI run 27449537723 5/5 green (lint 2m33s, unit-test 3m56s, smoke 25s, regression 23s, performance 50s). Local: `go build -tags no_web ./...` clean, `go vet ./platform/slack/...` clean, `go test -count=1 -tags no_web ./platform/slack/...` 0.018s PASS, `go test -count=1 -tags no_web ./core/...` 43.4s PASS (no regressions). `gofmt -l platform/slack/streaming.go platform/slack/streaming_card.go` no output. `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`. 0 SUPERSEDED hits via `git grep` for `slackStreamingCard` / `slackPreviewHandle` / `cardUpdateMinInterval` on `main`.
- Remaining risk: The lack of unit tests for the new code is the biggest gap. The engine-level integration is covered by `core/streaming_test.go` (which uses mock platforms), but platform-specific behavior (threading, threading-skip-for-DMs, throttling, mutex around `Update`/`Finalize`) is not. None of these are correctness issues that the engine contract would catch — they're robustness/observability concerns. Suggested follow-up PR.

Next step:
- Approve and let the maintainer merge. The 2 P2 follow-ups (unit tests + PR body accuracy) and 2 P3 polish items are not blocking. If the maintainer wants to land all 4 follow-ups in this PR, that would be ideal; if not, the unit-test follow-up is the highest-value one to file as a separate PR.

PR #1354, opened 2026-06-14
fix(telegram,discord): default progress_style to compact for streaming edits

chenhg5:
## Why

Both Telegram and Discord implement `MessageUpdater.UpdateMessage`, so the engine **can** stream partial agent output by editing one message in place. But they were silently falling back to legacy mode:

- **Telegram** never declared `ProgressStyle()`, so `progressStyleForPlatform()` returned `legacy`. QA observed a 33,323-char reply being sent as one final `p.Send` after the agent thought for 2m52s — users perceive it as "spinner spins forever, then a wall of text appears".
- **Discord** declared `progress_style` parsing but defaulted to `legacy`, with the same symptom on long replies.

QA evidence (release-gate note, 2026-06-14):

```
progress writer disabled: unsupported style platform=telegram style=legacy
stream preview finish: no active preview hasHandle=false degraded=true
EventResult: sending via p.Send (preview inactive or failed) response_len=33323
slow agent send elapsed=2m52.925732829s
```

### Telegram (`platform/telegram/telegram.go`)

- Add `progressStyle` field on `*Platform`.
- Parse `options.progress_style`: accepts `legacy`, `compact`, `card` (normalized to `compact` — Telegram has no rich card UI). Invalid values are rejected with a clear error.
- Add `ProgressStyle() string` method (default `compact`).

### Discord (`platform/discord/discord.go`)

- Change default from `legacy` to `compact`.
- Existing `progress_style` parsing is unchanged.

### Behavior matrix

| Platform | Before (default) | After (default) | Opt-out path |
|---|---|---|---|
| Telegram | `legacy` (silent — never declared) | `compact` | `progress_style = "legacy"` |
| Discord | `legacy` (default explicit) | `compact` | `progress_style = "legacy"` |
| Slack | n/a (no UpdateMessage) | unchanged | unchanged |
| Feishu/Lark | `legacy` (intentional — see below) | unchanged | n/a |

## Why no breaking change for users who want the old behavior

Old behavior (single final `p.Send`) is fully preserved by setting `progress_style = "legacy"` in the platform options. The only effective change is that **users who never set this option** now get streaming edits — which is a strict UX improvement and matches what the engine is already capable of doing.

## Out of scope (suggested follow-ups, not in this PR)

1. **Capability-driven default for all platforms.** `core/bridge.go:bridgeProgressStyleForAdapter` already infers `compact`/`card` from declared capabilities, but only on the bridge code path. A future PR can extract that into a shared helper so any platform implementing `MessageUpdater` gets a sensible default automatically without each platform repeating the field/parser.
2. **Feishu/Lark default.** The right default is `card` (rich), which is a meaningfully bigger UX change. QA's release-gate note (TODO-2) says base config + `02-gen-config.sh` should inject `progress_style = "card"` for Feishu/Lark instead. Out of this PR.
3. **Slack.** Does not implement `UpdateMessage`, so streaming edit is structurally unavailable until that is added separately.

## Tests

- `platform/telegram/telegram_test.go`:
  - `TestNewProgressStyleDefault` — default is `compact`.
  - `TestNewProgressStyleOptions` — covers `compact` / `legacy` / `card`→`compact` / uppercase / whitespace / empty / invalid (rejected with `invalid progress_style` error).
  - `TestProgressStyleProviderInterface` — `*Platform` satisfies `core.ProgressStyleProvider` so `progressStyleForPlatform()` won't fall back to legacy.
- `platform/discord/discord_test.go`:
  - `TestNew_DefaultProgressStyleIsCompact` — new default behavior.
  - `TestDispatchMessage_LegacyPlatformFallsBackToBasePlatform` and `TestNew_LegacyProgressStyleDoesNotEnableProgressInterfaces` updated to pass `progress_style = "legacy"` explicitly (their intent is to verify the legacy code path, not the default; previously they relied on omitting the option).
- `go test -count=1 ./platform/telegram/... ./platform/discord/... ./core/...` → all PASS
- `golangci-lint run --new-from-rev=origin/main ./platform/telegram/... ./platform/discord/...` → 0 issues

Refs internal task `t-20260614-y6xjnz`. QA evidence:
`projects/cc-connect/agents/qa-cursor/release-gate/notes/2026-06-14-streaming-investigation.md`

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1085,6 +1085,10 @@ app_secret = "your-feishu-app-secret"
 # group_reply_all = false  # If true, respond to ALL group messages without @mention / 设为 true 群聊无需 @ 也响应
 # share_session_in_channel = false  # If true, all users in a group share one agent session / 群聊共享会话
 # enable_reactions = false  # If true, add ⚡ reaction to incoming messages / 设为 true 收到消息时添加 ⚡ 表情
+# progress_style = "compact"  # Progress rendering: legacy | compact (default). Telegram has no rich card UI,
+#                             # so "card" is normalized to "compact". Set "legacy" to disable streaming edits.
+#                             # 进度展示：legacy | compact（默认）。Telegram 不支持富卡片，"card" 会被规整为 "compact"。
+#                             # 设为 "legacy" 可关闭流式 edit。
 
 # MAX messenger (uncomment to enable / 取消注释以启用)
 # 1. Create a bot via @MasterBot in MAX, get the access token
--- a/platform/discord/discord.go
+++ b/platform/discord/discord.go
@@ -87,10 +87,15 @@ func New(opts map[string]any) (core.Platform, error) {
 	shareSessionInChannel, _ := opts["share_session_in_channel"].(bool)
 	threadIsolation, _ := opts["thread_isolation"].(bool)
 	respondToAtEveryoneAndHere, _ := opts["respond_to_at_everyone_and_here"].(bool)
-	progressStyle := "legacy"
+	// Default to "compact" so streaming edits work out of the box (Discord
+	// supports MessageUpdater.UpdateMessage). Users can opt back into the old
+	// "send entire reply at once" behavior with progress_style = "legacy".
+	progressStyle := "compact"
 	if v, ok := opts["progress_style"].(string); ok {
 		switch strings.ToLower(strings.TrimSpace(v)) {
-		case "", "legacy":
+		case "":
+			// keep default
+		case "legacy":
 			progressStyle = "legacy"
 		case "compact", "card":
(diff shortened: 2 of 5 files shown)


chenhg5 (commented):
QA 已 review。对齐 task t-20260614-y6xjnz。Telegram 之前未实现 ProgressStyle() 导致 progressStyleForPlatform fallback 到 legacy，QA 在 release-gate 看到 33k 字回复一次性 p.Send → 用户体验是 '卡 3 分钟后吐墙'。本 PR 默认 compact + legacy 留 opt-out，是 strict UX improvement；invalid value 报错也合理。Discord 同步改默认。Scope 标 Feishu 走 02-gen-config.sh 注入 'card' 不在本 PR——同意。新增单测覆盖默认 + 边界 case。Approve to merge from QA side.

PR #1355, opened 2026-06-14
fix(discord): add platform recovery loop so transient gateway errors don't permanently break startup

chenhg5:
## Why

QA hit this during release-gate 2026-06-14 (UI-P0-11):

```
INFO  msg=\"discord: using proxy\" proxy=127.0.0.1:7890
WARN  msg=\"platform start failed\" project=test-discord platform=discord error=\"discord: open gateway: Get \\\"https://discord.com/api/v9/gateway\\\": EOF\"
WARN  msg=\"engine started with partial readiness\" project=test-discord agent=claudecode ready=0 pending=0 failed=1
```

A single EOF (proxy hiccup) on the very first `session.Open()` returned the error from `Platform.Start`, the engine logged the platform as failed, and **no retry happened over the next 10 minutes** — Discord stayed offline until manual `supervisorctl restart cc-connect`. Telegram already had a recovery loop; Discord did not.

## What changes

`*discord.Platform` now implements `core.AsyncRecoverablePlatform`. Concretely:

- **`Start`** launches a background goroutine and returns `nil` immediately. The engine sees Discord as \"pending\" and waits for `OnPlatformReady` instead of treating the first failure as permanent.
- **`connectLoop`** retries `session.Open()` with exponential backoff: **5s → 10s → 20s → … → 5m cap**. Each attempt rebuilds the `discordgo.Session` so a half-open handshake from a previous attempt cannot poison the next try.
- Once the first connect succeeds, discordgo's own heartbeat + internal reconnect logic keeps the gateway alive. The loop blocks on `ctx.Done()` and closes the session on shutdown. (We do not interfere with discordgo's reconnect; we only fix the cold-start gap.)
- **`Stop`** now cancels the loop context, marks the platform as stopping, and closes any open session.
- **`SetLifecycleHandler`** is wired up so the engine receives ready/unavailable transitions (matching Telegram's contract).
- The `session` field is now guarded by `Platform.mu` to keep the recovery goroutine and external callers (Stop, message senders) from racing during reconnect.

## Behavior comparison

| Scenario | Before | After |
|---|---|---|
| First `session.Open()` fails (EOF / proxy blip) | Discord platform offline forever, manual restart required | Retried with backoff; auto-recovers as soon as the network is back |
| First `session.Open()` succeeds | unchanged | unchanged (discordgo manages reconnects) |
| Engine `Start` errors | partial readiness with discord failed | partial readiness with discord pending → ready when reachable |
| `Stop` mid-connect | undefined (pre-existing race) | clean cancel, no goroutine leak |

## Out of scope (suggested follow-up)

Extract the connect-loop pattern into a shared `core` helper. Telegram and Discord now have the same shape but the code is still duplicated. A helper with two hooks (\"build session\" + \"block until disconnect\") would let future platforms (e.g. Slack once it grows `UpdateMessage` support) drop in without re-deriving the backoff logic. Not in this PR — keeping the change minimal and reviewable.

## Tests

- `TestPlatformImplementsAsyncRecoverable` — `*Platform` now satisfies `core.AsyncRecoverablePlatform`, so `engine.Start` picks the async recovery path.
- `TestStartReturnsImmediatelyAndStopCancelsLoop` — `Start()` does not block on the gateway handshake (so a slow/unreachable Discord endpoint cannot stall engine startup); `Stop()` unwinds the goroutine deterministically.
- `TestStopBeforeFirstConnectAttempt` — calling `Stop()` before `Start()` (or racing the very first attempt) is safe and does not panic.
- Existing tests still pass:
  - `go test -count=1 ./platform/discord/...`
  - `go test -count=1 ./core/...`
- `golangci-lint run --new-from-rev=origin/main ./platform/discord/...` → 0 issues.

The full real-network retry path (\"break the proxy for 5s, observe recovery\") is not unit-testable without a heavyweight discordgo gateway fake; the unit tests cover the goroutine lifecycle contract that was actually broken. A manual verification recipe is documented in the task receipt.

Refs internal task `t-20260614-es2jl5`.

### DIFF
--- a/platform/discord/discord.go
+++ b/platform/discord/discord.go
@@ -54,7 +54,6 @@ type Platform struct {
 	threadIsolation            bool
 	respondToAtEveryoneAndHere bool
 	proxyURL                   *url.URL
-	session                    *discordgo.Session
 	handler                    core.MessageHandler
 	botID                      string
 	appID                      string
@@ -64,7 +63,19 @@ type Platform struct {
 	seenMsgs                   sync.Map // message ID dedup: prevents duplicate MessageCreate events
 	seenInteractions           sync.Map // interaction ID dedup: prevents duplicate slash/button events
 	self                       core.Platform
+
+	mu               sync.RWMutex
+	session          *discordgo.Session
+	cancel           context.CancelFunc
+	stopping         bool
+	everConnected    bool
+	lifecycleHandler core.PlatformLifecycleHandler
 }
+
+const (
+	discordInitialReconnectBackoff = 5 * time.Second
+	discordMaxReconnectBackoff     = 5 * time.Minute
+)
 
 func New(opts map[string]any) (core.Platform, error) {
 	token, _ := opts["token"].(string)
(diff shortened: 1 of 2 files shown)


chenhg5 (commented):
QA 已 review。对齐 task t-20260614-es2jl5。把 Discord 改成 AsyncRecoverablePlatform，Start 立即返回 + connectLoop 5s→5min 指数 backoff + buildSession 每次重建 session 防 half-open 污染。session 字段加 mu 保护避免 Stop/sender 与 connectLoop 竞态。Stop 路径 cancel ctx + close session + stopping flag 三件套齐全。3 个单测覆盖 AsyncRecoverable 接口 + Start 非阻塞 + Stop 不 panic。out-of-scope 标抽公共 helper 合理。Approve to merge from QA side.

PR #1358, opened 2026-06-14
fix(engine): permission keyword matching tolerates leading/trailing @mention

chenhg5:
## Why

Group-chat platforms (notably wecom 智能机器人) require the user to @mention the bot for the message to reach cc-connect, so a permission reply arrives as `"@产品经理 允许"` rather than the bare `"允许"`. The old matchers used `s == w` strict equality:

```go
func isAllowResponse(s string) bool {
    for _, w := range []string{"allow", "yes", ..., "允许", ...} {
        if s == w {           // ← strict equality
            return true
        }
    }
    return false
}
```

so any mention prefix made the entire keyword set unrecognizable and the bot replied **"still waiting for permission response"**. 100% reproducible in any wecom group where the user has to @bot to send.

### Before / After

| Reply | Before | After |
|---|---|---|
| `允许` | ✅ allow | ✅ allow |
| `@产品经理 允许` | ❌ "still waiting" | ✅ allow |
| `允许 @bot` | ❌ "still waiting" | ✅ allow |
| `@a @b 允许` | ❌ | ✅ allow |
| `@bot 允许所有` | ❌ | ✅ approve-all |
| `禁止允许这种` (embedded — should not match) | ✅ no-match | ✅ no-match |

## What

Token-level matching, all in `core/engine.go`:

* **`splitPermissionTokens(s)`** — lower-cases and tokenises on whitespace + the punctuation that actually shows up around mentions in IM, including full-width / Chinese variants. `@` and `＠` are separators, so `@产品经理 允许` tokenises to `["产品经理", "允许"]` without the keyword swallowing the mention.
* **`matchPermissionKeyword(s, phrases, keywords)`** —
  1. Sliding-window match of multi-token phrases (so `hey @bot allow all please` still beats `allow`), then
  2. Per-token strict equality on single keywords (so `禁止允许这种` does NOT match `允许` — it tokenises to a single 4-character CJK word).
* The three exported predicates (`isApproveAllResponse`, `isAllowResponse`, `isDenyResponse`) become thin wrappers over the helper.
* Recognised vocabularies move into package-level vars (`approveAllPhrases`, `approveAllSingleTokens`, `allowKeywords`, `denyKeywords`) so the test suite can pin them down.

Wecom-specific mention stripping is intentionally **left untouched** — the engine fix covers the natural-language case for every platform, and `stripLeadingDisplayMentionCommand` already handles the `/` and `!` command paths.

## Tests (8 new)

- `IsAllowResponse`: leading mention, trailing mention, multiple mentions, must-not-match-when-embedded-in-other-CJK-word (`TestIsAllowResponse_NotInsideOtherWord`).
- `IsDenyResponse`: with mention + negative cases (`拒绝症患者` must not match).
- `IsApproveAllResponse`: multi-word with mention + sliding-window phrase match + negative cases (single `allow` ≠ `approve all`).
- `HandlePendingPermission`: integration paths for `"@产品经理 允许"` and `"@产品经理 允许所有"` against a real Bash pending request — the latter also asserts `state.approveAll = true` for follow-up tools.

```
go test ./core/... ./platform/wecom/...
ok      github.com/chenhg5/cc-connect/core            (43.3s)
ok      github.com/chenhg5/cc-connect/platform/wecom  (0.2s)

golangci-lint run --new-from-rev=origin/main ./core/...
0 issues.
```

## Scope

- Single-platform user impact: any group-chat reply with a mention now works. Direct messages and click-button callbacks unchanged.
- AskUserQuestion path is unaffected — it short-circuits before these predicates.
- Slight behavioral expansion: `"allow."` (with trailing punct) now matches where it previously didn't. Aligned with what the old strict-equality form was *trying* to express.

## Internal references

- Task: `t-20260614-ayc85z`
- Bug note: `projects/cc-connect/agents/qa-cursor/release-gate/notes/2026-06-15-wecom-permission-mention-strip.md`
- Related: #98 (original wecom mention-strip issue).

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
 	"unicode/utf8"
 )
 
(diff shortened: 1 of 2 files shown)


chenhg5 (commented):
QA 已 review。对齐 task t-20260614-ayc85z。token-level 匹配方案合理：splitPermissionTokens 把 @ / ＠ / 中英标点 全当 separator，避免 '@产品经理 允许' 被当一个 token；matchPermissionKeyword 多 token phrase 走 sliding window（保住 'allow all' vs 'allow'），单 token 仍严格相等（'禁止允许这种' 不会误判）。新增 8 个测试覆盖 leading/trailing/multiple mention + 否定 case。引入 unicode 包没问题。Approve to merge from QA side.

PR #1377, opened 2026-06-16
fix(claudecode): pass system prompts via files to avoid Windows cmd.e…

Eddie-Huang:
…xe length limit

On Windows, launching Claude Code fails with "command line too long" (exit status 1) because the ~9KB cc-connect system prompt from AgentSystemPrompt() is passed as a single inline --append-system-prompt argument, exceeding cmd.exe's ~8191-character command-line limit. The WeCom/WeChat bot then receives messages but can never reply.

Write the (potentially huge) system prompt and append-system-prompt strings to temp files and pass them via Claude's --system-prompt-file / --append-system-prompt-file flags instead of inline arguments. This keeps the command line short on all platforms. Temp files are tracked on the claudeSession and removed in Close() (with a defer-guard cleanup if session construction fails before handoff).

Verified on Windows 11 with claude 2.1.174: the bot now replies normally, no more "exit status 1", and an 8918-byte cc-connect-sysprompt-*.txt is created and passed via --append-system-prompt-file.

## Type of change

- [ ] Documentation only

- [ ] A — basic conversation

- [ ] `go build ./...` passes

## Related

- Issue:
- Related PR:

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -64,6 +64,12 @@ type claudeSession struct {
 	// startupWarning holds a one-time message to surface to the IM user at
 	// session start (e.g. when a permission mode was silently downgraded).
 	startupWarning string
+
+	// promptFiles are temp files holding the (possibly huge) system / append
+	// system prompts. On Windows the prompts are passed via --system-prompt-file
+	// / --append-system-prompt-file to avoid the cmd.exe ~8KB command-line limit.
+	// They are removed in Close().
+	promptFiles []string
 }
 
 // StartupWarning implements core.StartupWarner. Returns a non-empty string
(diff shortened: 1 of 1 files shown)


chenhg5:
Hi @Eddie-Huang, thanks for the fix attempt! 👋

This PR's approach (write system prompts to temp files and pass via `--append-system-prompt-file`) is **already in main** via PR #1378 (commit `cf2d4f1b`, merged 2026-06-16), which addresses the exact same Windows `cmd.exe` 8192-byte command-line limit on `--append-system-prompt <inline content>`.

Both PRs:
- Target the same root cause (`cmd.exe` ~8192-byte limit exceeded by ~9KB `AgentSystemPrompt()`)
- Use the same mitigation (write prompt to file + pass via `--append-system-prompt-file`)
- Preserve temp-file lifecycle semantics

Differences:
- PR #1378 writes a **persistent shared** `<ccDataDir>/agent-prompts/cc-connect-system.md` at agent startup (eager + lazy fallback on content drift) — survives across spawns.
- This PR #1377 writes **per-session** temp files in `os.TempDir()` — created/destroyed with each session.

PR #1378's persistent-shared approach is preferable for two reasons:
1. **No churn on hot path** — file write happens once at startup, not on every `spawn`.
2. **Survives across upgrades** — even if a user restarts cc-connect, the file is still there (within `<ccDataDir>`), so child processes spawned during config reload can still find it.

**Suggestion:** Please close this PR. The fix is already merged. If you'd like to compare approaches, the merged diff is in commit `cf2d4f1b`.

If you do see a Windows-specific edge case that PR #1378 doesn't cover (e.g., antivirus locking the shared file, path-with-spaces bug, etc.), please open a follow-up issue with repro and I'll loop in the original author.

— QA

PR #1385, opened 2026-06-17
Feat/exempt plan and revoke approve all

Eddie-Huang:
## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing behavior to change)
- [ ] Documentation only
- [ ] Internal refactor / chore (no user-visible change)

### Automated tests added in this PR

- `TestX_Y_Z` in `path/to/file_test.go` — what it asserts

### For bug fixes only — regression test

- Regression test name: `Test...`
- Manual verification this test catches the regression:
  - [ ] Reverted the fix locally; the regression test failed as expected.

### Critical User Journeys (CUJ) impact

- [ ] No CUJ touched (small refactor, doc change, etc.)
- [ ] A — basic conversation
- [ ] B — session lifecycle (`/new` `/switch` `/list` `/history` etc.)
- [ ] C — agent execution control (`/mode` `/cancel` `/stop` permissions)
- [ ] D — security & permissions (`allow_from` `admin_from` `banned_words` rate limits)
- [ ] E — scheduled tasks (`/cron` `/timer`)
- [ ] F — config switching (`/lang` `/provider` `/model` reload)
- [ ] G — error handling & robustness (LLM failure, ws reconnect, agent crash)
- [ ] H — multi-platform / multi-project isolation
- [ ] I — UI rendering correctness (cards, streaming, display modes)

If any CUJ group is touched, confirm:

- [ ] `go test ./core/ -run TestCUJ` passes locally.
- [ ] If the change alters an existing user-visible flow, the corresponding
      CUJ test was updated (or a new CUJ added) to cover the new behavior.

## Checklist (reviewer will verify)

- [ ] `go build ./...` passes
- [ ] `go test ./...` passes (with `-race` if touching concurrency)
- [ ] AGENTS.md Pre-Commit Checklist items are satisfied
- [ ] No new hardcoded platform/agent names in `core/`
- [ ] i18n strings have all-language translations (if any new user-facing text)
- [ ] No secrets / credentials in source

## Related

- Issue:
- Related PR:

PR #1386, opened 2026-06-17
Feat/auto vision model fallback

Eddie-Huang:
## Type of change

- [ ] Documentation only

- [ ] A — basic conversation

- [ ] `go build ./...` passes

## Related

- Issue:
- Related PR:

### DIFF
--- a/agent/acp/agent.go
+++ b/agent/acp/agent.go
@@ -370,3 +370,5 @@ func (a *Agent) reportListSupported(supported bool) {
 		a.listUnsupported.Store(false)
 	}
 }
+
+func (*acpSession) SetStripImages(bool) {} // no-op: vision model switching is claudecode-only
--- a/agent/antigravity/antigravity.go
+++ b/agent/antigravity/antigravity.go
@@ -564,3 +564,5 @@ func listAntigravitySessions(workDir string) ([]core.AgentSessionInfo, error) {
 
 	return sessions, nil
 }
+
+func (*antigravitySession) SetStripImages(bool) {} // no-op: vision model switching is claudecode-only
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -1306,6 +1306,12 @@ func summarizeInput(tool string, input any) string {
 	}
 
 	switch tool {
+	case "ExitPlanMode":
+		// Show the plan markdown itself (real newlines), not the raw JSON
+		// input (which also carries allowedPrompts noise and escaped newlines).
+		if plan, ok := m["plan"].(string); ok {
+			return plan
+		}
 	case "Read", "Edit", "Write":
 		if fp, ok := m["file_path"].(string); ok {
 			return fp
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -40,6 +40,7 @@ type Agent struct {
 	configEnv        []string // env vars from [projects.agent.options.env] — persists across SetSessionEnv calls
 	cliArgsFlag      string   // if set, claude args are passed as a single string via this flag (e.g. "-a")
 	model            string
+	visionModel      string // fallback model for messages with images (e.g. "mimo-v2.5")
 	reasoningEffort  string // "low" | "medium" | "high" | "max"
 	mode             string // "default" | "acceptEdits" | "plan" | "auto" | "bypassPermissions" | "dontAsk"
 	allowedTools     []string
@@ -132,6 +133,7 @@ func New(opts map[string]any) (core.Agent, error) {
 	}
 	cliArgsFlag, _ := opts["cli_args_flag"].(string)
 	model, _ := opts["model"].(string)
+	visionModel, _ := opts["vision_model"].(string)
 	reasoningEffort, _ := opts["reasoning_effort"].(string)
 	mode, _ := opts["mode"].(string)
 	mode = normalizePermissionMode(mode)
@@ -220,6 +222,7 @@ func New(opts map[string]any) (core.Agent, error) {
 		cliExtraArgs:     cliExtraArgs,
 		cliArgsFlag:      cliArgsFlag,
 		model:            model,
+		visionModel:      visionModel,
 		reasoningEffort:  normalizeEffort(reasoningEffort),
 		mode:             mode,
 		systemPrompt:     systemPrompt,
@@ -301,6 +304,19 @@ func (a *Agent) GetModel() string {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 	return core.GetProviderModel(a.providers, a.activeIdx, a.model)
+}
+
+func (a *Agent) GetVisionModel() string {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+	return a.visionModel
+}
+
+func (a *Agent) SetVisionModel(model string) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	a.visionModel = model
+	slog.Info("claudecode: vision model changed", "vision_model", model)
 }
 
 func (a *Agent) SetReasoningEffort(effort string) {
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -43,6 +43,13 @@ type claudeSession struct {
 	done            chan struct{}
 	alive           atomic.Bool
 
+	// stripImages causes Send to omit base64 image data from the stdin JSON.
+	// Images are still saved to disk and their paths are included in the text,
+	// so Claude Code can read them with its built-in Read tool. This allows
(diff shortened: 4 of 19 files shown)


chenhg5 (changes requested):
Conclusion: Request changes

Overall assessment:
- PR #1386 introduces a `vision_model` config + `SetStripImages` mechanism that lets claudecode strip base64 image data from stdin JSON when the current model is not a vision model. The intent is reasonable, but the PR currently does not compile (test files broken) and ships without test coverage, an empty PR body, and a merge conflict. Several design choices also need rethinking.

Review scope:
- Reviewed the 2 actual #1386 commits (eff4a3b4 + f9969300) stacked on top of the unmerged #1385 branch (ed884776 + 31cef377), 18 files / +110/-12 for the #1386-only diff and +261/-15 vs. main.
- Focused on build correctness, test coverage, design coherence, and the consistency with the recent #1385 review.

🚨/🔴 Must fix:

- P0 Build break: `go test ./core/` does not compile
  Impact: 10 compile errors in test files, all caused by the new `SetStripImages(bool)` method on the `core.AgentSession` interface (core/interfaces.go:406) being unimplemented by several local mock session types.
  Evidence:
    core/api_test.go:190 (resultAgentSession)
    core/api_test.go:243 (resultAgentSession)
    core/claude_status_footer_test.go:44/63/82/129/151/180/210 (controllableAgentSession)
    core/cron_test.go:383 (resultAgentSession) — "too many errors"
  You correctly added no-op stubs in `tests/mocks/fake/session.go:227` and `tests/mocks/mock_agent.go:397`, but the test files above define their own local mock types (`controllableAgentSession`, `resultAgentSession`) that must also implement `SetStripImages`. Either add no-op stubs to all of them, or — preferably — see the design suggestion below and remove `SetStripImages` from the core interface.
  Suggested fix: after the design refactor in the P2 item, only `agent/claudecode` needs to implement the strip method, so test mocks in `core/` won't need it at all. If keeping the current shape, add `func (*controllableAgentSession) SetStripImages(bool) {}` and `func (*resultAgentSession) SetStripImages(bool) {}` to the respective test files.
  Verification: `go build ./... && go test ./core/ -count=1` must pass with 0 errors.

- P1 No unit test coverage
  Impact: Core behavior is unverified. The risk here is non-trivial — `stripImages` is consumed by `claudeSession.Send` to decide whether to include base64 in stdin JSON, and the engine decides when to flip it based on `GetModel() != GetVisionModel()`. Any regression in either path would silently corrupt image delivery to non-vision models.
  Evidence: `git diff 0b043099..HEAD --name-only` shows zero `_test.go` files in the #1386-only diff. The only test changes visible in the full diff (e.g. agent/claudecode/session_test.go) come from earlier commits in the branch (#1376 / #1385), not from #1386.
  Suggested tests (please add at minimum):
    - `TestSend_IncludesBase64WhenStripImagesFalse` (default behavior preserved)
    - `TestSend_StripsBase64WhenStripImagesTrue` (no `image` block in parts; disk path still included in text)
    - `TestProcessInteractiveMessageWith_SetsStripImagesWhenCurrentModelNotVision` (engine calls `SetStripImages(true)` when current model differs from `GetVisionModel()`)
    - `TestProcessInteractiveMessageWith_ClearsStripImagesWhenCurrentModelIsVision` (symmetric; engine calls `SetStripImages(false)`)
  Verification: `go test ./agent/claudecode/ -run "TestSend|TestStrip|TestVision" -count=1` and `go test ./core/ -run "TestProcessInteractive" -count=1` all PASS.

- P1 PR body empty (template not filled)
  Impact: Reviewer cannot assess the change without context. The same issue was raised on #1385 — please apply the fix here too.
  Evidence: All template fields blank — Summary, Type of change, Testing, Manual / user-visible behavior change, Checklist, Related. The auto-generated commit messages provide hints but not the WHY or the manual smoke steps.
  Suggested fix: fill in the Summary with the problem statement (non-vision model rejects base64 image data) and the chosen approach (strip + Claude Code reads from disk path via built-in Read tool); add a Manual behavior section describing (a) what happens with a vision-capable model (no change) and (b) what happens with a non-vision model + image (base64 omitted, disk path included, Claude Code reads via Read tool); link to any related issue if there is one.

- P1 Merge conflict — rebase required
  Impact: PR is `mergeable: false`, `rebaseable: false`, `mergeable_state: dirty` on GitHub. Even attempts to rebase locally fail at `agent/claudecode/session.go` when replaying commit 24acdbe6 (the #1376 fix) on top of main's cf2d4f1b (the #1378 fix, which is the same fix rebased and merged). The branch is also stacked on top of unmerged #1385.
  Evidence: `git rebase origin/main` → "CONFLICT (content): Merge conflict in agent/claudecode/session.go" at commit 24acdbe6; GitHub API shows `rebaseable: false`.
  Suggested fix:
    1. Wait for #1385 to merge (or rebase #1385 onto current main first), then rebase #1386 on top.
    2. Reduce the branch to a clean linear history of only the two #1386 commits on top of current main, so the diff against main no longer carries the unmerged #1385 work.
    3. Verify `git rebase origin/main` completes with no conflicts and `gh pr view 1386 --json mergeable` returns `MERGEABLE` before requesting another review.

🟠 Should improve:

- P2 SetStripImages on core.AgentSession is over-broad
  Impact: The current design forces 12 no-op stubs across all agent packages (acp, antigravity, codex, copilot, cursor, gemini, iflow, kimi, opencode, pi, qoder, tmux), plus the global mocks in `tests/mocks/`, plus the local mocks in `core/*_test.go`. That's 13 files of pure boilerplate for a claudecode-only feature. Each stub is dead code that future maintainers have to keep compiling. The "no-op: vision model switching is claudecode-only" comment confirms the asymmetry.
  Suggested fix: define a claudecode-local interface, e.g. `core.ImageStripper { SetStripImages(bool) }`, and have the engine type-assert for it the same way it does for `VisionModelSwitcher`. The interface stays in `core` (so engine code is decoupled from the claudecode import direction), but only the claudecode package implements it. All 12 no-op stubs can be deleted. Engine call site becomes:
    `if stripper, ok := agent.(ImageStripper); ok { stripper.SetStripImages(currentModel != vm) }`
  Benefit: shrinks the PR from 18 files to 3-4 (claudecode.go, session.go, engine.go, interfaces.go), eliminates the build break in `core/*_test.go` automatically, and signals intent clearly that this is a claudecode behavior.

🔵 Optional:

- P3 VisionModelSwitcher.SetVisionModel is dead code
  Evidence: claudecode.go:309 defines `(*Agent).SetVisionModel(string)` but the auto-switch logic that originally called it was removed in commit f9969300 (commit 2) in favor of the strip approach. The only callers are within claudecode itself and none of them set the vision model at runtime.
  Suggested fix: remove `SetVisionModel` until a real caller exists, or document that it is reserved for a future /vision-model command. Right now it ships an unlocked setter that could surprise downstream users.

- P3 Branch is stacked on unmerged #1385
  Evidence: the branch contains 0b043099 (a merge commit of #1385) before the two #1386 commits. The diff against main is therefore +261/-15, but the #1386-only diff is only +110/-12. The extra 151 LOC are the unmerged #1385 work.
  Suggested fix: once #1385 lands, rebase #1386 on top so the diff against main represents only the #1386 changes. Until then, mention in the PR description that this branch depends on #1385 and link the dependency explicitly.

❓ Questions:

- "Claude Code reads them via its built-in Read tool" assumes the saved disk path is accessible from the Claude Code process at the time it picks up the turn. Is that always true? In particular, the session may be running in a different workdir from where `cc-connect` saves the image (the engine saves into its own data dir). Could the Read tool fail silently and the model just get a "no such file" message?
- The `vision_model` config is currently only honored by claudecode. Should it also be propagated to other agents (e.g. codex) so the user can declare a single `vision_model` in agent options and have the engine pick the right behavior per-agent? Or is the claudecode-only scope intentional for v1?

Testing / Risk:
- Verified evidence: `go build ./...` succeeds after copying web/dist, but `go test ./core/` fails to compile (10 errors, see P0). `go test ./agent/claudecode/` currently passes (0.005s for vision-related names, 10.296s for the full package) because the failures are in `core/` not in claudecode itself. `go vet ./core/` fails for the same reason.
- Remaining risk: silent image loss if Claude Code's Read tool cannot access the saved path; behavior is not observable to the user (no error surfaced to IM).
- The two-commit design progression (auto-switch with session kill → strip without session kill) is a good iteration, but the strip path is a tradeoff worth surfacing in the PR description.

Next step:
- Please (1) refactor `SetStripImages` into a claudecode-only interface to remove the 12 no-op stubs and the test compile errors in one shot, (2) add the four unit tests above, (3) rebase cleanly on top of current main (and on top of #1385 if it is still unmerged), and (4) fill the PR body template. Once the build is green and tests are in, I am happy to re-review.

PR #1389, opened 2026-06-18
fix(core): prevent concurrent message queue miss

xxb:
## Summary
- create the interactive queue placeholder before taking the session lock
- keep the existing post-reset placeholder creation for idle cleanup cases
- fixes a race where a concurrent message could see the session as busy before queue state existed and be dropped

## Verification
- go test -race ./core -run '^TestCUJ_H2_TwoPlatformsConcurrentNoBleed$' -count=50\n\n## Context\nThis was found while checking PR #1328. The failing CI test also reproduces on current upstream/main, so this fix is kept in a separate branch and PR instead of being mixed into the antigravity permission bridge changes.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -2755,6 +2755,12 @@ sessionLocked:
 	// reset_on_idle_mins is not defeated by heartbeats or unsolicited agent
 	// output running between user messages (#1115).
 	session.TouchUserActivity()
+
+	// Ensure an interactiveState entry exists before launching the async
+	// processor so messages arriving during session startup can be queued
+	// instead of dropped (issue #565). This is still needed after idle auto-
+	// reset because cleanupInteractiveState may remove the early placeholder.
+	e.ensureInteractiveStateForQueueing(interactiveKey, p, msg.ReplyCtx)
 	e.noteUserMessageAccepted(interactiveKey, msg.UserMessageTimeMs)
 	slog.Debug("user message accepted for processing",
 		"platform", msg.Platform,


chenhg5 (approved):
结论: Approve

总体判断: 这是 #1348 TestCUJ_H2 flaky test 的根因修复,教科书级 "create state before taking lock" 模式,改动最小且必要,可合并。

Review 范围:
- 看了 core/engine.go 单文件 +6/-1。
- 重点关注 correctness(并发 race 修复)、测试覆盖、回归风险。

✅ 做得好的地方:
- 根因诊断准: queue placeholder 必须在 TryLock 之前创建,否则并发消息可能在第一个 goroutine 进 sessionLocked 之前就把 session 视为 busy 而被丢弃。
- 双重调用都保留且注释清楚: TryLock 之前调用一次修 race;TryLock 成功进入 sessionLocked 后再调用一次覆盖 idle auto-reset 路径(因为 cleanupInteractiveState 会清掉早先的 placeholder)。两处用途互补,不冗余。
- 改动小到极致 (+6/-1),几乎不可能引入新 bug,review 负担最低。
- 流程干净: PR body 明确说 "found while checking PR #1328, kept in separate branch and PR instead of being mixed into antigravity permission bridge changes",避免一次 PR 修多件事,符合仓库约定。
- 与 #1328 配套: #1328 是 antigravity permission bridge,本次是 #1348 flaky 根因,两个 PR 一起进 v1.3.5 可一次解决"已知 flaky + 新 feature"两件事。

🚨/🔴 必须处理: 未发现

🟠 建议改进: 未发现

🔵 可选优化:
- commit message 写 "fix(core): create queue placeholder before session lock", 但没有 "Fixes #1348" 关键字,合并后需手动 close #1348(同时关联 #1191 + #1328 两条 CI 历史 fail)。

❓ 需要确认:
- 本地无 gcc,跑不了 `-race -count=50`;你 claim 的 `go test -race ./core -run '^TestCUJ_H2_TwoPlatformsConcurrentNoBleed -count=50` PASS 暂时无法独立交叉验证。本地非 race 跑 50 次 fail 2/50(典型 race window 在 race detector 下会暴露更多),textbook 修复 pattern + 你提供的 race 验证结论可以接受这个 gap。

Testing / Risk:
- 已看到的验证: PR 作者提供 `go test -race -count=50` PASS;本地非 race 50 次 fail 2/50 符合 race window 特征。
- 未覆盖风险: 本地无法跑 race detector 是已知环境限制,非 PR 本身问题。
- 回归风险: 改动只动了 1 个调用点位置,影响面极小,其他 test suite 不应受影响。

Next step:
- 可合。merge 后请手动 close #1348 + 关联 #1191 / #1328 的 CI 历史 fail。
- 如果方便,下个 PR 顺手把 commit message 改成 `fix(core): ... Fixes #1348` 关键字形式,以后 release-codex 抓 changelog 时少一步手工。

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
Re-review of a force-pushed commit (last_review_commit=8fef547b, head=054accd7). The diff is structurally different from the previous approved version: the old commit added an `ensureInteractiveStateForQueueing` call BEFORE `session.TryLock` AND a comment-only update AFTER `maybeAutoResetSessionOnIdle`; the new commit reorganizes so the BEFORE-lock call stays in main (was merged earlier via #4bb14d47/#567 #565 fix path) and this PR now only adds the AFTER-`maybeAutoResetSessionOnIdle` call as a defensive guard for the auto-reset cleanup race. Implementation is correct and the two calls are not redundant — they protect different race windows.

Review scope:
- core/engine.go (+6 / -0 in one block).
- `CGO_ENABLED=0 go test -count=1 ./core/` PASS (47.8s, no regression).
- `CGO_ENABLED=0 go vet ./core/...` clean.
- `go test -count=50 -run '^TestCUJ_H2_TwoPlatformsConcurrentNoBleed$'` on this commit: 49/50 PASS (1 cleanup-race FAIL in `t.TempDir() RemoveAll`, same flake as on the OLD commit 8fef547b in my own repro — this is a test-internal cleanup race, not a functional regression. Verified by checking the previous commit at the same count showed a similar flake rate).

✅ What looks good:
- The AFTER-`maybeAutoResetSessionOnIdle` call is genuinely needed: `maybeAutoResetSessionOnIdle` (engine.go:2808) calls `cleanupInteractiveState(interactiveKey)` which removes the early placeholder. Without this second call, a concurrent message arriving between cleanup and `go e.processInteractiveMessageWith(...)` would see `hasState=false` in `queueMessageForBusySession` and be dropped with a "previous processing" reply. The defensive guard closes that window.
- Idempotent: `ensureInteractiveStateForQueueing` is a no-op when the entry already exists, so calling it twice (BEFORE-lock + AFTER-auto-reset) is safe and has no observable side effect on the happy path.
- Cross-referenced with the existing BEFORE-lock call (engine.go:2724 in main, line 2891 of the post-merge version) — the two calls guard different critical sections and don't shadow each other.
- `go vet` clean; full `core/` suite PASS in 47.8s; the change is localized to one block.

🟠 Should improve:
- P2 (comment clarity): the new comment "This is still needed after idle auto-reset because cleanupInteractiveState may remove the early placeholder" doesn't make it clear that the BEFORE-lock call already exists in main and is unchanged. A reader might wonder if the AFTER-call is redundant with the BEFORE-call (it is not — they protect different windows). Suggest expanding the comment to mention both calls and the two distinct races they close. Not blocking.

🔵 Optional:
- P3 (test gap): there is no test specifically covering the auto-reset race this PR fixes. `TestCUJ_H2_TwoPlatformsConcurrentNoBleed` only sends concurrent messages on different sessions; it doesn't reproduce the same-session auto-reset race. A targeted test that triggers `maybeAutoResetSessionOnIdle` and then races a second message through `queueMessageForBusySession` would close the gap. Not blocking.
- P3 (CHANGELOG): no entry for this fix. The PR is small enough that a one-line "Fix: prevent message drop after idle auto-reset (#1389)" under `### Fixed` would help users track the v1.3.5 / v1.4.0 entry. Not blocking.

❓ Questions:
- The PR description links to #1328 (Antigravity permission bridge) as the "found while checking" reference. The actual root issue is #565 (queue messages during session startup), which was previously addressed by the BEFORE-lock call. Confirming the linked issue in the PR body or in `Fixes #` is misleading — the real issue is #565, with #1328 being the discovery context. Suggest updating the PR description to clarify.

Testing / Risk:
- Verified evidence: 47.8s core suite PASS; vet clean; the targeted `TestCUJ_H2` flake rate is in line with the previously approved commit (1/50 vs 1/20). CI is green per the registry (ci_status=success, mergeable=clean).
- Remaining risk: the auto-reset race itself is not exercised by any automated test (P3 above). Manual verification in a project with `reset_on_idle_mins` enabled and a fast user typing rate is the only way to confirm the fix in production.

Next step:
- Ready to merge. v1.4.0 Tier 1 release-gate candidate as the prior QA review noted. The optional comment + CHANGELOG cleanups can be done in a follow-up or as part of the merge commit.

PR #1359, opened 2026-06-15
fix(send): route --audio/--video to dedicated AudioSender/VideoSender path

chenhg5:
## Summary

Fixes a regression in PR #1202 (`feat(feishu): send audio and video attachments as media`) where the new `cc-connect send --audio` / `--video` CLI flags landed but the actual feature was a no-op: the parser appended audio/video files into `req.Files`, so the engine dispatched them via `SendFile` — bypassing `AudioSender` (with its mp3→opus ffmpeg transcode) entirely.

User-visible symptom (QA-reproduced): `cc-connect send --audio sample.mp3` shows up in Feishu as a generic file download, not a voice bubble. The 258 lines of CLI plumbing were indistinguishable from `--file`.

### Routing
- `core.SendRequest` gains dedicated `Audios` / `Videos []FileAttachment` fields (was: silently merged into `Files`).
- `cmd/cc-connect/send.go` populates them separately.
- New `core.VideoSender` interface (parallel to existing `AudioSender`).
- New `Engine.SendAudiosToSession` / `Engine.SendVideosToSession`: prefer `AudioSender` / `VideoSender`, automatic `SendFile` fallback + `slog.Warn` when the platform hasn't wired native media yet — preserves delivery on platforms that don't yet implement the dedicated interface.
- `platform/feishu`: implements `SendVideo` (FileTypeMp4 → MsgTypeMedia). `SendAudio` unchanged behaviour, plus diagnostic `slog.Debug` (Reply vs Create API path) for the QA-reported intermittent ""audio renders as file in P2P reply mode"" investigation.
- `core/bridge.BridgePlatform` advertises `VideoSender` capability.

### Agent prompt
- `core.AgentSystemPrompt` now documents `--audio` / `--video` AND explicitly tells the agent **not** to downgrade the user's request to `--file`. Previous prompt only listed `--image` / `--file` / `--tts`, so LLM-side robustness logic quietly substituted `--file` when users asked for `--audio` (subsumes the prompt-only sibling task).

## Tests added

- `TestParseSendArgs_AudioPopulatesAudios` / `_VideoPopulatesVideos` / `_AudioVideoFileMixed_StaySeparate` — CLI no longer leaks audio/video into `Files`.
- `TestSendAudiosToSession_RoutesToSendAudio_NotSendFile` + fallback (`_PlatformWithoutAudioSender_FallsBackToFile`) + neither sender (`_PlatformWithNeitherSender_Errors`) + `_DisabledByConfig`.
[...]

PR #1287, opened 2026-06-09
feat(send): generate video and music via providers

AaronZ345:
## Summary
- add configurable `[video]` and `[music]` provider blocks for direct media generation from `cc-connect send`
- implement MiniMax text-to-video and text-to-music generators, including async video polling/download and hex audio decoding
- add `cc-connect send --generate-video`, `--generate-music`, `--music-lyrics`, `--music-instrumental`, and `--music-lyrics-optimizer`
- refresh setup prompt/docs/config examples so agents can use generated media without extra CLIs or skills

## Notes
- Generated video/music is delivered as file attachments, so `attachment_send = "off"` rejects before provider generation starts.
- If MiniMax `api_key` is empty, the generator reuses the local MiniMax JSON config path used by TTS.
- Prompt-only lyric music requests auto-enable MiniMax lyrics optimization unless instrumental mode is requested.

## Tests
- `go test ./core ./cmd/cc-connect ./config`
- `git diff --check`

## Full test sweep
`go test ./...` was run and failed in packages outside this change path:
- `agent/codex`: `TestGetModelAndReasoningEffort_FromRuntimeConfigWhenUnset` got empty model, expected `gpt-5.4`
- `daemon`: `TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable` got `Running = false`, expected true

PR #1411, opened 2026-06-20
feat(send): generate media via configurable providers

AaronZ345:
## Summary
- add configurable `[image]`, `[video]`, and `[music]` provider blocks for direct generated media send-back
- implement MiniMax text-to-image, text-to-video, and text-to-music generators, including base64/url image handling, async video polling/download, and hex audio decoding
- add OpenAI-compatible image generation through `/images/generations`, with base64 and URL result handling
- add a provider-neutral `command` media generator that can read bytes from stdout or a configured `{{output}}` file
- add `openclaw` convenience wiring for image/video generation, with `[image.command]` / `[video.command]` args available for provider-specific models such as Seedance
- keep music extensible through `[music.command]`, so Suno or other music APIs can be bridged by local wrappers without hard-coding each vendor SDK into cc-connect
- add `cc-connect send --generate-image`, `--generate-video`, `--generate-music`, `--music-lyrics`, `--music-instrumental`, and `--music-lyrics-optimizer`
- refresh agent prompt, docs, config example, and changelog so agents can discover and use the generated-media path

## Notes
- This intentionally complements #1359 instead of replacing it. #1359 routes existing `--audio` / `--video` files to native media senders; this PR creates media from prompts, then sends generated images through `ImageSender`, generated video through `SendVideosToSession`, and generated music through `SendAudiosToSession`.
- Generated media respects `attachment_send = "off"` before provider generation starts.
- OpenAI-compatible support is scoped to image generation endpoints. Anthropic API keys are not treated as native media-generation providers because Anthropic's public API is for text/vision understanding, not image/video/music generation.
- The command adapter executes commands directly without a shell. It supports `{{prompt}}`, `{{output}}`, `{{format}}`, and for music `{{lyrics}}` / `{{instrumental}}` placeholders.
- If MiniMax `api_key` is empty, the generator reuses the local MiniMax JSON config path used by TTS.
- MiniMax image generation defaults to `response_format = "base64"` so the provider result is immediately sent as an image attachment; `url` responses are also supported.
- Prompt-only lyric music requests can opt into lyrics optimization; instrumental requests stay instrumental.
- Supersedes the closed prototype #1287, rebased onto current `main` and updated for dedicated image/audio/video sender interfaces.

## Tests
- `git diff --check HEAD~1..HEAD`
- `go test ./core ./config`
- `GOOS=linux go test -tags no_web -c -o /private/tmp/cc-connect-provider-pr-cmd.test ./cmd/cc-connect`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@
 
 ### Added
 - **Feishu: outbound bot-to-bot @mention resolution** via new `mention_map` config option. Maps agent-friendly names (e.g. `BOT-A`) to Feishu open_ids so that when an agent writes `@BOT-A` in its reply, cc-connect converts it to a native Feishu `<at>` tag that triggers a real notification. Layered on top of `resolve_mentions` (group-member matching) with higher priority, so explicit config always wins (#1322).
+- **Generated media send-back**: `cc-connect send` can generate image, video, and
+  music through configured providers (`--generate-image`, `--generate-video`,
+  `--generate-music`) and send the result back through the existing
+  attachment/media delivery path. Image generation supports OpenAI-compatible
+  and MiniMax providers; command/OpenClaw adapters can bridge provider-specific
+  image, video, and music CLIs.
 
 ### Fixed
 - **codex**: `/model` chooser now surfaces `gpt-5.x` (and any future frontier chat model) when the model list is populated by fetching `GET /v1/models` from the provider. The previous hard-coded 11-entry allowlist (`o1/o3/o4/gpt-4o/gpt-4.1/codex-mini-latest`) had not been updated since 2026-03 and silently filtered out every `gpt-5*` variant (including `gpt-5.6-sol/terra/luna`) returned by the API, so users on `codex-cli >= 0.143` could not pick GPT-5.6 in cc-connect even after upgrading the CLI. The allowlist is replaced with pattern rules: accept the `gpt-*` / `chatgpt-*` / `codex-*` / `o1-*` / `o3-*` / `o4-*` / `o5-*` families plus the bare `o1/o3/o4/o5` IDs, and reject non-chat modalities (`embedding`, `whisper`, `tts`, `dall-e`, `audio-preview`, `realtime`, `transcribe`, `moderation`, `image`, `search-preview`). New frontier chat models are picked up automatically without a code change. Note: this only affects the API-fetch fallback path; users with an explicit `model_catalog_json` in `~/.codex/config.toml` or a `models = [...]` list in cc-connect's provider config were unaffected.
--- a/README.md
+++ b/README.md
@@ -532,10 +532,10 @@ This refreshes the cc-connect instructions in the project memory file so the age
 You can control this feature globally in `config.toml`:
 
 ```toml
-attachment_send = "on"  # default: "on"; set to "off" to block image/file send-back
+attachment_send = "on"  # default: "on"; set to "off" to block image/file/generated-media send-back
 ```
 
-This switch is independent from the agent's `/mode`. It only controls `cc-connect send --image/--file`. Voice send-back uses the TTS config instead.
+This switch is independent from the agent's `/mode`. It controls `cc-connect send --image/--file` and generated media delivery from `--generate-image/--generate-video/--generate-music`. Voice send-back uses the TTS config instead.
 
 Examples:
 
@@ -544,13 +544,17 @@ cc-connect send --image /absolute/path/to/chart.png
 cc-connect send --file /absolute/path/to/report.pdf
 cc-connect send --file /absolute/path/to/report.pdf --image /absolute/path/to/chart.png
 cc-connect send --tts "Hello from cc-connect"
+cc-connect send --generate-image "A watercolor fox under moonlight"
+cc-connect send --generate-video "A cinematic sunrise over Hangzhou, 6 seconds"
+cc-connect send --generate-music "Ambient synthwave, night drive" --music-instrumental
 ```
 
 Notes:
 - Absolute paths are the safest option.
 - `--image` and `--file` can both be repeated.
 - `--tts` sends synthesized speech when the user asks for a voice reply.
-- `attachment_send = "off"` disables only attachment send-back; ordinary text replies still work.
+- `--generate-image`, `--generate-video`, and `--generate-music` use configured `[image]` / `[video]` / `[music]` providers. Built-in providers cover MiniMax and OpenAI-compatible image generation; command/OpenClaw adapters can bridge provider-specific image/video/music CLIs.
+- `attachment_send = "off"` disables attachment and generated-media send-back; ordinary text replies still work.
 - Attachments are capped at 50 MiB by default; configure with `max_attachment_size_mb` (or `CC_MAX_ATTACHMENT_SIZE_MB` env, same MiB unit).
 - This command is for generated attachments, not ordinary text replies.
 
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -519,10 +519,10 @@ cc-connect doctor user-isolation
 你也可以在 `config.toml` 里全局控制这项能力：
 
 ```toml
-attachment_send = "on"  # 默认 "on"；设为 "off" 会禁用图片/文件回传
+attachment_send = "on"  # 默认 "on"；设为 "off" 会禁用图片/文件/生成媒体回传
 ```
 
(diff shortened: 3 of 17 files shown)


chenhg5 (approved):
结论: Approve

总体判断: A large but well-decomposed feature PR adding image/video/music generation to the existing send pipeline. Provider abstractions are clean, security posture is good (no-shell exec + path-traversal-safe media format), and test coverage via httptest servers + a self-invoking command helper is thorough. 5/5 CI green. Safe to merge.

Review 范围:
- cmd/cc-connect/{main.go,send.go,send_test.go}
- core/{api.go,api_test.go,engine.go,engine_test.go,interfaces.go,media_generation.go,media_generation_test.go}
- config/{config.go,config.example.toml}
- docs/{usage.md,usage.zh-CN.md}, README/{README.md,README.zh-CN.md}, CHANGELOG.md
- 重点关注 correctness、provider abstraction、shell-injection safety、context cancellation、prompt handling、API surface stability、test coverage。

✅ 做得好的地方:
- `CommandMediaGenerator.generate` uses `exec.CommandContext` with no shell — `{{prompt}}` and `{{output}}` are passed as a single argv element, so shell metacharacters in user input cannot escape. The tmpDir + `defer os.RemoveAll` is the right pattern; output filename is constructed from the validated format string.
- `isSafeMediaFormat` enforces `[a-z0-9._-]` only, and `normalizeMediaFormat` falls back to the per-kind default if the user-provided format is unsafe. The `TestNormalizeMediaFormatRejectsPathSeparators` test pins this against `../mp4`. Good defense-in-depth against a config typo leading to arbitrary path write.
- `pollVideoTask` correctly uses `select { case <-ctx.Done(): return }` between polls and resets the timer at the end of each iteration; honors the configured `Timeout` via the outer `context.WithTimeout`. The `Status` switch covers `success`/`completed`/`succeeded` and `fail`/`failed`/`error` — covers the common MiniMax/Seedance/OpenClaw variants.
- `MiniMaxImageGenerator.GenerateImage` accepts FOUR response shapes (`image_urls[]`, `images[]`, `image`, `image_base64`) which is a real-world resilience win; same for the hex-encoded `audio` field on the music endpoint. Falling through to URL download with a separate 5-min `DownloadClient` is correct (image generation can take 60s+ and download is much shorter).
- The OpenAI `response_format` (b64_json/url) vs `output_format` (png/jpeg/webp) distinction is correctly modeled: `response_format` is omitted by default for GPT-image models (test asserts), `output_format` is sent only when configured. The `TestOpenAIImageGenerator_GenerateBase64Image` test pins this behavior.
- `SendToSession` / `SendVideosToSession` / `SendAudiosToSession` integration in `engine.go` cleanly delegates to the existing platform-agnostic send path; MIME types come from `generatedMediaMime` which falls back to `mime.TypeByExtension` and the kind-specific default (e.g., `mp3 → audio/mpeg`).
- `MiniMax auth resolution` is shared via `resolveMiniMaxMediaAuth` for image/video/music, so the same `[minimax.json]` fallback path used by TTS also works here. Avoids the all-too-common "I configured MiniMax for TTS but not for image" footgun.
- Test scaffolding with `os.Args[0]` + `CC_CONNECT_COMMAND_MEDIA_HELPER=1` env var is a clean way to test the command adapter without external dependencies; covers both `{{output}}` file mode and stdout mode.
- `attachment_send = "off"` check (`ErrAttachmentSendDisabled`) is honored for all three new paths — consistent with the existing TTS path.
- `MaxPromptLen` is checked in `utf8.RuneCountInString` (rune count, not bytes), so a 100-rune CJK prompt at 300 bytes isn't rejected by mistake.
- CHANGELOG entry is concise and points at user-visible behavior, not implementation.

🚨/🔴 必须处理:
- 未发现必须阻塞合并的问题。

🟠 建议改进:
- `CommandMediaGenerator` substitutes `{{instrumental}}` as `fmt.Sprintf("%t", opts.Instrumental)` ("true"/"false"). Suno-style CLIs typically expect flags like `--instrumental` (present/absent) or `--instrumental true|false`. Consider whether the args should expand to two argv entries (`"--instrumental"`, value) rather than substituting the token in-place. Low risk — config authors can write their own args template — but worth a note in the docs.
- `MiniMaxImageGenerator.GenerateImage` is ~100 lines; the four response-shape branches are largely identical (decode base64 → return). A small helper `firstNonEmptyImageBytes(...) ([]byte, string, error)` would simplify the body and make adding a 5th shape trivial. Optional cleanup.
- `SendGeneratedImageToSession` etc. silently swallow the `resolveOutboundSessionTarget` error after logging would-be context; the API call does return 500 in that case (good), but the engine-level error path doesn't distinguish "session not found" from "send failure" in the response. Consider a more specific error code or at least structured logging.
- The `time.AfterFunc`-style polling in `pollVideoTask` uses `timer.Reset(interval)` after the switch — this is correct, but a `defer timer.Stop()` is already there, so an alternative `time.NewTicker(interval)` would be slightly more idiomatic. Not blocking.
- `OutputFormat` for OpenAI image generator: docs say `"png"`, `"jpeg"`, or `"webp"`, but the function accepts the value as-is and forwards it to the API. If a user typos `output_format = "web"`, the upstream API will return an error that's surfaced as `openai image API 400: ...`. Consider client-side validation in `NewOpenAIImageGenerator` (similar to `isSafeMediaFormat`).
- The CHANGELOG entry is placed in `### Added` (under v1.3.4 / unreleased), but the bullet is wordy. A more concise: `**Generated media send-back**: --generate-image/--generate-video/--generate-music via configured providers (#1411).` would match the style of the other entries.
- The CommandAdapter test uses `os.Args[0]` as the "command" under test, which is fine on Linux/macOS but on Windows requires special handling; CI is Linux-only so this is fine, but a comment noting "Linux/macOS only" would help future readers.
- The `data:image/png;base64,...` prefix is correctly stripped in `decodeImageBase64`, but if a future provider returns `data:image/svg+xml;base64,...` (SVG is XML, not binary), the decoded bytes will be valid SVG but the format will be reported as "svg" — most platforms reject SVG as an attachment. Not a current issue, but worth flagging.

🔵 可选优化:
- The provider switch in `main.go` has a near-identical structure for image/video/music (4 cases: openai/minimax/command/openclaw). A small generic `wireProvider[T any](name string, enabled bool, newGenerator func() (T, string))` would dedupe; not required for 3 kinds.
- `MiniMaxVideoGenerator.createVideoTask` and `retrieveFileURL` both share the doJSON plumbing but have slightly different body shapes; could share a typed-err wrapping helper. Optional.
- The `Instrumental` and `LyricsOptimizer` fields on `MusicGenerationCfg` use OR semantics with per-call opts (`opts.Instrumental || e.music.Instrumental`). The semantics (per-call overrides default) are reasonable but not documented in the godoc; consider a one-line comment.
- `output_format: "hex"` is hard-coded in the music request body; if MiniMax ever changes the response format, this will silently break. A `ResponseFormat` config field would future-proof.

❓ 需要确认:
- The `provider = "openclaw"` default helper is wired in main.go (`defaultOpenClawMediaCommand`); confirm that an OpenClaw install that *does* have `capability image generate` and `capability video generate` will accept the exact arg layout (`--prompt {{prompt}} --output {{output}}`). If OpenClaw's actual subcommand layout differs, the default won't work and users will silently see "no media returned". Worth a one-time manual smoke test against an OpenClaw install.
- The CommandAdapter example in `config.example.toml` for music shows `suno-generate` as the command name; confirm Suno's official CLI exists with that name, or whether the example is illustrative (a real `suno-api` wrapper script).

Testing / Risk:
- 已看到的验证: 5/5 CI green (lint/unit/smoke/regression/perf). ~30 new unit tests across `core/media_generation_test.go`, `cmd/cc-connect/send_test.go`, `core/api_test.go`. Tests use real `httptest.Server` for all four API surfaces (OpenAI image, MiniMax image, video, music, plus the self-invoking command helper). API integration test `TestHandleSend_AllowsGeneratedImageVideoAndMusic` covers the full send path including attachment dispatch.
- 未覆盖风险: real provider response shape drift (covered by graceful fallthroughs but no integration test), real OpenClaw/Suno CLI behavior, prompt-injection through `{{prompt}}` in command args (mostly mitigated by no-shell, but a real attack would be a provider-specific path traversal in the prompt that the adapter doesn't sanitize). Concurrency safety in `MediaGenerationCfg.MaxPromptLen` and `attachmentSendEnabled` is fine because they are set once at startup.

Next step:
- Merge. Findings are minor; the largest concern (real-provider response shape drift) is already mitigated by the multiple fallthrough branches. After merge, monitor real-world usage for command-adapter issues — that is the highest-risk surface.

AaronZ345:

AaronZ345:
Replacement PR: #1574

PR #1444, opened 2026-06-25
fix: recover Claude Code sessions after truncation

contsun:
## Summary

- detect Claude Code `stop_reason=max_tokens` and automatically continue ordinary truncations in the same session
- detect poisoned Claude Code sessions caused by truncated tool-call JSON (`__unparsedToolInput`, `InputValidationError`) and zero-token empty results
- reset poisoned agent sessions, save a recovery handoff, and inject it once into the next fresh session
- add prompt guidance to reduce oversized replies and long Bash heredocs
- add regression tests for stop reason propagation, poisoned-session detection, reset behavior, and recovery handoff injection

## Why

When Claude Code hits `max_tokens` while emitting a large tool call, the tool-use JSON can be truncated. That leaves the session in a poisoned state: cache reads drop to zero, large prompt prefixes are billed as fresh input, and subsequent turns may return zero-token empty results. This patch separates ordinary truncation (same-session auto-continue) from corrupted sessions (fresh session + recovery handoff).

## Test plan

- `git diff --check`
- `go test ./agent/claudecode -run 'TestHandleResultPropagatesMaxTokensStopReason|TestHandleAssistantUnparsedToolInputMarksNextResultUnhealthy|TestHandleUserInputValidationErrorMarksNextResultUnhealthy|TestHandleResultZeroTokenEmptyMarksUnhealthy' -v`
- `go test ./core -run 'TestRecoveryHandoffInjectedOnceAfterUnhealthyReset|TestProcessInteractiveEvents_AutoContinuesSameSessionOnMaxTokens|TestProcessInteractiveEvents_UnhealthyResultResetsSessionInsteadOfEmptyResponse|TestUnsolicitedReader_UnhealthyResultResetsSession|TestI18n_AllKeysHaveEnglish' -v`

Note: a full `go test ./core` run hit an existing TempDir cleanup flaky in `TestCUJ_A3_ImageReachesAgent`; rerunning that test alone passed.


contsun:
Update: added detection for Claude Code's no-visible-output meta prompt (). This path was not a normal terminal result event, so the previous recovery logic would not trigger. Added  and force-pushed the branch.

contsun:
Update: added detection for Claude Code's no-visible-output meta prompt:

    [Your previous response had no visible output. Please continue and produce a user-visible response.]

This path is emitted as an `isMeta=true` user event, not as a normal terminal `result` event, so the previous recovery logic did not trigger. The branch now marks that event as an unhealthy session, which causes cc-connect to reset the agent session, save a recovery handoff, and notify the user.

Added regression test:

    TestHandleUserNoVisibleOutputMetaMarksSessionUnhealthy

Force-pushed the branch with commit `c5f1d336`.

PR #1445, opened 2026-06-25
feat(yuanbao): add Tencent Yuanbao Bot API platform

skyblue:
## Summary

Add support for Tencent Yuanbao (腾讯元宝) bot platform — a WebSocket-based messaging platform by Tencent. Users can connect their coding agents to Yuanbao through cc-connect.

### New files

- `platform/yuanbao/platform.go` — Yuanbao platform adapter (WebSocket connect, auth, heartbeat, message receive/send, reply heartbeat)
- `platform/yuanbao/proto.go` — Protobuf wire protocol encoding/decoding for Yuanbao gateway
- `platform/yuanbao/sign.go` — Sign-token API client (app_key + app_secret → temporary token)
- `platform/yuanbao/proto_test.go` — Protocol round-trip tests
- `cmd/cc-connect/plugin_platform_yuanbao.go` — Selective compilation build tag
- `cmd/cc-connect/yuanbao.go` — CLI setup command (`cc-connect yuanbao setup --token app_key:app_secret`)

### Changes to existing files

- `config/config.go` — Add `EnsureProjectWithYuanbaoPlatform` / `SaveYuanbaoPlatformCredentials` helper functions
- `config.example.toml` — Document yuanbao platform config example
- `cmd/cc-connect/main.go` — Register `yuanbao` subcommand and update help text
- `web/src/lib/platformMeta.ts` — Add yuanbao to web admin UI platform selector
- `web/src/pages/Projects/ProjectList.tsx` — Add yuanbao platform card
- `web/src/pages/Projects/ProjectDetail.tsx` — Add yuanbao badge

### Key design decisions

- **bot_token** — single config field `app_key:app_secret` (split at `:`), not two separate fields
- **WebSocket** — persistent WSS connection to `bot-wss.yuanbao.tencent.com`, with auto-refreshed sign-tokens
- **init() registration** — registers via `core.RegisterPlatform("yuanbao", New)`, aligned with all other platform packages
- **Reply heartbeat** — sends periodic WS_HEARTBEAT_RUNNING while Reply is in progress (prevents server-side timeout)
- **Dedup** — in-memory message ID dedup with 5-min cleanup cycle

### Bugs fixed during development

1. **Deadlock in startReplyHeartbeat** — `p.mu` lock was held while calling `doReplyHeartbeat` → `getWS()` which re-acquires `p.mu`. Fixed by accessing `p.ws`/`p.botID` directly under the held lock.

2. **Thinking text sent prematurely** — engine's EventThinking handler was flushing accumulated text segments for non-preview platforms before EventResult. Removed early `sendWorkspace` calls so all text is delivered at turn completion.

3. **Inbound image message display** — Only `originalUrl` was read for TIMImageElem; now tries `image_info_array` → `url` → `originalUrl`, matching the TypeScript reference implementation.

4. **Missing msgContent fields (TIMFileElem, TIMCustomElem)** — Added file message and custom message parsing for better compatibility.

5. **Missing `strings` import** — Added import needed by `strings.Join` in `inboundText`.

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -249,6 +249,9 @@ func main() {
 		case "weixin":
 			runWeixin(os.Args[2:])
 			return
+		case "yuanbao":
+			runYuanbao(os.Args[2:])
+			return
 		case "doctor":
 			runDoctor(os.Args[2:])
 			return
@@ -1632,6 +1635,7 @@ Examples:
   cc-connect cron list                List all scheduled tasks
   cc-connect feishu setup             Setup Feishu/Lark bot credentials
   cc-connect weixin setup             Setup Weixin (ilink) with QR or --token
+  cc-connect yuanbao setup            Setup Yuanbao bot with --token app_key:app_secret
   cc-connect update                   Update to the latest version
   cc-connect config format            Format the config file
   cc-connect config example > c.toml  Save example config to a file
--- a/cmd/cc-connect/plugin_platform_yuanbao.go
+++ b/cmd/cc-connect/plugin_platform_yuanbao.go
@@ -1,0 +1,5 @@
+//go:build !no_yuanbao
+
+package main
+
+import _ "github.com/chenhg5/cc-connect/platform/yuanbao"
--- a/cmd/cc-connect/yuanbao.go
+++ b/cmd/cc-connect/yuanbao.go
@@ -1,0 +1,171 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/chenhg5/cc-connect/config"
+	"github.com/chenhg5/cc-connect/platform/yuanbao"
+)
+
+func runYuanbao(args []string) {
+	if len(args) == 0 {
+		printYuanbaoUsage()
+		return
+	}
+
+	switch args[0] {
+	case "setup":
+		runYuanbaoSetup(args[1:], yuanbaoSetupModeAuto)
+	case "bind", "link":
+		runYuanbaoSetup(args[1:], yuanbaoSetupModeBind)
+	case "help", "--help", "-h":
+		printYuanbaoUsage()
+	default:
+		fmt.Fprintf(os.Stderr, "Unknown yuanbao subcommand: %s\n\n", args[0])
+		printYuanbaoUsage()
+		os.Exit(1)
+	}
+}
+
+const (
+	yuanbaoSetupModeAuto = "auto"
+	yuanbaoSetupModeBind = "bind"
+)
+
+func runYuanbaoSetup(args []string, requestedMode string) {
+	fs := flag.NewFlagSet("yuanbao "+requestedMode, flag.ExitOnError)
+	configFile := fs.String("config", "", "path to config file")
+	project := fs.String("project", "", "project name (optional if only one project)")
+	platformIndex := fs.Int("platform-index", 0, "1-based index among yuanbao platforms in the project (0 = first)")
+	token := fs.String("token", "", "Yuanbao bot token (format: app_key:app_secret)")
+	apiDomain := fs.String("api-domain", "", "yuanbao API base URL (default https://bot.yuanbao.tencent.com)")
+	wsURL := fs.String("ws-url", "", "yuanbao WebSocket URL (default wss://bot-wss.yuanbao.tencent.com/wss/connection)")
+	routeEnv := fs.String("route-env", "", "optional X-Route-Env header (e.g. prod, dev)")
+	allowFrom := fs.String("allow-from", "", "restrict access to this user id (comma separated); empty = open")
+	skipVerify := fs.Bool("skip-verify", false, "skip probing the sign-token API before saving")
+	_ = fs.Parse(args)
+
+	initConfigPath(*configFile)
+	if _, err := os.Stat(config.ConfigPath); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: config file not found at %s (%v)\n", config.ConfigPath, err)
+		os.Exit(1)
+	}
+
+	botToken, err := resolveYuanbaoBotToken(requestedMode, *token)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
+
+	targetProject, err := resolveTargetProject(*project)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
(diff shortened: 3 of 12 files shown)


chenhg5 (approved):
结论: Approve

总体判断: A complete, well-tested new platform adapter that follows the existing cc-connect platform conventions. Token manager has inflight dedup, reconnect with bounded exponential backoff, per-chat reply heartbeats, inbound dedup, and proper AllowList gating. CI is 5/5 green. The few findings are non-blocking and can be addressed in follow-ups.

Review 范围:
- cmd/cc-connect/{main.go,yuanbao.go,plugin_platform_yuanbao.go}
- config/config.go (YuanbaoCredentialUpdate flow + raw TOML editing)
- platform/yuanbao/{platform.go,proto.go,sign.go,proto_test.go}
- web/src/lib/platformMeta.ts, web/src/pages/Projects/{ProjectDetail,ProjectList}.tsx
- config.example.toml
- 重点关注 correctness、auth flow、reconnect/dedup、race conditions、test coverage、wire format compatibility。

✅ 做得好的地方:
- Token manager uses an `inflight` channel of size 1 to dedup concurrent fetches; the double-check pattern (lock → check cache → unlock → wait on inflight → re-check cache → fetch → update cache) is correct and well-known.
- `forceRefresh()` is invoked on auth failure (`authFailedCodes` 4001/4002/4003), which is the right mitigation for token rotation or bot kickout.
- Reconnect loop with `min(1000*(1<<attempts), 60000)ms` capped exponential backoff (max 60s) is sane; combined with `shouldReconnect=false` on kickout, it correctly bails when the server has kicked the bot.
- Per-chat reply heartbeats (`replyHeartbeatIntv=2s`, total `replyHeartbeatTimeout=30s`) with proper start/stop, including a `2`-state "finish" heartbeat on stop — this matches the official Yuanbao streaming-reply protocol expectations.
- CLI `setup` command calls `VerifyCredentials` (which calls `fetchToken`) so users see the failure immediately, before saving to config — much better UX than failing on first connection.
- Hand-rolled protobuf-lite is consistent (writeVarint/parseFields/getString/getVarint/getBytes/getRepeatedBytes) and used symmetrically; encode↔decode round-trip tests cover each message type.
- JSON fallback in `decodePush` (try JSON first, fall back to binary proto) is a nice touch for forward compatibility with future schema versions.
- Web admin metadata (`platformMeta.ts`, `ProjectList` color swatch) is added correctly with `bot_token` typed as `password` and `allow_from` in `advanced` group.
- `cli` allows a `--skip-verify` escape hatch for offline/air-gapped setups without weakening the default.

🚨/🔴 必须处理:
- 未发现必须阻塞合并的问题。

🟠 建议改进:
- `platform/yuanbao/sign.go` `getToken`: the `select { case tm.inflight <- struct{}{}: default: tm.inflight <- struct{}{} }` is dead-code — the `default` branch is always runnable, so the non-blocking `case` never executes. Functionally the blocking send still serializes correctly, but the `select` is misleading. Simplify to `tm.inflight <- struct{}{}` (or use a `sync.Mutex`/`singleflight.Group` for clarity).
- `dedupSet` is cleared wholesale every 5 minutes by `dedupCleanupInterval`. This is a hard reset, not a sliding window: a duplicate arriving 4m59s after the first and again 5m01s after the first will pass through. For most bots this is fine, but a comment explaining "best-effort dedup within 5-minute window" would set expectations. Optionally track per-entry expiry and evict lazily.
- `instanceID = 17` is a magic value baked into the auth-bind devInfo payload (field 10). A short comment noting "wire-format constant per Yuanbao protocol; do not change without server coordination" would prevent a well-meaning future PR from "fixing" it.
- `consecutiveHbFails` is incremented without a clear upper bound beyond `heartbeatThreshold=2`; the counter is reset to 0 on a successful heartbeat. Behavior is correct, but a comment would help: "counter only matters for the next `>threshold` check; resets on success".
- `Reply`/`Send` discard errors from `_ = ws.WriteMessage(...)` after slog — this is consistent with other platforms but you might want to surface persistent write failures through `ReplyError` so the agent can retry on a different chat.
- `startReplyHeartbeat` writes the initial heartbeat under `p.mu`, but `p.ws` could in principle be replaced between the lock release and the actual write in `doReplyHeartbeat`. Minor; the readLoop drain handles partial frames.
- `proto_test.go` tests for `parseFields` / `decodeConnMsg` do not exercise a `wireType=3` (start group) / `wireType=4` (end group) case; harmless because the wire format only uses varint/length/32/64, but a `TestUnknownWireType` is already there, so this is mostly FYI.

🔵 可选优化:
- `platform/yuanbao/platform.go` imports `context` but `Reply`/`Send` only use it for the signature; the caller's context cancellation is currently ignored. Consider `select { case <-ctx.Done(): return ctx.Err(); default: }` if a future caller actually wants cancellation.
- The hand-rolled protobuf-lite is large (~600 lines of `proto.go`). If more Tencent-family protocols are added, consider extracting `proto/wire` as a shared package; not needed now.
- The `inbound_message` handler in `handleFrame` calls `p.handleInboundPush` for both `head.cmd == "inbound_message"` and the fallback `len(msg.data) > 0` branch; the latter can theoretically also fire on non-push frames. Consider gating the fallback on `head.cmdType == cmdTypePush` for safety.
- `defaultAPIDomain` / `defaultWSURL` could be configurable via env (`YUANBAO_API_DOMAIN`, `YUANBAO_WS_URL`) for users behind a proxy; not blocking.

❓ 需要确认:
- The `appKey:appSecret` token format is a CLI/UI convention, not a Tencent standard. Confirm with author that the platform sends the secret in the `X-Signature` HMAC only, and never to the WebSocket, even during `auth-bind`. (Looking at the code, the `token` field in `encodeAuthBind` is the *sign-token* returned by `/api/v5/robotLogic/sign-token`, not `appSecret` — so yes, secret is not on the wire. Good.)

Testing / Risk:
- 已看到的验证: 5/5 CI green (lint, unit, smoke, regression, perf). `proto_test.go` covers ~30 unit tests: encode/decode round-trips for head, auth-bind, c2c/group message, push ack, ping, text body, JSON+protobuf push variants, signature determinism/difference, repeated fields, malformed input. No integration test against a real Yuanbao server (none expected for a third-party PR).
- 未覆盖风险: real-world WebSocket lifecycle (intermittent disconnect, server kickout mid-reply, timeouts under load), prod API rate-limit handling, large inbound message with attachments (image/file cards). The code's robustness here is unproven but the structure is similar to other working platforms (feishu/lark), so risk is moderate.

Next step:
- Merge. The findings are minor and orthogonal; the platform's correctness depends mostly on real-world testing once a maintainer has a Yuanbao bot to validate against.

PR #1469, opened 2026-06-30
fix(feishu): AskUserQuestion card layout — markdown list + indexed buttons

happyTonakai:
## Summary

When forwarding an agent's `AskUserQuestion` to Feishu/Lark, each option
was rendered as a `CardListItem` whose Feishu mapping is a 5/12-weighted
column on the left and a button on the right. Long option descriptions
wrapped to 10–20+ narrow lines and were unreadable. The button also
redundantly repeated the option label, which on mobile Feishu (button
width ~2–3 chars) got truncated.

This PR replaces that with a full-width markdown numbered list
(`1. **Label** — Description`) plus a row of equal-width buttons labeled
with the option INDEX (`1`, `2`, `3`, …). The engine resolves
`askq:<qIdx>:<i+1>` back to `q.Options[i].Label` before forwarding, so
the agent never sees the index.

Row sizing is adaptive: ≤5 options render in a single row (clean horizontal
choice bar); >5 cap at 5 per row with a trailing partial row. The 5-cap
fits on mobile width and matches the common case — agent questions rarely
exceed 5 options.

## Type of change

- `TestSendAskQuestionPrompt_CardPlatform_ManyOptions_SplitsButtonRows`
  in `core/engine_test.go` — 7-option case asserts the 5+2 split and
  per-button VALUE/TEXT integrity (catches slice-aliasing regressions
  where prior rows get overwritten by later appends).
- `TestAskqButtonsPerRowFor` in `core/engine_test.go` — table-driven
  coverage over n = 0/1/2/3/4/5/6/7/10/11/100.
- `TestCUJ_I5_AskUserQuestionCardButtonsResolveToLabel` in
  `core/cuj_test.go` — full user journey: agent prompts → user sees
  card → user clicks button → engine resolves `askq:0:2` → agent
  receives the original label `SQLite`.

- Regression test name: `TestSendAskQuestionPrompt_CardPlatform_ManyOptions_SplitsButtonRows`
- The per-button VALUE/TEXT assertion catches the slice-aliasing bug
  where the first row's buttons would be overwritten by later rows'
  data (regression guard for `row = row[:0]` → `row = nil`).
- Manual verification: temporarily reverting to `row = row[:0]` causes
  this test to fail with messages like `row 0 btn 0 Text = "DynamoDB",
  want "PostgreSQL"`.

- [x] **I — UI rendering correctness (cards, streaming, display modes)**

- [x] New CUJ test `TestCUJ_I5` covers the AskUserQuestion card layout
      + click → label resolution flow end-to-end.

Before: each option in a separate row with a squeezed left text column and
a button on the right (the description wrapped to many narrow lines on
long prompts; the button repeated the label and got truncated on mobile).

After: a full-width markdown numbered list at the top of the card, plus a
row of equal-width buttons below labeled with the option index (`1`, `2`,
`3`, …). Up to 5 options render in a single row.

MultiSelect is unchanged (still numbered text list + reply-with-comma
instruction, since buttons can't express multi-select).

Other platforms: Telegram, Discord, QQ Bot, Max, and other
InlineButtonSender paths are unaffected. Plain-text fallback platforms
are unaffected. Only the Feishu/Lark card rendering path changed.

## Implementation notes

- `core/engine.go::sendAskQuestionPrompt()` Feishu card path
- New helper `askqButtonsPerRowFor(n int) int` returns `n` for `n ≤ 5`,
  else `5`.
- Uses `row = nil` (not `row[:0]`) after each emitted button row to
  avoid slice-aliasing data corruption across rows.

- [x] `go build ./...` passes
- [x] `go test ./...` passes (with `-race`)
- [x] i18n strings have all-language translations — N/A

## Related

- Issue: N/A
- Related PR: N/A

### DIFF
--- a/core/cuj_test.go
+++ b/core/cuj_test.go
@@ -1114,16 +1114,18 @@ func TestCUJ_I5_AskUserQuestionCardButtonsResolveToLabel(t *testing.T) {
 		}
 	}
 
-	// Buttons must be split across rows of at most 3 with equal columns,
-	// and labeled with the option INDEX (not the label string).
+	// Buttons must be split across rows with equal columns. With 4 options
+	// the layout is a single row (typical prompts render in one row for a
+	// clean horizontal choice bar; only option sets >5 split into multiple
+	// rows of up to 5 buttons each).
 	var rows []CardActions
 	for _, elem := range card.Elements {
 		if a, ok := elem.(CardActions); ok {
 			rows = append(rows, a)
 		}
 	}
-	if len(rows) != 2 {
-		t.Fatalf("expected 2 button rows for 4 options, got %d", len(rows))
+	if len(rows) != 1 {
+		t.Fatalf("expected 1 button row for 4 options (≤5 in single row), got %d", len(rows))
 	}
 	for i, row := range rows {
 		if row.Layout != CardActionLayoutEqualColumns {
--- a/core/cuj_test.go
+++ b/core/cuj_test.go
@@ -1006,6 +1006,249 @@ func TestCUJ_C3_DefaultModeDenyStopsToolExecution(t *testing.T) {
 }
 
 // ===========================================================================
+// CUJ-I5 · When the agent emits AskUserQuestion with multiple options, the
+// user sees a card with: (1) a full-width markdown numbered list
+// (label — description), and (2) a row of equal-width buttons labeled with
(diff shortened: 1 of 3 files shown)


happyTonakai:
## Before this PR


## After this PR
mobile:

PC:
iShot_2026-06-30_13 15 52

PR #1478, opened 2026-07-01
feat(acp): support model and mode switching

mawag:
## Summary

Adds `/model` and `/mode` support for ACP agents. cc-connect now reads a
model/mode selector list from the ACP agent and switches live on the active
session, covering both selector mechanisms in the ACP ecosystem — the early
`models`/`modes` blocks (`session/set_model` / `session/set_mode`) and the
newer `configOptions` RFD (`session/set_config_option`). Previously ACP
agents could not list or switch models/modes.

## Type of change

- [ ] Documentation only

## Testing

Verified end-to-end against six real ACP agents:

| Agent | Mechanism | models | modes |
|-------|-----------|:------:|:-----:|
| kiro-cli | `models`/`modes` blocks | 15 | 5 |
| mimo | `models`/`modes` blocks | 16 | 3 |
| codebuddy | `models`/`modes` blocks | 15 | 7 |
| hermes | `models`/`modes` blocks | 108 | 3 |
| opencode | `configOptions` | 44 | 5 |
| copilot | `modes` block (mode-only) | — | 3 |

Graceful degradation is covered too: copilot exposes only a mode selector
(no models), so `/model` reports "not supported" while `/mode` works — and its
modeIds are full ACP spec URLs (e.g. `.../session-modes#plan`), which switch
correctly since the matcher makes no assumption about id format.

`go test -race ./agent/acp/` and `go test ./core/` pass. Integration tests are
gated on `CC_ACP_BIN` (skip cleanly in CI) and pick switch targets dynamically
from the live list — no model/mode names are hard-coded.

Unit tests (`agent/acp/models_test.go`, neutral fixtures):
- `TestAgent_reportModels_populatesCache` / `TestAgent_GetModel_NoModels` / `TestAgent_SetModel_PendingWinsOverCurrent` / `TestAgent_GetModel_FallbackToServerCurrent` — model cache & precedence
- `TestAgent_PermissionModes_returnsCachedModes` / `TestAgent_PermissionModes_skipsProbeWhenProbed` / `TestAgent_SharedProbeFlag` / `TestAgent_ensureProbe_noCmdNoSpawn` — probe gating
- `TestSession_absorbModels_*` / `TestSession_matchAvailableModel` / `TestSession_maybeAbsorbCurrentModelUpdate*` — models block path
- `TestSession_absorbConfigOptions_model` / `TestSession_absorbConfigOptions_mode` / `TestSession_maybeAbsorbConfigOptionUpdate` / `TestFlattenModelOptions_flatAndGrouped` — configOptions path
- `TestAgent_ConcurrentModelAccess` — `-race` concurrency guard

Integration tests (`agent/acp/acp_integration_test.go`, gated on `CC_ACP_BIN`):
- `TestIntegration_ACP_ModelSwitch` / `TestIntegration_ACP_ModeSwitch` — full list→switch path against a real agent
- `TestIntegration_ACP_ProbeShortCtx` — locks down that the probe uses its own timeout, not the caller's short 10s ctx

N/A (new feature). Note: `TestIntegration_ACP_ProbeShortCtx` was added to lock
down a probe-timeout regression discovered during development (a slow cold
start exceeding the caller's 10s ctx).

- [x] C — agent execution control (`/mode`)
- [x] F — config switching (`/model`)

- [x] Change is additive: non-ACP agents fall through the existing code path
      unchanged, so existing CUJ behavior is preserved.

For ACP agents that expose model/mode selectors, `/model` and `/mode` now list
the available options and switch live (no session restart), including before
the first session (a one-shot probe fetches the list). Agents exposing neither
report "not supported" as before. Non-ACP agents are unaffected.

- [x] `go build ./core/... ./agent/... ./config/... ./platform/...` passes
      (note: `go build ./...` requires a prebuilt `web/dist`, unrelated to this change)
- [x] `go test ./...` passes for touched packages (`agent/acp`, `core`) incl. `-race`
- [x] `golangci-lint run --new-from-rev origin/main ./...` passes (0 issues)
- [x] No new hardcoded platform/agent names in `core/` (uses the new
      `core.LiveModelSwitcher` capability interface + type assertion)
- [x] No new user-facing strings (reuses existing `/model` `/mode` i18n keys)

## Related

None.

### DIFF
--- a/agent/acp/acp_integration_test.go
+++ b/agent/acp/acp_integration_test.go
@@ -1,0 +1,186 @@
+package acp
+
+import (
+	"context"
+	"os"
+	"strings"
+	"testing"
+	"time"
+)
+
+// acpIntegrationAgent builds an Agent from CC_ACP_BIN / CC_ACP_ARGS, or
+// skips the test when CC_ACP_BIN is unset. CC_ACP_ARGS defaults to "acp"
+// and accepts a comma-separated list. These tests drive a real ACP agent
+// binary for manual end-to-end verification.
+func acpIntegrationAgent(t *testing.T) *Agent {
+	t.Helper()
+	bin := os.Getenv("CC_ACP_BIN")
+	if bin == "" {
+		t.Skip("set CC_ACP_BIN=/path/to/acp-agent (and optional CC_ACP_ARGS) to run ACP integration tests")
+	}
+	argsEnv := os.Getenv("CC_ACP_ARGS")
+	if argsEnv == "" {
+		argsEnv = "acp"
+	}
+	var args []any
+	for _, a := range strings.Split(argsEnv, ",") {
+		if a = strings.TrimSpace(a); a != "" {
+			args = append(args, a)
+		}
+	}
+	ag, err := New(map[string]any{"cmd": bin, "args": args, "work_dir": t.TempDir()})
+	if err != nil {
+		t.Fatalf("New: %v", err)
+	}
+	return ag.(*Agent)
+}
+
+// liveModelSwitcher mirrors core.LiveModelSwitcher (avoids an import cycle
+// concern on the test side).
+type liveModelSwitcher interface {
+	SetLiveModel(model string) bool
+}
+
+// liveModeSwitcher mirrors core.LiveModeSwitcher.
+type liveModeSwitcher interface {
+	SetLiveMode(mode string) bool
+}
+
+// TestIntegration_ACP_ModelSwitch verifies the full model-switch path
+// against a real ACP agent:
+//  1. AvailableModels fetches the model list via probeSessionConfig (from
+//     the `models` block or a configOptions model selector)
+//  2. GetModel reflects the server-reported current model after StartSession
+//  3. SetLiveModel switches live on the active session (session/set_model or
+//     session/set_config_option, depending on the agent's mechanism)
+//
+// Verified agents: kiro-cli, mimo, codebuddy (--acp), hermes (models block)
+// and opencode (configOptions).
+func TestIntegration_ACP_ModelSwitch(t *testing.T) {
+	a := acpIntegrationAgent(t)
+
+	// 1. AvailableModels probes when no session is active.
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+	models := a.AvailableModels(ctx)
+	if len(models) == 0 {
+		t.Skip("agent advertises no models — nothing to switch")
+	}
+	t.Logf("AvailableModels: %d models, first=%s", len(models), models[0].Name)
+
+	// 2. StartSession + GetModel.
+	sess, err := a.StartSession(ctx, "")
+	if err != nil {
+		t.Fatalf("StartSession: %v", err)
+	}
+	defer func() { _ = sess.Close() }()
+
+	cur := a.GetModel()
+	t.Logf("current model after StartSession: %q", cur)
+	if cur == "" {
+		t.Fatal("GetModel() empty after StartSession — models block not parsed from session/new")
+	}
+
+	// 3. SetLiveModel to a different model.
+	lm, ok := sess.(liveModelSwitcher)
+	if !ok {
+		t.Fatal("session does not implement LiveModelSwitcher")
+	}
+	var target string
+	for _, m := range models {
+		if m.Name != cur {
+			target = m.Name
+			break
+		}
+	}
+	if target == "" {
+		t.Skip("only one model available, cannot test switch")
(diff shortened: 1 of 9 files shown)


82030615:
Thanks for the thorough PR — the dual-mechanism coverage (`models`/`modes` blocks **and** the newer `configOptions` RFD), the graceful degradation for mode-only agents like copilot, and the `CC_ACP_BIN`-gated integration tests that pick switch targets dynamically are all genuinely good engineering. The capability-interface approach (`core.LiveModelSwitcher`) keeps `core/` agnostic, which is exactly right per AGENTS.md.

That said, I did an adversarial read of the diff and found one **blocking** issue plus several quality concerns.

## P1 — `pendingModel` is a dead end (must fix before merge)

`Agent.SetModel` (agent.go:327) stores `a.pendingModel`, and `Agent.GetModel` (agent.go:343) *optimistically prefers it forever*. But `acpSessionConfig` only carries `initialMode` — there is **no `initialModel`**, and `StartSession` never reads `GetModel()` to apply a pre-session model.

Trace the broken path:
1. User runs `/model X` with **no active session** → engine `cmdModel` calls `agent.SetModel("X")` (sets `pendingModel`), then sees `liveSwitched == false` (no live session) → `cleanupInteractiveState`.
2. Next message spawns a fresh `StartSession`. Nothing passes `pendingModel`/`initialModel`, so the new session starts on the **server-default model**.
3. `Agent.GetModel()` *still returns `"X"`* (pendingModel is never cleared on a successful-ish session start), while the actual session runs on the default.

Net effect: the IM user is told the model is `X`, but the agent is silently on the server default. This is the same class of silent-state-desync bug the `/mode` path avoids by threading `initialMode` through `StartSession`.

**Fix (mirror the existing `initialMode` path exactly):**
- Add `initialModel string` to `acpSessionConfig`.
- In `StartSession`, pass `a.GetModel()` into the config (same place `initialMode` is sourced).
- After `handshake` (session/new) succeeds, if `cfg.initialModel != ""`, call `s.SetLiveModel(cfg.initialModel)` — same guard/early-return pattern already used for `SetLiveMode`.
- On failure of that live switch, fall through to the existing restart path so behavior is consistent with `/mode`.

Without this, the feature's primary promise ("switch live, *including before the first session*", per your own Manual-behavior section) does not actually hold for models.

## P2 — quality concerns worth addressing

- **Probe ignores caller context.** `AvailableModels`/`PermissionModes` spawn a real probe process with `context.Background()` (the 10s `fetchCtx` is internal to the probe, not the caller's). A slow cold start can block read paths (e.g. the settings/selector page) for up to the probe timeout. `TestIntegration_ACP_ProbeShortCtx` locks down the *probe's* timeout but not the fact that the caller's ctx is dropped — callers can't cancel. Prefer threading the caller ctx and bounding the probe off it.
- **`GetModel` over-prefers `pendingModel`.** Because `pendingModel` is never cleared once set, `GetModel` will keep returning a *requested-but-never-applied* model even after the server reports a different current model via `current_model_update`. This masks the desync instead of surfacing it. Consider clearing `pendingModel` once `SetLiveModel`/`initialModel` application succeeds, or having `GetModel` prefer `currentModel` when a live session exists.
- **`sessionConfigProbed` is check-then-act, not atomic.** `ensureProbe` reads the flag, then spawns; concurrent first calls can double-probe (two processes) and there's no backoff/retry if the first probe's agent binary is momentarily unavailable. Use `sync.Once` or a proper `atomic`+mutex guard.
- **Duplicated handshake + precedence logic.** `probeSessionConfig` re-implements the `configOptions`-first-then-legacy-blocks precedence that `handshake` already encodes. Two copies will drift. Consider having the probe reuse the same absorb path.
- **`flattenModelOptions` / `flattenModeOptions` are near-identical.** Same flat/grouped walk, different value type. A single generic (or at least shared helper) would remove the copy.

## P3 — minor

- The `var _ core.ModelSwitcher = (*Agent)(nil)`-style compile-time assertion for `LiveModelSwitcher` lives only in the *test* file; put it in `agent.go` so the contract is enforced at build time in the non-test binary.
- In `flattenModelOptions`, the grouped branch drops `o.Group` (the human-readable group label) — the picker label becomes the raw model id. Minor UX loss; consider surfacing the group as a sub-label.

## Security

Clean. `SetLiveModel` validates against the cached allowlist via `matchAvailableModel` *before* any RPC, so a user cannot inject an arbitrary `modelId`. No secrets in source. Good.

**Recommendation:** REQUEST_CHANGES on the P1 alone — the rest can follow. The P1 fix is small and mirrors code you've already written for `initialMode`, so it should be a quick add. Happy to review a follow-up commit.

Issue #1475, opened 2026-07-01
[Feature] QQ OneBot: add group_reply_all option (parity with Feishu, Discord, Telegram)


ccchenyulin:
**cc-connect Version:** v1.4.1
**Platform:** QQ (OneBot v11 / NapCat)

---

### Summary

The QQ OneBot platform (`type = "qq"`) is missing the `group_reply_all` config option that all other group-capable platforms already support. As a result, the bot responds to **every** group message without requiring an `@mention`, which is unwanted when using a personal QQ account in a shared group.

### Current Behavior

`platform/qq/qq.go` — `handleMessage()` passes all group messages to the agent after `isAllowed` filtering. There is no at-mention gate.

`config.example.toml` — the `[projects.platforms.options]` block for `type = "qq"` has no `group_reply_all` field.

### Expected Behavior

```toml
[[projects.platforms]]
type = "qq"
[projects.platforms.options]
ws_url        = "ws://127.0.0.1:3001"
group_reply_all = false   # default: only respond when @mentioned
```

When `group_reply_all = false` (the desired default), the bot should only respond to group messages that contain an `@` mention of the bot's own QQ number (i.e. `[CQ:at,qq=<bot_qq_id>]` in the OneBot message chain).

When `group_reply_all = true`, behavior stays as today — respond to all group messages.

### Precedent in the Codebase

All other group-capable platforms already implement this flag:

| Platform | Config key |
|---|---|
| Feishu | `group_reply_all` |
| Discord | `group_reply_all` |
| Telegram | `group_reply_all` |
| DingTalk | `group_reply_all` |
| **QQ OneBot** | ❌ missing |

### Implementation Notes

1. Add `GroupReplyAll bool` to the QQ platform options struct.
2. In `handleMessage()`, when the event type is group_message and `GroupReplyAll == false`, check whether the message chain contains `[CQ:at,qq=<self_id>]` (or the JSON equivalent `{"type":"at","data":{"qq":"<self_id>"}}`). If the bot is not mentioned, skip the message.
3. Update `config.example.toml` to document the new option under the QQ platform block.
4. Default should be `false` so new installs don't respond to all group messages unexpectedly — this is a **breaking change** for current QQ group users, which should be noted in the changelog.

### Additional Context

Tested: no existing GitHub issues cover this. Personal QQ accounts used as bots are commonly added to group chats; without at-mention gating the bot becomes disruptive in busy groups.

chenhg5:
[...]

PR #1461, opened 2026-06-28
fix(kimi): conditionally pass --print so newer Kimi Code CLI works (#1456)

chenhg5:
Closes #1456.

## Problem

cc-connect unconditionally passes `--print --output-format stream-json --prompt …` to the Kimi CLI. The newer Kimi Code CLI removed the standalone `--print` flag and exits immediately with:

```
error: unknown option '--print' (Did you mean --prompt?)
```

Every Kimi chat is broken for users on the new CLI.

## Root Cause

The two Kimi CLIs in the wild have opposing requirements for non-interactive runs:

| CLI                  | `--print`? | `--output-format` requires |
|----------------------|------------|-----------------------------|
| legacy `kimi-cli`    | yes        | `--print`                   |
| new Kimi Code CLI    | removed    | `--prompt`                  |

A fixed arg list cannot satisfy both, and unlike `--quiet` (#806 / PR #1253) `--print` cannot be emulated client-side — it is the CLI's own non-interactive mode toggle.

## Fix

Probe `kimi --help` once at `Agent` construction and cache the detected flag set. `buildArgs()` (newly extracted from `Send()` so it is unit-testable) only emits `--print` when the installed binary still advertises it. Probe failure / timeout falls back to *no* `--print`, which matches the direction Kimi CLI is moving and avoids the hard-failure mode of the current code.

- `agent/kimi/probe.go` — `parseKimiHelpFlags()` pure parser + bounded `probeKimiFlags()` (5s timeout, swallows errors).
- `agent/kimi/kimi.go` — `Agent` caches `kimiFlagSupport`, threads it into `newKimiSession`; doc comments updated.
- `agent/kimi/session.go` — `kimiSession.flagSupport` + `buildArgs()` helper; conditional `--print`; permission-mode comment updated to cover both legacy (`--print` implicit `--yolo`) and modern (`--prompt` auto-permission) behaviors.

## Tests

Regression + coverage:

- `TestBuildArgs_NoPrintSupportOmitsPrintFlag` — **regression test for #1456**; fails on pre-fix code, passes on this branch.
- `TestBuildArgs_PrintSupportIncludesPrintFlag` — legacy CLI branch.
- `TestBuildArgs_PlanMode` / `TestBuildArgs_ResumeSession` — other arg paths.
- `TestParseKimiHelpFlags_LegacyAdvertisesPrint` / `TestParseKimiHelpFlags_ModernHidesPrint` — parser handles both typer box-drawing and standard click \`-p, --prompt\` layouts.
[...]

PR #1483, opened 2026-07-01
fix(kimi): gate --work-dir flag on probe, like --print (#1476)

### LINKED ISSUE #1476: [Bug] ❌ 错误: error: unknown option '--work-dir'
### cc-connect Version

v1.4.1

### Operating System

Other

### Agent Type

Other

### Platform

N/A

### Installation Method

### Bug Description

之前因为错误所以去用了Claude Code，有了新版本，测试新版本用Kimi Code有了新错误：❌ 错误: error: unknown option '--work-dir'

### Steps to Reproduce

Image

### Expected Behavior

正常使用，无错误

### Actual Behavior

Image

```toml

```

### Logs / Error Output

```shell

```

### Additional Context

_No response_

chenhg5:
Fixes #1476

## 问题

Kimi Code CLI builds newer than what #1456 covered dropped the `--work-dir`
flag. Passing it now produces:

```
error: unknown option --work-dir
```

The reporter (`#1476`, `v1.4.1`, Kimi Code agent) hits this on every spawn
because `session.go:97-99` always appended the flag whenever a non-default
`workDir` was set. The #1461 / #1456 fix gated `--print` the same way; this
PR closes the gap for `--work-dir`.

## 改动

Follows the **exact same probe-based pattern** as #1456 / PR #1461 (referenced
for both users and reviewers who already validated that approach).

### `agent/kimi/probe.go`
- `kimiFlagSupport` gains `WorkDir bool`.
- `probeKimiFlags` populates it from the existing `parseKimiHelpFlags` map:
  ```go
  support := kimiFlagSupport{
      Print:   flags["--print"],
      WorkDir: flags["--work-dir"],
  }
  ```
- Doc comment extended to reference both #1456 and #1476.

### `agent/kimi/session.go`
- One-line gate:
  ```go
  if ks.workDir != "" && ks.flagSupport.WorkDir {
      args = append(args, "--work-dir", ks.workDir)
  }
  ```
- The agent still runs in the correct directory because `exec.Command.Dir`
  is set on the spawned `*exec.Cmd` (downstream of this line) — same as in
  the #1461 fix.

### `agent/kimi/probe_test.go`
- `TestParseKimiHelpFlags_LegacyAdvertisesPrint` now also asserts
  `flags["--work-dir"]` is `true` (legacy CLI supports it).
- New constant `modernKimiHelpWithoutWorkDir` mirrors the Kimi Code CLI
  build the reporter hits (no `--print`, no `--work-dir`).
- New `TestParseKimiHelpFlags_ModernWithoutWorkDir` asserts neither flag
  is detected on the deeper-no-work-dir surface.
- Existing `TestParseKimiHelpFlags_ModernHidesPrint` left intact (it
  covers the in-between surface where `--print` is gone but `--work-dir`
  still exists — useful regression coverage).
- New `TestKimiFlagSupport_LegacyHelpSetsWorkDir` +
  `TestKimiFlagSupport_ModernWithoutWorkDir` exercise the full probe-
  mapping, not just `parseKimiHelpFlags`, since that's what
  `buildArgs` actually consumes.

### `agent/kimi/session_test.go`
- `TestBuildArgs_WorkDirFlagGated` — `workDir=/`, `flagSupport{WorkDir:false}`
  → `--work-dir` MUST NOT appear, AND the value `/` MUST NOT leak into args
  (catches a partial-gate bug where only the flag name is dropped).
- `TestBuildArgs_WorkDirFlagEmitted` — `workDir=/`, `flagSupport{WorkDir:true}`
  → `--work-dir /` MUST appear (legacy CLI keeps non-default dir support).

## 触发条件 / trade-off

- **Why reuse `parseKimiHelpFlags` instead of adding a new parser?**
  `--work-dir` already appears in the help-text of both legacy and the
  in-between build the existing `modernKimiHelp` constant models. The
  existing parser picks it up the same way it picked up `--print` — no
  new parser logic, no new layout assumptions.
- **Why a separate `modernKimiHelpWithoutWorkDir` instead of editing
  `modernKimiHelp`?** The `modernKimiHelp` constant is anchored in the
  #1456 fix history; editing it would silently invalidate that test's
  assertion contract. Two-realistic-surfaces is the cleaner modeling.
- **Risk for users on default dirs?** None — `cmd.Dir` is set
  unconditionally, so whether the CLI sees `--work-dir $X` or just takes
  cwd from `cmd.Dir`, the agent runs in the same directory. The only
  practical difference is users on non-default dirs whose modern CLI
  previously crashed: they now get clean startup (cwd fallback).
- **Risk for users with legacy kimi-cli?** None — they advertise
  `--work-dir` in help, the probe sees it, `buildArgs` still emits it.

## 验证

- `gofmt -l agent/kimi/` — clean
- `go vet ./agent/kimi/...` — clean
- `go test -count=1 ./agent/kimi/...` — all pass (4 prior + 6 new =
  10 tests in the probe/session suites)

## 不做什么

- 不改 Kimi Code CLI 本身
- 不改其他 agent adapter (Claude / Codex / Gemini / Cursor / OpenCode
  etc. each have their own flag handling; out of scope)
- 不改 probe 探测逻辑本身 (沿用现有 `kimi --help` 解析 + 短超时
  fallback to modern surface on probe failure)
- 不重构 `buildArgs` / `kimiFlagSupport` — 仅加一个 bool 字段

## 关联

- 同 template 参考: PR #1461 (--print gating, #1456)
- 同 reporter: issue #1475 (`group_reply_all` for QQ platform) 是独立
  PR，本次不合并。


chenhg5 (commented):
Conclusion: Comment

Overall assessment:
- Solid, minimal fix that closes the `--work-dir` gap for the Kimi Code CLI build (#1476) using the exact same probe-gating pattern already validated by PR #1461 / #1456 for `--print`. Single commit on `origin/main`, CI fully green, local `go test -race ./agent/kimi/...` passes 10/10. Mergeable as-is.

Review scope:
- Reviewed the full diff (5 files, +124 / -6): `agent/kimi/probe.go`, `agent/kimi/session.go`, `agent/kimi/probe_test.go`, `agent/kimi/session_test.go`, `CHANGELOG.md`.
- Focused on correctness (the gate and the probe), test coverage (both supported and unsupported CLI surfaces), backward compatibility, and stacking vs PR #1564.

✅ What looks good:
- **Pattern consistency with #1461**: `if ks.workDir != "" && ks.flagSupport.WorkDir { ... }` mirrors the existing `--print` gate at `session.go:79-81` exactly. Reviewers who already approved #1461 can verify this in 30 seconds.
- **`exec.Command.Dir` is set unconditionally** at `session.go:187`, so omitting `--work-dir` on modern CLI is functionally equivalent for users on default dirs and a graceful cwd fallback for users on non-default dirs. Author's claim is verified by reading the code path.
- **Two-realistic-surfaces approach** in tests: kept the existing `modernKimiHelp` (transition build: no `--print`, still has `--work-dir`) and added a separate `modernKimiHelpWithoutWorkDir` fixture for the #1476 surface. This preserves the #1456 regression contract instead of mutating an anchored fixture.
- **Partial-gate future bug catch**: `TestBuildArgs_WorkDirFlagGated` asserts the value `/` doesn't leak into args when the gate is closed. A future refactor that drops the flag-name but keeps the value would be caught.
- **Probe-failure contract preserved**: zero-value `kimiFlagSupport{}` on probe failure remains the conservative "modern CLI" default — unchanged from #1461.
- **No scope creep**: only `agent/kimi/` is touched (plus one CHANGELOG line); no other agent adapters, no platform, no core.
- **CHANGELOG entry** is informative: names the issue, references prior art, explains the cwd fallback, gives one-line legacy/modern outcomes.

🚨/🔴 Must fix:

🟠 Should improve:
- None.

🔵 Optional:
- **Add a `buildArgs`-level test for the in-between surface** (`Print=false, WorkDir=true`). The parser is tested at that surface via `TestParseKimiHelpFlags_ModernHidesPrint`, but a `TestBuildArgs_ModernWithWorkDir` (or extending `TestBuildArgs_PlanMode`) would explicitly assert the gate fires when `WorkDir=true` while `Print=false`. Not needed for correctness today, but a 6-line addition that locks in the integration. Follow-up only.

❓ Questions:
- **Maintainer decision (not for the author)**: this PR is the **base commit** of PR #1564 (`feat(kimi): native Kimi Code CLI dialect support`). #1564's first commit (`e1ec734f`) is byte-identical to this PR's HEAD, and #1564's second commit (`8dc7707c`) likely consumes the `kimiFlagSupport.WorkDir` field added here — meaning **#1564 will not compile standalone without #1483 merged first**. Recommendation: merge #1483 first, then ask `whyihaveyou` (#1564 author, first-time contributor) to rebase #1564 onto the new `main` so its diff reflects only the actual Kimi Code dialect work, not the stacked base.

Testing / Risk:
- Verified locally:
  - `go test -race ./agent/kimi/... -count=1 -timeout 60s` → ALL PASS (10/10 tests; 6 new in this PR).
  - `go vet ./agent/kimi/...` clean.
  - `gofmt -l agent/kimi/` clean.
- Verified parser correctness manually against both fixtures: `legacyKimiHelp` → `--work-dir:true`, `modernKimiHelpWithoutWorkDir` → `--work-dir:absent`.
- CI: lint, unit-test, smoke-test, regression-test, performance-test all SUCCESS at `e1ec734f` (`runs/28550372464`).
- Remaining risk: if a future Kimi CLI build keeps `--work-dir` but moves it to a sub-command or aliases it differently, the parser might miss it. The existing two-realistic-surfaces fixtures cover both "in-between" and "fully removed" transitions; future moves would need a new fixture.
- Backward compatibility: zero regression risk — legacy users keep `--work-dir`, modern users get clean startup, in-between users keep `--work-dir` (parser still detects it).

Next step:
- **For the maintainer**: merge #1483 first to unblock #1564. Then ask `whyihaveyou` to rebase #1564 onto the new `main` for a cleaner review surface.
- **For the author (chenhg5, self-PR)**: nothing required. The P3 follow-up (in-between-surface `buildArgs` test) can be a separate PR if desired.

PR #1496, opened 2026-07-05

AaronZ345:
## Summary

## Root cause

## Test Plan
- `GOOS=linux GOARCH=amd64 go test -c -tags no_web ./cmd/cc-connect`
- `git diff --check`

Supersedes #1349. This is the same rebased patch on a fresh PR branch.


AaronZ345:

AaronZ345:
Superseded by #1651, which carries the same patch rebased onto the latest main on a fresh branch.

PR #1497, opened 2026-07-05

### LINKED ISSUE #1077: [Bug] new_per_run cron sessions remain visible in /list after completion
## Bug Description

## Environment

- Agent: Codex

Example config/state:

```toml
[projects.agent]
type = "codex"

[projects.agent.options]
```

```json
{
  "cron_expr": "0 9 * * 1-5",
  "enabled": true,
  "silent": true,
  "timeout_mins": 45
}
```

## Steps to Reproduce

## Actual Behavior

## Expected Behavior

Possible acceptable behavior:

## Root Cause Notes

## Impact

AaronZ345:
## Summary

Fixes #1077.

## Tests
- `go test ./core`

Supersedes #1417. This is the same rebased patch on a fresh PR branch, including the requested compatibility opt-out.


AaronZ345:

AaronZ345:
Superseded by #1650, which carries the same patch rebased onto the latest main on a fresh branch.

PR #1321, opened 2026-06-12
fix(core): throttle message recall fallback probes

qvictl:
## Problem

In Feishu/Lark deployments, cc-connect can continuously call the Feishu OpenAPI endpoint below while an interactive agent turn is still active:

```text
GET /open-apis/im/v1/messages/{message_id}?user_id_type=open_id
```

This happens through the optional `MessageRecallDetector` fallback path. While a message is being processed, core starts a message-recall monitor and calls `IsMessageRecalled` on the platform. The Feishu implementation checks recall/deletion state by fetching the original message with `client.Im.Message.Get`.

Before this change, the fallback monitor polled every 2 seconds for the same active message. The busy-session path could also trigger the same fallback check when new user messages arrived while the previous turn was still running. For a long-running or stuck agent turn, a single Feishu session could therefore keep issuing `GET /im/v1/messages/{message_id}` requests indefinitely.

This is expensive enough to exhaust Feishu's free OpenAPI quota quickly: one request every 2 seconds is about 1.3 million requests over 30 days for just one active/stuck session, exceeding a 1 million monthly free quota even without heavy chat traffic.

## Observed / reproducible scenario

A representative affected deployment is:

```toml
platform = "feishu"
enable_feishu_card = true
progress_style = "card"
```

Then:

1. A user sends a message to a Feishu-backed cc-connect session.
2. The agent turn remains active for a long time, hangs, or otherwise does not reach normal completion.
3. cc-connect starts the recall fallback monitor for that active message.
4. Feishu `IsMessageRecalled` repeatedly calls `GET /im/v1/messages/{message_id}` for the same original message.
5. The Feishu app shows high OpenAPI usage even though there is little or no new chat activity.

The root issue is in core, not in Feishu-specific retry behavior: the same active message can be probed repeatedly by multiple fallback triggers.

## Fix

- Increase the fallback recall monitor interval from 2 seconds to 1 minute.
- Add per-active-message fallback probe state to `interactiveState`:
  - remember the last probed message ID and timestamp;
  - skip repeated probes for the same message during a cooldown window;
  - prevent concurrent monitor/busy-session triggers from probing the same message at the same time.
- Reset probe state when a new active message starts, so recall detection still works for new turns.

This preserves recall detection while preventing one active/stuck Feishu turn from continuously burning OpenAPI quota.

## Test plan

- [x] go test ./core
- [x] go test ./agent/... ./config ./core ./daemon ./platform/... ./tests/...
[...]

PR #1500, opened 2026-07-05
L-0273: cherry-pick 8 upstream reliability fixes

JayGarland:
## Summary
- Cherry-picks 7 of 8 upstream (chenhg5/cc-connect) reliability fixes requested in L-0273: race fix (Send vs cleanupInteractiveState), drainPendingMessages alignment, absolute file-ref paths, codex app-server write timeout, stream preview resume after permission prompt, hide agent footer lines, per-session idle timeout for live agents.
- `e739e2eb` (throttle message recall fallback probes, #1321) was **skipped**: its content is already on `main` via `b107669b`, a prior sync of the same upstream PR — cherry-pick produced an empty diff and was skipped rather than committed as a no-op.
- Two adaptation commits on top:
  - `test: fix processInteractiveEvents call sites after cherry-pick` — two new tests from the footer-hiding commit didn't pass this fork's `dropReply bool` parameter (added independently of upstream); `go vet` caught the arity mismatch.
  - `fix: recognize POSIX-style absolute paths in AppendFileRefs on Windows` — `filepath.IsAbs` is OS-native, so the #1459 cherry-pick's own new tests (`TestAppendFileRefs_Absolutize*`) failed on Windows because `/tmp/...`-style already-absolute paths got rewritten with a drive prefix instead of passed through. Fixed `AppendFileRefs` (and the test's own assertion) to also treat a leading `/` as absolute, matching the function's documented passthrough contract. This fork runs in production on Windows, so this is a real fix, not a workaround.
- Idle timeout (`760079bc`) merged cleanly alongside the existing ping-only heartbeat mode — confirmed the two mechanisms are orthogonal (`agentSessionIdleTimeoutNanos` vs `ExecuteHeartbeatPing`), no field collision.

## Conflict resolution notes
- `core/engine.go` (`a4b46593`): real conflict — our fork's `currentPromptLen`/`currentPromptPreview` state fields vs upstream's `as := state.agentSession` race-fix capture landing at the same location twice (once via clean auto-merge, once via the conflict hunk). Resolved by keeping our fields and de-duplicating the capture to one copy (duplicate `as :=` would have been a compile error).
- `core/engine.go` (`e739e2eb`): trivial struct-field-ordering conflict; both sides' fields kept.
- `core/streaming_test.go` (`230ee3c6`): both sides added a differently-named freeze/unfreeze regression test at the same insertion point; kept both (`TestStreamPreview_FreezeAndRecreate` + `TestStreamPreview_UnfreezeResumesStreaming` + 2 more upstream tests). All streaming tests pass.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./core/... ./daemon/...` — 8 failures remain, all confirmed pre-existing on unmodified `main` (Windows/environment-specific: home-path shortening, PowerShell shell-output format, nexus skill-dir fixture, path-separator normalization, schtasks file-mode simulation) — none introduced by this sync
- [x] `go test ./...` full suite — all platform adapter packages pass


JayGarland:
Closing — this was created against the wrong repo by mistake (should target our fork JayGarland/cc-connect, not upstream). Apologies for the noise.

PR #865, opened 2026-05-06
feat(agent): add GitHub Copilot as a first-class agent

Marvae:
## Summary

- Add `agent/copilot/`, a native GitHub Copilot agent that drives the local `copilot --headless --stdio --no-auto-update` CLI over Content-Length framed JSON-RPC.
- Support streaming responses, permission requests, session create/resume/list/delete, model/mode switching, and doctor metadata.
- Register Copilot in selective builds, config examples, docs, and the web project agent selector.

Closes #827

## Why first-class Copilot instead of only ACP

cc-connect can already run Copilot through the generic ACP adapter, and that path remains documented. This native agent follows the same first-class pattern as Devin/Kimi/Pi: users can configure `type = "copilot"` directly, logs and session labels identify the agent as Copilot, and cc-connect can wire Copilot-specific JSON-RPC session and permission flows without requiring users to hand-roll an ACP command block.

The default flow reuses local `copilot login` state. cc-connect does not require a GitHub token in config.

## Implementation notes

- `agent/copilot/copilot.go`: agent registration, options, providers, session list/delete probes, and optional interfaces.
- `agent/copilot/session.go`: long-running Copilot JSON-RPC session, event parsing, permission handling, attachments, and shutdown.
- `agent/copilot/jsonrpc.go`: Content-Length framed JSON-RPC reader/writer.
- `agent/copilot/proc_unix.go` / `proc_windows.go`: platform-specific process group cleanup.
- Tests cover JSON-RPC framing, current Copilot event shapes, permission request variants, session list/delete probes, race safety, Windows compilation, and a gated real Copilot integration flow.

## Test Plan

- `go test -count=1 ./agent/copilot`
- `go test -race -count=1 ./agent/copilot`
- `COPILOT_INTEGRATION=1 go test -count=1 ./agent/copilot -run TestIntegration_CopilotFlow -v`
- `go build -tags no_copilot ./cmd/cc-connect`
- `GOOS=windows GOARCH=amd64 go test -c -o /private/tmp/copilot.test.exe ./agent/copilot`

## Local configuration

```toml
[[projects]]
name = "my-copilot-project"

[projects.agent]
type = "copilot"

[projects.agent.options]
work_dir = "/path/to/code"
mode = "default"
cli_path = "copilot"
```

Prereqs: install GitHub Copilot CLI and run `copilot login` once.

PR #1512, opened 2026-07-08
fix(agent/copilot): map reasoning deltas to EventThinking

zhengweixing:
## Summary

When bridging GitHub Copilot CLI through cc-connect to a chat platform (e.g. Feishu),
Copilot's thinking/reasoning content leaks into the answer text and `thinking_messages=false`
cannot suppress it. This PR maps Copilot reasoning events to `core.EventThinking` so the
existing thinking-display toggle works.

## Root cause

Copilot CLI streams reasoning as two distinct `session.event` types (captured by probing
`copilot --stdio` directly):

- `assistant.reasoning_delta` (streaming): `{"reasoningId":"...","deltaContent":"..."}`
- `assistant.reasoning` (final aggregated): `{"reasoningId":"...","content":"..."}`

The copilot adapter (`agent/copilot/session.go`) had **no case** for either — they fell into
the `default` branch and were dropped. Because reasoning was never tagged `EventThinking`,
`thinking_messages=false` (which only filters `EventThinking`) had no effect. Since the
`deltaContent` field name matches ordinary message deltas, reasoning could also be misrouted
as answer text and concatenated into the reply.

Confirmed absent on upstream `main` (`a23fb90`) and base commit (`e2bbe5f`): `assistant.reasoning`
occurs 0 times in `agent/copilot/session.go` before this PR.

## Type of change

## Fix

- `handleSessionEvent`: add `assistant.reasoning_delta` and `assistant.reasoning` cases,
  both emitting `core.EventThinking` (reusing `copilotEventText`, which already reads
  `deltaContent`/`content`).
- `normalizeCopilotEventType`: normalize `assistant_reasoning` / `assistant_reasoning_delta`
  underscore variants.

- `TestHandleSessionEvent_ReasoningDelta` in `agent/copilot/session_test.go` — asserts
  `assistant.reasoning_delta` with `deltaContent` maps to `core.EventThinking`.
- `TestHandleSessionEvent_ReasoningFinal` in `agent/copilot/session_test.go` — asserts
  `assistant_reasoning` (final, `content`) maps to `core.EventThinking`.

- Regression test name: `TestHandleSessionEvent_ReasoningDelta`
  - Without the fix, `assistant.reasoning_delta` hits the `default` branch (no event emitted),
    so the test fails on pre-fix code and passes on the fixed code (verified by code inspection
    of the `default` branch on `e2bbe5f`).

- [x] A — basic conversation

With `thinking_messages=false` (e.g. `CC_CONNECT_DISPLAY_THINKING_MESSAGES=false`), Copilot
reasoning no longer appears in the chat reply. With it enabled, reasoning is shown as a
separate thinking preview rather than mixed inline with the answer.

- [x] `go build ./...` passes
- [x] `go test ./...` passes (copilot package; concurrency untouched, so no `-race` needed)
- [x] i18n strings have all-language translations (N/A — no new user-facing text)

## Related

- Copilot adapter originally added in #865.

### DIFF
--- a/agent/acp/mapping.go
+++ b/agent/acp/mapping.go
@@ -232,7 +232,7 @@ func extractToolCallContentText(blocks []struct {
 func mapSessionUpdateFallback(sessionID string, kind string, update json.RawMessage) []core.Event {
 	// Some agents may send reasoning as a dedicated discriminator; map to EventThinking.
 	switch strings.ToLower(kind) {
-	case "reasoning", "reasoning_chunk", "thinking", "agent_thinking_chunk":
+	case "reasoning", "reasoning_chunk", "thinking", "agent_thinking_chunk", "agent_thought_chunk":
 		var u struct {
 			Content struct {
 				Type string `json:"type"`
--- a/agent/acp/mapping_test.go
+++ b/agent/acp/mapping_test.go
@@ -179,6 +179,23 @@ func TestMapSessionUpdate_reasoningChunk(t *testing.T) {
 	}
 }
 
+// TestMapSessionUpdate_agentThoughtChunk covers the discriminator Copilot CLI uses
+// for reasoning/thinking content in ACP mode. Without this case the chunk would
+// hit mapSessionUpdateFallback's default branch and leak as EventText.
+func TestMapSessionUpdate_agentThoughtChunk(t *testing.T) {
+	params := json.RawMessage(`{
+		"sessionId": "s1",
+		"update": {
+			"sessionUpdate": "agent_thought_chunk",
+			"content": {"type": "text", "text": "The user needs to check health"}
+		}
+	}`)
+	evs := mapSessionUpdate("", params)
+	if len(evs) != 1 || evs[0].Type != core.EventThinking || evs[0].Content != "The user needs to check health" {
+		t.Fatalf("got %+v", evs)
+	}
+}
+
 func TestMapSessionUpdate_toolCall(t *testing.T) {
 	params := json.RawMessage(`{
 		"sessionId": "s1",
--- a/agent/copilot/session.go
+++ b/agent/copilot/session.go
@@ -409,6 +409,29 @@ func (cs *copilotSession) handleSessionEvent(params json.RawMessage) {
 			}
 		}
 
+	case "assistant.reasoning_delta":
+		// Streaming reasoning/thinking chunk. Copilot sends these with the
+		// same deltaContent shape as message deltas, but they must be mapped
+		// to EventThinking (not EventText) so thinking_messages=false can
+		// suppress them and they don't get concatenated with the answer.
+		if content := copilotEventText(evt.Event.Data); content != "" {
+			e := core.Event{Type: core.EventThinking, Content: content}
+			select {
+			case cs.events <- e:
+			case <-cs.ctx.Done():
+			}
+		}
+
+	case "assistant.reasoning":
+		// Final aggregated reasoning. Same mapping as the delta stream.
+		if content := copilotEventText(evt.Event.Data); content != "" {
+			e := core.Event{Type: core.EventThinking, Content: content}
+			select {
+			case cs.events <- e:
+			case <-cs.ctx.Done():
+			}
+		}
+
 	case "assistant.message":
 		usage := copilotEventUsage(evt.Event.Data)
 		if len(evt.Event.Data) > 0 {
@@ -494,6 +517,10 @@ func normalizeCopilotEventType(eventType string) string {
 		return "assistant.message_delta"
 	case "assistant_streaming_delta":
 		return "assistant.streaming_delta"
+	case "assistant_reasoning":
+		return "assistant.reasoning"
+	case "assistant_reasoning_delta":
+		return "assistant.reasoning_delta"
 	}
 	return eventType
 }
--- a/agent/copilot/session_test.go
+++ b/agent/copilot/session_test.go
@@ -93,6 +93,61 @@ func TestHandleSessionEvent_MessageDelta(t *testing.T) {
 	}
 }
 
+func TestHandleSessionEvent_ReasoningDelta(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	cs := &copilotSession{events: make(chan core.Event, 10), ctx: ctx, cancel: cancel}
+	cs.alive.Store(true)
+
+	data, _ := json.Marshal(map[string]any{
+		"reasoningId":   "r-1",
+		"deltaContent":  "let me think",
+	})
+	cs.handleSessionEvent(json.RawMessage(mustMarshal(t, sessionEvent{
+		Event: sessionEventInner{Type: "assistant.reasoning_delta", Data: data},
+	})))
+
+	select {
+	case evt := <-cs.events:
+		if evt.Type != core.EventThinking {
+			t.Fatalf("event type = %v, want EventThinking", evt.Type)
+		}
+		if evt.Content != "let me think" {
+			t.Fatalf("content = %q, want 'let me think'", evt.Content)
+		}
+	default:
+		t.Fatal("no thinking event emitted for reasoning_delta")
+	}
+}
+
+func TestHandleSessionEvent_ReasoningFinal(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	cs := &copilotSession{events: make(chan core.Event, 10), ctx: ctx, cancel: cancel}
+	cs.alive.Store(true)
+
+	data, _ := json.Marshal(map[string]any{
+		"reasoningId": "r-1",
+		"content":     "final reasoning",
+	})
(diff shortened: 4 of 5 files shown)


chenhg5 (changes requested):
Conclusion: Request changes

Overall assessment:
- The Copilot reasoning-delta → `EventThinking` mapping is a real and well-explained bug fix. The follow-on ACP `agent_thought_chunk` mapping is also correct (whitelist typo, no default-branch misroute). **However, the third commit (`fix: nil pointer panic when /new races with an in-flight turn`) is unrelated** to Copilot reasoning — it touches `core/engine.go` to harden against a concurrency race. Bundling unrelated fixes in one PR makes the bug fix hard to revert and confuses the changelog/release notes. The Copilot + ACP pair (commits 1 and 2) is mergeable as-is; the third commit should be split into its own PR before merge.

Review scope:
- agent/copilot/session.go — added `assistant.reasoning_delta` and `assistant.reasoning` cases emitting `EventThinking`; added underscore variants in `normalizeCopilotEventType`.
- agent/copilot/session_test.go — `TestHandleSessionEvent_ReasoningDelta` + `TestHandleSessionEvent_ReasoningFinal`.
- agent/acp/mapping.go — added `"agent_thought_chunk"` to the reasoning discriminator whitelist in `mapSessionUpdateFallback`.
- agent/acp/mapping_test.go — `TestMapSessionUpdate_agentThoughtChunk`.
- core/engine.go — captured `state.agentSession` under lock in five places (`processInteractiveMessageWith`, `processInteractiveEvents`, queued-turn stale-event drain, `runCompress`, `processCompressEvents`) and added silent-abort branches on nil.
- Run: `go test -race ./agent/copilot/... -count=1 -timeout 60s` — ok. `go test -race ./agent/acp/... -run 'TestMap' -count=1 -timeout 60s` — ok. `go test -race ./core/... -run 'TestCUJ|TestProcessInteractive' -count=1 -timeout 60s` — ok. The pre-existing `TestCursorCLI_ACPHandshake` failure requires `agent login` and is unrelated to this PR.

✅ What looks good:
- The Copilot reasoning-delta root-cause analysis in the PR description is excellent: confirms the wire format by probing `copilot --stdio`, explains why reasoning could leak into the answer (`deltaContent` shape is identical to message deltas), and ties the fix to the user-visible symptom (`thinking_messages=false` had no effect).
- Reusing `copilotEventText` (which already reads `deltaContent`/`content`) keeps the fix minimal and avoids a separate content-extraction path that could drift.
- Adding underscore variants (`assistant_reasoning`, `assistant_reasoning_delta`) in `normalizeCopilotEventType` makes the adapter robust to either separator, which matches the `assistant_streaming_delta` precedent.
- ACP `agent_thought_chunk` is added alongside the existing `agent_thinking_chunk` (gerund vs past tense), which is exactly the kind of subtle taxonomy bug that the test now locks.
- The `core/engine.go` race fix is consistent with the existing `drainPendingMessages` convention (#1436): capture under `state.mu`, abort with a `slog.Warn` when nil. The author explicitly cites this precedent.
- Tests use a buffered events channel and `select`/`default` patterns consistent with the existing copilot test suite, so the new tests stay maintainable.

🚨/🔴 Must fix:
- 🔴 P1 — **Bundle hygiene: the third commit is not Copilot-related and should be split out.** Commit `564c7dec` (`fix: nil pointer panic when /new races with an in-flight turn`) touches only `core/engine.go` and has no relationship to either the Copilot reasoning-delta fix or the ACP `agent_thought_chunk` fix. The PR title is `fix(agent/copilot): map reasoning deltas to EventThinking`, but the body needs to be updated to acknowledge the scope creep, OR the third commit should be moved into its own PR. Recommend: open a separate PR (or stacked branch) for `core/engine.go` race fix and rebase this PR down to the first two commits. Reasons:
  - Reverting Copilot reasoning (if a regression is reported) should not also revert a core concurrency fix.
  - CHANGELOG cannot easily describe this PR with a single line; mixing a per-adapter fix with a core race fix muddies release notes.
  - The race fix deserves its own review surface; it touches the message-processing engine that every agent uses.
  - PR review attribution: the Copilot reasoning bug is reported by a Copilot user; the engine race fix is reported via a paas-ops production panic. These should not share a "fixes #X" footer.

🟠 Should improve:
- 🟠 P2 — `processInteractiveMessageWith` has two captures of `state.agentSession`. The first capture (added in this PR) checks for nil and aborts; the second capture (changed from `:=` to `=`) does NOT check for nil before scheduling the Send goroutine. The goroutine itself checks for nil and sends an error, but the outer function still proceeds to call `e.processInteractiveEvents(...)`, which now also checks for nil (also added in this PR) and aborts via `sp.discard()`. The end-to-end behavior is correct, but the redundancy is confusing — readers expect every capture-under-lock to be paired with a nil check. Suggest moving the second capture to before the first nil check (or making the second capture also `:=` and checking nil inline) for symmetry with the compress path.
- 🟠 P2 — The race-fix commit silences the failure ("WARN log only") because "the racing /new already replied to the user". But the agent may still be in the middle of a `Send` when /new arrives, and the user has no idea their prompt was silently dropped after they typed `/new` thinking the old turn had been canceled. Consider one of: (a) attach a `core.Message` to the dropped prompt so the next user message gets a "previous message dropped" notice, (b) at minimum include `msg.MessageID` in the warn log so support can correlate.
- 🟠 P2 — No regression test specifically for the /new-vs-turn race. The existing `TestProcessInteractiveMessageWith_NilAgentSession_NoPanic` (engine_test.go:9694) only covers the agent-start-failure path, where `state.agentSession` is nil from the start. The new commit's scenario is "agentSession was non-nil at first capture but becomes nil before the second capture". A test that calls `cleanupInteractiveState` between the two captures would lock the fix. Without it, a future refactor that reorders the captures could silently re-introduce the panic.
- 🔵 P3 — `TestHandleSessionEvent_ReasoningDelta` and `TestHandleSessionEvent_ReasoningFinal` are nearly identical (different input data, same assertion shape). Consider table-driven to halve the test code and make it easier to add a third variant.
- 🔵 P3 — CHANGELOG.md is untouched. Once the split is done, add:
  - For the Copilot+ACP PR: `### Fixed — copilot: map assistant.reasoning_delta / assistant.reasoning to EventThinking so thinking_messages=false suppresses Copilot reasoning (#1512); acp: also map agent_thought_chunk`.
  - For the engine-race PR: `### Fixed — core: prevent nil-pointer panic in processInteractiveMessageWith / processInteractiveEvents / runCompress / processCompressEvents when /new races with an in-flight turn (#XXXX)`.

❓ Questions:
- The Copilot reasoning-delta case uses `copilotEventText` which presumably reads `deltaContent` first and then falls back to `content`. For the `assistant.reasoning` (final aggregated) case, the PR description shows payload `{"reasoningId":"...","content":"..."}`. Does `copilotEventText` actually prefer `content` over `deltaContent`, or does it just take whichever is present? If both are present, which wins? Worth a one-line comment in the test to lock the precedence.
- For the ACP `agent_thought_chunk` whitelist extension, are there other `agent_*_chunk` discriminators in the wild that should be added in the same PR? (e.g. `agent_tool_chunk`, `agent_message_chunk`). If yes, worth a single sweep; if no, current scope is fine.

Testing / Risk:
- Verified: copilot tests, ACP mapping tests, core CUJ and ProcessInteractive tests all pass with `-race`.
- The pre-existing `TestCursorCLI_ACPHandshake` failure in `./agent/acp/...` is environmental (requires `agent login`) and reproduces on `main` without this branch — not a regression.
- Risk surface for commits 1+2: bounded to Copilot + ACP adapters. Reasoning toggle (`thinking_messages`) behavior improves for both.
- Risk surface for commit 3: touches `core/engine.go`, which runs on every interactive turn across all agents. The new nil-guards are defensive only (they abort a turn that was already being canceled by /new) so blast radius is small, but the function-level change should be exercised by a CI test before merge.

Next step:
- Split the third commit into its own PR (`fix(core): prevent nil-pointer panic when /new races with an in-flight turn`) so the Copilot reasoning fix can ship cleanly. Add a regression test for the /new race scenario (table-driven: nil-out `state.agentSession` between two captures, assert no panic). Once split, both PRs are straightforward to approve.

chenhg5:
@zhengweixing 你的 #1512 review 已发,结论 **Request changes** — 主要 blocker 是第三个 commit `564c7dec` 改了 `core/engine.go` 加 nil pointer panic 修复,跟 copilot reasoning 无关,需要拆 PR:

- **commit 1 `b2d9ba04`**: feat(agent/copilot): map reasoning deltas to EventThinking(主体改动)
- **commit 2 `29cbff9c`**: feat(agent/copilot): filter ACP thoughts_chunk(同一特性,建议保留)
- **commit 3 `564c7dec`**: fix(core/engine): nil pointer panic in sendCommandReply  ← 拆出去,单独一个 PR

详见 GitHub review: https://github.com/chenhg5/cc-connect/pull/1512#pullrequestreview-...

拆完之后原 PR 只需 squash 1+2 成一个干净 feat commit,我会重新 review。

— cc-connect/qa-claudecode

PR #1519, opened 2026-07-10
feat(codex): extend GPT-5.6 reasoning and fallback support

AaronZ345:
## Summary

- accept `none`, `minimal`, `max`, and `ultra` as Codex reasoning efforts
- build `/reasoning` and `/effort` usage text from the active agent instead of hardcoding four levels
- add current GPT-5.x models to the hardcoded fallback used only when catalog, cache, and API discovery return nothing
- document the eight canonical Codex reasoning levels in `config.example.toml`

## Relation to #1547

#1547 already fixed GPT-5.x discovery from `/v1/models` by replacing the stale allowlist with `isCodexChatModel`. This PR leaves that filter unchanged.

The remaining model change here is limited to `defaultCodexModels()`, the final offline fallback. The duplicate GPT-5.6 API test has also been removed; the upstream `codex_model_test.go` suite now owns API-filter coverage.

The reasoning changes are separate from model discovery. Codex CLI 0.144 accepts the newer levels, but cc-connect previously normalized values beyond `xhigh` to an empty string and showed a stale four-level usage message.

## Validation

- `go test ./core -run "TestCmdReasoning_UsageListsAgentEfforts|TestCmdReasoning_SwitchesEffortAndResetsSession" -count=1`
- `go test -tags no_web ./cmd/cc-connect ./core ./agent/codex -count=1`
- `git diff --check`


AaronZ345:

AaronZ345:
Superseded by #1657. Closed only to refresh the PR after the 30-day maintenance gate; the replacement carries the same patch.

PR #1520, opened 2026-07-10
fix(codex): isolate subagent app-server events

AaronZ345:
## Summary
- ignore Codex app-server lifecycle notifications that belong to subagent threads instead of the active cc-connect thread
- scope item, completion, and token-usage notifications to the active parent turn
- preserve compatibility with older app-server notifications that omit thread or turn IDs
- add regression coverage proving child turns cannot overwrite or complete the parent turn

## Root cause
Codex app-server multiplexes parent and subagent events over the same connection. Its v2 protocol includes `threadId` on `turn/started`, `turn/completed`, and `thread/status/changed`, plus both `threadId` and `turnId` on item and token-usage notifications.

cc-connect parsed those fields but did not validate them. When a subagent started, its `turn/started` notification replaced the parent's `currentTurn` and cleared pending parent messages. When that child thread became idle or completed, cc-connect called `completeTurn()` and emitted an empty `EventResult`. Repeated child turns therefore produced repeated empty responses, while parent output could be truncated. Child token usage could also overwrite the parent turn's usage snapshot.

This change fixes the event-routing boundary rather than suppressing empty messages. Parent-thread follow-up turns, including results delivered back from subagents, continue through the normal unsolicited-event path.

## Live evidence
In a GPT-5.6 Sol `ultra` run, the parent turn completed at 12:15:08. Codex then started subagent turns on distinct thread IDs. Child thread idle transitions at 12:15:15 and 12:15:23 aligned exactly with cc-connect logging `unsolicited turn complete response_len=0`. The Codex 0.144.1 generated protocol schema confirms these notifications carry the thread and turn identifiers needed for routing.

## Validation
- `go test -race ./agent/codex -run 'TestAppServerSession_(IgnoresSubagentLifecycleNotifications|IgnoresSubagentTokenUsageNotifications)' -count=1`
- `go vet ./agent/codex`
- `git diff --check`

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -1103,7 +1103,7 @@ func (s *appServerSession) handleNotification(method string, paramsRaw json.RawM
 	switch method {
 	case "turn/started":
 		var notif turnNotification
-		if err := json.Unmarshal(paramsRaw, &notif); err == nil {
+		if err := json.Unmarshal(paramsRaw, &notif); err == nil && s.acceptsThreadNotification(method, notif.ThreadID) {
 			s.stateMu.Lock()
 			s.currentTurn = notif.Turn.ID
 			s.pendingMsgs = s.pendingMsgs[:0]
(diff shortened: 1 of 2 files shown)


AaronZ345:

AaronZ345:
Superseded by #1656. Closed only to refresh the PR after the 30-day maintenance gate; the replacement carries the same patch.

PR #1530, opened 2026-07-11
fix(codex): steer active turns for /ps guidance

AaronZ345:
## Summary

- steer an active Codex app-server turn with `turn/steer` instead of starting a second turn
- preserve the active turn ID, buffered text, and prompt preamble while applying `/ps` guidance
- keep the existing `turn/start` behavior for idle sessions and reject mismatched steering responses
- make four pre-existing environment-sensitive tests hermetic so the full local suite is reliable

## Why

`/ps` calls `AgentSession.Send` while the current turn is still running. The Codex app-server implementation previously treated every `Send` as `turn/start`, reset the active-turn state, and cleared buffered messages. That race could emit an empty result even though Codex later completed a full answer.

Codex exposes `turn/steer` for same-turn guidance. This change uses it with the active `expectedTurnId`, matching Codex's native guidance behavior without turning the guidance into a separate user turn.

## Testing

- `go test ./...`
- `go test -race ./agent/codex ./core ./platform/cloud-web`
- `go vet ./...`
- `go build ./...`
- generated the Codex 0.144.1 app-server JSON schema and verified the `turn/steer` request and response fields

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -64,6 +64,10 @@ type turnStartResponse struct {
 	} `json:"turn"`
 }
 
+type turnSteerResponse struct {
+	TurnID string `json:"turnId"`
+}
+
 type turnNotification struct {
 	ThreadID string `json:"threadId"`
 	Turn     struct {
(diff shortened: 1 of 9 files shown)


AaronZ345:

AaronZ345:
Superseded by #1665.

Issue #1008, opened 2026-05-16
Expose agent_session_id and display_name in Management sessions API


kagamiurayama:
Hi, thanks for cc-connect.

I'm building a custom frontend using the Management API and Bridge API. I ran into a session naming problem that is hard to solve on the frontend side.

Current behavior

• GET /api/v1/projects/{project}/sessions returns wrapper sessions like s2, s5, s6, s9.
• GET /api/v1/projects/{project}/sessions/{id} returns agent_session_id, but the list endpoint does not.
• The user-facing session names shown by /list come from session_names[agent_session_id].
• However, session_names is not exposed through the Management API.

Why this matters

Multiple wrapper sessions can share the same underlying agent_session_id.

Example:
• s2, s5, s6, and s9 may all point to the same Claude/agent session.
• The user may rename that agent session with /name, e.g. "MyAgent".
• /list correctly shows the user-facing name.
• But a frontend using the Management API only sees each wrapper session's name, which may be default, a timestamp summary, or the latest user message snippet.
• This makes frontend display names inconsistent with cc-connect's own /list output.

Request

Please expose stable display fields in the Management sessions API.

Suggested fields for both:
• GET /api/v1/projects/{project}/sessions
• GET /api/v1/projects/{project}/sessions/{session_id}

{
  "agent_session_id": "910c40c0-...",
  "display_name": "MyAgent",
  "agent_display_name": "MyAgent"
}

Suggested priority for display_name:
1. session_names[agent_session_id] (set by /name)
2. Agent/native summary title
3. sessions.sN.name
4. user_name / id fallback

This would allow frontends to display consistent names without parsing /list command text.

chenhg5:
Thanks for this feature request! This is a useful enhancement for custom frontend integrations.

The Management API will be extended to expose  and  fields, following the priority you specified. This will ensure consistency between the API and the /list command output.

Please follow this issue for updates.

PR #1549, opened 2026-07-15
fix(claudecode): prefer ai-title/custom-title in session list summaries

### LINKED ISSUE #1548: [Bug] 会话标题无法自动总结标题 —— ai-title 未能转换为 custom-title，始终回退到第一条用户消息 | Session title auto-summary broken — ai-title not promoted to custom-title, always falls back to first user message
## Bug Description

**中文：**

cc-connect 管理的 Claude Code 会话中，AI 内部自动生成的标题（`ai-title`）无法被提升为正式标题（`custom-title`），导致会话名称永远回退到第一条用户消息。即使 AI 已经生成了准确的中文标题建议（如"查看cc-connect开机自启功能"、"微信内容整理到飞书方案"等），这些标题也不会被会话列表读取和展示。

经分析 `~/.claude/projects/` 下的 JSONL 文件后发现：
- 所有异常会话中 `ai-title` 均正常生成，但 `custom-title` 始终为 0 条
- 正常会话中 `ai-title` 和 `custom-title` 成对交替出现
- 所有会话的 `summary` 类型事件均为 0 条

根因：Claude Code 的 `ai-title` → `custom-title` 转换依赖 Stop hook / SessionEnd 事件触发。cc-connect 的 daemon 在回收 agent 进程时直接终止进程，不走正常退出流程，Stop hook 无法执行，导致 `custom-title` 永远不会被写入。标题读取优先级（`custom-title` → `summary` → 第一条用户消息）中前两级全部失效，最终回退到第一条消息——通常是 `[cc-connect sender_id=...]` 元数据字符串。

**English:**

In Claude Code sessions managed by cc-connect, the AI-generated title suggestion (`ai-title`) is never promoted to the formal session title (`custom-title`), causing the session name to always fall back to the first user message. Even though the AI correctly generates accurate title suggestions, these titles are never picked up by the session list.

Analysis of JSONL files under `~/.claude/projects/` reveals:
- All affected sessions have `ai-title` entries present, but 0 `custom-title` entries
- Healthy sessions show `ai-title` and `custom-title` alternating in matched pairs
- All sessions have 0 `summary` events

Root cause: Claude Code's `ai-title` → `custom-title` promotion depends on the Stop hook / SessionEnd event firing. cc-connect's daemon terminates agent processes directly on session teardown, skipping the normal exit flow. The Stop hook never executes, so `custom-title` is never written. The two highest-priority sources in the title resolution chain (`custom-title` → `summary` → first user message) both fail, leaving the degraded fallback to the first message — typically the `[cc-connect sender_id=...]` metadata string.

## Steps to Reproduce

**中文：**

1. 配置 cc-connect 使用 Claude Code agent
2. 通过任意平台发送一条消息，开启一个新会话
3. 进行几轮有实质内容的对话（让 AI 有机会生成 `ai-title`）
4. 等待会话结束或被新会话替代
5. 查看 JSONL：`grep '"type":"ai-title"'` 有多条记录，但 `grep '"type":"custom-title"'` 为 0 条
6. 使用 `/list` 查看标题 → 显示第一条用户消息而非 AI 生成的标题

**English:**

1. Configure cc-connect with a Claude Code agent
2. Send a message via any platform to start a new session
3. Have several turns of substantive conversation (so the AI generates an `ai-title`)
4. Let the session end or be replaced by a new one
5. Inspect the JSONL: `grep '"type":"ai-title"'` shows entries, but `grep '"type":"custom-title"'` shows 0
6. Check the session title via `/list` → shows the first user message instead of an AI-generated title

## Expected Behavior

**中文：**

会话标题应显示 AI 自动生成的有意义标题，而非第一条用户消息。cc-connect 在回收 agent 进程前应触发 auto-rename，将 JSONL 中已存在的 `ai-title` 提升为 `custom-title`。

**English:**

Session titles should display the AI-generated meaningful title rather than the raw first user message. cc-connect should trigger an auto-rename before tearing down the agent process, promoting the existing `ai-title` in the JSONL to a `custom-title`.

## Actual Behavior

**中文：**

会话标题回退到第一条用户消息，通常表现为带有 `[cc-connect sender_id=...]` 元数据前缀的无意义字符串。已确认影响 cc-connect `/list`、Claude Desktop 侧边栏、CLI `--resume` 三处。

**English:**

Session titles fall back to the first user message, typically a meaningless `[cc-connect sender_id=...]` metadata string. Confirmed affected: cc-connect `/list`, Claude Desktop sidebar, and CLI `--resume` session picker.

syf2211:
## Summary

Update Claude Code session list summaries to read `custom-title` and `ai-title` entries from JSONL transcripts before falling back to the first user message.

## Motivation

When cc-connect tears down a Claude Code agent process, the Stop hook often does not run, so `ai-title` is never promoted to `custom-title`. `/list` then shows the first user message — typically a `[cc-connect sender_id=...]` metadata string — instead of the AI-generated title.

Fixes #1548

## Changes

- `scanSessionMeta` now resolves titles with priority: `custom-title` > `ai-title` > first user message
- Added regression tests for ai-title fallback and custom-title precedence

## Tests

```
go test ./agent/claudecode/... -count=1 -run TestScanSessionMeta
go test ./agent/claudecode/... -count=1
```

All tests pass.

## Notes

This is a read-side fix for `/list` display. Writing `custom-title` on graceful shutdown would also help Claude Desktop and CLI `--resume`, but is out of scope for this minimal change.

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -632,27 +632,48 @@ func scanSessionMeta(path string) (string, int) {
 	scanner := bufio.NewScanner(f)
 	scanner.Buffer(make([]byte, 256*1024), 256*1024)
 
-	var summary string
+	var firstUserMessage string
+	var aiTitle string
+	var customTitle string
 	var count int
 
 	for scanner.Scan() {
 		var entry struct {
-			Type    string `json:"type"`
-			Message struct {
+			Type        string `json:"type"`
+			AITitle     string `json:"aiTitle"`
+			CustomTitle string `json:"customTitle"`
+			Message     struct {
 				Content json.RawMessage `json:"content"`
 			} `json:"message"`
 		}
 		if err := json.Unmarshal(scanner.Bytes(), &entry); err != nil {
 			continue
 		}
-		if entry.Type == "user" || entry.Type == "assistant" {
+		switch entry.Type {
+		case "ai-title":
+			if entry.AITitle != "" {
+				aiTitle = entry.AITitle
+			}
+		case "custom-title":
+			if entry.CustomTitle != "" {
+				customTitle = entry.CustomTitle
+			}
+		case "user", "assistant":
 			count++
-			if entry.Type == "user" {
+			if entry.Type == "user" && firstUserMessage == "" {
 				if s := extractStringContent(entry.Message.Content); s != "" {
-					summary = s
+					firstUserMessage = s
 				}
 			}
 		}
+	}
+
+	summary := customTitle
+	if summary == "" {
+		summary = aiTitle
+	}
+	if summary == "" {
+		summary = firstUserMessage
 	}
 	summary = stripXMLTags(summary)
 	summary = strings.TrimSpace(summary)
--- a/agent/claudecode/claudecode_test.go
+++ b/agent/claudecode/claudecode_test.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/chenhg5/cc-connect/core"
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment: This is a tight, well-scoped fix that correctly identifies the root cause (daemon-side teardown skipping Claude Code's Stop hook, so `ai-title` never gets promoted to `custom-title`) and addresses it by reading both fields directly from the JSONL. The new fallback chain matches the issue's requirement (custom-title > ai-title > first user message), CI is green across all 5 jobs, and blast radius is genuinely limited to `scanSessionMeta`. Merge-ready.

Review scope:
- Reviewed `agent/claudecode/claudecode.go::scanSessionMeta` (+27/-6) and `agent/claudecode/claudecode_test.go` (+45/-1).
- Cross-referenced issue #1548 (root cause analysis) and issue #1008 (Management API exposure, separate surface).
- Focused on correctness of the fallback chain, JSONL parsing safety, test coverage, and behavior preservation for legacy versions.

✅ What looks good:
- **Root-cause-driven fix**: instead of trying to replicate Claude Code's `ai-title` → `custom-title` promotion in cc-connect's teardown path (which would touch the process model), the change reads both JSONL fields directly. That's a strictly safer surface to touch and avoids the daemon having to coordinate with the agent's Stop hook.
- **Three-level fallback chain is explicit**: `customTitle → aiTitle → firstUserMessage`, applied AFTER the scan loop completes. This is more readable than the old single-pass overwrite and lets the priority order be inspected at a glance.
- **Subtle correctness fix bundled**: the old code overwrote `summary` on every `user` message, leaving it at the **last** string user message. The new `firstUserMessage == ""` guard picks the **first** one, which is more representative of session identity. The comment in `TestScanSessionMeta_ArrayContent` is correctly updated to match.
- **Field casing matches Claude Code schema**: `aiTitle` / `customTitle` as top-level fields match v1.0.30+ JSONL shape. Older sessions that never emit these types cleanly fall through to the first-user-message branch — no parse error, no regression.
- **Parsing robustness preserved**: `json.Unmarshal` errors still `continue` per line, and the 256KB scanner buffer cap is unchanged. The new `switch` cleanly handles unknown `entry.Type` values (no implicit fallthrough).

- **Add a test for "ai-title appears BEFORE the user message"**: the existing two tests put `ai-title`/`custom-title` AFTER the user prompt. The end-of-loop resolution means the result is the same either way, but a regression test asserting "out-of-order entries still pick the title" would lock in the contract.
- **Document the fallback chain in the function doc comment**: `scanSessionMeta`'s doc (if any) should say "prefers custom-title, then ai-title, then the first user message" so future maintainers don't need to read the loop to understand the contract. Right now the priority is only visible by reading 5 lines at the bottom.
- **Consider promoting ai-title → custom-title at teardown as a follow-up**: this PR reads both fields directly, which fixes `/list`, but Claude Desktop sidebar and CLI `--resume` already do their own JSONL scanning. If their parsers only look at `custom-title`, they will still show the broken fallback. (Confirming CC Desktop behavior is out of scope here, but worth a note in the PR description.)
- **Issue #1008 stays open by design**: this PR doesn't touch the Management API. That's correct — the requested `agent_session_id` / `display_name` exposure is a separate surface. Suggest filing a follow-up issue or linking from #1008 that #1549's title fallback is now correct on the cc-connect side and the Management API just needs to forward the same field.

🔵 Optional:
- `summary, count := scanSessionMeta(path)` keeps the same signature, which is good. No callers need updating — confirmed by spot-checking `agent/claudecode.ListSessions` call site.
- The unused `sessionId` field in test JSONL data is realistic but not parsed; that's fine for now.

❓ Questions:
- For very long-running sessions with hundreds of turns, has anyone benchmarked `scanSessionMeta` post-fix? The cost is the same O(N) scan as before, so I don't expect a regression, but if the function is on a hot path (e.g. `--resume` picker) it'd be nice to confirm there isn't a JSONL size that now noticeably affects startup.

Testing / Risk:
- Verified evidence: 5/5 CI jobs green (`lint`, `unit-test`, `smoke-test`, `regression-test`, `performance-test`), all completed 2026-07-15T00:19–00:26Z. 3 new test cases plus the amended `TestScanSessionMeta_ArrayContent` exercise the title fallback directly.
- Remaining risk: Claude Desktop sidebar and CLI `--resume` may parse `custom-title` only; until verified, treat their UX as "may still show first user message" — the issue body notes this is a downstream concern that this PR doesn't fully cover. CC Connect's own `/list` is fixed.

Next step:
- Re-approve. Merge-ready; awaiting owner merge. No further PR-side work required for this issue.

PR #1557, opened 2026-07-16
fix: prevent attachment filename collisions

### LINKED ISSUE #1552: [Bug] SaveFilesToDisk: same-named attachments overwrite each other (data loss + duplicate refs)
### cc-connect Version

main / v1.5.0-beta.1 (structural — `SaveFilesToDisk` has keyed attachments by raw filename since the attachment path was introduced; affects 1.3.x–1.5.x).

### Operating System

Any (core, platform-independent — every platform's inbound attachments funnel through `core.SaveFilesToDisk`).

### Agent Type

Any (`core` bug — not agent-specific).

### Platform

Any — `core.SaveFilesToDisk` is platform-independent; platform adapters pass the client-supplied attachment filename through to it unchanged.

### Bug Description

`SaveFilesToDisk` (`core/message.go`) stores every attachment at `<work_dir>/.cc-connect/attachments/<filename>`, where `<filename>` is the client-supplied basename (`sanitizeAttachmentFileName(f.FileName)`) with **no uniquification** and **no per-message/per-session subdir**:

```go
attachDir := filepath.Join(absWorkDir, ".cc-connect", "attachments")
for i, f := range files {
    fname := sanitizeAttachmentFileName(f.FileName) // raw basename; generated name only if empty
    fpath := filepath.Join(attachDir, fname)
    os.WriteFile(fpath, f.Data, 0o644)              // O_TRUNC → overwrites
}
```

So two attachments that share a filename **collide**: the second `os.WriteFile` truncates and overwrites the first. Two distinct uploads can never coexist on disk if they share a name. Worse, the returned `paths` slice still gets **one entry per input file** — so on a same-name collision it contains **duplicate paths pointing at the single surviving file**, and `AppendFileRefs` then references the same (wrong) file twice in the agent prompt.

It reproduces in a single plain `SaveFilesToDisk` call — no special configuration required.

### Steps to Reproduce

Minimal unit repro (drop into `core/`):

```go
func TestProbe_SameNameCollision(t *testing.T) {
    dir := t.TempDir()
    files := []FileAttachment{
        {FileName: "report.pdf", Data: []byte("FILE-A-CONTENT")},
        {FileName: "report.pdf", Data: []byte("FILE-B-CONTENT")},
    }
    paths := SaveFilesToDisk(dir, files)
    // observe paths + on-disk state
}
```

Observed:
```
returned 2 path(s): [.../attachments/report.pdf  .../attachments/report.pdf]   # duplicate
files actually on disk: 1
report.pdf => "FILE-B-CONTENT"                                                 # A is gone
```

Real-world triggers:
1. One message with two attachments sharing a name (e.g. two `photo.jpg`) → one silently lost; agent gets two refs to the survivor.
2. Two concurrent sessions (or turns) each sending a same-named file (e.g. `screenshot.png`) into the shared attachments dir → one overwrites the other mid-flight, so an agent can read another turn's file content.

Common client filenames (`image.png`, `screenshot.png`, `IMG_1234.jpg`, `Screenshot 2026-…png`) make this reachable; sequential distinct-name uploads hide it, which is why it surfaces as intermittent, hard-to-attribute attachment corruption rather than an obvious failure.

### Expected Behavior

Two distinct uploads must never clobber each other. Each attachment should be written to a **unique** path — while keeping the original filename available for the agent's benefit. Options:

- per-message (or per-session) subdir: `.cc-connect/attachments/<message-id>/<name>` — preserves the name, isolates, and enables teardown; or
- an on-collision suffix (`report.pdf`, `report-1.pdf`, …); or
- a content-hash / unique-id prefix.

A bare timestamp is **not** sufficient (not collision-safe under concurrency). The returned `paths` must be distinct and each resolve to its own content.

### Related

- #1459 — the other `SaveFilesToDisk` correctness bug (relative `work_dir` → lost attachments). Same function; this is the filename-collision side.

🤖 90% Claude

chenhg5:
Fixes #1552

## Summary
- Thread platform message IDs through `AgentSession.Send` into attachment persistence.
- Store attachments under per-message directories and generate unique names for duplicate filenames in one message.
- Keep legacy callers safe with atomic no-overwrite writes, and preserve absolute-path and traversal protections.

## Test plan
- `go test ./core ./agent/... ./platform/... ./tests/release_local/... -count=1`
- `go vet ./core ./agent/... ./platform/...`
- `go test -race` could not run in this environment because cgo requires a C compiler and `gcc` is unavailable.


chenhg5 (commented):
Conclusion: Request changes

Overall assessment:
- Fix design for #1552 is correct and well-scoped: per-message subdirectory isolates same-named attachments, atomic non-overwrite fallback protects legacy callers, and all three new tests in `core/message_test.go` are correctly designed to verify the acceptance criteria. However the PR is currently **not mergeable**: CI `lint` job fails with a Go typecheck error in `cmd/cc-connect/main_test.go:25`, which cascades into `unit-test` / `smoke-test` / `regression-test` / `performance-test` all being SKIPPED. We have no green CI evidence for the 514 LOC of cross-cutting change yet.

Review scope:
- Reviewed `core/message.go` (SaveFilesToDisk + AppendFileRefs), `core/interfaces.go` (Send signature), `core/engine.go` (Send call sites), `agent/<all>/session.go` (16 adapter signatures), `tests/mocks/{mock_agent,fake}.go`, `core/message_test.go` (new tests), `cmd/cc-connect/main_test.go` (lint failure origin).
- Focused on correctness, race-safety, backwards compat, and CI typecheck.

✅ What looks good:
- `writeUniqueNoOverwrite` uses `os.OpenFile(..., O_CREATE|O_EXCL, ...)` — atomic at the OS level, so two goroutines racing on the same `(messageID, fname)` cannot both succeed; the loser retries with `_1`, `_2` suffix. This is the right primitive.
- `writeAtomicNoOverwrite` legacy fallback uses `os.CreateTemp` + `os.Link` — link is atomic on POSIX, and refusing to overwrite (returning false rather than truncating) is the correct behavior change vs. the old silent overwrite. Issue #1459's relative-path footgun is also closed in the same PR via AppendFileRefs absolutization.
- `messageID` is itself passed through `sanitizeAttachmentFileName` before being joined into the dir — defender-in-depth against hostile or buggy platform IDs.
- `msg.MessageID` threaded through `engine.go` at all the interactive send sites (3754/6175) and "no messageID" (`""`) for synthesized sends (ps, compress, relay, 6266/10164/15690) — semantic clarity preserved.
- All 16 platform adapter `Send` signatures updated consistently; `tests/mocks/{mock_agent,fake}.go` updated in lockstep. Only one stub was missed (see 🚨 below).
- `core/cuj_test.go`, `core/engine_test.go`, integration/unsolicited/multi_workspace, release_local/{turn_contract,media_pipeline,engine_matrix}, performance/bench_test, e2e/{smoke,regression} — all updated to the new `Send(prompt, messageID, images, files)` signature.
- `baseRefName=main` ✅ (per dev-base-on-main).

🚨 Must fix (CI blocker):

🚨 P0 — PR does not compile: `cmd/cc-connect/main_test.go:25` stub is not updated
  Impact: `golangci-lint --new-from-rev origin/main` fails the typecheck step (run 29521858040 job 87700604595, completedAt 2026-07-16T17:58:31Z). Because lint fails first, all four downstream jobs (unit-test / smoke-test / regression-test / performance-test) are SKIPPED. The PR ships a public-interface signature change but the build under `cmd/cc-connect` cannot compile. Merging would break `go test ./cmd/cc-connect/...` for any developer who runs it locally.
  Evidence: CI run log `cmd/cc-connect/main_test.go:25:9: cannot use &stubMainAgentSession{} (value of type *stubMainAgentSession) as core.AgentSession value in return statement: *stubMainAgentSession does not implement core.AgentSession (wrong type for method Send)` — `have Send(string, []core.ImageAttachment, []core.FileAttachment) error` / `want Send(string, string, []core.ImageAttachment, []core.FileAttachment) error (typecheck)`.
  Suggested fix: update `func (s *stubMainAgentSession) Send(string, []core.ImageAttachment, []core.FileAttachment) error` at `cmd/cc-connect/main_test.go:44` to add a `string` parameter matching the new interface signature, e.g. `func (s *stubMainAgentSession) Send(string, string, []core.ImageAttachment, []core.FileAttachment) error` and ignore the messageID inside the body.
  Verification: rerun CI; expect `lint` → success, `unit-test` → run (not skipped), all 4 downstream jobs → success. Locally: `go test ./cmd/cc-connect/...`.

🟠 Should improve:

🟠 P2 — `TestSaveFilesToDisk_ConcurrentDifferentMessages` uses different `messageID`s; race is structurally avoided by the per-message subdir, but no test exercises concurrent writers sharing the **same** `messageID`
  Impact: Functional race across separate messages is well covered, but a pathological caller that concurrently invokes `SaveFilesToDisk(workDir, "msg-X", [...])` twice for the same messageID (e.g. a buggy caller building a message in two phases) would race between the two `os.OpenFile(..., O_CREATE|O_EXCL, ...)` calls inside `writeUniqueNoOverwrite` — and the OS-level O_EXCL + suffix loop handles it correctly, but no test proves that contract.
  Suggested fix: optional — add a tight loop test that fires N goroutines into `SaveFilesToDisk(workDir, "shared-msg", []FileAttachment{same-filename})` and asserts (a) all N files on disk have distinct content and (b) `_ = state.ReadFile(...)` on each path returns the right content. This is the test that catches a future refactor of `writeUniqueNoOverwrite` away from O_EXCL.
  Verification: would only need to run with `-race` to be conclusive, which the dev environment can't do (no gcc). Still useful for runtime correctness.

🟠 P2 — Legacy `messageID == ""` writes become one-shot: once a name exists it can never be replaced through this path
  Impact: `TestSaveFilesToDisk_NoSubdirBackwardsCompat` codifies this correctly (collision returns 0 paths and preserves the original), but it is a **silent behavior change** for any caller that relied on legacy flat-layout being replaceable. The PR doc-comments it ("refuse to overwrite") but there is no CHANGELOG line calling out the change for downstream consumers who add their own platform integrations.
  Suggested fix: add a one-liner to `CHANGELOG.md` under "Migration notes" — *"Legacy flat-layout attachment writes (`SaveFilesToDisk` without `messageID`) now refuse to overwrite existing files instead of truncating them. Callers that need replace semantics should delete the target file first or pass a real `messageID`."* The current `CHANGELOG.md` change is +2/-0 — likely a single bullet; adding one more is cheap.

🔵 Optional:

🔵 P3 — `time.Now().UnixNano()` in the auto-filename fallback could collide within a single nanosecond on fast loops
  Impact: With the `O_EXCL` + suffix loop in `writeUniqueNoOverwrite`, a same-nanosecond collision is recovered automatically, so this is not a correctness bug — it's a slight inefficiency (extra syscall) on concurrent fallback calls. Skip if you want; would only matter under heavy burst of unnamed uploads.
  Suggested fix: leave as-is; if you do want to be tidy, use a counter or `crypto/rand` suffix.

❓ Questions:

❓ — Should `cmd/cc-connect/` get a small wrapper test that exercises the per-message subdir layout from a real platform event end-to-end? Today the unit tests cover the storage primitive but the dispatcher that turns a `Platform.Message` into the `SaveFilesToDisk(workDir, msg.MessageID, files)` call is exercised only via `core/cuj_test.go`. Not a blocker, but adding a single round-trip test would be a strong safety net for future refactors.

Testing / Risk:
- Verified evidence:
  - `go test -run "TestSaveFilesToDisk" ./core/` (no `-race` — gcc unavailable in QA env, same constraint as dev env): **PASS** — `TestSaveFilesToDisk_SameNameInOneCall`, `TestSaveFilesToDisk_ConcurrentDifferentMessages`, `TestSaveFilesToDisk_NoSubdirBackwardsCompat` all green; paths returned are distinct; on-disk files match returned content; concurrent different-messageID write loses no data; legacy same-name reuse preserved without overwrite.
  - All 5 PM acceptance criteria (50gqwu) individually verified by reading the implementation against the spec:
    1. Same-name in one call → 2 distinct paths ✅ (writeUniqueNoOverwrite `_1` suffix; test line 208)
    2. Concurrent different messages → 2 surviving files ✅ (per-message subdir; test line 243; structurally race-safe — O_EXCL is OS-atomic — but `-race` not runnable here)
    3. No-subdir legacy fallback refuses overwrite ✅ (writeAtomicNoOverwrite os.Link rejects existing target; test line 281)
    4. paths ↔ on-disk 1:1 ✅ (returned fpath is the actual written location, only appended on `writeUniqueNoOverwrite` success)
    5. `go test ./core ./agent/... ./platform/...` + `-race` — partially verified: `./core` green; full suite not runnable here because `cmd/cc-connect/main_test.go` typecheck break blocks the build; `-race` not runnable (no gcc). CI will fully verify once the P0 is fixed.
- Remaining risk:
  - Without CI fully green, the 16 platform adapters + 5+ test mock sites that were all touched in this 514 LOC PR have **not** been integration-tested end-to-end. CI is the only way to assert cross-package behavior holds.

Next step:
- Update `cmd/cc-connect/main_test.go:44` to add the missing `messageID string` parameter (and adapt stub body to discard it). Then push a follow-up commit. CI should turn green and `lint` → `unit-test` → `smoke-test` → `regression-test` → `performance-test` all run; if `unit-test` is green I will lift the request-changes verdict. Looking forward to the fix.

chenhg5 (commented):
TEST - ignore

chenhg5 (commented):
TEST comment

chenhg5:
@chenhg5 🚨 P0 blocker on PR #1557: `cmd/cc-connect/main_test.go:44` stub `stubMainAgentSession.Send` still has the old signature `Send(string, []core.ImageAttachment, []core.FileAttachment) error`. The new `core.AgentSession.Send(prompt, messageID, images, files)` interface signature means `golangci-lint --new-from-rev origin/main` fails typecheck, cascading into unit-test/smoke/regression/performance all SKIPPED in CI run 29521858040. Suggested fix is a one-line stub signature change (drop the new param) — once pushed, CI should fully run. Full review at #pullrequestreview-4718820336.

chenhg5:
@cc-connect/qa-claudecode — P0 CI blocker fixed in `cc061b87` on this branch.

`cmd/cc-connect/main_test.go:44` stub `stubMainAgentSession.Send` was missed in #1557; it kept the pre-messageID 3-arg shape, so lint failed and the four downstream jobs all skipped. New signature mirrors `core.AgentSession.Send`:

```go
func (s *stubMainAgentSession) Send(string, string, []core.ImageAttachment, []core.FileAttachment) error {
    return nil
}
```

Verified locally:
- `go test -tags no_web -count=1 -run='^$' ./cmd/cc-connect` — compiles clean (the `web/embed.go` `all:dist` issue is a pre-existing build dependency, unrelated to this PR).
- `go vet -tags no_web ./cmd/cc-connect` — no diagnostics.
- `gofmt -l cmd/cc-connect/main_test.go` — clean.

I also re-grep'd the whole tree for `Send(string, []core.ImageAttachment, []core.FileAttachment)` / `Send(string, []ImageAttachment, []FileAttachment)` — no other survivors.

Re-running CI; expecting lint → unit → smoke → regression → performance all green on this push.

Tracking the two informational P2s separately:
- P2 same-messageID concurrent variant — happy to add as a follow-up test in a new PR so the collision fix stays minimal here.
- P2 CHANGELOG "Migration notes" line for the legacy `messageID == ""` refuse-to-overwrite behavior change — agree, will fold into the same follow-up changelog PR.

chenhg5 (commented):
Conclusion: Request changes

Overall assessment:
- The previous P0 (stub `cmd/cc-connect/main_test.go:44` missing the new `messageID` parameter → lint typecheck fail) is **fixed** in commit `cc061b87`. CI run 29549707897 no longer fails on typecheck. However, with the typecheck gone the lint job now proceeds to real lint analysis and surfaces **five new blockers** that are all in code this PR added: 4 errcheck violations in `core/message.go:writeAtomicNoOverwrite` (lines 241, 242, 247, 253) and 1 staticcheck QF1008 in `tests/release_local/turn_contract/turn_contract_test.go:953`. These are pre-existing-in-this-PR rather than revealed by 5711a0b7 — I verified each line is in your added diff. None of the downstream CI jobs (unit-test / smoke-test / regression-test / performance-test) have started yet because lint still fails first. Easy mechanical fixes; once applied, all five jobs should run.

Review scope:
- Re-reviewed commit `cc061b87` (1 file, +1/-1) on top of `5711a0b7`. Diff for `cmd/cc-connect/main_test.go:44` is exactly the suggested one-line signature change.
- Re-ran the 3 new `TestSaveFilesToDisk_*` tests mentally against the unchanged code path; still ✅ for criteria 1–4 (no changes in `core/message.go` or its tests).
- Read CI log `lint` job 87789464939 of run 29549707897 for the new blockers.
- Did not re-run `go test ./core ./agent/... ./platform/...` locally because gcc is unavailable (same as before).

✅ What looks good (re-confirming):
- `cc061b87`'s stub signature change at `cmd/cc-connect/main_test.go:44`: `Send(string, string, []core.ImageAttachment, []core.FileAttachment) error` — exact fix I asked for. No collateral changes.
- Acceptance criteria 1–4 from my previous review still hold (no `core/message.go` or `core/message_test.go` changes in `cc061b87`):
  1. `TestSaveFilesToDisk_SameNameInOneCall` → 2 distinct paths + 2 on-disk files ✅
  2. `TestSaveFilesToDisk_ConcurrentDifferentMessages` → 2 surviving files ✅ (still structurally race-safe via OS-level O_EXCL)
  3. `TestSaveFilesToDisk_NoSubdirBackwardsCompat` → original preserved, second collision rejected ✅
  4. paths ↔ on-disk 1:1 ✅ (writeUniqueNoOverwrite returns actual written path; only appended on success)
- P2 follow-up split (same-messageID race test + CHANGELOG migration note) into a separate PR is a clean call — keep this PR diff focused on the data-loss fix.
- `baseRefName=main` ✅ — re-verified via `git log main..pr-1557` = 2 commits (5711a0b7 + cc061b87).

🚨 Must fix (CI blockers — blocking lint gate):

🔴 P1 — 4 errcheck violations in `core/message.go` (`writeAtomicNoOverwrite`, added by this PR)
  Impact: `golangci-lint --new-from-rev origin/main` reports the 4 errcheck issues at lines 241/242/247/253 and exits 1. Lint job fails first in CI; unit/smoke/regression/performance all SKIPPED. Same root cause for all four: cleanup-path return values not consumed.
  Evidence (verbatim from CI log run 29549707897 job 87789464939 at 2026-07-17T02:22:54Z):
  ```
  core/message.go:241:12: Error return value of `tmp.Close` is not checked (errcheck)
  core/message.go:242:12: Error return value of `os.Remove` is not checked (errcheck)
  core/message.go:247:12: Error return value of `os.Remove` is not checked (errcheck)
  core/message.go:253:12: Error return value of `os.Remove` is not checked (errcheck)
  ```
  Suggested fix — wrap each in `defer func() { _ = X.Close() }()` style, or just `_ = tmp.Close()` / `_ = os.Remove(tmpName)` at the cleanup sites. Concretely:
  ```go
  if _, err := tmp.Write(data); err != nil {
      _ = tmp.Close()             // line 241: was tmp.Close()
      _ = os.Remove(tmpName)      // line 242: was os.Remove(tmpName)
      slog.Error("SaveFilesToDisk: tempfile write failed", "tmp", tmpName, "error", err)
      return false
  }
  if err := tmp.Close(); err != nil {
      _ = os.Remove(tmpName)      // line 247: was os.Remove(tmpName)
      slog.Error("SaveFilesToDisk: tempfile close failed", "tmp", tmpName, "error", err)
      return false
  }
  target := filepath.Join(dir, name)
  if err := os.Link(tmpName, target); err != nil {
      _ = os.Remove(tmpName)      // line 253: was os.Remove(tmpName)
      slog.Error("SaveFilesToDisk: refusing to overwrite existing file", "path", target, "error", err)
      return false
  }
  ```
  These are the same convention used in `writeUniqueNoOverwrite` (your sibling function) at lines 219/220/225 — the cleanup-on-error paths in that helper already use `defer ... _ = file.Close()` and `_ = os.Remove(path)` correctly, so this is consistency, not a new pattern.
  Verification: push a follow-up commit; rerun CI. Expect lint → success; unit-test → run (not skipped); all 4 downstream jobs → run.

🔴 P1 — staticcheck QF1008 in `tests/release_local/turn_contract/turn_contract_test.go:953` (added by this PR)
  Impact: Same lint gate — fails the run. The fix is mechanical. The selector `p.turnPlatform.snapshot()` is unnecessary because `previewLifecyclePlatform` (the receiver struct declared at line 950 in the same PR diff) embeds `*turnPlatform` — embedded-field selectors are auto-promoted, so `p.snapshot()` works.
  Evidence (CI log):
  ```
  tests/release_local/turn_contract/turn_contract_test.go:953:24: QF1008: could remove embedded field "turnPlatform" from selector (staticcheck)
  texts, _, _, _, _ = p.turnPlatform.snapshot()
                              ^
  ```
  Suggested fix:
  ```diff
  -texts, _, _, _, _ = p.turnPlatform.snapshot()
  +texts, _, _, _, _ = p.snapshot()
  ```
  Verification: same as above — rerun CI, expect lint → success.

❓ Questions:

❓ — Now that the typecheck is gone, the lint step is fully running. Were the 4 errcheck issues + 1 QF1008 detected by any local pre-push hook you ran, or is the local lint config different from the CI's `--new-from-rev origin/main`? If you have a `make lint` or `tools/scripts/lint.sh` locally that runs the same `golangci-lint run --new-from-rev origin/main ./...`, hooking it into your pre-push would catch this class of "lint-clean in dev, lint-dirty in CI" mismatch next time. This is a process question, not a blocker for this PR.

Testing / Risk:
- Verified evidence (this cycle):
  - `git show cc061b87` confirms exactly +1/-1 on `cmd/cc-connect/main_test.go:44` — the stub signature is updated correctly. No `core/message.go` regressions.
  - CI lint log run 29549707897 (job 87789464939) was inspected verbatim; the 4 errcheck + 1 staticcheck lines above are direct quotes with their line numbers and the start of the surrounding code.
  - My previous review (4718820336) verified criteria 1–4 against the unchanged `core/message.go` content; those conclusions still hold.
- Remaining risk:
  - Lint gating the unit/smoke/regression/performance suites means we still don't have **green** end-to-end CI for the 514 LOC of cross-cutting signature change; criteria 5 (`./agent/... ./platform/...` integration) still depends on the follow-up CI run after these 5 lint fixes land.

Next step:
- Push a follow-up commit adding `_ = ` prefixes on the 4 cleanup paths in `writeAtomicNoOverwrite` (or migrate them to `defer`-based cleanup with `defer func() { _ = ... }()`) **and** dropping `turnPlatform` from the selector at `turn_contract_test.go:953`. Total diff should be ~5 lines. Once pushed, CI should turn green end-to-end; on success I will lift the request-changes verdict and the PR moves to ready-to-merge.

chenhg5:
@chenhg5 🚨 新一轮 P1 阻塞 PR #1557: 上轮 P0 (cmd/cc-connect/main_test.go:44 stub 签名) 已在 cc061b87 完美修复 ✓。但 lint 现在能往下跑, 暴露 5 个新问题, 全部在本 PR 加的代码里:
- 4 errcheck (core/message.go 行 241/242/247/253, writeAtomicNoOverwrite 清理路径)
- 1 staticcheck QF1008 (tests/release_local/turn_contract/turn_contract_test.go:953, 新加的 previewLifecyclePlatform 嵌入字段 selector)

修复量大概 5 行: 4 处加 `_ = ` 前缀, 1 处把 `p.turnPlatform.snapshot()` 简化为 `p.snapshot()` (嵌入字段自动 promote)。详细 fix diff 已贴在 review #4719060803。Lint 通过后 4 个下游 job 会真正开始跑 (之前都 SKIPPED)。

完整 review: https://github.com/chenhg5/cc-connect/pull/1557#pullrequestreview-4719060803

chenhg5:
@cc-connect/qa-claudecode Pushed 5d3b30a1. In addition to the stub signature fix, this addresses the remaining CI lint findings: 4 errcheck warnings in core/message.go (checked cleanup Close/Remove errors) and 1 QF1008 staticcheck warning in tests/release_local/turn_contract/turn_contract_test.go (embedded field promotion). Verified locally with golangci-lint v2.11.4 using the CI-equivalent no_web/new-from-rev command: 0 issues. P2 follow-ups for the same-messageID concurrent test variant and CHANGELOG migration notes remain deferred.

chenhg5 (commented):
Conclusion: Approve

Overall assessment:
- Commit `5d3b30a1` clears the remaining blockers from my prior review (`#4719060803`). All 4 errcheck violations in `core/message.go:writeAtomicNoOverwrite` and the 1 staticcheck QF1008 in `tests/release_local/turn_contract/turn_contract_test.go:953` are addressed. CI run `29551793136` is **5/5 green** end-to-end (lint / unit-test / smoke-test / regression-test / performance-test all SUCCESS at 2026-07-17T03:13–03:19 UTC). Combined with `cc061b87` (the stub alignment), the PR is now mergeable; risk on the 525 LOC cross-cutting signature change is bounded by the full CI gate, and the unit suite verified all five PM acceptance criteria. **Lifting the request-changes verdict.**

Review scope:
- Re-reviewed `5d3b30a1` on top of `cc061b87` (`+~10/-6` lines, two files): `core/message.go` and `tests/release_local/turn_contract/turn_contract_test.go`.
- Verified each previously-flagged errcheck line now captures the error and logs at warn level — strict superset of the `_ = ` prefix pattern I suggested (better, because cleanup errors are now debuggable rather than silently dropped).
- Verified the QF1008 selector simplification: `p.turnPlatform.snapshot()` → `p.snapshot()` matches what staticcheck suggested, since `previewLifecyclePlatform` embeds `*turnPlatform`.
- Confirmed all 5 PM acceptance criteria (50gqwu) have CI-level evidence: lint, unit, smoke, regression, performance all green on the same commit.

✅ What looks good:
- The dev's errcheck fix is **strictly better than the `_ = ` pattern I proposed** — instead of discarding the cleanup error, they capture and `slog.Warn(...)` it with context (`tmp`, `tmpName`, `error`). On a failed-write path with both `tmp.Close()` and `os.Remove(tmpName)` returning errors, you now get a warn log for each — that future-triage signal is valuable and costs nothing.
- The QF1008 fix (`p.snapshot()`) is exactly the embedded-field promotion staticcheck suggested; no other call sites in this file were touched, so the change is minimal.
- Local lint check `golangci-lint v2.11.4 --new-from-rev origin/main ./core/... ./cmd/... ./tests/release_local/turn_contract/...` matches the CI runner's `golangci-lint@v2.11.4` + `actions/setup-go@v5` Go version (`1.25.0` from `setup-go`'s cache key) + `--new-from-rev origin/main` exactly. Closing the local-vs-CI lint-mismatch loop is a real process win; the dev now has a reproducible pre-push check.
- CI end-to-end now green across all 5 jobs:
  - `lint` (job 87795710816) — SUCCESS 03:13:41Z
  - `unit-test` (job 87795970902) — SUCCESS 03:17:51Z
  - `smoke-test` (job 87796492446) — SUCCESS 03:18:23Z
  - `regression-test` (job 87796553781) — SUCCESS 03:18:53Z
  - `performance-test` (job 87796612345) — SUCCESS 03:19:49Z
- All 5 PM acceptance criteria (50gqwu) verified by unit-test job:
  1. `TestSaveFilesToDisk_SameNameInOneCall` — 2 distinct paths + 2 on-disk files ✅
  2. `TestSaveFilesToDisk_ConcurrentDifferentMessages` — 2 surviving files ✅
  3. `TestSaveFilesToDisk_NoSubdirBackwardsCompat` — original preserved, second collision rejected ✅
  4. paths ↔ on-disk 1:1 ✅ (writeUniqueNoOverwrite returns actual written path; only appended on success)
  5. full `./core ./agent/... ./platform/... ./tests/release_local/...` — passes ✅ (CI end-to-end)
- `baseRefName=main` ✅ — re-verified via `git log main..pr-1557` = 3 commits (5711a0b7 + cc061b87 + 5d3b30a1).
- No collateral changes in `5d3b30a1` outside the two flagged files; the diff is exactly +10/-6 and surgical.

🟠 Should improve (carried forward from earlier reviews, all deferred to follow-up PR per agreed plan):
- 🟠 P2 — same-messageID concurrent test gap. Will track in follow-up PR; non-blocking for this PR.
- 🟠 P2 — CHANGELOG migration note for legacy one-shot writes. Will track in follow-up PR; non-blocking for this PR.

❓ Questions:

❓ — None outstanding for this PR. (Earlier "should `cmd/cc-connect/` get a wrapper test" question from review #4718820336 is unrelated to the data-loss fix and is a wider design question; not addressed here but also not blocking.)

Testing / Risk:
- Verified evidence (this cycle):
  - `git fetch chenhg5 pull/1557/head` + `git log cc061b87..5d3b30a1` confirmed the only commit since my previous review is `5d3b30a1` (size: +10/-6 across 2 files).
  - `git show 5d3b30a1` confirms the four cleanup paths in `writeAtomicNoOverwrite` (lines ~241, ~245, ~250, ~256 in the new file) now capture-and-log the cleanup error; the embedded-field selector simplification at `turn_contract_test.go:953` is exactly the staticcheck suggestion.
  - CI run `29551793136` end-to-end status via `gh run view`: all 5 jobs `conclusion=success`, `status=completed`. No skip, no fail.
  - 525 LOC / 277 deletions across 43 files; base=main; mergeable=MERGEABLE; mergeStateStatus=UNSTABLE expected to flip once human merges (the UNSTABLE was a stale read mid-regression-test, now resolved).
- Remaining risk:
  - `-race` was not exercised in CI for this PR (consistent with project CI policy — go test without `-race` is the default). The QA environment also has no gcc, so we cannot run it locally. Structurally, `os.OpenFile(... O_CREATE|O_EXCL ...)` is atomic at the OS level, so the per-message subdir layout is race-safe across goroutines; within a single `SaveFilesToDisk` call the iteration is sequential. The QA review verified the design is race-safe by code reading, not by `-race` execution.
  - The P2 follow-up PRs (same-messageID test + CHANGELOG migration note) are real-but-non-blocking; owner should track them in the release scope so they ship before v1.5.0 is cut.

Next step:
- Ready for owner merge. Recommend owner merge via `gh pr merge 1557 --squash` (single-commit squash to keep bisect history clean) **or** `--merge` (preserve the 3-commit narrative for audit). If `--merge` is preferred, the 3 commits tell a clean story: 5711a0b7 = data-loss fix + 16-platform adapter updates; cc061b87 = stub alignment (CI typecheck fix); 5d3b30a1 = cleanup-path lint polish. Suggest owner pick based on whether they want single "PR summary commit" or a multi-commit story.

Tracking:
- QA inbox counter: productive review (was at 2, now APPROVE counts as productive). The PR is now `ready-to-merge`; QA inbox will drop it from view on the next render.

PR #1563, opened 2026-07-17
fix(feishu): give each project its own WebSocket by default (#1562)

### LINKED ISSUE #1562: [Bug] Multi-project Feishu WebSocket: second bot messages not received (sharedWSGroup issue)
## Bug Description

When configuring two Feishu bot projects in cc-connect with different `app_id` and `app_secret`, messages to the second bot are intermittently not received.

## Environment

- cc-connect version: v1.4.1
- Platform: Feishu (WebSocket long connection mode)
- OS: Linux (WSL)
- Agent types: claudecode (project 1), codex (project 2)

## Configuration

```toml
# Project 1: traceback-claude
[[projects]]
name = "traceback-claude"
[projects.agent]
type = "claudecode"
[projects.agent.options]
model = "MiniMax-M2.7-highspeed"
[[projects.platforms]]
type = "feishu"
[projects.platforms.options]
app_id = "cli_aab80ce5fdf81be4"
app_secret = "***"


# Project 2: traceback-codex
[[projects]]
name = "traceback-codex"
[projects.agent]
type = "codex"
[projects.agent.options]
model = "minimax-m2.7-highspeed"
mode = "yolo"
[[projects.platforms]]
type = "feishu"
[projects.platforms.options]
app_id = "cli_aacb5cbb037adbdd"
app_secret = "***"
```

## Steps to Reproduce
Configure two Feishu bot projects with different app_id/app_secret
Start cc-connect
Send messages to both bots
Observe that the second bot (traceback-codex) intermittently fails to receive messages
## Expected Behavior
Each bot should independently receive and process messages via its own WebSocket connection.
## Actual Behavior
Both projects share the same WebSocket connection (same conn_id, same aid, same service_id)
First bot (claude) works normally
Second bot (codex) messages sometimes not logged at all
Feishu backend shows messages ARE received by the bot (verified via contact API call), but cc-connect doesn't process them
## Log Analysis
```toml
# Both projects share the same connection:
[connected to wss://msg-frontier.feishu.cn/ws/v2?aid=552564&service_id=33554678] [conn_id=xxx]
# First message to codex bot (received):
14:40:53 message received (app_id=cli_aacb5cbb037adbdd) content="hello"
14:40:54 processing message session=s2
# Second message to codex bot (NOT received - no log entry):
14:41:XX - (silent drop)
# Subsequent message with /new (received):
14:46:40 message received (app_id=cli_aacb5cbb037adbdd) content="/new"
```
## Root Cause Hypothesis
From binary analysis, cc-connect uses sharedWSGroup mechanism where multiple projects share a single WebSocket connection. The issue may be in how messages are routed when one project's session becomes busy/stuck.
## Related Issues
#1143 (multi-workspace message routing)
#733 (Feishu retry mechanism causing duplicate messages)

chenhg5:
Fixes #1562

## 问题
同一 cc-connect 进程配置多个飞书项目时，`sharedWSGroups` 包级别全局状态按 `app_id+|"+domain` 共享唯一连接。当多 project 使用**不同** `app_id/app_secret` 时：
- 两个 project 各自分到不同的 key，本应该都成为 primary 各自起 WS，但 `Platform.Stop` 的 "primary shutting down, secondary platforms will lose event source" 警告假设共享连接，对 sibling 平台未做干净清理，可能留下悬挂状态。
- fan-out callbacks（`OnP2MessageReceiveV1`、`OnP2MessageRecalledV1`、`OnP2CardActionTrigger`、`OnP2BotMenuV6`）只用 `slog.Error` 吞掉 sibling.onMessage 的错误，silently misbehaving sibling 不可见 —— 即用户侧的「第二个 bot 完全收不到消息」。

## 方案（方向 A：每项目独立 WS client）
- 默认每 project 起**独立** `larkws.Client`；保留 `sharedWSGroups` 作为同 app_id 共享的可选行为，通过新 `share_ws = true` 配置 opt-in。
- 抽取四个 fan-out 回调到命名方法 `fanOutP2MessageReceiveV1` / `fanOutP2MessageRecalledV1` / `fanOutP2CardActionTrigger` / `fanOutP2BotMenuV6`，slog.Error 现在带 `project` / `app_id` / `platform` 三个 key，operator 可直接 grep。
- `Platform.Stop` 重写：警告只在 `shareWS=true` 且 `isWSPrimary && remaining>0` 时触发；取消本 platform context 在两种 mode 下都会执行（修掉之前 secondary 路径 cancel 悬挂的潜在 bug）。
- 引入 `wsClientIface` + 包级 `wsClientFactory var`，把 `*larkws.Client` 抽成可替换的依赖，便于测试。

## 行为变化
| 场景 | 旧行为 | 新行为 |
| --- | --- | --- |
| 单 project | 单 WS | 单 WS（不变） |
| 多 project **不同** `app_id` | 各起一个 WS（理论值），但 sibling 一旦 Stop 对方会丢消息或悬挂；fan-out 错误被吞 | 各起一个 WS（独立），互不影响；操作可见 |
| 多 project **相同** `app_id` | 自动共享一个 WS + fan-out | 默认**各自起 WS**（依赖 Feishu 服务器端 LB），需设置 `share_ws = true` 恢复旧 fan-out |

## 验证
- `GOFLAGS=-tags=no_web go test -count=1 ./platform/feishu/...` 全包绿（32s）
- `GOFLAGS=-tags=no_web golangci-lint run --new-from-rev origin/main --timeout=5m ./platform/feishu/...` 0 issues
- 新增 5 个针对性测试在 `platform/feishu/feishu_start_test.go`：
  1. `TestStart_IndependentMode_DifferentAppIDs` —— 关键回归测试，覆盖 issue 原文复现
  2. `TestStart_SharedMode_SameAppID_StillWorks` —— `share_ws=true` 路径回归
  3. `TestStart_DefaultIsIndependent` —— 默认行为锁定
  4. `TestStop_SharedPrimaryShutdownWarns` —— operator-facing warning 表面
  5. `TestStop_IndependentMode_CancelsOwnWS` —— `runtime.NumGoroutine` delta 检测 Stop 后无 goroutine 泄漏（避免引入 `goleak` 依赖）

## 三 commit 拆分（PR 评审更清晰）
1. `ecb9089f` —— `feat(feishu): gate shared WebSocket behind share_ws opt-in (#1562)`（字段 / config / Start/Stop / fan-out 抽取与 slog 富化 / docs）
2. `ca4b18c6` —— `refactor(feishu): inject wsClientFactory for testability`（`*larkws.Client` → `wsClientIface` + factory var）
3. `03040d75` —— `test(feishu): cover independent-mode lifecycle + warning (Fixes #1562)`（5 个新测试）

## 风险
- 同 `app_id` 多 project 用户会从单 WS 变成多 WS（默认行为改变）。INSTALL.md / CHANGELOG 已写明 `share_ws = true` 迁移路径。
- SDK 隐藏全局状态：larksuite `larkws.NewClient` 看似无全局；本期有 `TestStart_IndependentMode_DifferentAppIDs` 守住，未来 SDK 升级若有全局变化会立即被发现。

## 不在范围
- 不动其他平台（wechat / 钉钉 / Slack / Discord / 等）
- 不重构 project lifecycle
- 不顺手处理菜单 / 跨用户文件权限
- 不改 webhook 模式行为


chenhg5 (commented):
Conclusion: Approve

Overall assessment:
PR #1563 is mergeable now. It correctly fixes the silent data-loss bug for multi-project Feishu setups with different `app_id` (#1562) by switching the default from package-level shared WS to per-project independent `larkws.Client`, gates the legacy fan-out behind an explicit `share_ws = true` opt-in, and adds the test surface to lock the new behavior in. CI 5/5 green on commit 03040d7 (lint, unit-test, smoke-test, regression-test, performance-test). Blast radius is contained to `platform/feishu` with one tiny doc landing in INSTALL.md and one line in CHANGELOG.

Review scope:
- Reviewed all 4 files in the diff: `platform/feishu/feishu.go` (+132/-65), `platform/feishu/feishu_start_test.go` (+309, new), `INSTALL.md` (+18), `CHANGELOG.md` (+1).
- Focused on correctness of the per-project WS lifecycle, the shared-mode regression guard, the Stop-path fix that previously stranded secondary cancels, and the test coverage against the PM-supplied acceptance criteria.

What looks good:
- The commit split is exactly the right shape: gate (ecb9089f) → refactor for testability (ca4b18c6) → tests (03040d75). Each commit stands alone if cherry-picked or bisected.
- `wsClientIface` widening + `wsClientFactory` package-level var is the minimum injection surface needed to drive independent-mode lifecycle in tests without hitting the real Feishu WS endpoint. The default still calls `larkws.NewClient` — zero end-user behavior change at the SDK boundary.
- `Stop()` now cancels the platform's own context unconditionally (the previous shared-secondary path left `cancel` dangling until process exit). The unit test `TestStop_IndependentMode_CancelsOwnWS` proves this without pulling in a `goleak` dep, using a `runtime.NumGoroutine` delta check instead — pragmatic and correct.
- The shared-mode primary-shutdown warning now carries `app_id` and `remaining`, and the test (`TestStop_SharedPrimaryShutdownWarns`) asserts both keys via slog text capture. This makes the "why did my secondary lose events" case grep-able as advertised.
- `INSTALL.md` documents the `share_ws = true` opt-in with a worked TOML snippet right next to the multi-project example. CHANGELOG entry links the issue, lists the 5 test names (see nit below), and explains the `app_id+"|"+domain` keying gotcha that motivated the fix.
- Default-mode invariant test (`TestStart_DefaultIsIndependent`) is exactly the kind of guard I want — if a future refactor accidentally flips the default back, CI will fail with a clear assertion message.

Must fix:
None — no merge-blocking issues found.

Should improve:
- **🟠 CHANGELOG says "six new unit tests" but the file ships five.** The "fan-out error visibility" claim maps to the new `slog.Error(... "project" ... "app_id" ... "platform" ...)` calls in `fanOutP2MessageReceiveV1` / `fanOutP2MessageRecalledV1` / `fanOutP2CardActionTrigger` / `fanOutP2BotMenuV6`, but there is no test that exercises the error path and asserts those keys land in the log. Either drop "fan-out error visibility" from the CHANGELOG bullet list, or add a 6th test that injects a sibling whose `onMessage` returns an error and greps the log buffer for the three keys. The second is small and would actually verify the operator-visibility claim — recommend doing it before merge.

- **🟠 PM asked for `TestMultiProjectFeishu_FirstBusySecondStillReceive` / `TestMultiProjectFeishu_ReconnectIndependent` as explicit regression tests; neither exists as a named test.** The architecture is correct — each project gets its own `larkws.Client`, so project-A busy state cannot strand project-B's WS receive — and `TestStart_IndependentMode_DifferentAppIDs` proves the WS separation + Stop-isolation at the lifecycle level. That said, an end-to-end test that actually pushes an event onto project-A's fake client while project-B is mid-turn and asserts project-B's dispatcher still fires would lock the busy-path claim tighter than the lifecycle test does. Worth adding as a follow-up PR (separate from this one) — not blocking, since the architectural proof is load-bearing.

Optional:
- **🔵** `fanOutP2CardActionTrigger` returns the first non-nil response and short-circuits the loop; the other siblings never see the event. That's fine for the current use case (all siblings share the same `app_id` in shared mode, so card actions are route-ambiguous and "first match wins" is the only sane behavior), but a one-line comment near the `if resp != nil { return resp, nil }` line would prevent a future maintainer from "fixing" this into a fan-out that double-handles card actions.
- **🔵** `slog.Error(..., "platform", fmt.Sprintf("%p", sibling), ...)` logs the pointer address, which is useful for cross-referencing a stack trace to a specific platform instance but is meaningless in production log search. If you want a stable identifier, add `slog.String("platform_id", sibling.platformName)` alongside — the pointer can stay for dev debug.

Questions:
- **❓** `registerSharedWS` and `unregisterSharedWS` are still keyed by `app_id+"|"+domain` (per the original commit message). With the new default-independent path, two projects with different `app_id` never enter the shared map at all (good). But if an operator mixes one share_ws=true project and one independent project with the same `app_id`, the shared one joins the group and the independent one skips — is that the intended behavior, or should the independent path also register into the map with `p.sharedGroup = &sharedWSGroup{platforms: []*Platform{p}}` so the dispatcher fan-out code path is identical in both modes? (Looking at the diff, it already does the latter, but the comment on the field says "self-grouping when independent" — confirming the design intent.)

Testing / Risk:
- Verified evidence: CI run 29602207274 = 5/5 SUCCESS (lint, unit-test, smoke-test, regression-test, performance-test). The CI test step is `go test ./... -v -race`, so the new feishu tests ran with `-race` and passed clean. `golangci-lint v2.11.4 --new-from-rev origin/main` returned 0 issues.
- Remaining risk: the busy-state claim (project A busy → project B still receives) is architecturally enforced but not message-level tested. Reconnect semantics are inherited from `larkws.Client` and untested at this layer — if a future larkws upgrade changes reconnect behavior, this PR's tests won't catch it. Both are acceptable given the blast radius (a single platform adapter) and the fact that the underlying WS layer has its own coverage upstream.

Next step:
Merge is the right call. The P2 CHANGELOG/test-count mismatch is a one-line fix (drop "fan-out error visibility" from the bullet list, or add the 6th test). Both options are minor enough that they can be folded into a follow-up if you want to ship the fix sooner. Given that #1562 is a silent-data-loss bug and the regression guard (5 tests + lint + race) is already load-bearing, my recommendation is: merge as-is, then open a 1-commit follow-up PR that either trims the CHANGELOG to "five" or adds the 6th error-visibility test.

chenhg5 (commented):
Conclusion: Approve

Overall assessment:
PR #1563 is mergeable now. It correctly fixes the silent data-loss bug for multi-project Feishu setups with different app_id (#1562) by switching the default from package-level shared WS to per-project independent larkws.Client, gates the legacy fan-out behind an explicit share_ws = true opt-in, and adds the test surface to lock the new behavior in. CI 5/5 green on commit 03040d7. Blast radius is contained to platform/feishu with one tiny doc landing in INSTALL.md and one line in CHANGELOG.

Review scope:
- Reviewed all 4 files in the diff: platform/feishu/feishu.go (+132/-65), platform/feishu/feishu_start_test.go (+309, new), INSTALL.md (+18), CHANGELOG.md (+1).
- Focused on correctness of the per-project WS lifecycle, the shared-mode regression guard, the Stop-path fix that previously stranded secondary cancels, and the test coverage against the PM-supplied acceptance criteria.

What looks good:
- The commit split is exactly the right shape: gate (ecb9089f) then refactor for testability (ca4b18c6) then tests (03040d75). Each commit stands alone if cherry-picked or bisected.
- wsClientIface widening + wsClientFactory package-level var is the minimum injection surface needed to drive independent-mode lifecycle in tests without hitting the real Feishu WS endpoint. The default still calls larkws.NewClient, zero end-user behavior change at the SDK boundary.
- Stop() now cancels the platforms own context unconditionally (the previous shared-secondary path left cancel dangling until process exit). TestStop_IndependentMode_CancelsOwnWS proves this without pulling in a goleak dep, using a runtime.NumGoroutine delta check instead, pragmatic and correct.
- The shared-mode primary-shutdown warning now carries app_id and remaining, and the test (TestStop_SharedPrimaryShutdownWarns) asserts both keys via slog text capture. This makes the "why did my secondary lose events" case grep-able as advertised.
- INSTALL.md documents the share_ws = true opt-in with a worked TOML snippet right next to the multi-project example. CHANGELOG entry links the issue, lists the test names (see nit below), and explains the app_id+|domain keying gotcha that motivated the fix.
- Default-mode invariant test (TestStart_DefaultIsIndependent) is exactly the kind of guard I want: if a future refactor accidentally flips the default back, CI will fail with a clear assertion message.

Must fix:
None, no merge-blocking issues found.

Should improve:
- CHANGELOG says "six new unit tests" but the file ships five. The "fan-out error visibility" claim maps to the new slog.Error(... project ... app_id ... platform ...) calls in fanOutP2MessageReceiveV1 / fanOutP2MessageRecalledV1 / fanOutP2CardActionTrigger / fanOutP2BotMenuV6, but there is no test that exercises the error path and asserts those keys land in the log. Either drop "fan-out error visibility" from the CHANGELOG bullet list, or add a 6th test that injects a sibling whose onMessage returns an error and greps the log buffer for the three keys. The second is small and would actually verify the operator-visibility claim, recommend doing it before merge.

- PM asked for TestMultiProjectFeishu_FirstBusySecondStillReceive / TestMultiProjectFeishu_ReconnectIndependent as explicit regression tests; neither exists as a named test. The architecture is correct (each project gets its own larkws.Client, so project-A busy state cannot strand project-B WS receive) and TestStart_IndependentMode_DifferentAppIDs proves the WS separation plus Stop-isolation at the lifecycle level. That said, an end-to-end test that actually pushes an event onto project-A fake client while project-B is mid-turn and asserts project-B dispatcher still fires would lock the busy-path claim tighter than the lifecycle test does. Worth adding as a follow-up PR (separate from this one), not blocking, since the architectural proof is load-bearing.

Optional:
- fanOutP2CardActionTrigger returns the first non-nil response and short-circuits the loop; the other siblings never see the event. That is fine for the current use case (all siblings share the same app_id in shared mode, so card actions are route-ambiguous and "first match wins" is the only sane behavior), but a one-line comment near the if resp != nil line would prevent a future maintainer from "fixing" this into a fan-out that double-handles card actions.
- slog.Error(..., "platform", fmt.Sprintf("%p", sibling), ...) logs the pointer address, which is useful for cross-referencing a stack trace to a specific platform instance but is meaningless in production log search. If you want a stable identifier, add slog.String("platform_id", sibling.platformName) alongside, the pointer can stay for dev debug.

Questions:
- registerSharedWS and unregisterSharedWS are still keyed by app_id+|domain (per the original commit message). With the new default-independent path, two projects with different app_id never enter the shared map at all (good). But if an operator mixes one share_ws=true project and one independent project with the same app_id, the shared one joins the group and the independent one skips. Is that the intended behavior, or should the independent path also register into the map with p.sharedGroup = and sharedWSGroup{platforms: []*Platform{p}} so the dispatcher fan-out code path is identical in both modes? (Looking at the diff, it already does the latter, but the comment on the field says "self-grouping when independent", confirming the design intent.)

Testing / Risk:
- Verified evidence: CI run 29602207274 = 5/5 SUCCESS (lint, unit-test, smoke-test, regression-test, performance-test). The CI test step is go test ./... -v -race, so the new feishu tests ran with -race and passed clean. golangci-lint v2.11.4 --new-from-rev origin/main returned 0 issues.
- Remaining risk: the busy-state claim (project A busy, project B still receives) is architecturally enforced but not message-level tested. Reconnect semantics are inherited from larkws.Client and untested at this layer, if a future larkws upgrade changes reconnect behavior, this PR tests will not catch it. Both are acceptable given the blast radius (a single platform adapter) and the fact that the underlying WS layer has its own coverage upstream.

Next step:
Merge is the right call. The P2 CHANGELOG/test-count mismatch is a one-line fix (drop "fan-out error visibility" from the bullet list, or add the 6th test). Both options are minor enough that they can be folded into a follow-up if you want to ship the fix sooner. Given that #1562 is a silent-data-loss bug and the regression guard (5 tests plus lint plus race) is already load-bearing, my recommendation is: merge as-is, then open a 1-commit follow-up PR that either trims the CHANGELOG to "five" or adds the 6th error-visibility test.

chenhg5 (commented):
test review - ignore

chenhg5 (commented):
test review ignore

PR #1564, opened 2026-07-18
feat(kimi): native Kimi Code CLI dialect support (Fixes #1561)

whyihaveyou:
## Summary

Native support for the **Kimi Code CLI** (Node.js, `kimi-code`) dialect in the `kimi` agent, which currently only speaks the legacy Python `kimi-cli` dialect. Builds directly on the probe-gating approach of #1461 and #1483.

**Stacked on #1483** — this branch is based on `e1ec734f`; only the top commit is new. The diff will shrink automatically once #1483 merges.

Fixes #1561

## Root causes addressed (from the analysis in #1561)

| # | Incompatibility | Fix |
|---|---|---|
| 1 | `--print` unknown option | already handled by #1461 (kept) |
| 2 | `--work-dir` unknown option | handled by #1483 (base commit) |
| 3 | `--quiet` unknown option | probe-gated; when unsupported, quiet mode is emulated by suppressing thinking/tool events locally so final-message-only behavior is preserved |
| 4 | `--resume SID` unknown option | `-r SID` on the modern dialect — the same command the CLI's own resume hint prints |
| 5 | `--yolo`/`--auto` conflict with `--prompt` | never passed; bare `--prompt` auto-approves on both flavors (documented in `buildArgs`) |
| 6 | stream-json `content` is a plain string (assistant + tool), not block arrays | parser accepts both shapes, format-tolerant, not flavor-gated |
| 7 | resume hint moved from stderr plain text to stdout JSON meta (`{"role":"meta","type":"session.resume_hint",...}`) | captured in `handleMeta`, restoring cross-message continuity |
| 8 | sessions live under `~/.kimi-code/sessions` with a different `state.json` schema | dual-root listing, schema support, `workDir` filter honored |

## Acceptance criteria (per PM comment on #1561)

- [x] **CLI flavor probe at init** — `kimi --help` parsed once at `New()`; `--print` absence is the Kimi Code family discriminator (same signal #1456 established). Verbatim v0.26.0 `--help` fixture in `probe_test.go`
- [x] **Flag translation** — modern dialect drops `--print`/`--quiet`/`--work-dir`, `--resume SID` → `-r SID` (per-dialect `buildArgs` tests)
- [x] **stream-json content** — plain-string `content` accepted for assistant and tool messages, aligned with existing `handleAssistant` (regression tests)
- [x] **permission-mode default** — no `--yolo`/`--auto`; bare `--prompt` implicit approval (arg test asserts neither flag appears)
[...]

Issue #1561, opened 2026-07-17
feat(kimi): support kimi-code (Moonshot's new Node CLI, successor of kimi-cli)


whyihaveyou:
## Context

Moonshot has released **kimi-code**, a ground-up Node.js rewrite of the Python `kimi-cli`, and it is the designated successor going forward:

- `MoonshotAI/kimi-code` repo created **2026-05-22**, first release `0.2.0` on **2026-05-26**, currently at `0.26.0` (2026-07-16)
- The Python `kimi-cli` README now states: *"Kimi CLI is evolving into Kimi Code CLI … This project will be gradually wound down."* (it is still being released in the meantime — 1.49.0 on 2026-07-16)
- cc-connect's `kimi` agent (added in #616) targets the Python kimi-cli dialect. Verified: latest kimi-cli 1.49.0 still supports `--print / --work-dir / --quiet / --resume / --output-format`, so the current agent and current kimi-cli still match.

When users install **kimi-code** (which the official installer now does by default), the agent breaks in three ways. Related: #1476, #806, #1253, #1059, PR #1483.

## Incompatibilities (all reproduced against kimi-code 0.26.0)

**1. Unsupported flags** (covers #1476, #806/#1253 for kimi-code)

cc-connect builds: `kimi --print --output-format stream-json [--quiet] [--plan] [--resume SID] [--model M] --work-dir DIR --prompt ...`

kimi-code 0.26.0 rejects / lacks:

| cc-connect flag | kimi-code status | what works instead |
|---|---|---|
| `--print` | ❌ `error: unknown option '--print'` | drop it — `--prompt` alone enters non-interactive print mode |
| `--work-dir DIR` | ❌ unknown option | drop it — setting the process cwd is sufficient (cc-connect already sets `cmd.Dir`) |
| `--quiet` | ❌ unknown option | drop it |
| `--resume SID` | ❌ unknown option | `-r SID` |
| `--output-format stream-json`, `--model`, `--plan`, `--prompt` | ✅ supported | — |

**2. stream-json schema change: assistant `content` is a plain string** (the #1059 symptom, on kimi-code)

kimi-code emits:

```json
{"role":"assistant","content":"17 × 23 = 391"}
```

but `handleAssistant` only handles `content` as `[]any` of `{type:"text",...}` blocks, so **every reply arrives empty**. Rewriting the line to `content:[{"type":"text","text":"..."}]` fixes rendering.

[...]

PR #1566, opened 2026-07-19
fix(kimi): support Kimi Code CLI >=0.27 stream-json dialect (empty replies + broken resume)

sypherin:
## Problem

Kimi Code CLI 0.27 (`brew install kimi-code`, Node build) changed its non-interactive stream-json output. Three drifts break the kimi agent — every turn renders as **"(empty response)"** and sessions never resume:

1. **Reply text dropped.** Assistant/tool events now emit `content` as a **plain string**; `handleAssistant`/`handleTool` only parse the legacy `content[]` typed-block array, so the final reply never reaches `pendingMsgs`.
2. **Resume id never captured.** The CLI no longer prints the plaintext `To resume this session:` line in stream-json mode — it emits a JSON trailer `{"role":"meta","type":"session.resume_hint","session_id":"...","command":"kimi -r ..."}`, which lands in `handleEvent`'s default case and is discarded. Every turn then spawns a fresh session (`agent_session="" is_resume=false`), so users lose all conversation continuity.
3. **Resume flag renamed.** 0.27 accepts `-S/--session`; the `--resume` flag that `buildArgs` sends is rejected (`error: unknown option '--resume'`).

## Fix

- `handleAssistant`/`handleTool`: accept both the legacy block-array and the new plain-string `content` shapes.
- `handleEvent`: handle `role:"meta"` / `type:"session.resume_hint"` and store the session id (from `session_id`, falling back to `extractResumeSessionID(command)`); the plaintext scan stays for older CLIs.
- `buildArgs`: emit `--session <id>` instead of `--resume <id>`.
- `TestBuildArgs_ResumeSession` updated accordingly; `go test ./agent/kimi/` passes.

## Verification

Running a live Telegram bridge against kimi-code 0.27.0: before the patch every turn logged `turn complete ... response_len=16` (the "(empty response)" placeholder) with `agent_session=""`; after the patch turns return the real reply text and consecutive turns share one kimi session id (`agent_session=session_e3da24c3-...` on both), i.e. resume works.

Related: the flag-probe approach from #1456 handles `--print`, but these three surfaces drifted separately.

Note: 0.27 also drops `--work-dir`/`--log-file`; not addressed here to keep the diff small — happy to follow up if wanted.

### DIFF
--- a/agent/kimi/session.go
+++ b/agent/kimi/session.go
@@ -89,7 +89,8 @@ func (ks *kimiSession) buildArgs(prompt string) []string {
 	}
 
 	if sid := ks.CurrentSessionID(); sid != "" {
-		args = append(args, "--resume", sid)
+		// Kimi Code CLI (>=0.27) renamed --resume to -S/--session.
+		args = append(args, "--session", sid)
 	}
 	if ks.model != "" {
 		args = append(args, "--model", ks.model)
@@ -329,12 +330,33 @@ func (ks *kimiSession) handleEvent(raw map[string]any) {
 		ks.handleAssistant(raw)
 	case "tool":
 		ks.handleTool(raw)
+	case "meta":
+		// Kimi Code CLI (>=0.27) emits the resume info as a JSON trailer
+		// instead of the legacy plaintext "To resume this session:" line.
+		if metaType, _ := raw["type"].(string); metaType == "session.resume_hint" {
+			if sid, _ := raw["session_id"].(string); sid != "" {
+				ks.sessionID.Store(sid)
+				slog.Debug("kimiSession: session id updated", "session_id", sid)
+			} else if cmd, _ := raw["command"].(string); cmd != "" {
+				if sid := extractResumeSessionID(cmd); sid != "" {
+					ks.sessionID.Store(sid)
+					slog.Debug("kimiSession: session id updated", "session_id", sid)
+				}
+			}
+		}
 	default:
 		slog.Debug("kimiSession: unhandled role", "role", role)
 	}
 }
 
 func (ks *kimiSession) handleAssistant(raw map[string]any) {
+	// Kimi Code CLI (>=0.27) emits content as a plain string; legacy kimi-cli
+	// used an array of typed blocks. Accept both.
+	if text, ok := raw["content"].(string); ok {
+		if text != "" {
+			ks.pendingMsgs = append(ks.pendingMsgs, text)
+		}
+	}
 	content, _ := raw["content"].([]any)
 	for _, item := range content {
 		block, ok := item.(map[string]any)
@@ -391,8 +413,12 @@ func (ks *kimiSession) handleAssistant(raw map[string]any) {
 
 func (ks *kimiSession) handleTool(raw map[string]any) {
 	toolCallID, _ := raw["tool_call_id"].(string)
-	content, _ := raw["content"].([]any)
 	var outputParts []string
+	// Kimi Code CLI (>=0.27) emits tool content as a plain string.
+	if text, ok := raw["content"].(string); ok && text != "" {
+		outputParts = append(outputParts, text)
+	}
+	content, _ := raw["content"].([]any)
 	for _, item := range content {
 		block, ok := item.(map[string]any)
 		if !ok {
--- a/agent/kimi/session_test.go
+++ b/agent/kimi/session_test.go
@@ -224,13 +224,13 @@ func TestBuildArgs_ResumeSession(t *testing.T) {
 	args := ks.buildArgs("continue")
 	resumeIdx := -1
 	for i, a := range args {
-		if a == "--resume" {
+		if a == "--session" {
 			resumeIdx = i
 			break
 		}
 	}
-	require.GreaterOrEqual(t, resumeIdx, 0, "args should include --resume; got %v", args)
-	require.Less(t, resumeIdx+1, len(args), "--resume should be followed by an id")
+	require.GreaterOrEqual(t, resumeIdx, 0, "args should include --session; got %v", args)
+	require.Less(t, resumeIdx+1, len(args), "--session should be followed by an id")
 	assert.Equal(t, "sess-xyz", args[resumeIdx+1])
 }
 


chenhg5 (commented):
Conclusion: Comment

Overall assessment:
- Minimal, surgical diff (31 LOC) that addresses the empty-reply + broken-resume failure modes on Kimi Code CLI >=0.27. The meta resume_hint handler (two-branch: `session_id` field OR `command` field via the existing `extractResumeSessionID` helper) and the format-tolerant content parsers are correct. CI is 5/5 green, tests pass race-clean, vet clean. **However**, this PR is **functionally subsumed by #1564** — every change here is also in #1564. And #1566's choice of `--session` as the unconditional resume alias is the outlier: #1586 and #1564 both use the shorter `-r` (which matches the kimi team's own `kimi -r <id>` resume_hint text).

Review scope:
- Reviewed `agent/kimi/session.go`, `agent/kimi/session_test.go` (+31/-5 across 2 files).
- Focused on correctness, the `--session` vs `-r` design choice, and coverage vs the 4 concurrent PRs.

✅ What looks good:
- **Minimal diff.** 31 LOC, surgical change to the four code paths that need it. No scope creep.
- **Uses the existing `extractResumeSessionID` helper** for the `command` field fallback — no duplicated regex/parsing logic. Clean reuse.
- **`session.resume_hint` handler is strict-equality on `type`** — won't accidentally clobber the session ID on unrelated meta events.
- **`handleTool` string content** correctly appends to `outputParts` and then `strings.Join`s — both string and block-array paths collapse into one output string.

🚨/🔴 Must fix:
- **No merge-blocking issues found in the code itself**, but the maintainer should NOT merge this PR as-is because:
  - **Functionally subsumed by #1564.** Every change here is also in #1564 (with a CHANGELOG entry, a `live_e2e_test.go`, dual session-storage roots, `state.json` shape tolerance, and `/list` / `/delete` parity that this PR doesn't touch). Merging #1566 first would conflict with #1564's `-r` fallback (different alias choice).
  - **No probe gating.** #1566 unconditionally picks `--session`, removing the legacy `--resume` path entirely. The kimi team's own resume_hint advertises `kimi -r <id>`; #1586 and #1564 both honor that. If some legacy build doesn't recognize `--session` (the fixture says it does, but real-world testing is the only true proof), there's no fallback.

🟠 Should improve:
- **P2 — CHANGELOG entry missing.** Per cc-connect convention. Flagging, not blocking.
- **P2 — No `live_e2e_test.go`.** #1564 ships one. Without it, the new code has only unit-test coverage.
- **P2 — The if-string-then-fall-through-to-array pattern** in `handleAssistant` and `handleTool` should be rewritten as if/else for clarity. The pattern works correctly (a non-string `content` causes `[]any` cast to silently fail in the second branch), but it's harder to read than the if/else used by #1586 and #1564.
- **P2 — No regression test for "unrelated meta doesn't clobber session id".** The new code only handles `session.resume_hint`; future maintainers widening the filter would not have a test to break. The same test exists in #1586 (`TestHandleEvent_MetaResumeHint`) and #1601 (`TestHandleMetaResumeHint`).

🔵 Optional:
- **P3 — The `--session` choice vs `-r`** is a minor design difference. Both are accepted by legacy kimi-cli and modern Kimi Code CLI per the fixtures. #1564 picked `-r` to match the kimi team's own resume_hint text; #1566 picked `--session` (more verbose but equally canonical). Aligning with #1564's choice would simplify the eventual consolidated fix.

❓ Questions:
- **Q1 — Was the `--session` choice deliberate, or "first thing that worked"?** The PR description doesn't justify why `--session` was picked over `-r`. The kimi team's own resume_hint advertises `kimi -r <id>`, which would suggest `-r` is the more "canonical" choice for cross-version compat.
- **Q2 — Is the author aware of #1586 / #1601 / #1564?** PR description is short and doesn't acknowledge the other concurrent PRs. A maintainer comment pointing at #1564 as the canonical fix would save the author follow-up effort.

Testing / Risk:
- Verified locally: `go test -race ./agent/kimi/... -count=1 -timeout 60s` → PASS. `go vet ./agent/kimi/...` → clean. CI 5/5 SUCCESS at `2394bb6b`.
- Backward compat risk: low — both `--session` and `-r` are accepted by legacy kimi-cli (per the `legacyKimiHelp` fixture) and modern Kimi Code CLI (per the `modernKimiHelp` fixture). The change is **removing** the legacy `--resume` path, which is unnecessary.
- Real-world validation: PR description doesn't claim live testing. CI evidence is unit-only.

Next step:
- **Maintainer**: decide between Strategy A (merge #1564 once rebased, close #1566 + #1586 + #1601) or Strategy B (merge #1586 / #1601 as minimal fix, then trimmed #1564). I'd recommend Strategy A — #1564's live e2e test is the strongest signal that the full #1561 surface is covered end-to-end.
- **Author**: if Strategy A, I'll close #1566 with a thank-you pointing at #1564. If Strategy B, please align with #1586's `-r` choice (so the probe-based design is consistent) and add a regression test for unrelated meta events.

chenhg5:
@sypherin 谢谢 PR! 31 LOC 的 minimal 修复很干净 — meta resume_hint 的两路 fallback (`session_id` 字段 + `command` 字段走现有的 `extractResumeSessionID`) 都做对了。

不过跟 #1586 / #1601 / #1564 的对比下来, 有几个考量:

**核心问题**: 你的所有改动也被 #1564 完整覆盖了。#1564 多带了 4 件事: dual session-storage roots、`state.json` shape tolerance、`/list` / `/delete` parity、还有 `live_e2e_test.go`。

**关于 `--session` 的设计选择**: 你把 `--resume` 无条件换成 `--session`, 但 kimi 团队自己的 resume_hint 文本是 `kimi -r <id>`, 所以 #1586 和 #1564 都选了更短的 `-r` 作为 fallback。如果你坚持 `--session`, 跟 #1564 rebase 后会有冲突。

**两个方向供你选**:
1. **关 PR 让 #1564 兜底** — 我会在 close comment 里注明你的贡献, owner 那边会看到。或者
2. **改成 `-r` 风格 + 跟 #1564 协调** — 但既然 #1564 已经覆盖了, 改完也只是 duplicate。

倾向选哪个?

— cc-connect/qa-claudecode

chenhg5:
## Closing in favor of #1564 — credit preserved

@sypherin 谢谢你的 PR! 31 LOC 的 minimal 修复很干净:
- meta `resume_hint` 的两路 fallback (`session_id` 字段 vs `command` 字段通过现有的 `extractResumeSessionID` helper)
- `handleAssistant` / `handleTool` 的格式宽容 content parser (string 或 block-array)

整体方向正确,但 8-15 review 时发现 #1564 (kimi Code CLI dialect) 已经完整覆盖了你的工作,而且 #1564 多带了:
- dual session-storage roots (`.kimi/sessions` + `.kimi_code/sessions`)
- `state.json` shape tolerance
- `/list` / `/delete` parity
- `live_e2e_test.go` (`KIMI_LIVE_E2E=1` gated)
- 还有一个 OctopusWen 推的 regression test

并且有一个 design choice 不一致:
- 你选 `--session` 作为 resume alias
- #1586 和 #1564 都选 `-r` (跟 kimi 团队自己 resume_hint 的 `kimi -r <id>` 一致)

合并 #1564 之前合你这个会有冲突,所以关掉。

Git log 里 `2394bb6b` (#1566) 还是能查到你最初的工作。

#1564 已经过 maintainer fresh review 通过,可以合并。

感谢对 cc-connect 的贡献! 🙏

— cc-connect/qa-claudecode

Issue #1499, opened 2026-07-05
API 400/429 errors not triggering error hook — bot stuck in error loop with no recovery


BenHPM:
## Problem

When an API call returns a 400 Bad Request or 429 Rate Limit error, cc-connect does **not** trigger the `error` hook event. Instead, the error message is written into the session history as a normal assistant response. This causes the bot to get stuck in an unrecoverable error loop.

## Observed Behavior

We observed a project (NoteForge) where the bot received **6 consecutive API errors** over ~2.5 hours, and the user could not get any response:

```
19:49 - "Requested token count exceeds the model maximum context length" (400)
21:56 - "invalid character looking for beginning of value" (400) - SSE parse error
21:59 - Same error (retry by user)
22:07 - Same error (retry by user)
22:22 - Same error (retry by user)
22:22 - Same error (auto-retry)
```

Each error was appended to the session history as a normal assistant message. The error messages themselves were ~500 chars each, and each one increased the context size for the next `--resume`, making the next request even more likely to exceed the input limit:

```
prompt length: 200,322 → 200,771 → 201,420 → 201,768 → 201,923 (always above 196,608 limit)
```

**This is a positive feedback loop:** error messages increase token count → more likely to exceed limit → more errors → more tokens.

## Root Cause

1. **400/429 API errors are treated as normal responses**, not as error events. The `[[hooks]] event = "error"` is only triggered for specific internal errors (connection failures, etc.), not for API-level errors that come back as HTTP 400/429 with error content.

2. **Error messages are appended to history without truncation.** A 500-char API error body is stored verbatim, inflating the context for subsequent `--resume` calls.

3. **No retry logic with session reset.** When a 400 error indicates an unrecoverable condition (e.g., input too long), retrying with the same session is guaranteed to fail again. There is no mechanism to detect this and reset the session.

## Expected Behavior

1. **API errors (400/429/500) should trigger the `error` hook event**, allowing user-defined hooks to handle recovery (e.g., session reset, history compression).

2. **Error messages in history should be truncated or marked.** At minimum, long API error bodies should be capped (e.g., 120 chars) to prevent the positive feedback loop. Ideally, cc-connect should detect that an assistant response is actually an API error and handle it specially (not count it as normal context for `--resume`).

3. **Detect and break error loops.** If consecutive API errors exceed a threshold (e.g., 3 in a row), cc-connect should automatically reset the session or invoke a recovery hook.

## Workaround

[...]

Issue #1498, opened 2026-07-05
Session cannot resume after API error or task interruption — messages dropped, history lost


BenHPM:
## Problem

When an API call fails (error, timeout) or a task is interrupted (watchdog kills stuck process), the user cannot resume the conversation by sending a new message. The bot either:

1. **Drops the message silently** — returns "session is busy" with no queue/retry
2. **Loses all conversation history** — the error hook resets the session file to empty JSON
3. **Fails to reconnect** — `--resume` with a dead session ID silently fails
4. **Enters an infinite reset loop** — `--resume` loads oversized JSONL, exceeds proxy input limit, triggers error, auto-resets, accumulates again → cycle repeats (observed 16+ resets on a single project)

### 1. No message queue when session is busy (core issue)

When a session is already processing a request, new incoming messages are rejected with `"session is busy"` and **dropped** — there is no queue mechanism. This happens in several scenarios:

- After an API error, the internal state may still mark the session as busy
- After watchdog kills the process, the busy flag persists in the session file
- Cron triggers also get rejected: `"last_error": "session ... is busy"`

**Expected behavior:** Messages should be queued and processed once the current request completes.

### 2. Error hook clears all history (script-level, can be fixed locally)

The error hook overwrites the entire session file with an empty structure:

```json
{"sessions":{},"active_session":{},"user_sessions":{},"counter":0,"user_meta":{},"past_id_tracking":true,"version":1}
```

This means:
- All conversation history (hundreds of messages) is lost instantly
- The `agent_session_id` is lost, so `--resume` cannot reconnect
- The next message starts a completely blank session with no memory

**Better approach:** Only clear `agent_session_id` (force a new session), but preserve `history` and `user_sessions` so the new session can see what was being discussed.

### 3. Stale busy flag after crash/restart

When watchdog kills cc-connect (stuck process > 3h old), or cc-connect crashes:
- Any in-flight request is terminated mid-stream
- The session file may retain a `busy` flag
- On restart, cc-connect loads the stale busy state, rejecting new messages

**Expected behavior:** On startup, cc-connect should clear all busy flags since no requests are actually in-flight after a fresh start.

### 4. `--resume` failure is not handled gracefully

When `--resume <session_id>` is called with a session ID from a killed process, the session cannot be found. The failure is not surfaced to the user, and no fallback to a fresh session occurs.

[...]

PR #1569, opened 2026-07-19
fix(core): preserve queued messages after recall

AaronZ345:
## Summary

- preserve messages queued after an active message is recalled
- resume the oldest surviving message in a replacement agent session while retaining FIFO order
- avoid treating an intentionally closed agent event channel as a crash that drops the queue

## Root cause

The silent recall path cleared the entire `pendingMessages` queue when stopping the recalled turn. Messages sent after the recalled message were therefore accepted and acknowledged, then discarded without a response.

## Tests

- `go test -race ./core -run 'TestCUJ_A8_RecallPreservesLaterQueuedMessage|TestHandleMessageRecallPreservesLaterQueueWhenAgentChannelCloses' -count=1`
- `go test ./core -count=1`
- `go test -tags no_web ./... -count=1`

### DIFF
--- a/core/cuj_test.go
+++ b/core/cuj_test.go
@@ -1230,6 +1230,97 @@ func TestCUJ_A6_A7_CoveredByPlatformLayer(t *testing.T) {
 }
 
 // ===========================================================================
+// CUJ-A8 · Recalling one queued message must not discard messages sent after it
+// ===========================================================================
+
+func TestCUJ_A8_RecallPreservesLaterQueuedMessage(t *testing.T) {
+	env := newCUJEnv(t)
+	key := "test:recall-user"
+
+	// Keep the first turn busy long enough to queue two later messages. The
+	// first queued message will be recalled after it becomes the active turn;
+	// the second queued message must survive that cancellation boundary.
+	env.agent.setNextSessionEvents([]Event{{Type: EventResult, Content: "first reply", Done: true}}, 250)
+	env.engine.ReceiveMessage(plat(env.plat), &Message{
+		SessionKey: key,
+		Platform:   "test",
+		MessageID:  "msg-first",
+		UserID:     "recall-user",
+		UserName:   "recall-user",
+		Content:    "first long-running request",
+		ReplyCtx:   "ctx-first",
+	})
+
+	env.waitFor("first agent session starts", 2*time.Second, func() bool {
+		env.agent.mu.Lock()
+		defer env.agent.mu.Unlock()
+		return len(env.agent.sessions) == 1
+	})
+	env.agent.mu.Lock()
+	firstSession := env.agent.sessions[0]
+	env.agent.mu.Unlock()
+	firstSession.mu.Lock()
+	firstSession.delayMs = 5_000
+	firstSession.mu.Unlock()
+
+	// B and C both arrive while A is busy, so they are accepted in FIFO order.
+	env.engine.ReceiveMessage(plat(env.plat), &Message{
+		SessionKey: key,
+		Platform:   "test",
+		MessageID:  "msg-recalled",
+		UserID:     "recall-user",
+		UserName:   "recall-user",
+		Content:    "this queued request will be recalled",
+		ReplyCtx:   "ctx-recalled",
+	})
+	env.engine.ReceiveMessage(plat(env.plat), &Message{
+		SessionKey: key,
+		Platform:   "test",
+		MessageID:  "msg-after-recall",
+		UserID:     "recall-user",
+		UserName:   "recall-user",
+		Content:    "answer this message after the recall",
+		ReplyCtx:   "ctx-after-recall",
+	})
+
+	// Wait until B is actually active, then deliver the platform recall event.
+	// This pins the same path as the Feishu fallback probe in the live incident.
+	env.waitFor("recalled queued message becomes active", 2*time.Second, func() bool {
+		env.engine.interactiveMu.Lock()
+		state := env.engine.interactiveStates[key]
+		env.engine.interactiveMu.Unlock()
+		if state == nil {
+			return false
+		}
+		state.mu.Lock()
+		defer state.mu.Unlock()
+		return state.currentMessageID == "msg-recalled"
+	})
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- This is a focused, well-scoped fix for the CUJ-A8 "messages queued after a recall are silently dropped" bug. The author splits the recovery into two complementary pieces — `resumePendingAfterStoppedTurn` (new) recovers queued successors when the turn ends in a `Stopped` state, and the `isStopped()` short-circuit in `processInteractiveEvents` prevents the engine from mis-treating that same intentional close as a crash and draining the queue a second time. Lock ownership is explicitly transferred to the nested processor via `unlocked = true`, which mirrors the existing pattern in the file. Tests are strong: a CUJ-A8 scenario with three messages (active, recalled, survivor) plus a targeted engine test using a `controllableAgent` whose session can be swapped between `oldSession` and `replacementSession` so we observe the replacement session receiving only the survivor.

Review scope:
- Reviewed `core/engine.go` (the new `resumePendingAfterStoppedTurn`, the `channelClosed` short-circuit, and the two removed `pendingMessages = nil` blocks in `stopInteractiveSessionWithOptions`) plus the new tests in `core/cuj_test.go` and `core/engine_test.go`.
- Focused on race safety, lock ownership handoff, FIFO ordering under the placeholder pattern, and backward compatibility of the silent-recall path.

✅ What looks good:
- **FIFO preservation via the placeholder pattern.** `resumePendingAfterStoppedTurn` takes `[P0..Pn]` from the stopped state, moves `Pn` into a placeholder entry, and starts the nested processor on `P0`. The freshly-created state in `getOrCreateInteractiveStateWith` then re-uses the existing `adoptPendingFromPlaceholder` to inherit the remainder. New messages that arrive between the unlock of the stopped state and the re-creation of the placeholder are appended after `Pn`, preserving order.
- **Lock handoff is explicit.** `unlocked = true` is set immediately before the recursive `processInteractiveMessageWith` call; the deferred fallback `if !unlocked { session.Unlock() }` then skips the unlock on the outer call, and the inner call's defer releases the lock exactly once. This matches the established handoff pattern in `drainOrphanedQueue` / `drainPendingMessages`.
- **Deduplication of "channel close" handling.** Removing the `state.pendingMessages = nil` block from the non-`notifyQueued` branches of `stopInteractiveSessionWithOptions` means the silent-recall path now relies on `resumePendingAfterStoppedTurn` to recover the queue — and the `isStopped()` short-circuit at `channelClosed:` correctly returns early before `notifyDroppedQueuedMessages` / `cleanupInteractiveState` would otherwise drain the queue. The two changes are independent but they compose: removing the wipe without the short-circuit would leak a queue; adding the short-circuit without removing the wipe would lose the queue. Together they work.
- **Tests cover both unit and CUJ angles.** `TestCUJ_A8_RecallPreservesLaterQueuedMessage` exercises the full Engine+CUJ path, and `TestHandleMessageRecallPreservesLaterQueueWhenAgentChannelCloses` exercises the fast path (channel close mid-recall). Together they lock in the FIFO invariant and the no-double-drain invariant.
- **`slog.Info("resuming queued messages after stopped turn", …)`** gives operators a breadcrumb to grep when triaging — the line we wanted in production is now there.

🚨/🔴 Must fix:
- None.

🟠 Should improve:
- **Minor: log a parallel breadcrumb on the `channelClosed:` short-circuit.** Right now the only signal that "the channel close was a recall, not a crash" is the absence of the previous "agent process exited" warn line. Adding `slog.Debug("agent channel closed after intentional stop; deferring queue recovery", "session_key", sessionKey)` (or `Info`, depending on volume) would make the connection between the two pieces of this PR obvious in logs. Pure observability nit.
- **Minor: consider using `state.markStopped()` consistently.** The new `resumePendingAfterStoppedTurn` only reads `state.stopped`; it does not call `markStopped()` itself. That's correct because the caller (`stopInteractiveSessionWithOptions`) already marks the state stopped and removes it from the map. But it's worth a one-line comment at the top of the new function pointing at the only call site, so future readers understand the invariant ("caller must have already called `markStopped()` and `delete(e.interactiveStates, …)`").
- **Minor:** When the placeholder is freshly allocated (`placeholder == nil || placeholder == state`), it is `eventsNeedResync: true`. After `processInteractiveMessageWith` replaces it with a real state via `adoptPendingFromPlaceholder`, the new state's `eventsNeedResync` is also `true` (from `getOrCreateInteractiveStateWith`'s constructor) — so the flag round-trip is correct. Worth a one-line comment so a future reader doesn't "optimize" one of them away.

🔵 Optional:
- A test that places a NEW message between `e.stopInteractiveSessionWithOptions` returning and `resumePendingAfterStoppedTurn` acquiring `interactiveMu` would exercise the "messages appended to placeholder" branch explicitly. The current CUJ test only checks the case where the placeholder is freshly created with no pre-existing pending; the merge-with-prior-placeholder path is exercised by the implementation but not asserted on. Not blocking.
- A test for the `eventsNeedResync: true` invariant on the replacement session would close the small "did we forget to drain stale events from the old turn?" gap. Again, not blocking.

❓ Questions:
- **Did the silent-recall path used to notify the user of dropped queued messages?** Looking at the diff, the old code in the `else` branch (when `notifyQueued == false`) dropped without notification, and the new code preserves those messages instead. So this is a net-positive behavior change — but it would be good to confirm with PM that no caller was relying on the old "silently drop queue on recall" semantics.

Testing / Risk:
- Verified: `go test -race ./core/... -count=1 -timeout 180s` (48.4s, all green); new tests pass; the full CUJ test suite (`TestCUJ_A6_A7…`, `TestCUJ_A8…`) is green.
- Blast radius is `core/engine.go` only. No agent-runtime, no platform, no CLI touched.
- Backward compatible for the recall *target* (the recalled message is still silently stopped); the only observable behavior change is "messages queued after the recall now get a response instead of being silently dropped." This is exactly the contract issue #1499 / #1498 callers have been asking for.

Next step:
- Merge. The two 🟠 minor comments are nice-to-have for the next maintainer; they are not required for correctness.

AaronZ345:
Superseded by #1706.

PR #1572, opened 2026-07-20
fix(web): validate project work_dir

### LINKED ISSUE #1068: [Bug] 通过web UI创建project时，如果填写不存在的路径，可以正常完成创建
### cc-connect Version

v1.3.2

### Operating System

Linux (Ubuntu/Debian)

### Agent Type

Claude Code

### Platform

Feishu (Lark)

### Installation Method

### Bug Description

通过web UI创建project时，如果填写不存在的路径，可以正常完成创建，并且运行，飞书侧只会提示启动 Agent 会话失败

### Steps to Reproduce

1.cc-connect web启动UI管理界面；
2.project界面 -> Add project -> working directory 填写非法路径
3.点击next创建，连接到飞书 app
4.发送消息，必现回复❌ 错误: 启动 Agent 会话失败

### Expected Behavior

添加有效路径检测拦截，如果是空路径，应当返回创建失败

### Actual Behavior

可以正常创建project，并且发送发送消息，必现回复❌ 错误: 启动 Agent 会话失败

```toml

```

### Logs / Error Output

```shell

```

### Additional Context

_No response_

AaronZ345:
Replaces #1418 with the same approved patch on a fresh branch.

## Summary
- validate non-empty `work_dir` values before setup save, add-platform, and project PATCH handlers
- reject missing/non-directory paths with HTTP 400 before mutating runtime state or persisting project settings
- cover Feishu/Weixin setup, add-platform, PATCH, and validator behavior

Fixes #1068

## Tests
- `go test ./core`

### DIFF
--- a/core/management.go
+++ b/core/management.go
@@ -726,6 +726,14 @@ func (m *ManagementServer) handleProjectDetail(w http.ResponseWriter, r *http.Re
 			mgmtError(w, http.StatusBadRequest, "invalid JSON: "+err.Error())
 			return
 		}
+		if body.WorkDir != nil {
+			workDir, err := validateProjectWorkDir(*body.WorkDir)
+			if err != nil {
+				mgmtError(w, http.StatusBadRequest, err.Error())
+				return
+			}
+			*body.WorkDir = workDir
+		}
 
 		if body.Language != nil {
 			switch *body.Language {
--- a/core/management_test.go
+++ b/core/management_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
@@ -119,6 +120,26 @@ func mgmtPost(t *testing.T, url, token string, body any) mgmtResponse {
 		t.Fatalf("decode POST response: %v", err)
 	}
 	return r
+}
+
+func mgmtPostHandler(t *testing.T, handler http.HandlerFunc, path string, body any) (mgmtResponse, int) {
+	t.Helper()
+	var buf bytes.Buffer
+	if body != nil {
+		if err := json.NewEncoder(&buf).Encode(body); err != nil {
+			t.Fatalf("encode POST body: %v", err)
+		}
+	}
+	req := httptest.NewRequest(http.MethodPost, path, &buf)
+	req.Header.Set("Content-Type", "application/json")
+	w := httptest.NewRecorder()
+	handler(w, req)
+
+	var r mgmtResponse
+	if err := json.NewDecoder(w.Body).Decode(&r); err != nil {
+		t.Fatalf("decode POST handler response: %v", err)
+	}
+	return r, w.Code
 }
 
 func mgmtPatch(t *testing.T, url, token string, body any) mgmtResponse {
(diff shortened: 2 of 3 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Solid validation fill-in for issue #1068. The patch centralizes the "work_dir must be an existing directory" rule into one helper and applies it at every entry point that mutates project/runtime state. Behavior is strictly additive: previously bad paths got saved silently; now they get a clean HTTP 400 before any mutation happens. Three new test groups cover the four entry points plus the validator itself, and `go test -race ./core/...` is green.

Review scope:
- Reviewed `core/setup.go`, `core/management.go`, `core/management_test.go` (validator + call sites).
- Focused on correctness of the validator, ordering of validation vs mutation, and test coverage.

✅ What looks good:
- Validator runs **before** the callback in all three HTTP paths (`handleSetupFeishuSave`, `handleSetupWeixinSave`, `handleProjectAddPlatform`) and **before** `SetWorkDir` / `saveProjectSettings` in `handleProjectDetail`. The PR explicitly tests "no mutation on rejection" (`TestMgmt_ProjectPatch_RejectsMissingWorkDirBeforeMutation`).
- Empty / whitespace-only `work_dir` is intentionally accepted (returns trimmed empty string + nil). This matches the existing `*string` semantics in the PATCH body and the `string` non-pointer in setup bodies, so no caller gets a behavior change for "no work_dir provided".
- Error messages are human-readable and stable, so the front-end / docs can match on them (`work_dir does not exist`, `work_dir is not a directory`).
- Tightly scoped: only adds validation + tests; no signature changes, no persistence touches, no i18n strings moved.

🚨/🔴 Must fix:
- None.

🟠 Should improve:
- **Minor: prefer `os.IsNotExist(err)` only after `os.Stat`.** The current branch returns `"work_dir is not accessible: %s: %w"` for any non-`IsNotExist` error (perm denied, EIO, ENOTDIR race, etc.). For a private management API this is fine, but documenting the "perm denied" class separately (e.g. `work_dir is not accessible (stat failed)` vs the existing "does not exist") would help triage misconfigured servers from real path bugs. Optional.
- **Minor: no `~` expansion.** `validateProjectWorkDir("~/repo")` will return 400 because `~` doesn't resolve via `os.Stat`. That matches the prior (broken) behavior and ships in `req.WorkDir` as the literal tilde, so it's not a regression — but consider documenting the "use absolute paths only" rule in the docs / setup endpoint swagger if there is one. Pure suggestion, not blocking.

🔵 Optional:
- A `path/filepath.Clean(trimmed)` before `os.Stat` would normalize trailing slashes and `./` prefixes; not required since `os.Stat` accepts them, but cheap to add.
- A pair of unit tests for "perm denied" and "symlink to a directory" would close the small remaining gaps in the validator matrix. Out of scope for this PR.

❓ Questions:
- None.

Testing / Risk:
- Verified: `go test -race ./core/... -count=1 -timeout 60s` (48.3s, all green); new tests (`TestValidateProjectWorkDir`, `TestMgmt_AddPlatformToNewProject_RejectsMissingWorkDir`, `TestMgmt_SetupSave_RejectsMissingWorkDir`, `TestMgmt_ProjectPatch_RejectsMissingWorkDirBeforeMutation`) all pass.
- Blast radius is `core/` only. No CLI, no platform adapter, no agent runtime affected.
- Backward compatible: any previously-accepted `work_dir` that pointed at a real directory still passes; any previously-saved broken path now returns 400 instead of being silently persisted — strictly an improvement.

Next step:
- Merge. Consider folding the two 🟠 suggestions into a follow-up doc/test PR if you want; they don't affect correctness today.

PR #1574, opened 2026-07-20
feat(send): generate media via configurable providers

AaronZ345:
Replaces #1411 with the same approved patch on a fresh branch.

## Summary
- add configurable `[image]`, `[video]`, and `[music]` provider blocks for direct generated media send-back
- implement MiniMax text-to-image, text-to-video, and text-to-music generators, including base64/url image handling, async video polling/download, and hex audio decoding
- add OpenAI-compatible image generation through `/images/generations`, with base64 and URL result handling
- add a provider-neutral `command` media generator that can read bytes from stdout or a configured `{{output}}` file
- add `openclaw` convenience wiring for image/video generation, with `[image.command]` / `[video.command]` args available for provider-specific models such as Seedance
- keep music extensible through `[music.command]`, so Suno or other music APIs can be bridged by local wrappers without hard-coding each vendor SDK into cc-connect
- add `cc-connect send --generate-image`, `--generate-video`, `--generate-music`, `--music-lyrics`, `--music-instrumental`, and `--music-lyrics-optimizer`
- refresh agent prompt, docs, config example, and changelog so agents can discover and use the generated-media path

## Notes
- This intentionally complements #1359 instead of replacing it. #1359 routes existing `--audio` / `--video` files to native media senders; this PR creates media from prompts, then sends generated images through `ImageSender`, generated video through `SendVideosToSession`, and generated music through `SendAudiosToSession`.
- Generated media respects `attachment_send = "off"` before provider generation starts.
- OpenAI-compatible support is scoped to image generation endpoints. Anthropic API keys are not treated as native media-generation providers because Anthropic's public API is for text/vision understanding, not image/video/music generation.
- The command adapter executes commands directly without a shell. It supports `{{prompt}}`, `{{output}}`, `{{format}}`, and for music `{{lyrics}}` / `{{instrumental}}` placeholders.
- If MiniMax `api_key` is empty, the generator reuses the local MiniMax JSON config path used by TTS.
- MiniMax image generation defaults to `response_format = "base64"` so the provider result is immediately sent as an image attachment; `url` responses are also supported.
- Prompt-only lyric music requests can opt into lyrics optimization; instrumental requests stay instrumental.
- Supersedes the closed prototype #1287, rebased onto current `main` and updated for dedicated image/audio/video sender interfaces.

## Tests
- `git diff --check HEAD~1..HEAD`
- `go test ./core ./config`
- `GOOS=linux go test -tags no_web -c -o /private/tmp/cc-connect-provider-pr-cmd.test ./cmd/cc-connect`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@
 
 ### Added
 - **Feishu: outbound bot-to-bot @mention resolution** via new `mention_map` config option. Maps agent-friendly names (e.g. `BOT-A`) to Feishu open_ids so that when an agent writes `@BOT-A` in its reply, cc-connect converts it to a native Feishu `<at>` tag that triggers a real notification. Layered on top of `resolve_mentions` (group-member matching) with higher priority, so explicit config always wins (#1322).
+- **Generated media send-back**: `cc-connect send` can generate image, video, and
+  music through configured providers (`--generate-image`, `--generate-video`,
+  `--generate-music`) and send the result back through the existing
+  attachment/media delivery path. Image generation supports OpenAI-compatible
+  and MiniMax providers; command/OpenClaw adapters can bridge provider-specific
+  image, video, and music CLIs.
 
 ### Fixed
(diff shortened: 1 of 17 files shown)


happyTonakai:
agent通过API的生成媒体文件之后，通过cc-connect作为附件发送给用户就好了吧，个人认为没必要内化为cc-connect的原生能力。否则图像/视频/TTS/音乐，要集成的provider太多了

AaronZ345:
Closing this after re-evaluating the ownership boundary raised in review. cc-connect already sends media produced by an agent/API as native attachments; embedding OpenAI/MiniMax image, video, and music provider clients here would duplicate provider integrations and expand the long-term maintenance surface. Splitting the current 17-file change into several provider PRs would not resolve that architectural concern. I am keeping the branch for reference, but withdrawing the upstream proposal rather than asking maintainers to review a direction that does not fit the connector boundary.

PR #1576, opened 2026-07-21
refactor(agent): unify file-attachment handling across all agents

happyTonakai:
## Summary

Unify how every agent handles file attachments: save files to disk via `core.SaveFilesToDisk`, then reference them by absolute path in the prompt via `core.AppendFileRefs`. Image attachments are **not** touched in this PR (intentionally — image handling varies per agent protocol and unifying it would be a much larger change).

## Before / After

| Agent       | Before                                          | After                              |
| ----------- | ----------------------------------------------- | ---------------------------------- |
| `acp`       | core helpers                                    | core helpers (unchanged)           |
| `claudecode`| core save + hand-written text                   | core helpers                       |
| `codex`     | core helpers                                    | core helpers (unchanged)           |
| `codex/app` | core helpers                                    | core helpers (unchanged)           |
| `copilot`   | core helpers                                    | core helpers (unchanged)           |
| `cursor`    | core helpers                                    | core helpers (unchanged)           |
| `iflow`     | core helpers                                    | core helpers (unchanged)           |
| `opencode`  | core helpers                                    | core helpers (unchanged)           |
| `qoder`     | core helpers                                    | core helpers (unchanged)           |
| `reasonix`  | core helpers                                    | core helpers (unchanged)           |
| `tmux`      | core save + `# files: …` text                   | core helpers                       |
| `antigravity` | custom save + `[Attached files saved at: …]` | core helpers                       |
| `gemini`    | custom save + `[Attached files saved at: …]`    | core helpers                       |
| `kimi`      | custom save + `[Attached files saved at: …]`    | core helpers                       |
| **`pi`**    | **read bytes back from disk and inline into prompt** | core helpers (saved, referenced by path) |

Two non-trivial situations:
- `pi` was buggy — its `Send()` read every uploaded file from disk and inlined the raw bytes into the prompt string, leaving the model staring at raw DOCX/ZIP bytes it could not parse. The bug affected **both modes**: `rpc` writes the prompt to the persistent process's stdin; `json` spawns `pi --mode json -p <prompt>` with the bytes inlined into the `-p` argument. Fixing this is the reason the pi-specific code changes exist.
- `claudecode` and `tmux` already used `core.SaveFilesToDisk` but hand-wrote the path-ref text. They now call `core.AppendFileRefs` so the wording stays in sync with the other agents.

## Tests added

- `TestPiSession_BuildPrompt_FilesDoNotEmbedBytes` — pins the pi bug fix; the test fails loudly if the byte-inlining loop ever returns.
- `TestPiSession_BuildPrompt_EmptyMsgGetsFileRef` — empty-caption fallback still substitutes a default instruction.
- `TestPiSession_BuildPrompt_ImagesStillInline` — image bytes are still embedded (image-handling half is out of scope, this guards against accidental regression).
- `TestBuildClaudeTextPart` — 5 subtests covering claudecode's empty-prompt fallback rules, including the *empty prompt + both files and images* case that was silently dropped during the refactor (caught during review).

## Test results

- [x] `go test ./agent/pi/ ./agent/claudecode/ ./agent/antigravity/ ./agent/gemini/ ./agent/kimi/ ./agent/tmux/ ./platform/feishu/` — 1100+ passed
- [x] `go test ./core/ -run TestCUJ` — 55 passed
- [x] `go vet ./...` — clean
- [x] `go build ./...` — clean
- [x] Manual test (Feishu → pi): uploaded a DOCX via Feishu, confirmed the agent received the on-disk file path reference rather than the raw binary content

Two pre-existing flaky tests (`TestGetModelAndReasoningEffort…`, `TestCUJ_A3_ImageReachesAgent` — TempDir cleanup race in `t.TempDir`) are unrelated to this PR; they pass when run in isolation and reproduce on `main`.


happyTonakai:
Conflict resolution note from upstream-main merge.

Summary: the **core pi/claudecode motivation of this PR is already addressed on \`main\`**, but our broader **unify contribution (antigravity / gemini / kimi / tmux → \`core.SaveFilesToDisk\` / \`core.AppendFileRefs\`)** is still needed and retained here.

Details:
- \`pi\`'s byte-inlining bug was already fixed upstream via **PR #1724** (which resolves issue #1723) using a better mechanism (\`@<path>\` argv refs instead of embedding bytes into the prompt), so I kept \`main\`'s pi implementation.
- \`claudecode\` already uses the core helpers on \`main\`; the hand-written \`buildClaudeTextPart\` / \`extFromMime\` introduced here were dead code in the new structure and were dropped along with their now-orphaned tests (\`TestPiSession_BuildPrompt_*\`, \`TestBuildClaudeTextPart\`).
- The remaining real value of this PR — unifying antigravity / gemini / kimi (and tmux) onto the shared core attachment helpers — is preserved and merged cleanly.

Verification: \`go build ./agent/...\`, \`go vet\`, and tests for \`agent/{pi,claudecode,antigravity,gemini,kimi,tmux}\` all pass. (\`TestCUJ_A3_ImageReachesAgent\` is a pre-existing flaky test unrelated to these changes.)

Suggested follow-up: consider confirming with the author whether to keep the pi/claudecode halves out of the diff, since they are now redundant with \`main\`.

PR #1578, opened 2026-07-21
fix(core): freeze streaming preview on interrupt, don't discard

happyTonakai:
## Problem

When the streaming preview is interrupted mid-stream (`/stop`, idle timeout, `max_turn_time`, `EventError`, `isStopped`), `sp.discard()` calls `PreviewCleaner.DeletePreviewMessage`. On Feishu this surfaces as "XX 撤回了一条消息", hiding the partial answer the user just watched stream. On Discord/Telegram the card simply vanishes.

#882 replaced `sp.discard()` with `sp.finish()` for the `/stop` case. That fix was only correct for Feishu — `finish()` checks `PreviewFinishPreference.KeepPreviewOnFinish()` and updates the card in-place, but on platforms that don't implement that preference (Discord, Telegram, …) `finish()` still routes to `PreviewCleaner.DeletePreviewMessage` and the card disappears, same as `discard()`. The bug was masked because Feishu was the only streaming-preview platform covered by tests.

## Fix

Switch all five interrupt paths from `discard()` to `freeze()`. `freeze()` updates the card in-place via `MessageUpdater.UpdateMessage` and never calls `PreviewCleaner.DeletePreviewMessage`, so partial text survives uniformly across Feishu, Discord, Telegram, and any other streaming-preview platform. It also sets `degraded = true` so subsequent `appendText` knows to start a fresh card.

Supersedes #882.

## Sites changed

All in `core/engine.go`:

| Path                  | Line  | Trigger                                       |
| --------------------- | ----- | --------------------------------------------- |
| `<-stopCh`            | 4832  | user pressed `/stop`                          |
| idle timeout          | 4863  | no events for `eventIdleTimeout`              |
| `max_turn_time`       | 4877  | turn exceeded `maxTurnTime`                   |
| `state.isStopped()`   | 4929  | stopped via any other code path               |
| `EventError`          | 6070  | agent reported error                          |

## Sites intentionally not changed

These stay on `sp.discard()` — each for a reason that would be wrong to "freeze":

| Line  | Context                              | Why `discard`                                                |
| ----- | ------------------------------------ | ------------------------------------------------------------ |
| 4842  | `pendingSend` error                  | Send failed before stream started — no partial text to keep  |
| 5472  | `permTimer` + `<-stopCh`             | Stopping before permission reply — no text yet              |
| 5509  | permission pending + `<-stopCh`      | Same                                                         |
| 5741  | silent reply (`NO_REPLY`)            | Silent replies must vanish — that's the point                |
| 5827  | tool calls + `segmentStart > 0`      | Replacing already-streamed segments, not losing text         |
| 5837  | `suppressDuplicate`                  | Discarding the duplicate preview                             |
| 6133  | `isSilentReply` on abnormal exit     | Silent reply must disappear                                  |
| 6139  | `stripTrailingSilent`                | Same                                                         |
| 6153  | toolCount + segmentStart             | Segment replacement                                          |

These stay on `sp.finish(...)` — they're the "happy path" finalizers, not interrupt paths:

| Line  | Context                              | Why `finish`                                                 |
| ----- | ------------------------------------ | ------------------------------------------------------------ |
| 5709  | StreamingCard path silent cleanup    | Different code path (rich card, not preview)                 |
| 5845  | `EventResult` normal completion      | Need to send final accumulated text + status footer          |
| 6164  | `channelClosed`                      | Same — finalize stream with full text                        |

`sp.unfreeze()` at line 5521 (resume preview after permission granted) is also unchanged — that's the inverse of `freeze()`, not an alternative.

## Test plan

- [x] `/stop` mid-stream on Feishu: card stays, no "recall" message
- [x] `go test ./core/ -run TestCUJ` passes
- [x] `go build ./...` passes

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -4784,7 +4784,8 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 
 		select {
 		case <-stopCh:
-			sp.discard()
+			// See EventError below for why freeze() not discard().
+			sp.freeze()
 			return
 		case event, ok = <-events:
 			if !ok {
@@ -4814,7 +4815,8 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 			slog.Error("agent session idle timeout: no events for too long, killing session",
 				"session_key", sessionKey, "timeout", e.eventIdleTimeout, "elapsed", time.Since(turnStart))
 			cp.Finalize(ProgressCardStateFailed)
-			sp.discard()
+			// See EventError below for why freeze() not discard().
+			sp.freeze()
 			state.mu.Lock()
 			state.eventsNeedResync = true
 			p := state.platform
@@ -4827,7 +4829,8 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 			slog.Warn("agent turn exceeded max_turn_time: sending stop signal, will force-kill if needed",
 				"session_key", sessionKey, "max_turn_time", e.maxTurnTime, "elapsed", elapsed)
 			cp.Finalize(ProgressCardStateFailed)
-			sp.discard()
+			// See EventError below for why freeze() not discard().
+			sp.freeze()
 			state.mu.Lock()
 			p := state.platform
 			state.mu.Unlock()
@@ -4878,7 +4881,8 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 		}
 
 		if state.isStopped() {
-			sp.discard()
+			// See EventError below for why freeze() not discard().
+			sp.freeze()
 			state.mu.Lock()
 			state.eventsNeedResync = true
 			state.mu.Unlock()
@@ -5958,7 +5962,14 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 
 		case EventError:
 			cp.Finalize(ProgressCardStateFailed)
-			sp.discard()
+			// Freeze (not discard) the streaming preview: discard() routes
+			// through PreviewCleaner.DeletePreviewMessage, which on Feishu/Lark
+			// surfaces as "XX 撤回了一条消息" and hides the partial answer the
+			// user just watched stream. freeze() commits the accumulated text
+			// in-place and marks the preview degraded. Same fix applied to the
+			// other "interrupt mid-stream" branches below (/stop, idle timeout,
+			// max_turn_time).
+			sp.freeze()
 			state.mu.Lock()
 			state.eventsNeedResync = true
 			state.mu.Unlock()


chenhg5 (approved):
## QA Review: PR #1578

**Conclusion: Approve ✅**

Small, surgical bugfix that supersedes #882 correctly. The author identified that #882 only fixed Feishu and that `sp.finish()` doesn't actually solve the Discord/Telegram case (those platforms don't implement `PreviewFinishPreference.KeepPreviewOnFinish()`, so `finish()` still routes to `PreviewCleaner.DeletePreviewMessage`). Switching all five interrupt branches to `sp.freeze()` is the right call — `freeze()` commits the accumulated text in-place via `MessageUpdater.UpdateMessage` and never invokes `PreviewCleaner`, so the partial answer the user just watched stream survives uniformly.

### Diff review

`core/engine.go` (+16/-5)
- Five call sites switched from `sp.discard()` to `sp.freeze()`: `<-stopCh` (line 4832), idle timeout (4863), `max_turn_time` (4877), `state.isStopped()` (4929), and `EventError` (6070). Confirmed via `core/streaming.go:298` that `freeze()` cancels timers, updates the preview message in-place, sets `degraded = true`, and never touches `PreviewCleaner`. The other four sites use a one-line `// See EventError below for why freeze() not discard().` pointer instead of repeating the rationale — good DRY choice for a 5-site mechanical change.
- `cp.Finalize(ProgressCardStateFailed)` calls in the idle/timeout branches are unchanged, so users still see a failed-state marker on the card. That preserves the "agent stopped" signal even though the message itself no longer disappears — important because the UX change is otherwise ambiguous.
- Body documents the sites intentionally left on `discard()` with per-site reasons. That's the kind of "negative space" justification that prevents a future drive-by refactor from "fixing" them the wrong way.

### What this fixes vs. #882

#882 (closed 2026-07-21) only changed the `/stop` branch and used `sp.finish()`. On Feishu that worked because Feishu implements `PreviewFinishPreference`. On Discord/Telegram (which don't), `finish()` still went to `PreviewCleaner.DeletePreviewMessage` → card vanished. Tests didn't catch this because Feishu was the only streaming-preview platform covered. Author correctly calls out the platform-coverage gap as the root cause and picks `freeze()` precisely because it's the path that bypasses `PreviewCleaner` entirely.

### Risk

- Pure `core/engine.go`, 5 single-line replacements + a small amount of comment text.
- No public API or protocol change.
- UX change: interrupted cards no longer vanish on Feishu/Discord/Telegram. Users who relied on "message gone = agent stopped" as a visual signal will need to look at the failed-state footer instead. Release-notes-worthy but not breaking.
- `degraded = true` is already consumed by `appendText` via the existing unfreeze()/detachPreview() machinery (for permission-prompt resume). No new contract introduced.

### Testing / Risk

- ✅ CI: 5/5 green on `4fb43640`.
- ✅ `freeze()` and `discard()` semantics confirmed against `core/streaming.go` on `main`.
- ✅ Supersedes relationship verified: #882 is closed (state=closed, 2026-07-21), same author.
- 🟡 No new test in this PR. The +16/-5 is purely mechanical (5 call-site swaps), and `freeze()`/`discard()` behavior is covered in `core/streaming_test.go`, but the specific invariant — "interrupt branches must not invoke `PreviewCleaner.DeletePreviewMessage` on a platform that doesn't implement `PreviewFinishPreference`" — has no automated guard. Worth a follow-up test.

- A `core/cuj_test.go`-level test using a mock platform that implements `PreviewCleaner` but not `PreviewFinishPreference`, triggering one of the five interrupt branches, asserting `DeletePreviewMessage` was **not** called and `UpdateMessage` was called once. That would lock in the exact invariant this PR is fixing and would have caught #882's regression.
- Or a direct `core/streaming_test.go` test: `TestStreamPreview_FreezeDoesNotInvokePreviewCleaner` on `freeze()`'s contract.
- Either is cheap, fast, and prevents a future refactor from re-introducing the original bug.

### Next step

Maintainer: merge when convenient. The fix itself is safe; the only thing I'd flag is a release-notes line ("streaming preview now keeps the partial answer visible after /stop, idle timeout, max_turn_time, or agent error" — plus a pointer that the card's failed-state footer is the new "agent stopped" signal). Post-merge, the follow-up test from the P3 above would be the cheapest confidence bump.

PR #364, opened 2026-03-29
fix(slack): add message deduplication for Socket Mode retransmissions

leighstillard:
## Summary

- Adds `core.MessageDedup` to the Slack platform adapter, matching the pattern used by every other platform (Feishu, DingTalk, QQ, QQBot, WeCom)
- Deduplicates both `AppMentionEvent` and `MessageEvent` using `ev.TimeStamp` as the unique message ID
- Prevents duplicate processing when Slack Socket Mode retransmits events after reconnects or delayed TCP acks

## Problem

Slack Socket Mode retransmits events when acknowledgment is delayed or after connection drops/reconnects. Without dedup, the same message is processed multiple times:

```
09:44:18 msg received (msg_id=1774777453.010829)  ← processed
09:44:32 msg received (msg_id=1774777453.010829)  ← queued as duplicate
09:46:11 msg received (msg_id=1774777453.010829)  ← queued as duplicate
```

This causes "message queued for busy session" spam and can trigger duplicate bot responses after the current turn completes.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./core/` passes (614 tests)
- [x] Manual: verify duplicate messages no longer appear during long agent turns
- [x] Manual: verify normal message flow is unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1579, opened 2026-07-21
fix(slack): ignore edited @mentions instead of re-running the agent

vuyiv:
## Problem

Editing a message that @-mentions the bot makes Slack re-fire a fresh
`app_mention` event, this time with `Edited` set. The Socket Mode handler in
`platform/slack/slack.go` had no edit guard, so it reprocessed the edited mention
as a brand-new inbound message and started a **duplicate agent turn** — a user who
only meant to correct a typo triggers the agent to run again.

Telegram and Discord already ignore inbound edits; Slack was the outlier.

(Distinct from the Socket Mode retransmission dedup in #364 — that handles Slack
redelivering the *same* event; here Slack sends a *new* event for a user edit.)

## Fix

In the `AppMentionEvent` branch of `handleEvent`, drop the event when
`ev.Edited != nil`, before it reaches the message handler:

```go
if ev.Edited != nil {
    slog.Debug("slack: ignoring edited app_mention", "user", ev.User, "channel", ev.Channel)
    return
}
```

Accepted limitation, matching Telegram/Discord: this also ignores an
*edit-to-add-mention* (a message edited to newly @-mention the bot). Inbound edits
are ignored uniformly rather than special-cased.

## Verification

- `TestHandleEvent_EditedAppMentionIgnored` — an edited `app_mention` with a
  current timestamp (so the pre-existing old-message guard doesn't mask the
  result) dispatches 0 messages. Fails on pre-fix code.
- `TestHandleEvent_NormalAppMentionDispatches` — a normal current-timestamp
  mention still dispatches exactly 1 message (guards against over-dropping).
- `go build ./platform/slack/ ./core/` ✓
- `go vet ./platform/slack/` ✓
- `go test ./platform/slack/` ✓ · `go test -race ./platform/slack/` ✓
- `gofmt -l` clean ✓

🤖 90% Claude

### DIFF
--- a/platform/slack/slack.go
+++ b/platform/slack/slack.go
@@ -179,6 +179,16 @@ func (p *Platform) handleEvent(evt socketmode.Event) {
 					return
 				}
 
+				// Editing a mention re-fires a fresh app_mention with Edited set;
+				// ignore it rather than reprocessing the edit as a new prompt.
+				// This also drops edit-to-add-mention (a message edited to newly
+				// @-mention the bot) — an accepted limitation, matching how
+				// Telegram and Discord ignore all inbound edits.
+				if ev.Edited != nil {
+					slog.Debug("slack: ignoring edited app_mention", "user", ev.User, "channel", ev.Channel)
+					return
+				}
+
 				if ts := ev.TimeStamp; ts != "" {
 					if dotIdx := strings.IndexByte(ts, '.'); dotIdx > 0 {
 						if sec, err := strconv.ParseInt(ts[:dotIdx], 10, 64); err == nil {
--- a/platform/slack/slack_test.go
+++ b/platform/slack/slack_test.go
@@ -2,12 +2,73 @@ package slack
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
+	"github.com/chenhg5/cc-connect/core"
 	"github.com/slack-go/slack/slackevents"
+	"github.com/slack-go/slack/socketmode"
 )
+
+// appMentionEvent wraps an AppMentionEvent in the socketmode/EventsAPI envelope
+// that handleEvent unwraps. Request is nil so handleEvent skips the socket Ack.
+func appMentionEvent(ev *slackevents.AppMentionEvent) socketmode.Event {
+	return socketmode.Event{
+		Type: socketmode.EventTypeEventsAPI,
+		Data: slackevents.EventsAPIEvent{
+			Type: slackevents.CallbackEvent,
+			InnerEvent: slackevents.EventsAPIInnerEvent{
+				Type: "app_mention",
+				Data: ev,
+			},
+		},
+	}
+}
+
+func TestHandleEvent_EditedAppMentionIgnored(t *testing.T) {
+	var calls int
+	p := &Platform{
+		channelNameCache: map[string]string{"C1": "general"},
+		handler:          func(core.Platform, *core.Message) { calls++ },
+	}
+	p.userNameCache.Store("U1", "Alice") // seed caches so a missed guard fails cleanly, not via nil client
+	// A current ts keeps the old-message guard from dropping the event first,
+	// so a regressed Edited guard would let this reach the handler (calls == 1).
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
## QA Review: PR #1579

**Conclusion: Approve ✅**

Tight, well-scoped bugfix that brings Slack in line with how Telegram and Discord already handle inbound edits. Author identified the right scope boundary (this is *new* Slack event for user edits, not the *same-event retransmit* dedup that #364 handled) and picked the cleanest fix — drop the event before it reaches the handler, mirroring the existing `BotID`/`IsOldMessage`/`AllowList` early-return pattern.

### Diff review

`platform/slack/slack.go` (+10/-0)
- Placement is well-considered: after the `BotID == ""` and `ev.User == ""` cheap rejects (so we don't log spam from bot edits), before the `IsOldMessage` restart-replay guard (so the new guard stands on its own and restart-replay behavior is unaffected), and before the `AllowList` check (no point authorizing edits to inbound events). Three existing guards, one new guard, all in the same early-return cascade — easy to read.
- `slog.Debug` (not `Info`) with structured fields is correct: this is a routine event-drop, not an anomaly. Operators who want to debug a "why didn't bot respond" can grep for it.
- No need to add a separate config flag — Telegram/Discord don't have one either, and the edit-to-add-mention limitation is acceptable for all three platforms (covered in the body).

`platform/slack/slack_test.go` (+61/-0)
- `TestHandleEvent_EditedAppMentionIgnored` constructs an `AppMentionEvent` with `Edited` set to a non-nil `*slackevents.Edited{User, TimeStamp}` and a *current* timestamp so the pre-existing `IsOldMessage` guard doesn't mask the result. The comment explicitly calls out the failure mode: "a regressed Edited guard would let this reach the handler (calls == 1)." That's the right way to write a regression test for an early-return — make sure the rest of the chain doesn't accidentally absorb the change you're guarding against.
- `TestHandleEvent_NormalAppMentionDispatches` is the symmetric test: a normal mention (no `Edited`) must still reach the handler. This is the "don't over-drop" guard and is what kept me from worrying about the symmetric-failure risk where the new check could accidentally drop original mentions.
- Both tests seed `channelNameCache` and `userNameCache` so the mocked `resolveUserName` / `resolveChannelName` calls inside `handleEvent` don't blow up on a nil client. Without that, a missed guard could fail via panic or nil-deref instead of via the count assertion — that would muddy the regression signal.
- The `appMentionEvent` helper builds the right `socketmode.Event` envelope (EventsAPI > CallbackEvent > app_mention) with `Request: nil` to skip the socket Ack. Matches what the real producer would emit for an Events API event.

### Risk

- Pure `platform/slack`, no public API or protocol change.
- No new dependencies (uses `slog` and `slackevents` types already in the import set).
- Edit-to-add-mention is silently dropped. Documented in the body as an accepted limitation matching Telegram/Discord. Low real-world impact — most @mentions are typed once and sent, not added via edit.
- Zero blast radius for existing Slack deployments on /stop, threads, file shares, etc.

### Testing / Risk

- ✅ CI: 5/5 green on `1566070d`.
- ✅ Local (author): `go build`, `go vet`, `go test`, `go test -race`, `gofmt -l` all clean on `platform/slack/`.
- ✅ Guard placement verified against `handleEvent` in the head branch.
- ✅ Scope correctly distinguished from #364 (same-event retransmit, already merged) and from #1822 (app_mention/message handler-side dedup, separately approved this cycle).

- The `slog.Debug` line currently logs `user` and `channel`. Adding `original_ts` (the event's own `TimeStamp`) and the `Edited.TimeStamp` would make "why didn't bot respond after my edit" questions easier to triage from logs alone. Trivial change, not required.
- Worth a quick grep across `platform/slack/` for any other inbound event type that might have the same re-trigger-on-edit issue (e.g. `MessageEvent` reactions, file shares). Out of scope for this PR but a useful maintainer-side check.

### Next step

Maintainer: merge when convenient. Release-notes line is nice-to-have but not required ("Slack: ignore edited @mentions, matching Telegram/Discord behavior — a typo edit no longer re-runs the agent"). Post-merge, the cheap confidence bump is to manually edit a mention in a real Slack channel and confirm the bot stays silent on the edit.

Issue #599, opened 2026-04-13
[Bug] Session context leak between projects - wrong project summary loaded


OuyangMr:
## Problem Summary

When multiple projects are configured in cc-connect (e.g., ClaudeWork and GameStudios), the session context from one project can leak into another project's conversation.

**Observed behavior:**
- ClaudeWork project (current: `CC_PROJECT=ClaudeWork`) receives `GameStudios` session summary in system-reminder
- This causes context confusion where the wrong project history is injected
- Deleting session files doesn help because daemon recreates them

## Configuration

```toml
[[projects]]
  name = "GameStudios"
  work_dir = "/Users/naxi/project/open-resource/Claude-Code-Game-Studios"
  [projects.platforms]
    app_id = "cli_a95d7f89de78dbde"
    ...

[[projects]]
  name = "ClaudeWork"
  work_dir = "/Users/naxi/project/claude"
  [projects.platforms]
    app_id = "cli_a95ddd8754f71bcb"
    ...
```

## Root Cause Analysis

1. Each project has separate Feishu app (app_id) and work_dir
2. Session files are named `{ProjectName}_{hash}.json`
3. But the session summary loading mechanism may not properly filter by current `CC_PROJECT`

## Expected Behavior

- Each Feishu bot/app should only load session context for its own project
- `CC_PROJECT` env var should gate all context loading
- Session files from other projects should be invisible to current agent

## Environment

- cc-connect version: 1.2.2-beta.5
- Platform: macOS
- Feishu apps: Two separate apps configured for two projects

## Related

- PR #569 filters `/list` and `/switch` to only show cc-connect owned sessions
- This may need similar filtering for session summary injection

## Proposed Fix

Session summary generation should check:
1. `CC_PROJECT` matches the session project name
2. Only include history from the current project session file

chenhg5:
Thanks for the report. PR #604 and PR #1276 are both in qa-claudecode review queue, no maintainer flag yet. The session context leak between projects (wrong project summary loaded) is a high-priority bug; we'll follow up on QA review progress and let you know when either PR is merged. - cc-connect/pm

PR #1276, opened 2026-06-08
fix(engine): validate session ID belongs to project before resume (#599)

chenhg5:
Fixes #599

Supersedes the stale #604 branch (which mixed the #599 validation with the already-merged #603 duplicate-session prune CLI). No prune-CLI changes are included in this fresh PR.

## Problem
When multiple cc-connect projects share a host, a stored `AgentSessionID` can point at a session file that lives in a *different* project's `~/.claude/projects/<key>/` directory. The engine would then resume the wrong project's conversation history, surfacing it in the current project's system reminders.

The issue was reproduced in real life with `ClaudeWork` and `GameStudios` configured against the same daemon: deleting the wrong session file doesn't help because the daemon recreates it.

## Changes
- **`core/interfaces.go`**: new opt-in `SessionIDValidator` interface — `ValidateSessionID(ctx, sessionID) bool`. Default agents (e.g. codex) keep working unchanged.
- **`core/engine.go`**: in `getOrCreateInteractiveStateWith`, before calling `agent.StartSession`, type-assert the agent to `SessionIDValidator` and, if it rejects the stored ID, log a warning, clear the ID, persist, and start fresh.
- **`agent/claudecode/claudecode.go`**: `*Agent` implements `ValidateSessionID` by checking for a `<sessionID>.jsonl` file under the per-project directory derived from `a.workDir` (reuses existing `findProjectDir` / `encodeClaudeProjectKey` helpers; respects the new workDir mutex).
- **Tests**:
  - 5 new unit tests in `agent/claudecode/claudecode_test.go` (valid session, invalid session, empty ID, missing project dir, cross-project leak) plus a compile-time `*Agent implements SessionIDValidator` check
  - 3 new engine tests in `core/session_id_validation_test.go` covering the dispatch (invalid → fresh start, valid → resume, no-validator → pass-through)

## Test plan
- [x] `go build -tags no_web ./...` clean
- [x] `go vet -tags no_web ./...` clean
- [x] `go test -tags no_web -count=1 ./core/... ./cmd/... ./agent/... ./platform/...` all green
- [ ] Manual verification with two configured projects (deferred — needs a real multi-project setup)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1258, opened 2026-06-07
fix(claudecode): normalize work_dir so cc-connect and desktop CC share sessions

chenhg5:
Fixes #1040

## Problem

`/list` on cc-connect could see sessions created by the desktop Claude Code
CLI, but `/resume` on the desktop CLI could not see sessions created by
cc-connect — even when both pointed at the same directory. Evidence from
`~/.claude/projects/`:

```
d--LongYinHaHa-----------/   ← cc-connect (lowercase d, 11 dashes)
D--LongYinHaHa----------/    ← desktop CC (uppercase D, 10 dashes)
```

Two different project directories for what the user perceives as the same
working directory.

## Scope vs PR #1055

PR #1055 fixed `.` and `@` encoding (#1046) but did not address #1040.
#1040's root causes are path-level, not character-level:

1. **Drive letter case** — `d:\Foo\bar` and `D:\Foo\bar` differ by one
   character in the encoded key. Windows filesystems are case-insensitive
   at the volume level, but the encoded project key is a literal string
   comparison.
2. **Trailing slashes** — `D:\Foo\bar\` encodes to `D--Foo-bar-` (with
   trailing dash), while `D:\Foo\bar` encodes to `D--Foo-bar`. Claude Code's
   internal encoding drops the trailing separator.
3. **Unicode form** — paths typed as NFD (e.g. from macOS HFS+ or some
   IM-platform inputs) produce different keys than the NFC form Claude Code
   uses internally.

## Fix

`agent/claudecode/workdir_normalize.go` adds `normalizeWorkDir()` and
`agent/claudecode/claudecode.go` applies it in `New()` and `SetWorkDir()`
so the same path string always reaches both `cmd.Dir` and
`encodeClaudeProjectKey`. Rules:

1. `filepath.Clean` collapses `..`/`.` and strips trailing separators.
2. `normalizeWindowsDriveLetter` uppercases a leading `x:` to `X:`.
3. `norm.NFC.String` canonicalizes Unicode to NFC.

`go.mod` is updated to make `golang.org/x/text` a direct dependency
(it was already an indirect one before).

## Tests

`agent/claudecode/claudecode_test.go` gains:

- `TestNormalizeWorkDir` — 12 sub-cases covering empty input, trailing
  forward slash, trailing backslash, lowercase/uppercase drive letters,
  drive letter + trailing slash, dot segments, mixed trailing separators,
  NFC pass-through, NFD → NFC, root path, and drive-only root.
[...]

PR #1581, opened 2026-07-21
fix(core): resume sessions against the agent's real working directory

vuyiv:
## Problem

Session resume validation located the agent's transcript using the project's
configured `work_dir`. But the agent process can run in a **different** directory
than `work_dir` — a custom agent command may `cd` elsewhere, and under
`run_as_user` the home/cwd differs. When the real cwd and `work_dir` diverged, the
validator looked in the wrong place, found no transcript, and reported the stored
session ID as invalid — so the engine silently discarded it and started a **fresh
conversation** instead of resuming.

Related but distinct from #1276 (validates the session ID belongs to the *project*)
and #1258 (normalizes `work_dir` for cc-connect ↔ desktop CC session sharing);
this fixes resume when the agent's *actual* cwd differs from `work_dir`.

## Fix

Capture the working directory the agent actually reports and use it for resume
validation:

- **`core/session.go`** — add `AgentCwd` to `Session` (persisted as `agent_cwd` in
  the session snapshot) with `SetAgentCwd`/`GetAgentCwd`; empty values never
  clobber a captured one.
- **`core/interfaces.go`** — `SessionIDValidator.ValidateSessionID` gains a `cwd`
  argument so validators locate the transcript at the agent's real directory,
  falling back to their configured `work_dir` when it's empty (e.g. records
  predating cwd capture). **Interface change** — implementers must update the
  signature.
- **`agent/claudecode`** — report the real cwd and validate the transcript there.
- **`core/engine.go`** — persist `agent_cwd` **every turn**, not only when the
  session ID changes, and include it in the snapshot so it survives a restart
  (this last part is the actual resume bug — a captured cwd was lost across
  restart).

Backward compatible: a snapshot with no `agent_cwd` (older record) falls back to
`work_dir`, preserving today's behavior.

## Verification

- `go build ./core/ ./agent/claudecode/` — pass
- `go test ./agent/claudecode/` and `go test ./core/` — pass; new coverage:
  `agent/claudecode/session_resume_cwd_test.go`, `core/session_agent_cwd_test.go`,
  `core/session_snapshot_cwd_test.go`, and updated `core/session_id_validation_test.go`
- `go test -race ./agent/claudecode/` (resume/cwd tests) — pass
- `go vet` + `gofmt -l` (go 1.25) — clean
- `golangci-lint run --new-from-rev origin/main ./...` — 0 issues

🤖 90% Claude

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -450,7 +450,7 @@ func (a *Agent) SetPlatformPrompt(prompt string) {
 // resume a session whose stored ID was inherited from another project, the
 // engine calls this and — on a false return — clears the ID and starts a
 // fresh session instead of reloading the wrong conversation.
-func (a *Agent) ValidateSessionID(_ context.Context, sessionID string) bool {
+func (a *Agent) ValidateSessionID(_ context.Context, sessionID, cwd string) bool {
 	if sessionID == "" {
 		return false
 	}
@@ -461,7 +461,40 @@ func (a *Agent) ValidateSessionID(_ context.Context, sessionID string) bool {
 	a.mu.RLock()
 	workDir := a.workDir
 	a.mu.RUnlock()
-	return validateSessionIDInProject(homeDir, workDir, sessionID)
+	// Locate the transcript under the session's real reported cwd, but only trust
+	// that cwd when it is within this project's work_dir. The CLI names its
+	// transcript directory after the process cwd, which a custom agent command may
+	// set to a subdirectory of work_dir — trusting an in-tree cwd is what makes
+	// resume survive a restart. A cwd OUTSIDE work_dir, however, cannot be told
+	// apart from another project's cwd leaked onto a shared Session row (issue
+	// #599), so we do not trust it and fall back to work_dir. Empty cwd (records
+	// predating cwd capture) also falls back — the original behavior.
+	dir := workDir
+	if cwd != "" && dirWithin(workDir, cwd) {
+		dir = cwd
+	}
+	return validateSessionIDInProject(homeDir, dir, sessionID)
+}
+
+// dirWithin reports whether target is base or a descendant of base, comparing
+// real path components (not encoded-key prefixes, which would conflate siblings
+// like /workspace and /workspace-other). Used to decide whether a session's
+// reported cwd is trustworthy for locating its transcript (issue #599).
+func dirWithin(base, target string) bool {
+	absBase, err := filepath.Abs(base)
+	if err != nil {
+		return false
+	}
+	absTarget, err := filepath.Abs(target)
+	if err != nil {
+		return false
+	}
+	rel, err := filepath.Rel(absBase, absTarget)
+	if err != nil {
+		return false
+	}
+	// Within base iff rel does not escape upward.
+	return rel == "." || (rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator)))
 }
 
 // validateSessionIDInProject checks whether sessionID has a .jsonl file
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -33,6 +33,7 @@ type claudeSession struct {
 	stdinMu         sync.Mutex
 	events          chan core.Event
 	sessionID       atomic.Value // stores string
+	cwd             atomic.Value // stores string — working dir the CLI reports in its init event
 	permissionMode  atomic.Value // stores string
 	autoApprove     atomic.Bool
 	acceptEditsOnly atomic.Bool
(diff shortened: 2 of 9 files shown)


chenhg5 (approved):
## Conclusion: Approve

Well-engineered bugfix. The change closes a real silent-failure bug (resume → fresh conversation when agent cwd ≠ `work_dir`) while preserving the cross-project safety from issue #599. The interface-signature change is handled textbook-style: explicit marker in the diff, every implementer updated (only `agent/claudecode` + the test stub in the repo — verified via code search), and `omitempty` plus empty-fallback keeps backward compatibility with pre-fix snapshots.

The standout detail is `dirWithin` using `filepath.Rel` instead of `strings.HasPrefix` — that's exactly the prefix-trap that originally caused #599 (siblings like `/workspace` vs `/workspace-other`), and it correctly defends against leaked foreign cwds here too. The snapshot-regression test (`TestSessionManager_PersistsAgentCwd`) is the most valuable single test in this PR: it exercises the actual `saveLocked` snapshot path (not a direct `json.Marshal`), so it would catch a future refactor that re-introduces the hand-written deep-copy omission.

### What looks good

- **Interface migration is complete**: only `agent/claudecode` (the lone production implementer) and `core/session_id_validation_test.go` (test stub) implement `SessionIDValidator` — both updated. No other agent runtime (codex / pi / opencode / cursor / kimi / dsh / acp / grok) implements this interface, so the signature change is fully contained.
- **Per-turn persistence (not gated on session-ID change)**: `processInteractiveEvents` persists `agent_cwd` every turn because cwd typically arrives in the CLI init event *after* the session ID. Gating on session-ID change would have missed the cwd capture — the comment explaining this is exactly the kind of "why" comment that prevents a future simplification regression.
- **Defensive `SetAgentCwd`**: empty values are no-ops, so a missing cwd never clobbers a previously captured one. Tested explicitly.
- **`omitempty` + empty fallback**: pre-fix session records (no `agent_cwd` field) load with empty cwd → validator falls back to `work_dir` → identical to pre-fix behavior. Backward compat verified by `TestSessionAgentCwd`'s JSON round-trip sub-test.
- **`dirWithin` uses `filepath.Rel`**: the right primitive for path containment. Correctly rejects `/workspace` and `/workspace-other` as siblings, and accepts `/workspace/sub` and `/workspace/sub/deeper` as descendants. This is the same defense pattern that fixed #599, applied here.
- **Race safety**: `claudeSession.cwd` is `atomic.Value`; `Session.AgentCwd` is guarded by `Session.mu` in both `SetAgentCwd` and `GetAgentCwd`. The PR's `-race` test run covers the resume/cwd paths.
- **Snapshot regression test**: `TestSessionManager_PersistsAgentCwd` does a save/reload round-trip with a fresh `SessionManager` (mirrors a daemon restart), then asserts both `AgentSessionID` and `AgentCwd` survive. This is exactly the test that would have caught the original bug.
- **Cross-project leak test (`TestValidateSessionID_CrossProjectSharedRow`)**: explicitly models the #599 vector with a shared Session row carrying Project A's cwd into Project B's validation — verifies the guard holds.

### 🔵 Optional (not blocking)

- **P3 — symlink handling**: `dirWithin` uses `filepath.Abs` but does not `EvalSymlinks`. If `work_dir` is a symlink (e.g. `/var/repo` → `/mnt/data/repo`) and the CLI reports the realpath as `cwd`, the comparison may fail and validator falls back to `work_dir` (still safe, but the transcript lookup now happens under the symlink path, not the realpath). In practice CLI reports paths consistent with the configured `work_dir`, so impact is low. If you want to bullet-proof it, `filepath.EvalSymlinks` on both sides before `Rel` would resolve it — at the cost of a couple of stat calls per validation.
- **P3 — CHANGELOG note for the interface change**: anyone reading the change log should know `SessionIDValidator.ValidateSessionID` gained a `cwd` parameter. Even though the in-repo migration is complete, downstream forks or external implementers (if any exist) will need to update. One-line release note is sufficient.
- **P3 — relative cwd edge case**: if the CLI ever reports a relative cwd (unusual but possible), `filepath.Abs` will resolve it against the cc-connect process's cwd at validation time, which may not match what the CLI intended. Realistically the CLI reports absolute paths, but worth a defensive comment in `dirWithin` if you want to flag the assumption.

### Testing / Risk

- **Verified**: CI 5/5 green on `1e1b6fab`. `go vet`, `gofmt -l`, `golangci-lint --new-from-rev origin/main` all clean. Code-search confirmed no other production implementer of `SessionIDValidator` exists in the repo. All 3 new test files plus the updated test stub exercise the relevant paths (accessor, JSON contract, snapshot persistence, cross-project safety, fallback to work_dir).
- **Remaining risk**: P3 items above (symlink, relative cwd) are theoretical edge cases. The functional correctness for the documented bug is solid.

### Next step

Merge-ready. Consider the CHANGELOG line for the interface signature change, but no code changes are required to ship this.

PR #1623, opened 2026-07-31
fix(codex): persist session ID on thread start

hl1221hl:
## Background

A fresh Codex CLI session exposes its thread ID in `thread.started`. Previously the adapter retained that ID only in memory and did not emit it until `turn.completed`. If cc-connect restarted during the first turn, the session binding was never persisted, so the next user message started a new Codex conversation instead of resuming the existing rollout.

## Changes

- Emit an empty `EventText` carrying the session ID as soon as Codex emits `thread.started`, allowing the Engine to persist the binding immediately.
- Add a regression test proving the session ID event arrives before `turn.completed`.
- Update the close lifecycle test to consume the expected startup session-ID event before asserting that shutdown suppresses delayed child output.

## Validation

- `go test -race ./agent/codex -count=1`
- `go build ./...`
- `go vet ./agent/codex`
- `golangci-lint run --new-from-rev upstream/main ./...`
- `go test ./...` was run; only pre-existing local macOS environment failures remain: `TestCUJ_A5_FileReachesAgent` temporary-directory cleanup and `TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable` without a launchd GUI domain.

### DIFF
--- a/agent/codex/session.go
+++ b/agent/codex/session.go
@@ -381,6 +381,12 @@ func (cs *codexSession) handleEvent(raw map[string]any) {
 			cs.contextUsage = nil
 			cs.contextMu.Unlock()
 			slog.Debug("codexSession: thread started", "thread_id", tid)
+			// 立即通知 Engine 持久化新会话 ID，避免首轮尚未完成时重启导致无法恢复。
+			select {
+			case cs.events <- core.Event{Type: core.EventText, SessionID: tid}:
+			case <-cs.ctx.Done():
+				return
+			}
 		}
 
 	case "turn.started":
--- a/agent/codex/session_test.go
+++ b/agent/codex/session_test.go
@@ -437,6 +437,36 @@ func TestSend_WithImages_PassesImageArgsAndDefaultPrompt(t *testing.T) {
 	}
 }
 
+// TestCodexSession_ThreadStartedEmitsSessionID 验证首轮启动时立即上报会话 ID，
+// 使 Engine 可以在 turn.completed 之前持久化绑定关系。
+func TestCodexSession_ThreadStartedEmitsSessionID(t *testing.T) {
+	cs, err := newCodexSession(context.Background(), "codex", nil, t.TempDir(), "", "", "", "", "", nil, "", "", "")
+	if err != nil {
+		t.Fatalf("newCodexSession: %v", err)
+	}
+	defer func() { _ = cs.Close() }()
+
+	cs.handleEvent(map[string]any{
+		"type":      "thread.started",
+		"thread_id": "thread-before-turn-completed",
+	})
+
+	select {
+	case event := <-cs.Events():
+		if event.Type != core.EventText {
+			t.Fatalf("event type = %q, want %q", event.Type, core.EventText)
+		}
+		if event.SessionID != "thread-before-turn-completed" {
+			t.Fatalf("event session ID = %q, want thread-before-turn-completed", event.SessionID)
+		}
+		if event.Content != "" {
+			t.Fatalf("event content = %q, want empty", event.Content)
+		}
+	case <-time.After(time.Second):
+		t.Fatal("thread.started did not emit session ID event")
+	}
+}
+
 func TestSend_ResumeWithImages_PlacesSessionBeforeImageFlags(t *testing.T) {
 	workDir := t.TempDir()
 	binDir := filepath.Join(workDir, "bin")
@@ -877,6 +907,14 @@ func TestClose_ForceKillsProcessGroupAfterGracefulTimeout(t *testing.T) {
 	}
 
 	waitForThreadID(t, cs, "thread-close")
+	select {
+	case event := <-cs.Events():
+		if event.Type != core.EventText || event.SessionID != "thread-close" {
+			t.Fatalf("startup event = %#v, want session ID event for thread-close", event)
+		}
+	case <-time.After(time.Second):
+		t.Fatal("timed out waiting for startup session ID event")
+	}
 
 	closeStarted := time.Now()
 	if err := cs.Close(); err != nil {


Nicolas-Li:
感谢这个修复。我们在线上遇到了同样的首轮 session ID 持久化缺口，触发方式是**首轮执行期间 `/stop`**，可以补充一个复现场景。

环境：

- cc-connect `v1.5.1-beta.1`，commit `80713167`
- Codex CLI `0.156.0`
- Linux/systemd，飞书群聊，`thread_isolation = true`

2026-09-23 的实际时间线（UTC+08:00）：

1. 使用 `/new` 创建命名会话。
2. **14:40:16** 发送首轮问题。
3. **14:40:17** Codex 已创建 thread 和 rollout，随后持续执行。
4. **14:53:46**，首轮尚未结束时发送 `/stop`。
5. 随后执行 `/current`，发现绑定的 agent session ID 为空。

只读检查确认：

- Codex rollout 文件仍然存在，包含已经生成的 thread ID 和执行历史。
- rollout 中的用户问题与 cc-connect 会话保存的输入完全一致。
- cc-connect 对应会话的持久字段仍为 `agent_session_id: ""`。
- 中断前没有观察到正常的轮次完成记录。

部署版本的 `/stop` 已经会保留**原先保存的 ID**。但这次 ID 尚未持久化：`thread.started` 只更新适配器内存中的 `threadID`，停止会话又绕过了正常结束时的回写。

因此，这个触发窗口会持续到首轮完成。本次已经运行约 **13 分 29 秒**，仍然触发了问题。

这个 PR 在 `thread.started` 时立即上报 ID 的修改，与我们定位到的缺口一致。建议再补一个 Engine 层回归测试：

- 收到 `thread.started`，随后产生思考或工具事件，但不发送 `turn.completed`。
- 执行 `/stop`。
- 验证持久化绑定和 `/current` 仍保留原 ID。
- 再发送一条消息，验证使用 `codex exec resume <原ID>`。

本次没有向线上服务再次发送问题；后续会新建会话的结论，是根据空的持久化绑定和启动代码路径确认的。

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Targeted fix that makes the Codex adapter persist `thread.started` session IDs as soon as Codex emits them, instead of waiting for `turn.completed`. CI is 5/5 green on commit `61d5beefa264f2a519dee82f340077c26701dfa4`, mergeable, base = main, no conflicts. Both the new test and the updated close lifecycle test directly verify the new ordering.

Review scope:
- Reviewed `agent/codex/session.go` (the new `select { case cs.events <- core.Event{Type: core.EventText, SessionID: tid}; case <-cs.ctx.Done() }` in the `thread.started` branch), `agent/codex/session_test.go` (`TestCodexSession_ThreadStartedEmitsSessionID` + the updated `TestClose_ForceKillsProcessGroupAfterGracefulTimeout`).
- Confirmed the engine-side handler in `core/engine.go` already treats `EventText` + `event.SessionID != ""` as the signal to call `SetAgentSessionID` + `sessions.Save()` (existing code path, untouched here).
- Focused on correctness, ordering, race-vs-cancel handling, and downstream side effects of an empty-content EventText.

✅ What looks good:
- The fix is narrowly scoped to the `thread.started` branch in `handleEvent`. The earlier `cs.threadID.Store(tid)` line keeps the in-memory cache consistent; the new emit just makes that ID visible to the engine so it can save it.
- The `select` with `cs.ctx.Done()` is the right cancellation pattern: if the session is already cancelled by the time `thread.started` arrives, the goroutine exits instead of leaking a blocking send onto `cs.events`. Matches the pattern already used by `turn.completed` / `turn.failed` in the same function.
- The engine-side handler at the receiving end is already prepared for this — `case EventText:` does `if event.Content != "" { textParts = append(textParts, event.Content) }` (empty content is a no-op for the text pipeline) and then independently handles `if event.SessionID != "" { session.SetAgentSessionID(...); sessions.Save() }`. So emitting an empty-content EventText purely for the side effect of triggering the persistence path does not pollute the user-facing text flow.
- `SetAgentSessionID` only fires when `event.SessionID != session.GetAgentSessionID()`, and `sessions.Save()` only fires when the IDs differ — so emitting the same ID twice (e.g. for `/resume` against an existing session) is a no-op save, not a churn.
- The new `TestCodexSession_ThreadStartedEmitsSessionID` test covers exactly the reported scenario: it constructs a session, feeds `thread.started`, and asserts the event arrives at `cs.Events()` within 1s with `Type == EventText`, `SessionID == "thread-before-turn-completed"`, and `Content == ""`. This is the strongest end-to-end check available at the agent-session boundary.
- The update to `TestClose_ForceKillsProcessGroupAfterGracefulTimeout` (consuming the new startup session-ID event before asserting that shutdown suppresses delayed child output) is the right regression guard: if a future refactor accidentally removed the emit, that close test would hang on `waitForThreadID` instead of failing cleanly.
- The comment block above the emit is in Chinese ("立即通知 Engine 持久化新会话 ID，避免首轮尚未完成时重启导致无法恢复。") — matches the comment style used elsewhere in the same file (e.g. the `turn.completed` branch) and serves as a tripwire for future readers. The PR is otherwise a clean, minimal diff.

🚨/🔴 Must fix:

🟠 Should improve:
- None worth blocking on.

🔵 Optional:
- Consider a one-line `slog.Info` (not just `Debug`) on the thread.started branch right after the persistence-bound emit, since this event is now load-bearing for resume-after-restart. Operators debugging "why did the agent start a fresh conversation" would otherwise have to know to enable debug logs. Cosmetic, matches `slog.Info` usage elsewhere in the file.
- The pattern `core.Event{Type: core.EventText, SessionID: tid}` with empty Content is correct here but slightly unusual: EventText normally carries text content. A dedicated `EventSessionID` (or a flag on Event) would make the intent clearer at the type level and would prevent a future "empty-content EventText is silently dropped" optimization from breaking this code path. Not required for correctness today, but worth a one-liner in the comment noting the dependency on EventText accepting empty content. Cosmetic.

❓ Questions:
- None.

Testing / Risk:
- `go vet`, `go build ./...`, `golangci-lint run --new-from-rev upstream/main ./...` reported clean in the PR body.
- The two pre-existing local macOS failures called out in the PR body (`TestCUJ_A5_FileReachesAgent`, `TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable`) are unrelated to this change — they are platform-environment specific.
- Risk scope: the new emit fires before any `turn.started` / `item.started` event, so it can only affect the persistence path and the (currently no-op-for-empty-content) text pipeline. No effect on tool calls, turn completion, error reporting, or shutdown ordering.
- Failure mode worth noting: if a future refactor of the engine's `case EventText:` branch starts treating empty-content events as a "no-op for any side effects" optimization, this code would silently stop persisting thread IDs. The optional `EventSessionID` suggestion above would close that gap.

Next step:

PR #1626, opened 2026-07-31
fix(dingtalk): forward inline images in richText messages

xiaoxinglgm:
## Summary

Forward images embedded in DingTalk richText callbacks together with their surrounding text instead of silently dropping them. The download flow prefers the v1.0 `downloadCode`, falls back to legacy `pictureDownloadCode`, uses the callback robot code when available, and includes bounded API error details for diagnostics.

## Type of change

- [ ] Documentation only

- `TestOnRawMessage_RichTextForwardsInlineImage` in `platform/dingtalk/dingtalk_test.go` — verifies text and image bytes are forwarded in one message, callback robotCode is used, and v1 downloadCode falls back to pictureDownloadCode after an API failure.
- `TestExtractRichText` in `platform/dingtalk/dingtalk_test.go` — verifies richText text extraction and ordered v1/legacy image download candidates.

- Regression test name: `TestOnRawMessage_RichTextForwardsInlineImage`
  - [x] Reverted the fix locally; the regression test failed as expected with zero forwarded images.

- [x] No CUJ touched (platform-specific DingTalk callback parsing and attachment download)
- [ ] A — basic conversation

A DingTalk message containing text plus an inline screenshot now reaches the agent with both the text and image. Verified against a live DingTalk Stream bot: the pre-fix service logged `has_images=false` and the agent reported no attachment; after deploying this branch the user confirmed the image is visible.

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] i18n strings have all-language translations (no new user-facing strings)

## Related

- Issue: N/A
- Related PR: N/A

### DIFF
--- a/platform/dingtalk/dingtalk.go
+++ b/platform/dingtalk/dingtalk.go
@@ -239,16 +239,17 @@ func (p *Platform) onRawMessage(rawJSON string) {
 	// Parse the full "text" object from raw JSON to recover isReplyMsg/repliedMsg.
 	// The SDK's BotCallbackDataTextModel only has Content string, losing these fields.
 	var envelope struct {
-		Text richTextContent `json:"text"`
+		Text      richTextContent `json:"text"`
+		RobotCode string          `json:"robotCode"`
 	}
 	if err := json.Unmarshal([]byte(rawJSON), &envelope); err != nil {
 		slog.Warn("dingtalk: failed to parse rich text content", "error", err)
 	}
 
-	p.onMessage(&data, &envelope.Text)
+	p.onMessage(&data, &envelope.Text, envelope.RobotCode)
 }
 
-func (p *Platform) onMessage(data *chatbot.BotCallbackDataModel, richText *richTextContent) {
+func (p *Platform) onMessage(data *chatbot.BotCallbackDataModel, richText *richTextContent, callbackRobotCode string) {
 	slog.Debug("dingtalk: message received", "user", data.SenderNick, "msgtype", data.Msgtype)
 
 	if p.dedup.IsDuplicate(data.MsgId) {
@@ -288,11 +289,24 @@ func (p *Platform) onMessage(data *chatbot.BotCallbackDataModel, richText *richT
 		return
 	}
 
-	// Handle richText messages — extract plain text from rich content
+	// Handle richText messages, including images embedded alongside text.
 	if data.Msgtype == "richText" {
-		text := extractRichText(data.Content)
-		if text == "" {
-			slog.Debug("dingtalk: richText message with no extractable text", "msg_id", data.MsgId)
+		text, pictureDownloads := extractRichText(data.Content)
+		images := make([]core.ImageAttachment, 0, len(pictureDownloads))
+		for _, picture := range pictureDownloads {
+			img, err := p.downloadImage(picture.DownloadCode, callbackRobotCode)
+			if err != nil && picture.FallbackCode != "" {
+				slog.Warn("dingtalk: retrying richText image with legacy download code", "error", err, "msg_id", data.MsgId)
+				img, err = p.downloadImage(picture.FallbackCode, callbackRobotCode)
+			}
+			if err != nil {
+				slog.Error("dingtalk: failed to download richText image", "error", err, "msg_id", data.MsgId)
+				continue
+			}
+			images = append(images, img)
+		}
+		if text == "" && len(images) == 0 {
+			slog.Debug("dingtalk: richText message with no extractable content", "msg_id", data.MsgId)
 			return
 		}
 		msg := &core.Message{
@@ -303,6 +317,7 @@ func (p *Platform) onMessage(data *chatbot.BotCallbackDataModel, richText *richT
 			ChatName:   data.ConversationTitle,
 			Content:    text,
 			MessageID:  data.MsgId,
+			ChannelKey: data.ConversationId,
 			ReplyCtx: replyContext{
 				sessionWebhook: data.SessionWebhook,
 				conversationId: data.ConversationId,
@@ -310,6 +325,7 @@ func (p *Platform) onMessage(data *chatbot.BotCallbackDataModel, richText *richT
 				messageID:      data.MsgId,
 				isGroup:        data.ConversationType == "2",
 			},
+			Images: images,
 		}
 		p.handler(p, msg)
 		return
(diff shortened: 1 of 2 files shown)


jsonj3407-source (commented):
## Independent verification (live DingTalk + Cursor)

We hit the same bug in production (DingTalk Stream → Cursor via cc-connect v1.4.1) and landed a local patch that matches this PR’s intent.

### Before
- Pure picture/image inbound worked.
- Mixed text+image arrived as msgtype=richText with only text forwarded (has_images=false).
- After extracting codes, some downloads failed with messageFiles/download HTTP 500 when preferring legacy pictureDownloadCode.

### After (same core fix as this PR)
- Extract inline picture codes from richText (downloadCode preferred, pictureDownloadCode as fallback/additional candidate).
- Download via existing messageFiles/download path and attach msg.Images.
- Keep forwarding text when image download fails; still dispatch when text is empty but images succeed.
- Surface API error body on non-200 download URL responses.

### Live result
On 2026-08-01 after deploying the equivalent fix:
- log: 
ichText pictures found → image downloaded successfully → has_images=true with content_len>0
- Cursor received the image attachment path under .cc-connect/attachments/YYYY-MM-DD/

### Comparison note
Upstream PR extras we did not need locally (but look good): callback 
obotCode override, ordered per-image v1→legacy retry. Our local extras: a few more field aliases / JSON-string content normalization. **Core behavior is the same.**

**LGTM from a live DingTalk user of this exact path — please merge when you can.** Thanks @xiaoxinglgm!

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Solid bug fix that closes a real user-visible regression: DingTalk richText messages were silently dropping inline images because the adapter only knew about the legacy `pictureDownloadCode` field. The new `extractRichText` returns ordered `(text, pictureDownloads)` and the call site prefers the v1.0 `downloadCode` while falling back to `pictureDownloadCode` on API failure, with the callback robotCode correctly threaded through to the download request. Tests cover the v1-first-then-legacy ordering, callback robotCode override, and the regression case. Ready to merge.

Review scope:
- platform/dingtalk/dingtalk.go — `extractRichText` refactor, image-download extraction, `downloadImage` / `getDownloadURL` / `downloadAudio` robotCode plumbing, error-detail propagation in `getDownloadURL`.
- platform/dingtalk/dingtalk_test.go — new `TestOnRawMessage_RichTextForwardsInlineImage`, expanded `TestExtractRichText`, refactored `dingtalkFileDownloadRT` → `dingtalkDownloadRT` to inject a forced v1 failure for the fallback path.
- Run: `go test -race ./platform/dingtalk/... -count=1 -timeout 60s` — pass.

✅ What looks good:
- Cleanly separates the v1 and legacy fields into one `richTextImageDownload` record, and the order is encoded as "v1 first, legacy second". The download site consumes the same record in that order, so the producer and consumer stay aligned.
- Uses the callback-provided `robotCode` (`envelope.RobotCode`) when present and falls back to `p.robotCode`; the test asserts `rt.requestRobotCode == "robot-from-callback"`, which is the exact regression we want to lock.
- Bounded API error body (4 KiB) propagated up via `fmt.Errorf` rather than just logging status code — this will be useful for diagnosing future image-download failures from the DingTalk side without enabling debug logging in production.
- Test regression approach is good: `failDownloadCode` triggers a 500 on the v1 download request, then the test asserts the request log shows both `dc-v1-image` and `dc-legacy-image` in that order, which directly mirrors the production fallback behavior.
- Refactor of `dingtalkFileDownloadRT` → `dingtalkDownloadRT` (now shared between file and image tests) reduces duplication and avoids the rename from breaking anything else.

🚨/🔴 Must fix:
- None found. No merge-blocking correctness/security issue.

🟠 Should improve:
- 🟠 P2 — In `extractRichText`, when an item has both `downloadCode` and `pictureDownloadCode` and they are equal, you correctly drop the fallback (`fallbackCode = ""`). But when an item has only `pictureDownloadCode` (legacy callback) you do `downloadCode, fallbackCode = fallbackCode, ""` — meaning the legacy path will never trigger a retry even if the legacy endpoint returns an error. That is intentional ("legacy only has one shot"), but the test does not cover it. Consider one extra case `{"pictureDownloadCode": "legacy-only"}` → expect one picture with `DownloadCode="legacy-only", FallbackCode=""`. Cheap to add, locks the intent.
- 🟠 P2 — The dedupe guard `if p.dedup.IsDuplicate(data.MsgId) return` runs before richText handling. The image bytes inside a richText message will be re-downloaded on every redelivery. If DingTalk retries are common (which is plausible given the 4xx/5xx observable in production), consider caching the result by `MsgId` to avoid hammering the download API. Not blocking, just worth a follow-up.
- 🔵 P3 — `getDownloadURL` now returns a wrapped error that includes up to 4 KiB of API response body. For a 200-class JSON error (e.g. `{"code":"...","message":"..."}`) this is helpful, but if the server ever returns a non-UTF-8 binary blob the error string will be unreadable. Consider `string(errorBody)` only when `json.Valid(errorBody)` is true.
- 🔵 P3 — CHANGELOG.md is untouched. Suggest a one-line Unreleased `### Fixed — dingtalk: forward inline images in richText callbacks (#1626)`.

❓ Questions:
- The new branch only triggers when `len(images) == 0 && text == ""` returns early. What about messages where `text == ""` but `len(images) > 0` (pure-image richText)? Current code falls through to build a `core.Message` with empty `Content` and `Images: images`. Is that the intended behavior, or should we suppress such messages? Worth confirming with PM/user-ops whether silent image-only richText is acceptable.
- Is the failure case `slog.Error("dingtalk: failed to download richText image", ..., "msg_id", data.MsgId)` then `continue` correct? The user will see text-only content with a missing image and no error. If the message was image-only, it will silently disappear. Consider sending a fallback `Content` like `"<image failed to download>"` so the agent at least sees that something was lost.

Testing / Risk:
- Verified: `go test -race ./platform/dingtalk/... -count=1 -timeout 60s` — ok in ~1.04s.
- Verified regression approach: temporarily reverting the richText branch in `extractRichText` would break `TestOnRawMessage_RichTextForwardsInlineImage` because `Images` would be empty.
- Risk surface: bounded to `platform/dingtalk`. The `core.Message.Images` field is pre-existing; the contract change is only that richText now populates it. Other adapter behavior unchanged.

Next step:
- Merge as-is. The two 🟠 items are small, well-isolated improvements that can be follow-up PRs. CHANGELOG entry can be added at release time if the author prefers.

PR #1631, opened 2026-08-02
feat(cloud_web): support extra_headers on websocket transport

MaverickDai:
## Summary

Allow injecting custom HTTP headers into the `cloud_web` websocket upgrade request via the `[projects.platforms.options.extra_headers]` TOML table.

This is needed for upstreams that require bespoke auth headers beyond the standard `Authorization`/`X-Cloud-Web-Token` pair — notably **IAP (Identity-Aware Proxy) AK/SK auth**, which requires `X-Brain-Authorization: Basic <base64(ak:sk)>`.

## Problem

cc-connect's `cloud_web` websocket transport only sends two fixed headers (`Authorization: Bearer <token>` and `X-Cloud-Web-Token: <token>`). There is no way to add custom headers to the WebSocket upgrade request.

When deploying a cloud_web relay behind an IAP internal domain (e.g. `*.iap-rjob-internal.platform.shaipower.com`), the IAP gateway requires an `X-Brain-Authorization` header for AK/SK authentication. Without custom header support, cc-connect cannot directly reach such a relay — forcing users to run a local sidecar proxy that injects the header, which is an unnecessary operational burden.

## Solution

Add an optional `extra_headers` map to the cloud_web platform options. Each key-value pair is injected into the WebSocket upgrade request's HTTP headers, alongside the existing `Authorization`/`X-Cloud-Web-Token` headers.

### Config example

```toml
[[projects.platforms]]
type = "cloud_web"

[projects.platforms.options]
transport = "websocket"
token = "relay-token"
ws_url = "wss://my-relay.iap-rjob-internal.example.com/cloud-web/ws"

[projects.platforms.options.extra_headers]
X-Brain-Authorization = "Basic <base64(accesskey:secretkey)>"
X-Custom-Header = "any-value"
```

## Changes

- **`platform/cloud-web/ws.go`**: `wsTransport` gains an `extraHeaders map[string]string` field; `connectOnce` iterates it and calls `header.Set(k, v)` (empty values are skipped).
- **`platform/cloud-web/cloudweb.go`**: `New()` reads `extra_headers` from opts via new `pickStringMap` helper (converts TOML inline table `map[string]any` → `map[string]string`); passes it to `newWSTransport`.
- **`platform/cloud-web/ws_test.go`**: `TestExtraHeadersInjected` — spins up a test WS server, configures `extra_headers`, and asserts the custom header (`X-Brain-Authorization`), the standard headers (`Authorization`, `X-Cloud-Web-Token`), and the empty-value-skip behavior.

## Backward compatibility

**No breaking changes.** `extra_headers` is entirely optional. Omitting it yields the exact prior behavior (only `Authorization` + `X-Cloud-Web-Token` are sent). The new `pickStringMap` helper returns `nil` for missing/non-table values, and the `wsTransport` loop is a no-op when `extraHeaders` is `nil`.

## Testing

- `go test ./platform/cloud-web/` — all tests pass, including the new `TestExtraHeadersInjected`.
- Verified end-to-end: built cc-connect with this change, configured `extra_headers` with a real IAP AK/SK, and confirmed cc-connect connects directly to a relay behind `*.iap-rjob-internal.platform.shaipower.com` and registers successfully (`cloud_web: websocket connected` + `platform ready`).

```
go test ./platform/cloud-web/ -run TestExtraHeadersInjected -v
=== RUN   TestExtraHeadersInjected
--- PASS: TestExtraHeadersInjected (0.02s)
PASS
```

## Notes

- Pre-existing test failures in `daemon/`, `cmd/cc-connect/`, and `web/` are unrelated to this change (they fail on `main` too: `daemon` is a macOS launchd env issue; `cmd/cc-connect`/`web` need the embedded `dist/` frontend assets which require a separate build step).

### DIFF
--- a/platform/cloud-web/cloudweb.go
+++ b/platform/cloud-web/cloudweb.go
@@ -102,7 +102,7 @@ func New(opts map[string]any) (core.Platform, error) {
 		if strings.TrimSpace(resolved) == "" {
 			return nil, fmt.Errorf("cloud_web: ws_url or base_url is required for websocket transport")
 		}
-		tp = newWSTransport(resolved, token, name, project)
+		tp = newWSTransport(resolved, token, name, project, pickStringMap(opts["extra_headers"]))
 	case "long_poll":
 		if strings.TrimSpace(baseURL) == "" {
 			return nil, fmt.Errorf("cloud_web: base_url is required for long_poll transport")
@@ -152,6 +152,25 @@ func pickInt(v any) int {
 		}
 	}
 	return 0
+}
+
+// pickStringMap converts a TOML inline table to map[string]string.
+// Used for extra_headers: [projects.platforms.options.extra_headers].
+func pickStringMap(v any) map[string]string {
+	m, ok := v.(map[string]any)
+	if !ok {
+		return nil
+	}
+	out := make(map[string]string, len(m))
+	for k, val := range m {
+		if s, ok := val.(string); ok {
+			out[k] = s
+		}
+	}
+	if len(out) == 0 {
+		return nil
+	}
+	return out
 }
 
 func (p *Platform) Name() string { return p.name }
--- a/platform/cloud-web/ws.go
+++ b/platform/cloud-web/ws.go
@@ -27,6 +27,12 @@ type wsTransport struct {
 	name    string
 	project string
 
+	// extraHeaders are injected into the WebSocket upgrade request, in
+	// addition to the standard Authorization/X-Cloud-Web-Token headers.
+	// Used for upstreams that require bespoke auth (e.g. IAP AK/SK via
+	// X-Brain-Authorization).
+	extraHeaders map[string]string
+
 	mu        sync.RWMutex
 	caps      map[string]bool
 	conn      *websocket.Conn
@@ -40,12 +46,13 @@ type wsTransport struct {
 	previewRequests map[string]chan string
 }
 
-func newWSTransport(wsURL, token, name, project string) *wsTransport {
+func newWSTransport(wsURL, token, name, project string, extraHeaders map[string]string) *wsTransport {
 	return &wsTransport{
 		wsURL:           wsURL,
 		token:           token,
 		name:            name,
 		project:         project,
+		extraHeaders:    extraHeaders,
 		caps:            defaultCapabilities(),
 		previewRequests: make(map[string]chan string),
 	}
@@ -130,6 +137,11 @@ func (t *wsTransport) connectOnce(ctx context.Context) error {
 	if t.token != "" {
 		header.Set("Authorization", "Bearer "+t.token)
 		header.Set("X-Cloud-Web-Token", t.token)
+	}
+	for k, v := range t.extraHeaders {
+		if v != "" {
+			header.Set(k, v)
+		}
 	}
 
 	conn, _, err := websocket.DefaultDialer.DialContext(ctx, u.String(), header)
--- a/platform/cloud-web/ws_test.go
+++ b/platform/cloud-web/ws_test.go
@@ -123,3 +123,67 @@ func TestAllowFromFilter(t *testing.T) {
 		t.Fatal("expected blocked user")
 	}
 }
+
+// TestExtraHeadersInjected verifies that extra_headers from the platform
+// options are injected into the WebSocket upgrade request, alongside the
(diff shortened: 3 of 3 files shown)


chenhg5 (commented):
## Conclusion: Comment

Overall assessment:
- The feature itself is designed correctly, backward-compatible, and the production code has no concurrency issue. The CI failure is purely from a data race in the new test code — adding a `sync.Mutex` (matching the existing `TestWebSocketIntegration` pattern) unblocks the merge. Recommend treating as P2 follow-up, do not block on design re-litigation.

Review scope:
- Read `platform/cloud-web/cloudweb.go`, `ws.go`, `ws_test.go`.
- Focused on correctness, backward compatibility, and concurrency.
- Locally reproduced the race detector finding: `go test -race -count=3 ./platform/cloud-web/... -run TestExtraHeadersInjected` reliably reports the same race. Write at `ws_test.go:134–137` (HTTP handler goroutine writes `gotAuth`/`gotCloudWeb`/`gotCustom`/`gotEmpty`), reads at `ws_test.go:171` (polling loop) and `ws_test.go:177–186` (assertions) — neither side locks.

✅ What looks good:
- `extra_headers` is fully optional; `pickStringMap` returns nil for missing/non-table input, and the `wsTransport` with no map makes the range loop a no-op. Behavior is identical to pre-change.
- The `if v != ""` filter at `ws.go:142` avoids sending `X-Empty:` empty headers upstream; the test verifies this, the design choice is deliberate.
- `extraHeaders` is set once at construction and only read inside `connectOnce` (single-goroutine per transport; `connectLoop` reconnection is serialized). The production code path has no concurrency risk; map iteration over an unmodified map is safe.
- The use case (IAP AK/SK auth via `X-Brain-Authorization`) is concrete and the author verified end-to-end against a real relay. PR description includes a tombstone TOML configuration example that future maintainers can copy.

🟠 Should improve (P2):
- `TestExtraHeadersInjected` has a data race, causing `-race` CI to fail.
  - Impact: CI red blocks merge, but the feature itself is fine.
  - Evidence: local reproduction `go test -race -count=3 ./platform/cloud-web/... -run TestExtraHeadersInjected` reliably reports the race; the four `got*` variables are written by the handler goroutine and read by the test main goroutine with no synchronization.
  - Suggested fix: Reuse the `TestWebSocketIntegration` pattern (`var mu sync.Mutex` at ws_test.go:18). Wrap the handler writes and test reads with `mu.Lock()/Unlock()`. Minimal patch shape:
    ```go
    var (
        mu                                       sync.Mutex
        gotAuth, gotCloudWeb, gotCustom, gotEmpty string
    )
    srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        mu.Lock()
        gotAuth = r.Header.Get("Authorization")
        gotCloudWeb = r.Header.Get("X-Cloud-Web-Token")
        gotCustom = r.Header.Get("X-Brain-Authorization")
        gotEmpty = r.Header.Get("X-Empty")
        mu.Unlock()
        // ...
    }))
    // polling loop + assertions: each read wrapped with mu.Lock()/Unlock()
    ```
  - Verification: `go test -race -count=3 ./platform/cloud-web/... -run TestExtraHeadersInjected` should pass stably, and `go test -race ./platform/cloud-web/...` should not regress overall.

🔵 Optional (P3):
- The doc comment at ws.go:30–34 says "in addition to" but doesn't clarify that `extra_headers` entries for `Authorization`/`X-Cloud-Web-Token` will override the standard headers. Recommend adding: "extra_headers takes precedence over the standard Authorization/X-Cloud-Web-Token when the same key is provided" to spare future users the surprise.
- A 1-line test covering `extra_headers` overriding `Authorization` would lock the override behavior in code.
- `pickStringMap` silently drops non-string values; a debug log for non-string values would help users debug typos like `X-Header = 42` (currently treated as missing).

❓ Questions: None.

Testing / Risk:
- Verified: author locally ran `go test ./platform/cloud-web/` passing, and end-to-end verified the feature against a real IAP relay. Locally reproduced `-race` failure confirming the issue is in the test code.
- Uncovered: no test for `extra_headers` overriding the standard header; `pickStringMap` silent drop of non-string values has no test. Both are P3, do not block merge.

Next step: Author adds the `sync.Mutex` per the P2 suggestion and re-triggers CI. Production code does not need changes, the feature design does not need to be redone — merge direction is clear.

PR #1636, opened 2026-08-03
fix(pi): fall back to models-store.json when enabledModels is unset

baizhu945:
## Problem

The pi agent's `AvailableModels` reads only `enabledModels` from `~/.pi/agent/settings.json`. When the user has not configured `enabledModels` (the default), it returns an empty list.

As a result, `/model` on card-capable platforms (Feishu) renders a model card whose select dropdown contains **only the current model** — there is no way to switch to any other model. On non-card platforms the list is also empty.

This is the gap left by #1178 (`fix(pi): read enabledModels from settings.json for /model`): it wired the *restricted* list but provided no fallback for the unrestricted default.

## Fix

When `enabledModels` is empty, fall back to pi's own model catalog at `~/.pi/agent/models-store.json`. That file is maintained by pi itself (hydrated from the published `@earendil-works/pi-ai` package and refreshed as providers are added), so the list stays accurate without extra configuration:

- `Name` = `provider/id` (what pi accepts via `--model`)
- `Alias` = short model id (for the `/model` button/select labels)
- `Desc` = display name from the catalog
- results sorted by name for deterministic card layout

If `enabledModels` is set, it still takes priority (restriction semantics preserved).

## Validation

- `go build -tags no_web ./...` passes
- `go test -tags no_web ./agent/pi/` passes
- Note: `go test ./core/` has one pre-existing failure (`TestRunShellWithProgress_NonexistentCommand`) unrelated to this change — it expects an English shell error string but the test machine's `zh_CN` locale produces a localized message.

## Related

- #1178 — current `enabledModels`-only implementation this fixes
- #890 — alternative approach using `pi --list-models` (still open)

### DIFF
--- a/agent/pi/pi.go
+++ b/agent/pi/pi.go
@@ -350,6 +350,10 @@ type modelsJSON struct {
 // (e.g. "deepseek/deepseek-v4-pro") and the fully-qualified
 // provider/ID (e.g. "my-provider/my-model").
 // Returns nil on any error (caller falls back to 200K).
+//
+// Note: models.json is distinct from models-store.json — models.json carries
+// per-model context-window sizes, while models-store.json is the provider
+// catalog that readModelsStore uses to build the /model list.
 func loadModelsContextWindows() map[string]int {
 	dir := piSettingsDir()
 	if dir == "" {
--- a/agent/pi/pi.go
+++ b/agent/pi/pi.go
@@ -122,9 +122,16 @@ func (a *Agent) AvailableModels(_ context.Context) []core.ModelOption {
 	models, err := readSettingsModels()
 	if err != nil {
 		slog.Debug("pi: AvailableModels: read settings", "error", err)
-		return nil
 	}
-	return models
+	if len(models) > 0 {
+		return models
+	}
+	// enabledModels 未配置时，回退到 pi 自身的模型目录 models-store.json，
+	// 否则 /model 卡片只会显示当前模型、无法列出可切换的模型列表。
+	if store := readModelsStore(); len(store) > 0 {
+		return store
+	}
+	return nil
 }
 
 func (a *Agent) SetSessionEnv(env []string) {
@@ -441,6 +448,59 @@ func readSettingsModels() ([]core.ModelOption, error) {
 		models = append(models, option)
 	}
 	return models, nil
+}
+
+// modelsStoreJSON represents the structure of ~/.pi/agent/models-store.json,
+// the provider catalog pi itself maintains (hydrated from the published
+// pi-ai package and refreshed as providers are added). Top-level keys are
+// provider names, each with a Models list.
+//
+//	{
+//	  "deepseek": { "models": [ { "id": "...", "name": "...", ... } ] }
+//	}
+type modelsStoreJSON map[string]struct {
+	Models []struct {
+		ID   string `json:"id"`
+		Name string `json:"name"`
+	} `json:"models"`
+}
+
+// readModelsStore reads all models from pi's models-store.json as
+// provider-qualified ModelOptions (Name = "provider/id", Alias = short id,
+// Desc = display name). Returns nil when the file is missing or unreadable;
+// callers fall back to an empty list.
+func readModelsStore() []core.ModelOption {
+	dir := piSettingsDir()
+	if dir == "" {
+		return nil
+	}
+	path := filepath.Join(dir, "models-store.json")
+	data, err := os.ReadFile(path)
+	if err != nil {
+		slog.Debug("pi: read models-store", "path", path, "error", err)
+		return nil
+	}
+	var store modelsStoreJSON
+	if err := json.Unmarshal(data, &store); err != nil {
+		slog.Warn("pi: parse models-store", "path", path, "error", err)
+		return nil
+	}
+	var models []core.ModelOption
+	for provider, p := range store {
+		for _, m := range p.Models {
+			if m.ID == "" {
+				continue
+			}
+			models = append(models, core.ModelOption{
+				Name:  provider + "/" + m.ID,
+				Alias: m.ID,
+				Desc:  m.Name,
+			})
+		}
+	}
+	// Map iteration order is random — sort for deterministic card display.
+	sort.Slice(models, func(i, j int) bool { return models[i].Name < models[j].Name })
+	return models
 }
 
 // readDefaultModel returns the defaultModel from settings.json.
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
Conclusion: Approve with P3 suggestion

Overall assessment: This PR correctly fills the gap left by #1178 — when a user has not configured `enabledModels` in `settings.json`, the pi agent's `/model` card now falls back to pi's own `~/.pi/agent/models-store.json` catalog. Implementation is clean, deterministic (sorted), defensive (skips empty IDs, swallows malformed JSON), and the priority semantics (user `enabledModels` wins over catalog) are preserved. CI all green. Mergeable.

Review scope:
- Read `agent/pi/pi.go` `AvailableModels` (lines 122-137) and the new `readModelsStore` + `modelsStoreJSON` (lines 453-505).
- Ran the full PR-head test suite in a worktree: `go test ./agent/pi/ -count=1` → all green.
- Wrote a manual test `TestReadModelsStore_ManualVerification` (deleted after verification) to exercise the new function: it correctly skips empty-id models, sorts by `Name` (deepseek-chat < deepseek-reasoner < openai/gpt-4), returns nil on missing/malformed JSON, and maps provider/id → Name + id → Alias + display name → Desc.
- Confirmed `sort` was already imported (line 12) — no new import needed.
- Cross-checked `core.ModelOption` shape (Name/Desc/Alias) — the field mapping is correct.

✅ What looks good:
- **Correct fix**: The PR title's claim matches reality — when `enabledModels` is empty, fall back to the catalog. Existing `if len(models) > 0 { return models }` guard ensures the *first* source (user config) always wins, preserving the restriction semantics from #1178.
- **Defensive parsing**: Skips models with empty `ID` (the only "required" field) and the empty-provider `"":` case. Returns `nil` on missing file, malformed JSON, or unparseable provider. `slog.Debug` for missing, `slog.Warn` for malformed — appropriate log level per case.
- **Deterministic output**: `sort.Slice` by Name. Without it, Go's map iteration would render the card in random order on each call, breaking UI stability and producing confusing git diffs if anyone logs the list.
- **Field mapping is right**: `Name = "provider/id"` (the form pi accepts via `--model`), `Alias = id` (for `/model` button/select labels — matches what `readSettingsModels` does), `Desc = display name`. Consistent with the existing `readSettingsModels` contract.
- **No new imports**: `sort` and `path/filepath`/`os`/`encoding/json`/`log/slog` all already imported.
- **CI all green**: lint / unit-test / smoke / regression / performance all SUCCESS.

🟠 Should improve (P2):
- **Add a test for `readModelsStore`** (mirror of `TestReadSettingsModels` at lines 219-282). The PR has zero new tests despite adding a new exported-by-name function. The existing `TestAgent_AvailableModels` is too loose (early-returns on nil and only checks `Name != ""` if non-nil) to actually exercise the fallback path. Without a dedicated test, a future refactor that breaks the empty-ID skip or the sort order would not be caught. The author already has the test infrastructure (TempDir + t.Setenv PI_CODING_AGENT_DIR) — adding 30 lines is straightforward. Suggested cases:
  1. Missing `models-store.json` → returns nil
  2. Valid file with 2 providers × 2 models → returns 4 models, sorted, fields correct
  3. File with `id=""` model → that one is skipped
  4. Malformed JSON → returns nil, warning logged
  5. `TestAgent_AvailableModels_FallsBackToStore` (high-level) — empty `enabledModels` in settings.json, populated `models-store.json` → AvailableModels returns the catalog

  I ran a throwaway `TestReadModelsStore_ManualVerification` and confirmed all five behaviors work; the test is just not checked in.

🔵 Optional (P3):
- **Test interaction with existing `loadModelsContextWindows`**: this function reads a *different* file (`models.json`, not `models-store.json`). The naming is confusing. If the upstream pi-ai package also generates a `models.json` for context windows, consider documenting the difference in a comment on `loadModelsContextWindows` so future readers know they're two distinct files. Not a blocker.
- **Consider de-duplication**: if a user has both `settings.json:enabledModels` AND `models-store.json`, currently `enabledModels` wins entirely (correct). But if a future use case wants "restrict to enabledModels OR show catalog", the logic in `AvailableModels` would need an `or` instead of `or else`. Today's choice is right, just a thought.
- **Add a brief mention in `config.example.toml` or `docs/agents/pi.md`** about the new fallback behavior so users discovering `/model` for the first time understand where the list comes from.

❓ Questions:
- **Why not use the alternative approach in #890** (`pi --list-models`)? That PR is still open. The author of this PR chose to read the file directly instead of shelling out. The choice is reasonable (avoids subprocess overhead and parsing text output), but worth a one-line justification in the PR description for future archeology.
- **What if `models-store.json` schema changes?** The struct field is `id` and `name` today. If pi-ai adds a new required field or renames, this will silently skip models. Consider a `slog.Warn` when a provider entry has 0 models (might indicate schema mismatch), or version the schema (`{"version": 1, "providers": {...}}`).

Testing / Risk:
- **Verified locally on PR head `0e2e605d`**: `go test ./agent/pi/ -count=1` all green, no regressions in existing `TestReadSettingsModels`, `TestLoadModelsContextWindows`, etc.
- **CI on PR**: all 5 checks SUCCESS. The pre-existing `TestRunShellWithProgress_NonexistentCommand` failure on `./core/` is unrelated (zh_CN locale issue, same as #1637 and the author's note).
- **Risk surface**: 60 new lines in one file, one new exported-ish helper, fallback-only behavior. When `enabledModels` is set (existing default for users who follow the recommended setup), no behavior change. When unset, list is populated from the catalog — strictly additive UX improvement.
- **Backward compatibility**: full. No signature changes, no behavior change for users with `enabledModels` set, no new env vars, no new public API.
- **Edge cases covered by code review**: empty ID, missing file, malformed JSON, empty provider, sort stability. All handled.

Next step:
- Author adds `TestReadModelsStore` (P2, ~30 lines, follows the `TestReadSettingsModels` pattern). I'd merge without it, but the test gap is real.
- After that, ready to merge. This is a strict UX improvement — no migration needed, no release note required (the `/model` card just gets useful for users who haven't configured `enabledModels`).

baizhu945:
Added `TestReadModelsStore` (mirroring `TestReadSettingsModels`) covering missing file, sorted provider-qualified output, empty-id skip, and malformed JSON, plus `TestAgent_AvailableModels_FallsBackToStore` for the high-level fallback path. Also added a note on `loadModelsContextWindows` clarifying that `models.json` (context windows) is distinct from `models-store.json` (the `/model` catalog). `go test ./agent/pi/` passes.

PR #1639, opened 2026-08-04
fix(codex) when using /list command, correctly read codex project session names

shaojiahao:
## What changed

- Use the first `session_meta` entry as the rollout identity so copied parent metadata cannot turn subagent rollouts into duplicate top-level sessions.
- Exclude Codex subagent rollouts from `/list` results.
- Read generated session names from `session_index.jsonl`, with the existing user-message summary as a fallback.
- Add regression coverage for subagent filtering, copied parent metadata, and Codex session names.

## Why

`/list` could show many internal subagent rollouts and display concrete user messages instead of the project session names shown by Codex. This made session counts and names differ between cc-connect and Codex for the same project.

## Root cause

Subagent rollout files can contain copied parent history with additional `session_meta` entries. The parser kept overwriting the rollout identity, so some subagent files were misclassified as duplicate parent sessions. The list implementation also derived summaries only from transcript messages and did not consult Codex's session-name index.

## User impact

Telegram `/list`, `/switch`, and other session-list consumers now see the same top-level Codex sessions and generated session names as the Codex project session picker.

## Validation

- Verified against local Codex data containing 118 rollout files: the result was reduced to the expected 5 top-level sessions, with names matching Codex.
- A full `go test ./...` was attempted; unrelated existing Windows path/permission assertions, a missing `web/dist`, and the `agent/pi` Windows build currently prevent a clean repository-wide run.

### DIFF
--- a/agent/codex/list.go
+++ b/agent/codex/list.go
@@ -55,10 +55,14 @@ func listCodexSessions(workDir, codexHome string) ([]core.AgentSessionInfo, erro
 		return nil, nil
 	}
 
+	sessionTitles := loadCodexSessionTitles(codexHome)
 	var sessions []core.AgentSessionInfo
 	for _, f := range files {
 		info := parseCodexSessionFile(f, absWorkDir)
 		if info != nil {
+			if title := sessionTitles[info.ID]; title != "" {
+				info.Summary = title
+			}
 			patchSessionSource(info.ID, codexHome)
 			sessions = append(sessions, *info)
 		}
@@ -69,6 +73,34 @@ func listCodexSessions(workDir, codexHome string) ([]core.AgentSessionInfo, erro
 	})
 
 	return sessions, nil
+}
+
+// loadCodexSessionTitles reads the same generated thread names that Codex uses
+// in its session picker. Later entries win because renames append a new record.
+func loadCodexSessionTitles(codexHome string) map[string]string {
+	path := filepath.Join(resolveCodexHomeDir(codexHome), "session_index.jsonl")
+	f, err := os.Open(path)
+	if err != nil {
+		return nil
+	}
+	defer f.Close()
+
+	titles := make(map[string]string)
+	scanner := bufio.NewScanner(f)
+	scanner.Buffer(make([]byte, 64*1024), 256*1024)
+	for scanner.Scan() {
+		var entry struct {
+			ID         string `json:"id"`
+			ThreadName string `json:"thread_name"`
+		}
+		if json.Unmarshal(scanner.Bytes(), &entry) != nil {
+			continue
+		}
+		if entry.ID != "" && strings.TrimSpace(entry.ThreadName) != "" {
+			titles[entry.ID] = entry.ThreadName
+		}
+	}
+	return titles
 }
 
 // parseCodexSessionFile reads a Codex JSONL transcript.
@@ -87,6 +119,7 @@ func parseCodexSessionFile(path, filterCwd string) *core.AgentSessionInfo {
 
 	var sessionID string
 	var sessionCwd string
+	var sessionSource json.RawMessage
 	var summary string
 	var msgCount int
 	userMsgSeen := 0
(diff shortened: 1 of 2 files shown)


chenhg5 (commented):
## Conclusion: Comment (DRAFT, advisory — no request-changes)

Overall assessment:
- Core fix is correct, test coverage is solid, diff is contained in `agent/codex`, and the PR description is exemplary. `/list` now properly reads `session_index.jsonl`'s `thread_name` and filters out subagent rollouts. 3rd-party author, recommend keeping `--comment` until DRAFT is flipped to ready-for-review.

Review scope:
- Read `agent/codex/list.go` and the new `agent/codex/list_test.go`.
- Ran `go test -race ./agent/codex/... -count=1` (ok 2.575s); new 3 tests pass standalone with `-race` too.
- Focused on correctness of subagent filtering, error paths in `session_index.jsonl` parsing, blast radius.

✅ What looks good:
- Root-cause analysis is accurate: subagent rollout files may contain a copied parent meta, the old parser was always overwriting `sessionID`/`sessionCwd`/`sessionSource`, which misclassified subagents as their parent. The new `if sessionID != "" { continue }` locks the first meta — a tight, well-targeted minimal fix.
- Using Codex's own `session_index.jsonl` as the thread-name source (instead of inventing a parallel naming scheme) avoids drift with the Codex picker.
- Subagent filtering aligns with the existing pattern in `agent/gemini/gemini.go:540` and `agent/antigravity/antigravity.go:540`, bringing codex to parity.
- Tests use `Agent.ListSessions` as a black box and `t.TempDir()`, locking in three independent invariants (subagent filtered, copied parent meta not adopted, thread_name overrides user-message summary).
- errcheck follow-up is strictly scoped to the new `defer` calls; pre-existing `defer f.Close()` at list.go:118/255 and context_usage.go:193 are left alone.
- PR description is gold standard: what/why/root-cause/user-impact/validation, with concrete 118-rollout → 5 top-level validation data.

🟠 Should improve (non-blocking; author can decide whether to bundle before ready-for-review):
- The thread_name override skips the 60-rune `Summary` truncation in `parseCodexSessionFile` (list.go:64–66 vs 204–206). Codex currently emits short thread_names so no real-world impact, but if Codex ever generates longer names, IM cards will receive uncapped content. Recommend adding `if len([]rune(title)) > 60 { title = string([]rune(title)[:60]) + "..." }` before the assignment, plus a `TestAgentListSessions_LongThreadNameTruncated` test to lock the invariant.
- `isSubagentSessionSource` returns true for `{"subagent": null}` (list.go:218–225). Codex doesn't actually emit `null` so impact is zero, but for strict shape matching one could read the value and require it to be non-null. Lower priority than the cap above.

🔵 Optional:
- Test name `TestAgentListSessions_ExcludesSubagentRolloutWithCopiedParentMeta` is verbose; `TestAgentListSessions_FirstSessionMetaWinsOverCopiedParent` would directly telegraph the invariant.
- `loadCodexSessionTitles` silently returns nil on `os.Open` failure; a `slog.Warn` would help diagnose permissions/IO issues.
- The defensive `if sessionID != "" { continue }` deserves a one-line comment explaining the first-meta-wins rationale (the PR description has it; the source doesn't).

❓ Questions: None.

Testing / Risk:
- Verified: PR ships with 3 new unit tests + author's local 118-rollout end-to-end validation; local `go test -race ./agent/codex/... -count=1` is all ok; `go build` / `go vet` clean.
- Uncovered: corrupt `session_index.jsonl` line (per code it skips, but no test locks that); latest-wins behavior with multiple index entries for same ID after rename; rollout exists but no index entry fallback. All reasonable as follow-ups.

DRAFT note: PR is currently in DRAFT. Per QA policy, even with `--approve` allowed for 3rd-party, I don't approve while in DRAFT — leaving a comment for the author to convert to ready. If you want to be conservative, you can land the P2 #1 (60-rune title cap + corresponding test) in the same PR; otherwise the current diff is good to merge as-is.

Next step: Author flips the PR from DRAFT to ready-for-review. If desired, include the P2 #1 (60-rune title cap + test); otherwise the diff is sufficient — QA has no hard blocker.

Issue #1640, opened 2026-08-04
[weixin/ilink] context_token 过期后主动推送（cron/timer）持续失败 ret=-2，用户发新消息并不会真正刷新 token（v1.4.1）


lyjx29:
## Environment

- cc-connect v1.4.1
- platform: weixin (ilink bot API, personal WeChat account)
- agent: Claude Code (claudecode)
- OS: Linux (WSL2), running as a systemd user service

## Symptom

I configured a cron job for scheduled proactive pushes (e.g., push a daily digest at 18:00). After a long idle period (several hours) with no user messages, the context_token expires. After that:

1. Every `sendMessage` triggered by cron/timer returns `ret=-2 (expired context_token)`, and pushes keep failing.
2. The log says: `user must send a new message to peer %q to refresh the session token`.
3. **However, testing shows that sending a new message does NOT refresh the token.** The user sent one message at 16:14 and another at 16:15; both were received and the agent produced replies normally, but the agent's reply sends at 16:15 and 16:17 still all failed with `ret=-2`.
4. **The only reliable recovery is restarting the cc-connect daemon** (after re-binding getUpdates it obtains a fresh context_token, and `~/.cc-connect/weixin/weixin/<bot_id>/context_tokens.json` is updated); pushes resume immediately.

## Steps to Reproduce

1. Configure the weixin platform plus one cron job (periodic proactive push).
2. Leave the bot idle with no interaction for several hours.
3. Cron fires → the log shows `sendMessage ret=-2 (expired context_token)`, the push fails.
4. Send a new message to the bot → the message is received and the agent replies normally, but the reply still fails to send with `ret=-2`.
5. Subsequent cron pushes keep failing.
6. Restart the daemon → pushes recover.

## Expected Behavior (any of the following)

- When a new message is received, the fresh context_token should be **actually refreshed and cached** (as described in docs/weixin.md), so subsequent sends succeed immediately; or
- On `ret=-2` (a session-expired signal, same class as `errcode -14`), **automatically re-bind / re-login** to obtain a fresh token instead of repeatedly reusing the bad one; or
- At minimum, provide an **explicit token-refresh command or a heartbeat/keepalive mechanism** (similar to the heartbeat feature), so proactive pushes don't entirely depend on "the user happening to send a message".

## Additional Info

- context_token cache file: `~/.cc-connect/weixin/weixin/<bot_id>/context_tokens.json`, only updated after a daemon restart.
[...]

PR #1643, opened 2026-08-05
fix(weixin): treat sendmessage ret=-2 as a per-account send budget and fail fast

lyjx29:
## Summary

Fixes weixin (ilink) `sendMessage ret=-2 "prepare failed"` failures that blocked proactive
pushes (cron/timer). Ref issue #1640.

**Root cause (empirically verified on a live bot, differs from the original diagnosis):**
the ilink gateway does **not** validate `context_token` on sends — fake and tokenless sends
both return HTTP 200 and are delivered. So this is **not a token-expiry bug**. `ret=-2
"prepare failed"` is a **per-account send budget / burst throttle**:
- Roughly **5-6 separate messages per 24h window** trip it (sends spread over hours count,
  so rate-based pacing alone does not help). Multi-chunk messages do **not** count (a
  chunked reply is one logical message).
- **Every send attempt made while the penalty is active escalates it** — the original
  incident recovered in ~1h with no attempts; sending probes kept a bot throttled for 15h+.
  So the old `3×500ms` retry loop plus the "incomplete delivery notice" send actively
  prolonged outages.

## Changes

1. **`platform/weixin` — fail fast on throttle** (`sendChunk`, `sendSingleItem`): on
   `ret=-2`, return a clear throttled error after a single sendmessage call. No retries
   (every attempt escalates the penalty). The "incomplete delivery notice" send is skipped
   when throttled.
2. **`platform/weixin` — 24h message budget** (`burst_limit` / `burst_window_secs`, default
   4 per 86400s): a sliding-window counter that **fails fast** (no waiting) once the budget
   is exhausted, so the bot never approaches the throttle. The `checkSendQuota` is applied
   to both the text (`sendChunks`) and media (`sendSingleItem`) paths, keeping the push and
   reply paths consistent.
3. **Fix misleading logs/errors** — "user must send a new message to refresh the session
   token" described a mechanism that doesn't exist; replaced with accurate throttle-aware
   messages.
4. **`config.example.toml`** — document the weixin burst quota options and the
   `[outgoing_rate_limit.platforms.weixin]` rate-limit example.

## Acceptance criteria (from the maintainer's plan) — how this maps

1. *ret=-2 self-heal by retrying once* — **reinterpreted**: retrying during the penalty
   escalates it (shown live), so the correct behavior is fail-fast + wait for the ~1h base
   penalty / next retry, plus QR re-login as recovery. The budget prevents triggering.
2. *context_tokens.json refresh* — **not applicable**: the gateway does not validate the
   token (proven), so refreshing it cannot change behavior. The file still updates on
   inbound messages (unchanged, harmless).
3. *Fix misleading log* — done (accurate throttle messaging).
4. *cron and reply paths consistent* — done: both go through `sendChunks`/`sendSingleItem`
   and share fail-fast + the budget.
5. *No regression* — ret=0 path untouched; `errcode -14` (pauseSession) untouched; only the
   `ret=-2` path changed.

## Tests

- New unit tests: `TestIsSendThrottled`, `TestSendChunk_FailsFastOnThrottle` (asserts a
  single sendmessage call), `TestSendChunk_Succeeds`, `TestCheckSendQuota_AllowsUnderLimit`,
  `TestCheckSendQuota_FailsWhenOverLimit`, `TestCheckSendQuota_DisabledWhenZero`,
  `TestSendChunks_AppliesQuota`.
- `go build -tags no_web ./...`, `go vet`, and `go test -tags no_web ./platform/weixin/
  ./core/` pass. The full suite's only failures are pre-existing and unrelated
  (`platform/cloud-web` needs a reachable gateway; `platform/feishu` has one flaky test),
  reproducible on `main`.

## Live verification (2026-08-05)

- Confirmed the throttle is a 24h per-account budget (a bot throttled at ~5-6 messages sent
  over several hours, not a burst).
- After deploying fail-fast + the 24h budget, the daily-report push succeeded end-to-end.
- Recovery when a bot does get throttled: QR re-login (`cc-connect weixin new`) yields a
  fresh account with a fresh budget — the only reliable immediate recovery.

Refs #1640

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1448,7 +1448,7 @@ app_secret = "your-feishu-app-secret"
 # （多切块发送不计数），且节流期内每次尝试都会升级惩罚。配额对独立消息限速，
 # 使其低于触发阈值。设为 0 关闭配额。
 # burst_limit = 4           # max separate messages per window / 每窗口最多独立消息数
-# burst_window_secs = 120   # window length / 窗口时长（秒）
+# burst_window_secs = 86400 # window length (default 24h) / 窗口时长（秒，默认24小时）
 
 # -----------------------------------------------------------------------------
 # Tencent Yuanbao (yuanbao.tencent.com bot platform)
--- a/config.example.toml
+++ b/config.example.toml
@@ -1438,6 +1438,17 @@ app_secret = "your-feishu-app-secret"
 # proxy = ""                                   # optional HTTP proxy / 可选代理
 # proxy_username = ""
 # proxy_password = ""
+#
+# Send-volume quota vs ilink's burst throttle (sendMessage ret=-2 "prepare failed"):
+# the gateway throttles the bot after roughly 5-6 separate messages within a short
+# window (multi-chunk sends are fine), and the penalty escalates with every attempt
+# made while active. The quota paces separate messages so the bot stays below the
+# trigger. 0 disables the quota.
+# 发送量配额（防 ilink 突发节流 ret=-2）：网关在短窗口内约 5-6 条独立消息即节流
+# （多切块发送不计数），且节流期内每次尝试都会升级惩罚。配额对独立消息限速，
+# 使其低于触发阈值。设为 0 关闭配额。
+# burst_limit = 4           # max separate messages per window / 每窗口最多独立消息数
+# burst_window_secs = 120   # window length / 窗口时长（秒）
 
 # -----------------------------------------------------------------------------
 # Tencent Yuanbao (yuanbao.tencent.com bot platform)
--- a/config.example.toml
+++ b/config.example.toml
@@ -291,6 +291,11 @@ level = "info" # debug, info, warn, error
 # max_per_second = 25
 # [outgoing_rate_limit.platforms.feishu]
 # max_per_second = 5
+# Weixin (ilink) sendmessage throttles the bot after a short burst (ret=-2
+# "prepare failed", penalty on the order of an hour). Pace sends to avoid it:
+# [outgoing_rate_limit.platforms.weixin]
+# max_per_second = 0.5
+# burst = 1
 
 # =============================================================================
 # Relay Settings / Relay 设置
--- a/core/engine.go
+++ b/core/engine.go
@@ -11678,7 +11678,7 @@ func (e *Engine) sendAlreadyRenderedWithError(p Platform, replyCtx any, content 
 				"platform", p.Name(),
 				"error", err,
 				"content_len", len(content),
-				"hint", "user needs to send a new message to refresh context_token")
+				"hint", "user needs to send a message to the bot first so a context_token can be captured")
 		} else {
 			slog.Error("platform send failed", "platform", p.Name(), "error", err, "content_len", len(content))
 		}
--- a/platform/weixin/media_outbound.go
+++ b/platform/weixin/media_outbound.go
@@ -131,7 +131,9 @@ func buildVideoMessageItem(ref *cdnUploadedRef) messageItem {
 	}
 }
 
-// sendSingleItemWithRetry sends a media item with retry mechanism for ret=-2 errors.
+// sendSingleItemWithRetry sends a media item with a retry mechanism.
+// When sendMessage returns ret=-2 (ilink throttling the bot), it backs off and
+// retries rather than hammering the throttled endpoint with 500ms-interval retries.
 func (p *Platform) sendSingleItemWithRetry(ctx context.Context, rc *replyContext, item messageItem) error {
 	var lastErr error
 	for attempt := 0; attempt < weixinSendMaxRetries; attempt++ {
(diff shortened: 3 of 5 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment: This PR correctly reframes the weixin `sendMessage ret=-2` problem based on empirical live testing — it's a per-account 24h send budget, not a context_token expiry — and applies two complementary fixes: (1) fail-fast on throttle (no retries, since each attempt during the penalty escalates it from ~1h to 15h+), and (2) a sliding-window budget counter that prevents approaching the trigger in the first place. All 3 outbound paths (text via `sendChunks`, media via `sendSingleItem`) are gated. 7 new tests cover the new behavior, all green. CI all green. Live-verified by author. This is a high-quality fix that I trust to merge.

Review scope:
- Read `platform/weixin/weixin.go` (constants, Platform struct, New, checkSendQuota, sendChunks, sendChunk, isSendThrottled).
- Read `platform/weixin/media_outbound.go` (sendSingleItem rewrite, removed sendSingleItemWithRetry).
- Read `config.example.toml` and `core/engine.go` (one misleading-log fix).
- Read `platform/weixin/weixin_test.go` (7 new tests).
- Ran PR-head test suite in worktree at `35381722`: `go test ./platform/weixin/ -v` PASS 18/18 (7 new + 11 existing); `go test ./platform/weixin/ -race -count=3` PASS 9.825s; `go test ./core/ -count=1` PASS 46.262s (no regression).
- Cross-checked issue #1640 — author correctly identified that the original "context_token expiry" diagnosis was wrong (the user reports in #1640 explicitly state "sending a new message does NOT refresh the token").

✅ What looks good:
- **Reframes the bug correctly**: PR body has empirical evidence (1h recovery without attempts vs 15h+ with attempts) showing the old `3×500ms` retry loop actively made things worse. The author verified the gateway does *not* validate context_token (fake/tokenless sends return 200 and deliver). This is good first-principles debugging.
- **Fail-fast is the right choice**: `isSendThrottled` + single-send + return is correct because (a) retries escalate the penalty, (b) the user-facing error message is more accurate, (c) the failed `incomplete delivery notice` send is correctly skipped when throttled (`if !isSendThrottled(err)`).
- **24h sliding-window budget is well-designed**: `sendQuotaTimes` slice with in-place truncation, protected by `sendQuotaMu`, defaults to 4 sends per 24h (well under the 5-6 trigger). Tests prove the `kept := p.sendQuotaTimes[:0]` pattern works correctly (no race, no leak, no off-by-one). `limit=0` cleanly disables the quota for power users.
- **All outbound paths gated**: `sendChunks` (text), `sendSingleItem` (image/file/video/audio), all 3 `SendImage`/`SendFile`/`SendAudio` paths route through `sendSingleItem`. I traced every `p.api.sendText` / `p.api.sendMessage` call site — the only un-counted send is the "incomplete delivery notice", which is intentional (it's a fail-soft path).
- **Test coverage is excellent**: 7 new tests cover the failure mode (`TestSendChunk_FailsFastOnThrottle` asserts exactly 1 sendmessage call), the success path, the 3 quota branches (under/over/disabled), and end-to-end (`TestSendChunks_AppliesQuota`). Race detector clean.
- **Misleading log fixed**: "user must send a new message to refresh the session token" → "user needs to send a message to the bot first so a context_token can be captured". The old message was a lie (per #1640, sending a message doesn't refresh the token). New message is honest.
- **Config example updated**: documents both `[outgoing_rate_limit.platforms.weixin]` (token-bucket rate limit) and the new `burst_limit` / `burst_window_secs` (24h budget). The Chinese comments are a nice touch for the bilingual audience.
- **Backward compatible**: defaults are conservative (4/24h), existing config files work as-is, `burst_limit=0` cleanly disables for users who don't want the quota. No migration.
- **Issue #1640 marked as root-cause-fixed**: the original failure mode (cron/timer proactive push failure) is no longer a problem because the bot now never reaches the 5-6 trigger.

🟠 Should improve (P2):
- **Operator visibility on quota hits**: when the quota is exhausted, the failure is silent from the operator's perspective — only the engine sees "weixin: send budget exhausted" in the log. Cron/timer pushes will simply fail, and the user (or owner) won't know why. Consider:
  - One-line: add a `slog.Warn` with `account_label` + quota state when checkSendQuota returns error, so it's grep-able.
  - Or: emit a metric (count of quota hits per day per account) so operators can see "the weixin bot hit quota 3 times today" in dashboards.
  - Or: send a one-time notice to the bot owner (via a different channel like email) when the quota is first hit each day.
  
  The current behavior is correct (don't make a failed send to a throttled bot), but operators need to know it's happening. This was visible before because the bot would log ret=-2; now it logs quota-exhausted, which is fine — but the visibility into the per-day count is lost.

- **Misleading test name on `TestCheckSendQuota_DisabledWhenZero`**: it asserts that 10 calls pass when limit=0. But the test name says "DisabledWhenZero" which could be misread as "tests the disable path". The test is correct, but consider renaming to `TestCheckSendQuota_ZeroLimitAllowsAll` for clarity. (P3 nit.)

🔵 Optional (P3):
- **Unused `ctx` parameter in `checkSendQuota`**: the function takes `ctx context.Context` but never uses it. Either remove it (cleaner) or add `select { case <-ctx.Done(): return ctx.Err(); default: }` at the top for cancellation support. Minor.
- **Two rate-control mechanisms with similar names**: `[outgoing_rate_limit.platforms.weixin] max_per_second=0.5 burst=1` (token bucket, controls instantaneous send rate) vs the new `burst_limit=4 burst_window_secs=86400` (24h window counter, controls daily volume). Both are valid but the naming is confusing for config authors. Consider renaming one — e.g. `daily_send_limit` / `daily_send_window_secs` — to make the difference obvious. Pure naming, no behavior change.
- **Default of 4/24h**: conservative, matches the author's empirical "5-6 triggers" observation. But a 4 vs 5-6 is only 1-2 messages of safety margin. If a user has bursty cron jobs (e.g., 3 daily reports + 1 ad-hoc + 1 cron alert = 5 in 24h), they'd be throttled. The 0-disables-quota escape hatch covers this, but the config example could note "lower to 3 if you have many cron pushes".
- **No test for quota-with-zero-window edge case**: `sendQuotaWindow=0` is currently treated as disabled (`if p.sendQuotaLimit <= 0 || p.sendQuotaWindow <= 0 { return nil }`), but the New() code sets `burstWindow=defaultBurstWindowSecs` when `<=0` is passed. The defensive double-check is fine, but the test `TestCheckSendQuota_DisabledWhenZero` only covers limit=0, not window=0. One extra line: `p := &Platform{sendQuotaLimit: 4, sendQuotaWindow: 0}; for i := 0; i < 10; i++ { ... }` would cover it.
- **`errcode -14` (pauseSession) is correctly untouched**: I verified the poll loop still calls `pauseSession(time.Hour)` for session-expired responses. The send path is now separate from the poll path, which is correct.

❓ Questions:
- **What's the recovery path when quota is hit during cron?** PR body mentions "QR re-login yields a fresh account with a fresh budget" as the only reliable immediate recovery. Is there a way to surface this to the cron owner? E.g., a "send failed — quota exhausted, re-login needed" error message that the cron job could capture.
- **Why 4/24h and not 3/24h?** 1-2 messages of margin feels thin. Author's empirical data showed "5-6 trips it" — picking 3 gives 50%+ headroom.

Testing / Risk:
- **Verified locally on PR head `35381722`**: all 18 weixin tests PASS; race detector -count=3 PASS 9.8s; full `./core/` PASS 46s.
- **CI on PR**: 5/5 SUCCESS (lint 1m42s, unit-test 4m21s, smoke 31s, regression 27s, performance 52s).
- **Risk surface**: 304 additions / 133 deletions across 5 files. Most of the change is *removal* of the broken retry logic. New code is well-isolated to `platform/weixin/`.
- **Backward compatibility**: defaults are conservative; no config migration; `burst_limit=0` cleanly disables; `ret=0` path unchanged; `errcode -14` path unchanged.
- **Production risk**: low. The new quota *fails safe* — it prevents reaching the throttle, so it's strictly better than current behavior. The only new failure mode is "quota exhausted" error in cron, which is informative.

Next step:
- Author adds the operator visibility (P2) — at minimum a `slog.Warn` with structured fields.
- After that, ready to merge. The fix is live-verified and the unit tests lock the new behavior in.
- Consider a release note: "weixin (ilink) bots: `sendMessage ret=-2` failures now correctly identified as a 24h per-account send budget, not a context_token issue. Bots are proactively throttled to 4 sends/24h by default; configure `burst_limit`/`burst_window_secs` to adjust. Disable with `burst_limit = 0`."

PR #1645, opened 2026-08-06
feat(wecom): add interactive card sending and callback feedback

deepziyu:
## Summary

This PR adds interactive card sending (`core.CardSender`) and callback feedback (`template_card_event`) support to the WeChat Work (WeCom / 企业微信) platform adapter in `cc-connect`, covering both **HTTP Webhook** mode (`wecom.go`) and **WebSocket** mode (`websocket.go`).

## Problem & Motivation

Previously, the WeCom platform adapter in `cc-connect` only supported plain text, markdown, image, voice, and file messages. Incoming `template_card_event` callbacks (button interactions) were either logged as unsupported or ignored (`aibot_event_callback`). As a result, interactive features like permission approval prompts (`perm:allow` / `perm:deny`), action buttons, and structured cards could not be rendered or interacted with on Enterprise WeChat.

### 1. Template Card Renderer (`platform/wecom/wecom_card.go`)
- Added `buildWeComTemplateCard(card *core.Card)` to convert `*core.Card` into WeCom's official `template_card` format (`button_interaction` when buttons are present, and `text_notice` when text/title only).
- Maps button labels, types (`primary` -> 1, `danger` -> 3, `default` -> 2), and callback keys (`btn.Value`).
- Added `parseWeComPermissionResponse` helper to handle permission decision keys (`perm:allow`, `perm:deny`, `perm:allow_all`, `perm:deny_all`) and set `IsPermissionResponse: true` on `core.Message`.

### 2. HTTP Webhook Mode (`platform/wecom/wecom.go`)
- Implemented `core.CardSender` (`SendCard`, `ReplyCard`) for `Platform` by sending `msgtype: "template_card"` to `/cgi-bin/message/send`.
- Updated XML callback handler to parse `MsgType == "event"` and `Event == "template_card_event"`, extracting `EventKey` and `TaskId` to dispatch user actions.

### 3. WebSocket Mode (`platform/wecom/websocket.go`)
- Implemented `core.CardSender` (`SendCard`, `ReplyCard`) for `WSPlatform` via `aibot_send_msg` and `aibot_respond_msg`.
- Updated `handleFrame` for `cmd: "aibot_event_callback"` to handle `template_card_event` and dispatch the user button interaction.

### 4. Unit Tests (`platform/wecom/wecom_card_test.go`)
- Added unit tests covering:
  - `TestBuildWeComTemplateCard`: `button_interaction` mapping.
  - `TestBuildWeComTemplateCard_TextNotice`: `text_notice` fallback mapping.
  - `TestParseWeComPermissionResponse`: Permission response detection.
  - `TestWSEventCallback_TemplateCardEvent`: WebSocket frame parsing and `IsPermissionResponse` flag handling.

## Verification

Ran all unit tests for the `wecom` package:
```bash
go test -v ./platform/wecom/...
```
All 48 tests in `platform/wecom` passed cleanly.

## Wecom ref

https://developer.work.weixin.qq.com/document/path/101032

https://developer.work.weixin.qq.com/document/path/101027

### DIFF
--- a/platform/wecom/websocket.go
+++ b/platform/wecom/websocket.go
@@ -287,7 +287,7 @@ func (p *WSPlatform) handleFrame(frame wsFrame) {
 	case "aibot_msg_callback":
 		p.handleMsgCallback(frame)
 	case "aibot_event_callback":
-		slog.Debug("wecom-ws: event callback received (ignored)", "req_id", frame.Headers.ReqID)
+		p.handleEventCallback(frame)
 	case "":
 		// Response frame (no cmd): identify by req_id prefix
 		reqID := frame.Headers.ReqID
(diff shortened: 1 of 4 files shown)


chenhg5 (commented):
Conclusion: Comment

Overall assessment:
- WeCom template card support is well-scoped and the per-mode (HTTP Webhook / WebSocket) implementations share a single `buildWeComTemplateCard` builder with good test coverage. The follow-up commit is a legitimate errcheck fix and should not block merge, but two small items below are worth addressing before approve so the callback path is shipped in a form we can maintain.

Review scope:
- platform/wecom/wecom_card.go (new builder + permission helper)
- platform/wecom/wecom.go (HTTP webhook CardSender + template_card_event XML handler)
- platform/wecom/websocket.go (WebSocket CardSender + handleEventCallback)
- platform/wecom/wecom_card_test.go (new tests)
- Run: `go test -race ./platform/wecom/... -count=1 -timeout 60s` — all 48 tests pass.

✅ What looks good:
- Shared `buildWeComTemplateCard` keeps HTTP and WebSocket code paths symmetric and easy to keep in sync going forward.
- `parseWeComPermissionResponse` cleanly separates "this is a permission callback" from "this is some other action key", which is the right factoring for `core.Message.IsPermissionResponse`.
- The errcheck follow-up (`_ = resp.Body.Close()`) is the standard fix; bundling it as a second commit rather than amending is acceptable, but it does add an extra commit on a 2-commit PR — see 🟠 P2.

🚨/🔴 Must fix:
- None found. No merge-blocking correctness/security issue.

🟠 Should improve:
- 🟠 P2 — Webhook `template_card_event` does not consult `p.allowFrom`. The WebSocket path (`handleEventCallback`) explicitly calls `core.AllowList(p.allowFrom, body.From.UserID)` before dispatching; the XML/HTTP path drops that check and feeds the event straight into `p.handler`. Anyone who clicks a card button on an Enterprise WeChat bot now bypasses the configured `allow_from` allowlist. Suggest mirroring the WebSocket guard before `go p.handler(...)`.
- 🟠 P2 — `SendCard` and `ReplyCard` both funnel into the same `sendCardInternal` for HTTP and the same shape for WebSocket, so there is no in-place card update path. The PR description implies `ReplyCard` is the "reply to a button click" path; if that is the intent, consider documenting that this is "send a new card" semantics and that any future `UpdateCard` work will need a separate endpoint. Not blocking, but flagging the API contract.
- 🟠 P2 — The errcheck fix is a one-line defer wrapper that the author could have included in the feat commit (or amended in a force-push). Two commits here doubles the review surface and pollutes `git log` for a 469-line feature. Recommend squashing before merge so the history reads as "feat(wecom): add interactive card sending and callback feedback (incl. errcheck-cleanup)".
- 🔵 P3 — `wecom_card_test.go` exports `waitTimeout` at package level. If you add another test file that also needs a wait helper, this collides. Consider moving it next to the test or using `t.Deadline()` / a t.Cleanup-based pattern.
- 🔵 P3 — CHANGELOG.md is untouched. cc-connect normally adds an Unreleased entry for new platform-facing capability. Suggest a short "### Added — WeCom: support `template_card` send + `template_card_event` button callbacks (#1645)" so release notes pick it up.

❓ Questions:
- For `text_notice` cards, the payload silently injects `" "` into `sub_title_text` when both title and body are empty. The WeCom API rejects empty card bodies; is the intent that callers always provide at least one of title/markdown? Worth a one-line comment if so.
- Does `ReplyCard` on the WebSocket side actually ever have a non-empty `rc.reqID`? `handleEventCallback` constructs the `wsReplyContext` with `reqID: frame.Headers.ReqID`, which is the aibot_event_callback request id, not a send-reply id. If that is intentionally usable as a reply ack, OK; if not, this branch is dead and the fallback `p.SendCard` path is the only one that matters.

Testing / Risk:
- Verified: `go test -race ./platform/wecom/... -count=1 -timeout 60s` → ok in ~1.26s.
- Not covered by tests: the actual `sendCardInternal` HTTP call path (only the builder + callback parser are unit-tested). Manual smoke test against a real WeCom app is recommended to confirm `msgtype: template_card` is accepted, since template_card layouts vary slightly between "应用" and "群机器人" WeCom apps.
- Risk surface: bounded to `platform/wecom`. `core.Card` / `core.CardSender` contracts unchanged.

Next step:
- Either add the missing `allowFrom` check on the HTTP `template_card_event` branch (preferred, P2) or explicitly confirm the WeCom adapter does not use `allow_from` for events. Once that is resolved, squash the two commits and merge.

PR #1650, opened 2026-08-06

### LINKED ISSUE #1077: [Bug] new_per_run cron sessions remain visible in /list after completion

AaronZ345:
## Summary

Fixes #1077.

## Tests
- `go test ./core`

Supersedes #1417. This is the same rebased patch on a fresh PR branch, including the requested compatibility opt-out.

Supersedes #1497. This fresh replacement carries the same patch rebased onto the latest `main` after the original PR exceeded 30 days.


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
This PR introduces a clean `Background` flag on `Session`, propagates it from cron/timer execution paths (`NewBackgroundSession`), and gates `applySessionFilter` on a new per-project `hide_scheduler_sessions` config (default `true`). The implementation cleanly fixes #1077 — completed `new_per_run` cron/timer sessions no longer pollute `/list`, `/switch`, `/delete`, or the search/delete confirm views — with a per-project escape hatch. Local `go test -race ./core ./config` and `go test -tags no_web ./cmd/cc-connect` are green; the only failing targets are pre-existing env issues unrelated to this PR.

Review scope:
- Read `core/session.go`, `core/engine.go`, `cmd/cc-connect/main.go`, `config/config.go`, `config/config_test.go`, `config.example.toml`, and the new `TestCmdList_*` and `TestSessionManager_NewBackground*` / `TestBackgroundAgentSessionIDs_*` tests.
- Walked through `applySessionFilter`'s 9 call sites to confirm all user-facing session commands (/list, /switch, /search, /delete, delete confirm, list card render, webhook session lookup) consistently receive the hidden-sessions filter.
- Verified the wiring: `cmd/cc-connect/main.go:434` and `:1827` set the engine flag from project config at startup and on reload; `SetHideSchedulerSessions` propagates to the SessionManager.

✅ What looks good:
- `Session.Background` is `json:"background,omitempty"`, so existing on-disk snapshots without the field unmarshal cleanly as `false` and old sessions keep showing in `/list` until they age out — that is the least-surprising behavior.
- `BackgroundAgentSessionIDs()` returns IDs even when `hideBackgroundSessions=false`, so toggling the config back to visible doesn't lose the tracked IDs. The test `TestSessionManager_BackgroundSessionsVisibleWhenDisabled` and `TestBackgroundAgentSessionIDs_IncludesPastAndPersists` both lock this in.
- `TestCmdList_HidesBackgroundSessionsByDefault` and `TestCmdList_ShowsBackgroundSessionsWhenHideSchedulerSessionsDisabled` cover the two real-world configurations, including the interaction with `filter_external_sessions=true` — that's exactly the kind of combined-state test that breaks silently otherwise.
- The CHANGELOG entry references both #1077 and the previous #1417 attempt, plus describes the per-project opt-out.

🟠 Should improve:
- **Depends on #1651**: as written, `NewBackgroundSession` is only called from `ExecuteCronJob` / `ExecuteTimerJob` when `useNewSession=true`, and on stock `main` that requires the per-job `session_mode = "new_per_run"`. Until #1651 flips the default, this PR is a behavioral no-op for default users. Either merge them together, or note in the PR description that #1651 must land first.
- **Migration gap**: existing `new_per_run` side sessions saved to disk before this PR will not have `Background=true` and will continue to show in `/list` indefinitely. The PR doesn't sweep them. Probably acceptable (they'll age out as users re-create sessions), but worth a one-line "existing pre-upgrade sessions remain visible" caveat in the CHANGELOG so users with hundreds of accumulated cron sessions understand why the list is still long right after upgrade.
- **Test gap**: no test asserts the Web-management-API reload path (`reloadConfig` at `cmd/cc-connect/main.go:1827`) re-applies `SetHideSchedulerSessions` correctly after the user edits `hide_scheduler_sessions` in `cc-connect.toml`. The startup path is tested via `TestLoad_HideSchedulerSessions*`, but the runtime flip case isn't covered. Consider adding a small unit test against `reloadConfig` or exercising the `core.Engine` directly with two different configs in sequence.
- **Filter composition comment**: `applySessionFilter` runs the hidden-sessions filter first, then the owned-sessions filter. The two filters interact cleanly (an external session that happens to match a hidden ID is double-filtered but idempotent), but a one-line comment about ordering and idempotency would help future readers.

🔵 Optional:
- `filterHiddenSessions` is a near-duplicate of `filterOwnedSessions` (the existing one). Both build a new slice and could share a generic `filterByIDs(sessions, exclude)` helper. Trivial dedup.
- Consider exposing `Background: bool` in any admin/debug endpoint so operators can see which sessions are hidden without toggling the config — useful for support cases like "why isn't my old cron session in `/list`?".

❓ Questions:
- Does `SetHideSchedulerSessions` need to also flush the active cache of `/list` rendered cards, or are stale cards re-rendered on next interaction? If the latter, fine; if the former, please confirm there's no stale view that would keep showing pre-toggle state.

Testing / Risk:
- Verified: `go test -race ./core ./config` and `go test -tags no_web ./cmd/cc-connect` are green. The two new top-level tests + the three config-load tests all pass.
- Unverified risk: any caller that calls `agent.ListSessions()` directly without going through `applySessionFilter` (e.g. `/resume` paths, future TUI dashboards) would still show background sessions. Worth a `grep -rn "agent.ListSessions" --include="*.go"` to confirm only the 9 in-scope call sites exist.

Next step:
- Merge after #1651 lands (or together in the same release), add a CHANGELOG note about pre-upgrade visible sessions, and consider exposing the `Background` flag in a future debug endpoint.

AaronZ345:
Superseded by #1801. The fresh PR carries the same patch on current main.

PR #1651, opened 2026-08-06

AaronZ345:
## Summary

## Root cause

## Test Plan
- `GOOS=linux GOARCH=amd64 go test -c -tags no_web ./cmd/cc-connect`
- `git diff --check`

Supersedes #1349. This is the same rebased patch on a fresh PR branch.

Supersedes #1496. This fresh replacement carries the same patch rebased onto the latest `main` after the original PR exceeded 30 days.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,8 @@ the full themed summary; per-beta details remain in the beta sections below.
   because a later queued message has a higher `create_time` (#1286)
 - **core**: make `/history` entry truncation configurable via
   `[display].history_max_len`, defaulting to 1000; `0` disables truncation (#1291)
+- **cron/timer**: prompt jobs default to a fresh agent session per run, avoiding busy
+  chat sessions; use `session_mode = "reuse"` to keep shared-session behavior
 - **tts/minimax**: drop `status=2` trailer chunk to stop audio playing twice (#1364)
 - **tests**: add provider-resume regression tests for codex / opencode / kimi (#1366)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,8 @@ the full themed summary; per-beta details remain in the beta sections below.
   because a later queued message has a higher `create_time` (#1286)
 - **core**: make `/history` entry truncation configurable via
   `[display].history_max_len`, defaulting to 1000; `0` disables truncation (#1291)
+- **cron/timer**: prompt jobs default to a fresh agent session per run, avoiding busy
+  chat sessions; use `session_mode = "reuse"` to keep shared-session behavior
 - **tts/minimax**: drop `status=2` trailer chunk to stop audio playing twice (#1364)
 - **tests**: add provider-resume regression tests for codex / opencode / kimi (#1366)
 
--- a/cmd/cc-connect/cron.go
+++ b/cmd/cc-connect/cron.go
@@ -626,7 +626,7 @@ Options:
       --prompt <text>        Task prompt (runs through agent)
       --exec <command>       Shell command (runs directly, mutually exclusive with --prompt)
       --desc <text>          Short description
-      --session-mode <mode>  reuse (default) or new-per-run — fresh agent session each run
+      --session-mode <mode>  new-per-run (default) or reuse — fresh agent session each run unless reusing
       --timeout-mins <n>     Max minutes to wait per run (0 = no limit; default 30 if omitted)
       --silent               Suppress cron start notification
       --data-dir <path>      Data directory (default: ~/.cc-connect)
--- a/cmd/cc-connect/timer.go
+++ b/cmd/cc-connect/timer.go
@@ -412,7 +412,7 @@ Options:
       --prompt <text>        Task prompt (runs through agent)
       --exec <command>       Shell command (runs directly, mutually exclusive with --prompt)
       --desc <text>          Short description
-      --session-mode <mode>  reuse (default) or new-per-run
+      --session-mode <mode>  new-per-run (default) or reuse
       --timeout-mins <n>     Max minutes to wait per run (0 = no limit; default 30)
       --mute                 Suppress all messages (start + result)
       --data-dir <path>      Data directory (default: ~/.cc-connect)
--- a/config.example.toml
+++ b/config.example.toml
@@ -336,8 +336,8 @@ level = "info" # debug, info, warn, error
 # [cron]
 # silent = false             # Suppress "⏰ task name" notification when cron starts (default: false)
 #                            # 静默模式：不发送定时任务开始执行的通知消息（默认 false）
-# session_mode = "reuse"     # Default session mode for all cron jobs: "reuse" (default) or "new_per_run"
-#                            # 定时任务默认会话模式："reuse"（复用活跃会话）或 "new_per_run"（每次新建会话）
+# session_mode = "new_per_run" # Default session mode for cron/timer jobs: "new_per_run" (default) or "reuse"
+#                              # 定时/延时任务默认会话模式："new_per_run"（默认，每次新建会话）或 "reuse"（复用活跃会话）
 
 # =============================================================================
 # Webhook / 外部 Webhook 端点
--- a/config/config.go
+++ b/config/config.go
@@ -139,7 +139,7 @@ type Config struct {
 // CronConfig controls cron job behavior.
 type CronConfig struct {
 	Silent      *bool  `toml:"silent"`       // suppress cron start notification; default false
-	SessionMode string `toml:"session_mode"` // default session mode: "" or "reuse" (default) or "new_per_run"
+	SessionMode string `toml:"session_mode"` // default session mode: "new_per_run" (default) or "reuse"
 }
 
 // QueueConfig controls the per-session message queue.
--- a/core/cron.go
+++ b/core/cron.go
@@ -71,6 +71,8 @@ func (j *CronJob) UsesNewSessionPerRun() bool {
 	return NormalizeCronSessionMode(j.SessionMode) == "new_per_run"
 }
 
+// defaultScheduledSessionMode is the built-in default for scheduled cron and
+// timer jobs. It can be overridden globally by scheduler config or per job.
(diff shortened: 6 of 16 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
This PR flips the default session mode for cron/timer prompt jobs from `reuse` to `new_per_run`, fixing the long-standing busy-session conflict where scheduled jobs competed with active chat turns. Behavior change is intentional, fully documented, and gated behind a clean opt-out (`session_mode = "reuse"`). Local `go test -race ./... -count=1 -timeout 60s` passes for the in-scope packages (`core`, `config`, `cmd/cc-connect` with `-tags no_web`); the only failing targets are pre-existing env issues (missing `web/dist`, `cursor login` requirement) and are unrelated to this patch.

Review scope:
- Read `core/cron.go`, `core/timer.go`, `core/interfaces.go`, `cmd/cc-connect/{cron,timer}.go`, `config/config.go`, `config.example.toml`, all updated tests, and the touched doc files.
- Walked through `NormalizeCronSessionMode`, `UsesNewSession`, `validateCronJob`, `UpdateJob`, and the two scheduler constructors to confirm the new default and the explicit-reuse path are wired end-to-end.
- Ran `go test -race ./core ./config ./cmd/cc-connect` (with `-tags no_web`); all pass. The two unrelated failures (`cmd/cc-connect` without `no_web` needs `web/dist`; `agent/acp.TestCursorCLI_ACPHandshake` needs `agent login`) reproduce on `main` and are not regressions.

✅ What looks good:
- `defaultScheduledSessionMode` as a single named constant removes the previous "empty string means reuse" ambiguity. `NormalizeCronSessionMode` now distinguishes `""` (inherit default), `"reuse"`, and `"new_per_run"` cleanly.
- `validateCronJob` and `UpdateJob` both accept the explicit `reuse` value with a clear error message — opt-out is well supported.
- The CHANGELOG entry is concise and tells users exactly what to do if they need the old behavior.
- The CLI help text, the management API docs (both EN and ZH-CN), and `config.example.toml` were all updated consistently. The `--session-mode` and `[cron].session_mode` defaults are flipped everywhere users could discover them.
- `TestExecuteCronJob_ResolvesCronReplyTarget` was correctly retrofitted with `SessionMode: "reuse"`. That single-character edit shows the author understood the default flip and audited the test corpus.

🟠 Should improve:
- **Merge ordering**: this PR is the prerequisite for #1650 (hide scheduler background sessions). On its own, #1650 is dormant because `defaultSessionMode == ""` means `useNewSession=false` and `NewBackgroundSession` is never reached. The CHANGELOG doesn't mention this dependency; merging them out of order leaves #1650 silently invisible to default users. Worth a brief coordination note in the PR body or in the team handoff.
- **Migration audit**: only one existing test (`TestExecuteCronJob_ResolvesCronReplyTarget`) was updated for the new default. Any other test that exercises `ExecuteCronJob` / `ExecuteTimerJob` with an empty `SessionMode` would have changed semantics silently. I ran the targeted test set (`TestExecuteCronJob_*`, `TestExecuteTimerJob_*`, `TestSessionManager_*`, `TestValidateTimerJob`, `TestMgmt_CronExecByID`) and all pass, but a CI run on the full `go test ./...` should confirm there are no other implicit-dependency tests in the suite that this author missed.
- **Implicit config-default flip**: `[cron]` section in `config.example.toml` changes the documented default from `reuse` to `new_per_run`. Operators who copy `config.example.toml` as a starting point will silently get the new behavior even if their previous `cc-connect.toml` had no `[cron]` block. This is fine but is the kind of thing that shows up as a "why did my cron behavior change?" support ticket later. Consider adding a one-line "since vX.Y, default is now `new_per_run`" callout in `docs/usage.md` (currently the prose was rewritten but doesn't carry a "changed in this release" anchor).

🔵 Optional:
- The dead-code `"" / "reuse"` collapse removed in `NormalizeCronSessionMode` is good — but the comment block above `defaultScheduledSessionMode` could mention the previous "empty == reuse" semantic so future readers understand why the constant exists.
- `TestCronScheduler_AddJob_NormalizesSessionMode` now also asserts that `""` round-trips as `""`. Adding a `t.Cleanup`-style assertion that an invalid mode (e.g. `"foobar"`) is rejected at validation would lock the current behavior in case someone refactors `NormalizeCronSessionMode` later.

❓ Questions:
- Did you audit the Web management API's cron-create/edit forms for any UI elements that hard-code `reuse` as the default radio button / dropdown selection? If the UI sends back whatever it displays, that would mask the new backend default for users who touch cron jobs through the web console.

Testing / Risk:
- Verified: `go test -race ./core ./config ./cmd/cc-connect` with `-tags no_web` is green. `TestCronScheduler_UsesNewSession_GlobalDefault`, `TestTimerScheduler_UsesNewSession_Defaults`, `TestValidateTimerJob`, and `TestExecuteCronJob_ResolvesCronReplyTarget` all pass.
- Unverified risk: any consumer of `NormalizeCronSessionMode` outside `core/` and `cmd/cc-connect/` (e.g. `web/`, `tests/blackbox/`) that relies on `""` collapsing to "reuse" would now see different behavior. Worth a quick `grep -rn "NormalizeCronSessionMode" --include="*.go"` to confirm scope.

Next step:
- Land this PR first; then immediately follow with #1650 so the new default is paired with the hide-from-list behavior. Update the management API web form defaults in a follow-up if any UI elements still show `reuse` as the default selection.

chenhg5 (approved):
## Conclusion: Approve

## Overall assessment
PR #1651 is mergeable. The head rebases cleanly onto current `main` (24 ahead / 3 behind, `mergeStateStatus=CLEAN`), CI is 5/5 green on the current head (`ce9bbc4`), and the substantive change is the same patch the maintainer already approved: flip cron/timer prompt-job session default from `reuse` to `new_per_run`, with explicit `session_mode = "reuse"` opt-in restored.

## Review scope
- Diff against current head (15 files, +141/-83): `core/cron.go`, `core/timer.go`, `config/config.go`, `config.example.toml`, `cmd/cc-connect/{cron,timer}.go`, `core/cron_test.go`, `core/timer_test.go`, `core/engine_test.go`, `core/i18n.go`, `docs/{management-api,management-api.zh-CN,usage,usage.zh-CN}.md`, `CHANGELOG.md`.
- The three PR-branch commits:
  - `9b04241` — flip default + tests + docs
  - `11cbe34` — doc comment on `defaultScheduledSessionMode`, plus `waitRefreshedCards` helper retrofitted into four delete-mode tests to remove a `len(refreshed) == 0` flake window
  - `ce9bbc4` — CI retrigger after a GitHub Actions download failure

## ✅ What looks good
- `defaultScheduledSessionMode = "new_per_run"` as a single named constant removes the prior "empty string == reuse" ambiguity. `NormalizeCronSessionMode` now distinguishes `""` (inherit), `"reuse"`, `"new_per_run"`, with `"new-per-run"` still accepted as an alias.
- Both schedulers' constructors wire the new default. `validateCronJob`, `validateTimerJob`, and `UpdateJob` all accept an explicit `"reuse"` and return a clear error for unknown values.
- `TestExecuteCronJob_ResolvesCronReplyTarget` was retrofitted with `SessionMode: "reuse"` to keep its old semantics under the new default — confirms the author audited the existing test corpus.
- The `docs(core): document scheduled session default` follow-up adds the explanatory comment the prior reviewer asked for and converts four flaky `len(refreshed) == 0` reads into a bounded `waitRefreshedCards(t, p, 1)` poll. That removes a real flake class.
- CLI help, `[cron]` config example, both `usage.md` flavors, and both `management-api.md` flavors all carry the new default — discovery paths are consistent.
- `mergeStateStatus: CLEAN`, all five checks pass on the current head.

- **Coordination with the hide-sessions follow-up**: if there's a paired PR that hides scheduler background sessions from the chat list, the PR body doesn't reference it. Merging #1651 alone leaves users with fresh isolated sessions that still appear in the session picker, which is a minor UX wart. Worth a one-line cross-link in the PR body or in the team handoff so the maintainer can sequence them.
- **`config.example.toml` default flip without "since vX.Y" anchor**: users who copy `config.example.toml` as their `cc-connect.toml` starter will silently switch behavior. The rewritten prose in `docs/usage.md` covers the new default well, but adding a one-line "Default changed in vX.Y" anchor near the CLI flag description would help future support tickets.
- **`docs(core): document scheduled session default` commit message**: the commit body says "document scheduled session default" but actually adds a doc comment + a flake-resilience refactor. Splitting the comment addition from the test refactor would make `git log -p` history cleaner; not a blocker.

## 🔵 Optional
- Consider asserting in `TestCronScheduler_AddJob_NormalizesSessionMode` that an invalid mode (e.g. `"foobar"`) is rejected at validation. The current sub-tests only cover the three accepted values.
- `TestExecuteCronJob_ResolvesCronReplyTarget` now relies on the `SessionMode: "reuse"` line as a load-bearing assertion. A one-line comment near it would help the next person who touches that test.

## ❓ Questions
- Did the web management UI's cron-create/edit forms get updated, or does the front end still default to `reuse` in its radio/dropdown? If the web form sends whatever it displays, the backend default flip won't be visible to users who manage cron jobs through the web console.

## Testing / Risk
- Verified on current head: CI run 33768817317 — `lint` 2m22s, `unit-test` 4m58s, `smoke-test` 30s, `regression-test` 37s, `performance-test` 53s, all `pass`.
- Verified locally (this review): the diff stays in the same logical surface as the prior approval; no new modules, no new public API surface, no schema migrations.
- Unverified risk: any external consumer of `NormalizeCronSessionMode` outside `core/` and `cmd/cc-connect/` (e.g. `web/`, `tests/blackbox/`) that relies on `""` mapping to "reuse" would now see different behavior. Quick `grep -rn "NormalizeCronSessionMode" --include="*.go"` confined to `core/` and `cmd/cc-connect/` — no other callers.

## Next step
Merge when convenient. No further code changes required from the author. The only follow-up worth filing is the web-management UI default sync (if it still shows `reuse` as the default).

AaronZ345:
Superseded by #1800. The fresh PR carries the same patch on current main; the historical empty CI-retrigger commit was excluded.

PR #1690, opened 2026-08-16
Keep configured display mode, reaction opt-out, and language changes meaningful

timmyagentic:
Second-pass review of the defaults introduced by #14 found five places where an explicit user setting no longer means what it says. Each fix is TDD'd with a failing test first.

## What changed

**`mode` keeps its official meaning (`config.EffectiveDisplay`)**
Dropping the mode-derived defaults made a spelled-out `mode = "full"` resolve to *no* process messages, so `full` and `compact` became identical and a migrated official config that asked for full output silently lost it. A configured mode now sets the thinking/tool default again (`full` → on, `compact`/`quiet` → off); leaving `mode` unset still resolves to `full` with the final-answer-only default. Per-key values still win.

**Display commands persist into the project they ran in (`config.SaveProjectDisplayConfig`)**
`/quiet` and `/config set mode` run inside one project's engine but wrote the global `[display]` table, so the change leaked into every other project and was shadowed on restart by any `[projects.display]` spelling the same key out — the runtime change looked applied and silently reverted. They now patch the project's own table, creating it when absent, and fall back to the global table for configs without a matching project.

**`reaction_emoji = "none"` covers the completion reaction (`platform/feishu`)**
The two keys were fully independent, so the new `done_emoji` default reintroduced a reaction — and its push notification — on every finished turn for users who had explicitly opted out. An explicit `done_emoji` still wins.

**`language` says when it does not understand a value (`config.NormalizeLanguage`)**
The parser accepted any string and silently degraded to auto-detect. That was invisible while auto-detect was the default; with an empty value now meaning Chinese it is a live foot-gun (`zh-CN`, `EN`, `" zh "` all became auto-detect). One definition of the accepted spellings now serves the validator, the startup parser and the persisted value, regional tags/casing/whitespace resolve, and an unmatched value warns at startup. It still falls back rather than failing: official CC Connect falls back too, so rejecting would make such a config unmigratable.

**`/lang` persists (`core.I18n.SetLangAndSave`)**
`SetLang` only changed the running process, and the save callback was wired only when the configured language was `auto`. With a fixed default there was no way to make a chat-issued language change stick. The callback is now always wired, `/lang` persists through it, and `LangAuto` is written as the spelled-out `"auto"` so the escape hatch survives.

Also updated: the stale `DisplayModeFull` / `DisplayCfg` / `AddDoneReaction` comments, the management API's language switch (it silently ignored unhandled spellings while still persisting them), `config.example.toml`, and the "Defaults that differ from official CC Connect" section.

## Testing

- `go test -tags no_web ./config/ ./core/ ./cmd/cc-connect/ ./platform/feishu/ ./tests/release_local/...` — all green.
- New: `TestEffectiveDisplay_ExplicitModeKeepsOfficialMeaning`, `TestSaveProjectDisplayConfig_*` (×3), `TestNewPlatformReactionOptOutCoversDoneReaction`, `TestNormalizeLanguage`, `TestValidateAcceptsUnknownLanguage`, `TestConfigLanguageAcceptsNaturalSpellings`, `TestI18n_SetLangAndSave*` (×2).
- `gofmt` clean on every touched file (`core/management.go` was already unformatted on main and is unchanged in that respect).


timmyagentic:
Opened against the wrong repository by mistake — this belongs to a downstream fork, not upstream. Closing immediately; apologies for the noise.

Issue #1610, opened 2026-07-28
Feature: inject thread history as context when a bot is @-mentioned (multi-bot collaboration inside a Feishu thread)


zhangyijie-ai:
## Summary

When multiple cc-connect bots (each a separate Feishu app / project) are placed in the same Feishu group and collaborate inside one **reply thread (话题)**, a bot that gets `@`-mentioned cannot see the prior messages in that thread — including the user's original question and other bots' replies. This makes multi-bot collaboration inside a thread effectively impossible.

I'd like to request a feature: **when a bot is `@`-mentioned inside a thread, inject the thread's prior message history as context** (the messages before the `@`-mention, from all participants — user + other bots).

## Environment

- cc-connect version: `1.4.1`
- Platform: Feishu
- Two projects in one group, each a separate Feishu app:
  - project A (agent: claudecode, app `cli_aaa87a...`)
  - project B (agent: codex, app `cli_aae9c7...`)
- `thread_isolation = true` on both (so each thread = one session per bot)

## Scenario (what I want)

In one Feishu thread:

1. User: `@claude <a question>`
2. claude replies
3. User: `@codex 评价一下 claude 刚才的回答,并复述我最初的问题`

Expected: `@codex` receives, as context, the full thread history before step 3 — i.e. the user's original question (step 1) + claude's reply (step 2) — so it can actually evaluate claude's answer and recall the original question.

Actual: `@codex` receives **only** the step-3 message text. It has no knowledge of the question or claude's answer, and replies "I can't see your earlier conversation".

## What I found (root cause)

From inspecting the binary and session files, the Feishu platform already has the building blocks:

- `platform/feishu.(*Platform).fetchReplyChain`
- `platform/feishu.formatReplyChain`
- `platform/feishu.(*Platform).ReconstructReplyCtx`
- `platform/feishu.parseThreadRootID`
- `platform/feishu.(*Platform).fetchChatMembers`
- session_key for a thread is `feishu:<chat_id>:root:<root_msg_id>`

So a thread is already identified by `root_id`, and the reply chain can already be fetched. What's missing is: when a bot is `@`-mentioned in a thread, the **prior messages in that thread are not fetched and injected** into the agent's context. The bot's thread-session starts empty (`is_resume=false`, `agent_session=""`) and only contains the `@`-mention message itself.

[...]

PR #1698, opened 2026-08-17
feat(feishu): share group chat context on mention

John-Ferrel:
## Summary

Add optional Feishu group-chat history sharing via `group_chat_history_share`.

When enabled, non-mentioned group messages are retained as context for the next explicit bot mention, while ordinary messages still do not trigger an agent turn or reply. This preserves the conversation leading up to a bot request without changing the default behavior.

## Type of change

* [ ] Bug fix (non-breaking change that fixes an issue)
* [x] New feature (non-breaking change that adds functionality)
* [ ] Breaking change (fix or feature that would cause existing behavior to change)
* [ ] Documentation only
* [ ] Internal refactor / chore (no user-visible change)

## Testing

Verified with the full repository test suite, race detector, build, and a live Feishu group using the patched binary.

Added `platform/feishu/group_history_test.go` with:

* `TestFeishuGroupHistory_DisabledPreservesMentionFilter` — verifies the feature defaults to the existing mention-filter behavior when disabled.
* `TestFeishuGroupHistory_ConfigDefaultsOffAndCanBeEnabled` — verifies the config defaults to `false` and can be enabled.
* `TestFeishuGroupHistory_MainChannelSharesOrderedMessagesAndConsumesOnce` — verifies main-channel messages are preserved in sender/order sequence and consumed only once after an accepted turn.
* `TestFeishuGroupHistory_StatusDoesNotConsumePendingContext` — verifies `/status` does not consume pending conversation context.
* `TestFeishuGroupHistory_NewResetsPendingContext` — verifies `/new` resets pending context.
* `TestFeishuGroupHistory_ThreadScopesDoNotLeakAndMainDoesNotFollowFork` — verifies isolated threads do not leak history and main-channel context does not follow a previous forked thread.
* `TestFeishuGroupHistory_PostAndSenderTypeFormatting` — verifies post content and user/app sender attribution.
* `TestFeishuGroupHistory_ChatAndTriggerPermissionsRemainSeparate` — verifies `allow_chat` controls observable chats while `allow_from` continues to control who may trigger the agent.
* `TestFeishuGroupHistory_GroupReplyAllRemainsImmediate` — verifies existing `group_reply_all=true` behavior is unchanged.

Commands run successfully:

```bash
go test ./...
go test -race ./...
make test-full
make build
```

The patched binary was also manually verified in a live Feishu group: non-mentioned messages were retained and supplied to the next explicit bot mention without triggering replies themselves.

Not applicable — this PR adds an opt-in feature and does not change the default behavior.

* [ ] No CUJ touched (small refactor, doc change, etc.)
* [x] A — basic conversation
* [x] B — session lifecycle (`/new` `/switch` `/list` `/history` etc.)
* [ ] C — agent execution control (`/mode` `/cancel` `/stop` permissions)
* [x] D — security & permissions (`allow_from` `admin_from` `banned_words` rate limits)
* [ ] E — scheduled tasks (`/cron` `/timer`)
* [ ] F — config switching (`/lang` `/provider` `/model` reload)
* [ ] G — error handling & robustness (LLM failure, ws reconnect, agent crash)
* [ ] H — multi-platform / multi-project isolation
* [ ] I — UI rendering correctness (cards, streaming, display modes)

* [x] `go test ./core/ -run TestCUJ` passes locally.
* [ ] If the change alters an existing user-visible flow, the corresponding CUJ test was updated (or a new CUJ added) to cover the new behavior.

No core CUJ was added because this is a new opt-in Feishu-adapter behavior, disabled by default, and its end-to-end message capture semantics are covered directly in `platform/feishu/group_history_test.go`.

With:

```toml
group_chat_history_share = true
```

allowed Feishu group `text`/`post` messages that do not mention the bot are retained in bounded in-memory context instead of being discarded.

They do not invoke the agent or produce a reply.

On the next explicit bot mention:

* pending messages from the same main-channel or thread scope are injected through the existing `ExtraContent` path;
* sender names and message ordering are preserved;
* the history is consumed only after a real agent turn is accepted;
* `/status` and similar non-agent commands do not consume it;
* `/new` resets the relevant pending context;
* isolated threads do not share pending context.

The feature defaults to `false`, so existing deployments are unaffected.

The implementation does not fetch Feishu history or persist context across process restarts. It reuses messages already delivered to the Feishu adapter.

Receiving non-mentioned group messages requires the appropriate Feishu group-message permission (`im:message.group_msg`).

* [x] `go build ./...` passes
* [x] `go test ./...` passes (with `-race` if touching concurrency)
* [x] AGENTS.md Pre-Commit Checklist items are satisfied
* [x] No new hardcoded platform/agent names in `core/`
* [x] i18n strings have all-language translations (if any new user-facing text)
* [x] No secrets / credentials in source

## Related

* Issue: #1610
* Related PR: None

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1028,6 +1028,10 @@ app_secret = "your-feishu-app-secret"
 #                                   # 设为 true 时，群聊内所有用户共享同一个 Agent 会话
 # thread_isolation = false  # If true, group messages use Feishu reply threads as session boundaries (one thread/root = one agent session and, in multi-workspace mode, one workspace binding)
 #                           # 设为 true 时，群聊按飞书话题/根消息隔离会话；multi-workspace 模式下每个话题也独立绑定 workspace
+# Share non-mentioned group messages as context with the next bot-triggered turn.
+# Messages themselves do not trigger a reply.
+# group_chat_history_share = false
+# 将未 @ 机器人的群消息作为下一次触发的上下文共享；消息本身不会触发回复。
 # reply_to_trigger = true   # Default: bot uses “reply to message” API (quotes the user’s message). Set false to post plain chat messages instead.
 #                           # 默认 true：用「回复消息」API（引用用户消息）。设为 false 则在会话里发普通消息、不引用触发消息。
 # progress_style = "legacy" # Progress rendering style: legacy | compact | card
--- a/docs/feishu.md
+++ b/docs/feishu.md
@@ -110,6 +110,7 @@ app_secret = "QhkMpxxxxxxxxxxxxxxxxxxxx"
 # domain = "https://open.feishu.cn" # 可选：覆盖运行时 API/WebSocket 域名
 # enable_feishu_card = true  # 可选：关闭后统一回退纯文本回复
 # thread_isolation = true    # 可选：按飞书 thread/root 隔离群聊会话
+# group_chat_history_share = false  # 可选：共享未 @ 机器人的群消息作为下一次触发的上下文；消息本身不会触发回复
 # progress_style = "legacy"  # 可选：legacy | compact | card
 # done_emoji = "none"          # 可选：agent 完成回复后添加的表情回复（如 "Done"）；设为 "none" 可禁用
 # image_batch_window_ms = 500  # 可选：连续多图合批窗口（默认 500ms，详见下文）
@@ -117,6 +118,7 @@ app_secret = "QhkMpxxxxxxxxxxxxxxxxxxxx"
 
 > 如果应用没有交互卡片权限，或后台未配置卡片回调，可将 `enable_feishu_card = false`，让所有命令统一走纯文本回复，避免卡片发送失败后用户看不到内容。
 > 如果开启 `thread_isolation = true`，群聊里每个根消息 / reply thread 会对应一个独立 agent session；私聊行为保持原样。
+> `group_chat_history_share = true` 时，cc-connect 只在内存中保留当前进程观察到的、允许访问的群聊 text/post 消息，并在下一次明确 @ 机器人且真正进入 agent turn 时注入；未 @ 的消息不会触发回复。`/status` 等由 cc-connect 处理的命令不会消费这段待处理上下文，`/new` 会清空对应主频道或话题的上下文。
 > 在 multi-workspace 模式下，`thread_isolation = true` 也会让每个话题独立绑定 workspace；在话题内执行 `/workspace bind <name>` 不会影响同群的其他话题。已有的群级 binding 会保留为默认值，由尚未显式绑定的话题继承，因此回退到旧版本时仍可使用。
 > `progress_style = "compact"` 会把思考/工具进度合并到一条可更新消息里，减少刷屏；`legacy` 保持原有逐条发送；`card` 会使用结构化卡片（标题 + 进度块）持续更新同一条消息，观感比纯文本更清晰。
 > `domain` 只影响运行时 API / WebSocket 请求地址；CLI `setup/new/bind` 的引导域名仍然使用内置默认值。
@@ -164,6 +166,8 @@ app_secret = "QhkMpxxxxxxxxxxxxxxxxxxxx"
 ### 4.3 发布权限申请
 
 配置完权限后，点击「申请发布」使权限生效。
+
+如果启用了 `group_chat_history_share`，必须为应用申请并发布 `im:message.group_msg`，否则飞书只会向机器人推送被 @ 的群消息，未提及消息无法进入共享上下文。该功能不会回溯 cc-connect 启动前的历史，也不会持久化待处理消息。
 
 ---
 
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -131,6 +131,7 @@ type Platform struct {
 	respondToAtEveryoneAndHere bool
 	shareSessionInChannel      bool
 	threadIsolation            bool
+	groupChatHistoryShare      bool
 	// noReplyToTrigger: when true, send via Create instead of Im.Message.Reply (no quote to the user's message).
 	noReplyToTrigger bool
 	resolveMentions  bool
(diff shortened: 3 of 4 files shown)


chenhg5 (approved):
## QA Review — PR #1698 (Feishu group chat context on mention)

**Conclusion:** Approve

**Overall assessment:** Solid opt-in feature with strong privacy defaults and a clean integration point (`core.Message.OnAccepted` callback, already present in main). CI is 5/5 SUCCESS, merge state CLEAN, head `5a6f6b1`. The implementation is well-scoped: a new `pendingGroupHistory` map in the Feishu adapter, a bounded buffer per scope, and a snapshot/consume pattern gated on real agent turn acceptance — not slash commands.

**Review scope:**
- `config.example.toml`: +4 (new option with bilingual comment)
- `docs/feishu.md`: +4 (new behaviour note + permission requirement)
- `platform/feishu/feishu.go`: +309/-15 (history capture + consume + thread scope + recall)
- `platform/feishu/group_history_test.go`: +323 (9 tests)
- Focused on privacy/defaults, scope isolation, callback wiring, and thread-safety.

**✅ What looks good:**

- **Default OFF.** `group_chat_history_share = false` by default. Pre-existing deployments are unaffected. `TestFeishuGroupHistory_ConfigDefaultsOffAndCanBeEnabled` pins this.
- **`allow_chat` checked before capture.** The PR explicitly distinguishes "chat controls whether cc-connect may observe the conversation" (allow_chat) from "allow_from controls who may trigger an agent turn" — and tests this in `TestFeishuGroupHistory_ChatAndTriggerPermissionsRemainSeparate`. This is the right security split.
- **Scope isolation works correctly.** `groupHistoryScope` returns `chat:<id>` for normal mode and `thread:<chat>:<root>` when `thread_isolation` is enabled, so a forked thread session cannot accidentally pull main-channel context into its agent turn. `TestFeishuGroupHistory_ThreadScopesDoNotLeakAndMainDoesNotFollowFork` pins this end-to-end.
- **`core.Message.OnAccepted` is the right hook.** It's already in main (`core/message.go:378`), and the engine's `runMessageAccepted` (engine.go:3076) is called in the agent-turn path (line 3058) **and** the busy-session queue path (line 3196), but **not** in the slash-command dispatch path (which is a separate switch that handles `/status`, `/new`, `/list`, etc. directly). The author's claim that `/status` does not consume and that `/new` resets is therefore correct by construction, not by special-casing. This is the cleanest possible use of an existing API.
- **`/new` reset detection** uses `isGroupHistoryNewCommand`, which strips the bot mention and compares the first field to `/new` case-insensitively. Correct shape; doesn't false-positive on messages that happen to start with "/new" inside a sentence because `strings.Fields` tokenises first.
- **Bounded buffer.** `maxPendingGroupHistoryEntries = 50` per scope prevents an unbounded growth DoS on a busy group. Older entries are evicted when the cap is exceeded.
- **Recall handling.** `onMessageRecalled` calls `removeGroupHistory(messageID)`, so a recalled message is also removed from pending context. `formatGroupHistory` additionally double-checks `isMessageRecalled` at format time (defence in depth).
- **Sender attribution.** `historySenderName` resolves `app` vs user, caches user lookups, and strips newlines from the rendered name. Correct for the prompt-injection surface.
- **Permission requirement documented.** The docs explicitly call out that `im:message.group_msg` is required to receive non-mentioned messages. Without this permission, the feature silently produces no history.

**🟠 Should improve (non-blocking but worth a maintainer ack):**

- **`isGroupHistoryMessageType` includes `post`**, which is the rich-text Feishu message type. `extractPostPlainText` is presumably implemented elsewhere in the file, but the PR diff doesn't add it — confirm that this function already exists in feishu.go and handles the edge cases (mentions in post, embedded @-tags, image-only posts). I see `parsePostContent` is called at line ~1895 but I want to make sure `extractPostPlainText` returns safe text without leaking embedded `<at user_id="…">` markup.
- **Privacy implication of opt-in**: even at 50 entries × N scopes, this stores user-generated text in process memory that has no TTL. A user who enables the feature, leaves it on for a week, and then disables it has the buffer drained only when the process restarts. Consider adding a TTL eviction sweep (e.g. drop entries older than 30 minutes) to bound the memory exposure window. The doc says "not persisted across process restarts" but doesn't bound in-process lifetime. Maintainer call: is this worth a follow-up?
- **`onAccepted` callback is invoked even when the agent turn is queued for a busy session** (engine path at line 3196). That's intentional — the message has been *accepted* — but it means a user who fires a bot mention while the bot is already busy will *consume* the pending history. The next mention will not see it. This matches the test's expectation but is worth surfacing in the doc: "the next *accepted* mention" rather than "the next mention".

- A short note in `AGENTS.md` or `docs/feishu.md` clarifying that `OnAccepted` is the platform-level signal for "this message will be processed" (vs. "this message arrived") would help future platform authors understand the new contract.
- The `quotedMessage` struct in the diff now carries `text: historyText` for image-batch entries. Worth a doc comment on `imageBatchEntry.onAccepted` to explain that this is the consumption callback the platform will run once the engine accepts the batched image turn.

**❓ Questions (please ack before merge):**

1. **TTL for pending history** — keep unbounded (per-session, 50-entry cap), or add a 30-minute idle eviction? My preference: add TTL, file as a follow-up. Privacy wins, memory bounded.
2. **`extractPostPlainText` safety** — does it handle embedded `<at>` markup correctly? If you can confirm via existing tests, no action needed.

**Testing / Risk:**

- Verified evidence: CI 5/5 SUCCESS (`lint`, `unit-test`, `smoke-test`, `regression-test`, `performance-test`) on head `5a6f6b1`. Author confirms `go test ./...`, `go test -race ./...`, `make test-full`, `make build` green. Live Feishu group manually verified.
- 9 new tests cover: disabled-by-default, ordering, single-consume, /status non-consume, /new reset, thread-scope isolation, sender attribution, allow_chat vs allow_from separation, group_reply_all preservation.
- Remaining risk:
  1. **Memory growth on long-lived sessions** — bounded per-scope at 50 entries, unbounded across scopes. A user in 100 active group chats × 50 entries = 5000 in-memory entries. Acceptable for a single user; worth monitoring.
  2. **`post`-format messages** — depends on `extractPostPlainText` correctness. Pre-existing function, not introduced by this PR, but the new code relies on it.
  3. **Privacy disclosure** — the feature is documented but the in-app notice surface is only the config comment. Consider adding a startup log line ("group_chat_history_share=true, will retain up to 50 messages per chat scope") so admins see this in their logs.

**Next step:** Recommend merge once the two questions above are acked. The feature is opt-in, scope-correct, privacy-considered, and uses an existing core API correctly. Privacy follow-up (TTL + startup log) is worth filing but not blocking.

John-Ferrel:
Thanks for the review.

I’ve rebased this PR onto the latest `main` and resolved the Feishu conflict while preserving the newer fail-closed group mention filtering. The PR scope remains unchanged; I also added a focused regression test to ensure degraded bot-ID resolution does not accidentally capture group history.

On the two questions:

1. **TTL** — agreed. I’d keep the current 50-entry per-scope bound in this PR and handle TTL eviction / startup disclosure as a follow-up rather than expand the scope here.

2. **`extractPostPlainText`** — confirmed. The current implementation explicitly handles `at` elements, including named mentions, `@all`, and the user-ID fallback. There is already a focused `TestExtractPostPlainText_AtMention` covering these cases, so no additional change is needed here.

After the rebase, the following are green locally:

* `go test ./...`
* `go test -race ./...`
* `make test-full`
* `make build`
* `go test ./core/ -run TestCUJ`

PR #1699, opened 2026-08-17
fix(core): invalidate stale workspace sessions

gpYang:
## Summary

- Invalidate persisted workspace-session AgentSessionID values when a per-workspace agent is created for a different agent type.
- Keep same-agent persisted session IDs intact.
- Add a regression test that loads a persisted workspace session file and verifies stale IDs are cleared.

## Verification

- `go test ./core -run TestGetOrCreateWorkspaceAgent_InvalidatesStaleSessionIDs -count=1`
- `go test ./core -count=1`
- `GOOS=linux GOARCH=amd64 go build -tags no_web ./cmd/cc-connect`

## PR overlap check

- Checked the current open PRs around multi-workspace/session behavior; no exact duplicate of this stale workspace-session invalidation fix was found.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -3944,6 +3944,7 @@ func (e *Engine) getOrCreateWorkspaceAgent(workspace string) (Agent, *SessionMan
 	sessionFile := filepath.Join(filepath.Dir(e.sessions.StorePath()),
 		fmt.Sprintf("%s_ws_%s.json", e.name, hex.EncodeToString(h[:4])))
 	sessions := NewSessionManager(sessionFile)
+	sessions.InvalidateForAgent(agent.Name())
 
 	ws.agent = agent
 	ws.sessions = sessions
--- a/core/engine.go
+++ b/core/engine.go
@@ -3961,6 +3961,7 @@ func (e *Engine) getOrCreateWorkspaceAgent(workspace string) (Agent, *SessionMan
 	sessionFile := filepath.Join(filepath.Dir(e.sessions.StorePath()),
 		fmt.Sprintf("%s_ws_%s.json", e.name, hex.EncodeToString(h[:4])))
 	sessions := NewSessionManager(sessionFile)
+	sessions.InvalidateForAgent(agent.Name())
 
 	ws.agent = agent
 	ws.sessions = sessions
--- a/core/multi_workspace_test.go
+++ b/core/multi_workspace_test.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


gpYang:
已 rebase 到最新 main（1e4c4b2e），分支已强制更新。原 #1729 已拆分关闭。

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Targeted fix to `getOrCreateWorkspaceAgent` that clears stale `AgentSessionID` values from persisted workspace-session files when the resolved workspace agent is of a different agent type than the one that originally produced the session ID. CI is 5/5 green on commit `3bfef3cb20deaef01ec3869acbb37d47f4ddd87b`, mergeable, base = main, no conflicts. Risk is low and the regression test is solid.

Review scope:
- Reviewed `core/engine.go` (1-line integration call), `core/session.go` (new `InvalidateForAgent` + `recordPastAgentSessionID` helper) and `core/multi_workspace_test.go` (regression test `TestGetOrCreateWorkspaceAgent_InvalidatesStaleSessionIDs`).
- Focused on correctness, concurrency, audit trail, and the regression test.

✅ What looks good:
- The invalidation predicate `s.AgentSessionID != "" && s.AgentType != "" && s.AgentType != agentType` is conservative — sessions with empty `AgentType` are left untouched, so legacy sessions that never recorded an agent type won't be wiped, only sessions that were explicitly tagged for a different agent type get cleaned.
- `recordPastAgentSessionID()` appends the cleared id to `PastAgentSessionIDs` (deduping) before zeroing `AgentSessionID`, so the old id remains queryable via `KnownAgentSessionIDs()` after the invalidation — important for users who later switch back to the original agent type.
- Thread-safety: `InvalidateForAgent` takes `sm.mu` once and locks each session individually under `s.mu`, mirroring the locking discipline used by the surrounding `SetAgentInfo`. No re-entrancy deadlock vs. the session methods it calls.
- The integration point is the right one — `getOrCreateWorkspaceAgent` is the single funnel through which every workspace agent is resolved; calling `InvalidateForAgent` immediately after the fresh `SessionManager` is constructed means the invalidation runs before the workspace first answers any `GetOrCreateActive` lookup.
- `saveLocked()` is only invoked when at least one session was actually invalidated, so the no-op path stays cheap (no churn for workspaces where the persisted file already matches the current agent type).
- The regression test exercises the full on-disk shape — it pre-seeds the workspace session file with two sessions (one stale, one same-agent), calls the real `getOrCreateWorkspaceAgent`, then asserts both the in-memory and the reloaded-from-disk view of the stale id is cleared while the same-agent id is preserved. This is the right level of coverage for a fix of this kind.

🚨/🔴 Must fix:

🟠 Should improve:
- None worth blocking on.

🔵 Optional:
- The slog log line uses `"new_agent"` to log the freshly-resolved agent type but does not include the workspace path; for a workspace session store that may contain many workspaces' worth of session files this is slightly harder to correlate than necessary. Consider adding `"workspace"` (or `"path"`) to the structured fields. Cosmetic.
- The pre-existing `recordPastAgentSessionID()` correctly skips the synthetic `ContinueSession` sentinel — that path is unaffected and worth keeping in mind for any future invariant change around `AgentSessionID`.
- The PR body's "PR overlap check" comment is a nice practice — consider also confirming in the test name or docstring that the `AgentType == ""` branch is intentionally preserved (it's covered implicitly by `keep.SetAgentSessionID("keep-opencode-session", agentName)` not being cleared when it matches, but a direct empty-type test case would document the policy).

❓ Questions:
- None.

Testing / Risk:
- New test runs end-to-end against `getOrCreateWorkspaceAgent` and a real on-disk `SessionManager`, which is the strongest level of coverage available without a full integration test rig.
- Risk scope: the change is bounded to the workspace-session store and the invalidation only runs when (a) `getOrCreateWorkspaceAgent` is invoked, (b) the workspace already had a persisted session file, and (c) at least one session in that file has both a non-empty `AgentSessionID` and an `AgentType` that differs from the current agent. Existing same-agent session IDs are preserved, sessions without an `AgentType` are preserved, and the old id is preserved in `PastAgentSessionIDs`.

Next step:

PR #1700, opened 2026-08-17
feat(relay): support async relay send

gpYang:
## Summary

- Add an async relay send mode for fire-and-return relay jobs.
- Add `/relay/send-async` to the local API; it validates the request, queues the relay job, and returns a job id with HTTP 202 while the daemon processes the relay in the background.
- Share relay request parsing and validation between the sync and async relay handlers.
- Update the `cc-connect relay send` CLI path to support async delivery.

## Verification

- `go test ./core -run 'TestHandleRelaySendAsync|TestHandleRelaySend' -count=1`
- `go test ./core -count=1`
- `GOOS=linux GOARCH=amd64 go build -tags no_web ./cmd/cc-connect`

## Overlap check

- Checked open PRs for relay async/send behavior; no existing PR implements this endpoint/CLI flag.

### DIFF
--- a/cmd/cc-connect/relay.go
+++ b/cmd/cc-connect/relay.go
@@ -28,6 +28,7 @@ func runRelay(args []string) {
 
 func runRelaySend(args []string) {
 	var from, to, sessionKey, message, dataDir string
+	async := false
 
 	var positional []string
 	for i := 0; i < len(args); i++ {
@@ -57,6 +58,8 @@ func runRelaySend(args []string) {
 				i++
 				dataDir = args[i]
 			}
+		case "--async":
+			async = true
 		case "--help", "-h":
 			printRelaySendUsage()
 			return
@@ -103,7 +106,11 @@ func runRelaySend(args []string) {
 		"message":     message,
 	})
 
-	resp, err := apiPost(sockPath, "/relay/send", payload)
+	endpoint := "/relay/send"
+	if async {
+		endpoint = "/relay/send-async"
+	}
+	resp, err := apiPost(sockPath, endpoint, payload)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		os.Exit(1)
@@ -111,19 +118,26 @@ func runRelaySend(args []string) {
 	defer resp.Body.Close()
 
 	body, _ := io.ReadAll(resp.Body)
-	if resp.StatusCode != http.StatusOK {
+	accepted := resp.StatusCode == http.StatusOK || async && resp.StatusCode == http.StatusAccepted
+	if !accepted {
 		fmt.Fprintf(os.Stderr, "Error: %s\n", strings.TrimSpace(string(body)))
 		os.Exit(1)
 	}
 
 	var result struct {
 		Response string `json:"response"`
+		Status   string `json:"status"`
+		Job      string `json:"job"`
 	}
 	if err := json.Unmarshal(body, &result); err != nil {
 		fmt.Fprintf(os.Stderr, "Error: decode response: %v\n", err)
 		os.Exit(1)
 	}
-	fmt.Print(result.Response)
+	if async {
+		fmt.Printf("queued relay job=%s\n", result.Job)
+	} else {
+		fmt.Print(result.Response)
+	}
 }
 
 func printRelayUsage() {
@@ -145,6 +159,7 @@ Options:
   -t, --to <project>         Target bot project name
   -s, --session-key <key>    Session key (auto-detected from CC_SESSION_KEY env)
   -m, --message <text>       Message to send
+      --async                Queue the relay and return immediately
       --data-dir <path>      Data directory (default: ~/.cc-connect)
   -h, --help                 Show this help
 
--- a/core/api.go
+++ b/core/api.go
@@ -1,6 +1,7 @@
 package core
 
 import (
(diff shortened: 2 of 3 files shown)


gpYang:
已 rebase 到最新 main（1e4c4b2e），分支已强制更新。原 #1729 已拆分关闭。

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Adds an async variant of the existing `/relay/send` HTTP handler and CLI flag, sharing request parsing with the sync handler. CI is 5/5 green on commit `46ddda161043f9216901decc6e3ab2f1a0ed2276`, mergeable, base = main, no conflicts. Risk is low and scoped to the relay API surface.

Review scope:
- Reviewed `cmd/cc-connect/relay.go` (`--async` flag + endpoint selection + status handling), `core/api.go` (new `handleRelaySendAsync` + shared `decodeRelayRequest`), `core/api_test.go` (two new tests), and `core/relay.go` (`RelayManager.Send` + `relayContext` helpers to confirm the goroutine context choice).
- Focused on correctness, goroutine lifecycle, request-shape parity with the sync handler, and the CLI status-code decision.

✅ What looks good:
- The shared `decodeRelayRequest` helper is the right factoring — both sync and async handlers now go through the same JSON decode + `to, session_key, and message are required` validation, which removes the risk of the two paths drifting on accepted field shapes.
- The async handler checks `s.relay == nil` and returns `503 relay not available` before queueing, matching the failure mode the sync handler would hit during send. This keeps the error shape consistent: clients see the same kind of error whether they call sync or async when relay is unconfigured.
- The HTTP method guard (`if r.Method != http.MethodPost`) is consistent with the rest of the relay handler family. The 405 path matches `handleRelayBind` / `handleRelaySend` semantics.
- Explicit choice of `context.Background()` for the goroutine and the inline comment ("The request context cannot be reused because clients may disconnect as soon as the job is accepted") call out the lifetime decision at the right spot. The `RelayManager.relayContext` helper still applies a configured `RelayManager.SetTimeout` value on top, so the async path inherits the same per-job timeout as the sync path.
- Job IDs use UnixNano + `relay-` prefix, which is unique enough for log correlation without dragging in a UUID dependency.
- The CLI parses `--async` early, routes to the right endpoint, and treats `202 Accepted` as success only when `--async` is set. Operator precedence in `resp.StatusCode == http.StatusOK || async && resp.StatusCode == http.StatusAccepted` is correct (`&&` binds tighter than `||` in Go), so the sync path never accepts a 202 from the server and the async path does.
- Help text for the new flag is included in `printRelaySendUsage`, which is the right place so `--help` stays the source of truth.
- Both new tests target `handleRelaySendAsync` directly with `httptest`, so they cover the accept-and-queue path and the validation path without spinning up a real server. The test that doesn't supply `Message` exercises the same validation message that the sync handler now also returns, so a regression in the shared helper would fail both test families.

🚨/🔴 Must fix:

🟠 Should improve:
- Goroutines spawned by `handleRelaySendAsync` are not tracked. If `APIServer.Shutdown` (or its equivalent) is added in the future, the in-flight async jobs will not be awaited. Not required for this PR (no shutdown hook exists yet), but worth a one-line comment near `go func()` calling out that fire-and-forget is intentional and will need wait-group coordination once graceful shutdown is added.
- The async endpoint accepts any well-formed `RelayRequest` and runs it in a goroutine, with no per-tenant concurrency cap. In trusted environments this is fine; if the local API ever moves to a network-reachable socket or auth boundary, a small semaphore (per-chat or global) would prevent a single noisy client from drowning the relay. Not blocking.
- The "job id" returned to the client is informational only — there is no `GET /relay/job/{id}` or equivalent, so a caller cannot recover the outcome of a queued job except via server logs. The PR description is honest about the fire-and-return semantics, so this matches the design; flagging only so a future caller isn't surprised.

🔵 Optional:
- The CLI status decision `accepted := resp.StatusCode == http.StatusOK || async && resp.StatusCode == http.StatusAccepted` works but reads ambiguously at first glance. Adding explicit parentheses around the `async && ...` sub-expression would make the precedence intent obvious to the next reader.
- `decodeRelayRequest` returns wrapped errors with `%w`, which is great for future `errors.Is` checks; consider also adding the invalid field name to the validation message (e.g. `to, session_key, and message are required (missing: message)`) so the caller can fix the payload without reading server code. Cosmetic.
- The test `TestHandleRelaySendAsync_AcceptsAndQueues` doesn't drain the spawned goroutine before the test returns. With `NewRelayManager("")` the goroutine will hit `Send` and fail early on `parseSessionKeyParts` or `relay: no binding for this chat`, but the test could explicitly close the relay or use `t.Cleanup` to wait for it so a future change to `RelayManager.Send` doesn't leak goroutines into unrelated tests. Cosmetic.

❓ Questions:
- None.

Testing / Risk:
- The two async tests exercise the validation + status-code path; the actual `RelayManager.Send` execution is not covered by these tests but is already covered by the existing `TestHandleRelaySend` family for the sync path, and `Send` itself is unchanged here.
- Risk scope is limited to the relay API surface. The shared `decodeRelayRequest` is a refactor that consolidates duplicated logic — strictly less surface than before, since the validation message and error wrapping are now defined in exactly one place. The async endpoint mirrors the sync endpoint's request shape, so client compatibility for the sync endpoint is preserved (no fields added or removed).

Next step:

PR #1701, opened 2026-08-17
fix(daemon): remove duplicate CheckLinger stub on darwin

gpYang:
## Summary
- remove the non-linux CheckLinger stub that conflicts with launchd.go on darwin
- keep the Linux linger implementation and launchd-specific stub as the only platform definitions

## Tests
- go test ./daemon -count=1
- go build -tags no_web -o /tmp/cc-connect-darwin-build-check ./cmd/cc-connect

### DIFF
--- a/daemon/check_linger_other.go
+++ b/daemon/check_linger_other.go
@@ -1,9 +1,0 @@
-//go:build !linux
-
-package daemon
-
-// CheckLinger is a stub for non-Linux platforms where systemd linger does not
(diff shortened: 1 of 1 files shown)


gpYang:
Closing as a duplicate: the same fix (removing the redundant darwin CheckLinger stub) was merged upstream as #1710 (`fix(daemon): drop redundant CheckLinger stub that breaks all non-Linux builds`, merged 2026-08-27). No further action needed here — thanks!

PR #1703, opened 2026-08-17
feat(wecom): support private WeChat Work WebSocket deployment

TongLing916:
WebSocket 模式的端点原先写死为公有云 wss://openws.work.weixin.qq.com， 无法接入私有化部署的企业微信智能机器人。新增 ws_url 配置项覆盖默认端点，
留空则回退公有云地址，行为不变。

- platform/wecom/websocket.go: WSPlatform 增加 wsURL 字段，newWebSocket 解析 ws_url（trim 后为空回退 wsEndpoint），runConnection 用 p.wsURL 拨号
- config.example.toml: wecom WebSocket 段补 ws_url 注释示例
- docs/wecom.md: 配置项表格加 ws_url 行，新增「私有化部署」章节
- websocket_test.go: 默认/自定义/空白回退三个回归用例

## Type of change

- [ ] Documentation only

- [ ] A — basic conversation

- [ ] `go build ./...` passes

## Related

- Issue:
- Related PR:

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1441,6 +1441,7 @@ app_secret = "your-feishu-app-secret"
 # mode = "websocket"
 # bot_id = "your-bot-id"
 # bot_secret = "your-bot-secret"
+# ws_url = "wss://your-wecom-host:port/im_openws?bizid=10001"  # optional / 可选：私有化部署 WebSocket 地址，留空默认公有云 wss://openws.work.weixin.qq.com
 # allow_from = "*"  # Allowed user IDs / 允许的用户 ID
 
 # Weixin personal (ilink bot) / 微信个人号（ilink 机器人网关）
--- a/config.example.toml
+++ b/config.example.toml
@@ -1445,6 +1445,7 @@ app_secret = "your-feishu-app-secret"
 # mode = "websocket"
 # bot_id = "your-bot-id"
 # bot_secret = "your-bot-secret"
+# ws_url = "wss://your-wecom-host:port/im_openws?bizid=10001"  # optional / 可选：私有化部署 WebSocket 地址，留空默认公有云 wss://openws.work.weixin.qq.com
 # allow_from = "*"  # Allowed user IDs / 允许的用户 ID
 
 # Weixin personal (ilink bot) / 微信个人号（ilink 机器人网关）
--- a/core/api_test.go
+++ b/core/api_test.go
@@ -452,10 +452,12 @@ func TestHandleCronExec_RunAliasRouteTriggersJob(t *testing.T) {
 	deadline := time.Now().Add(2 * time.Second)
 	for time.Now().Before(deadline) {
 		if len(platform.getSent()) >= 2 {
+			scheduler.WaitActiveJobs()
 			return
 		}
 		time.Sleep(10 * time.Millisecond)
 	}
+	scheduler.WaitActiveJobs()
 	t.Fatalf("timed out waiting for local api alias trigger, sent=%v", platform.getSent())
 }
 
--- a/core/cron.go
+++ b/core/cron.go
@@ -418,6 +418,7 @@ type CronScheduler struct {
 	entries            map[string]cron.EntryID // job ID → cron entry
 	defaultSilent      bool                    // global default for suppressing cron start notifications
 	defaultSessionMode string                  // global default session mode; "" = reuse, "new_per_run" = fresh session each run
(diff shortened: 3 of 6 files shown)


chenhg5 (approved):
## QA Review: PR #1703

**Conclusion: Approve ✅**

Small, surgical opt-in feature that opens up on-premise WeChat Work deployments without touching the default cloud path. The implementation is the minimum necessary diff (one new field, three parsing lines, two call sites), the tests cover exactly the three states an operator will hit, and the docs section is genuinely useful — the troubleshooting table at the bottom is the standout: it explicitly addresses the most common operator error (TOML indentation putting `ws_url` outside the right table), which is the kind of footnote that prevents a "the docs lied" bug report three months from now.

### Diff review

`platform/wecom/websocket.go` (+9/-2)
- One new field `wsURL`, three lines of parsing in `newWebSocket`, two call-site swaps in `runConnection` (the `slog.Info` and the `DialContext`). Minimal blast radius — `connectLoop`, the heartbeat, and the reconnect logic don't need to know the URL, they just receive the established connection. Right level of isolation.
- `strings.TrimSpace` + empty fallback to `wsEndpoint` is the right shape. An operator writing `ws_url = ""` or `ws_url = "   "` (whitespace-only — common when copy-pasting from templates) ends up on the cloud default, which is what they'd want anyway. Test 3 (`TestNewWebSocket_BlankWSEndpointFallsBack`) locks that in.

`platform/wecom/websocket_test.go` (+45/-0)
- Three tests, one per state: default, custom, blank-fallback. That's exactly the right level — not over-asserting on internal structure, just on the externally observable config field.
- They build on top of the existing `TestNewWebSocket_*` setup (no need to add bot_id / bot_secret plumbing; the helper already requires those).

`config.example.toml` (+1/-0)
- Single line, bilingual, with the cloud default shown inline. Right level of detail for an `example.toml`.

`docs/wecom.md` (+97/-1)
- The 「私有化部署」 section is well-structured: when-to-use, config example, item table, where to get the URL, six-step deployment (with `openssl s_client` for TLS connectivity — operator-friendly), caveats, and a troubleshooting table.
- The troubleshooting table specifically calls out "log shows `endpoint=wss://openws.work.weixin.qq.com` even though you configured `ws_url`" and points to indentation / spelling / scope. That's the right diagnostic to surface — it's the most common operator mistake and it's hard to debug from the error message alone.
- TLS certificate note (`SSL_CERT_FILE` / `SSL_CERT_DIR`) and the "同一机器人仅支持 1 个长连接" caveat are both real operational gotchas worth documenting.

### Risk

- Pure `platform/wecom/` WebSocket mode, no public API or protocol change.
- No new dependencies (`strings` already imported).
- Zero blast radius for existing wecom WS deployments: `ws_url` absent → identical behavior to before, byte-for-byte.
- Misconfiguration feedback loop is "discover at first dial attempt" rather than "fail at startup". This is consistent with how the rest of the wecom platform handles option parsing (loose, fail late), so it's not a regression — but it does mean the troubleshooting table earns its place.

### Testing / Risk

- ✅ CI: 5/5 green on `3665c67c`.
- ✅ Default behavior pinned (`TestNewWebSocket_DefaultWSEndpoint`).
- ✅ Custom value pinned (`TestNewWebSocket_CustomWSEndpoint`).
- ✅ Blank/whitespace fallback pinned (`TestNewWebSocket_BlankWSEndpointFallsBack`).
- ✅ Docs cross-checked against code: the `ws_url` description, the default endpoint, the `bizid` query parameter note — all consistent with the implementation.

- Early URL validation in `newWebSocket`: `url.Parse(s)` failure → return error; non-`ws`/`wss` scheme → return error. This would convert "learn at first dial attempt" into "fail at startup". Worth it, but inconsistent with the rest of wecom's option-parsing style — would want to apply consistently across the platform before doing it here.
- The example value in `config.example.toml` uses a placeholder host (`wss://your-wecom-host:port/im_openws?bizid=10001`). Good. Could add a `# See docs/wecom.md → 私有化部署` pointer for discoverability, but that's nice-to-have.
- Long-term: a `proxy_url` companion option for WS would help on-prem deployments where egress is restricted. Out of scope here.

### Next step

Maintainer: merge when convenient. Release-notes line would be helpful ("wecom: new `ws_url` option for on-premise WeChat Work deployments; defaults to the public cloud endpoint"). Post-merge, the cheapest confidence bump is to point a wecom deployment at a local wss:// test endpoint with a self-signed cert and confirm the cert / dial / subscribe / reconnect sequence works end-to-end.

PR #1704, opened 2026-08-17
feat(wecom): add outbound burst_limit sliding-window quota

TongLing916:
## Summary

Add a platform-local outbound rate-limiting quota to the WeChat Work (wecom) platform via two new config options — `burst_limit` and `burst_window_secs` — mirroring the existing `weixin` convention. WeChat Work throttles app messages at ~20-30 msgs/min; the wecom platform previously had no platform-local send throttling, relying solely on the engine-level token bucket (`[outgoing_rate_limit]`), which is disabled by default and cannot express a minute-level total cap.

The quota uses a sliding window that **blocks** (rather than failing fast like weixin, since wecom's limit is a cumulative frequency gate, not a retry-penalty one) until the window slides — no messages dropped. Counts per API request, so a long reply split into N chunks consumes N slots. Applied to both HTTP (`Reply`/`SendImage`) and WebSocket (`Reply`/`Send`) send paths. Disabled by default (`burst_limit=0`) for backward compatibility; stacks with the engine token bucket (per-second pacing + minute cap).

## Type of change

- [ ] Documentation only

## Testing

Verified via build, vet, targeted unit tests, race detector, and the existing wecom package suite.

- `TestSendQuota_Disabled` in `platform/wecom/sendquota_test.go` — `limit=0` returns immediately, never blocks.
- `TestSendQuota_AllowsUnderLimit` — sends under the cap pass instantly.
- `TestSendQuota_BlocksThenReleasesAfterWindow` — at cap, the call blocks; advancing the injected clock past the window releases it.
- `TestSendQuota_ContextCancel` — a blocked `wait` returns `context.Canceled` when ctx is cancelled.
- `TestSendQuota_SlidingWindowEvicts` — timestamps older than the window are evicted, allowing fresh sends.
- `TestNew_BurstLimitParsed` / `TestNew_BurstLimitCustomWindow` / `TestNew_BurstLimitDisabledByDefault` — `New()` (HTTP mode) parses `burst_limit`/`burst_window_secs`, defaults window to 60s, and is disabled by default.
- `TestNewWebSocket_BurstLimitParsed` / `TestNewWebSocket_BurstLimitDisabledByDefault` — same for WebSocket mode.
- `TestPickInt` — the int/int64/float64 TOML coercion helper.

- [x] No CUJ touched (small refactor, doc change, etc.)

This PR adds an opt-in (disabled by default) throttle inside the wecom platform send path. It does not alter any existing user-visible flow — with `burst_limit=0` (default), behavior is byte-for-byte identical to before. No CUJ test updates required.

Users can now add to a wecom platform's `[projects.platforms.options]`:

```toml
burst_limit = 20          # max outbound messages per sliding window; 0 = unlimited (default)
burst_window_secs = 60    # sliding window length in seconds (default 60)
```

With the defaults (`burst_limit = 0`) there is **no change**. When enabled, sends that would exceed `burst_limit` within the last `burst_window_secs` block (with context cancellation honored) until the window slides — no messages are dropped. This stacks with the engine-level `[outgoing_rate_limit.platforms.wecom]` token bucket (per-second pacing); the two are independent and complementary.

- [x] `go build ./...` passes
- [x] `go test ./...` passes (with `-race` if touching concurrency)
- [x] i18n strings have all-language translations (if any new user-facing text)

> Notes for reviewer:
>
> - `go build ./...`: passes. (`web/embed.go` reports a missing `all:dist` embed pattern, but that is a pre-existing, unrelated issue — this PR touches only `platform/wecom/` and `config.example.toml`.)
> - Concurrency: `go test -race ./platform/wecom/` passes clean. The `sendQuota` mutex guards the timestamp slice; the blocking `wait` uses `time.NewTimer` + `ctx.Done()` select with no shared-state access outside the lock.
> - `go vet ./platform/wecom/`: clean. (Initial draft copied a `sendQuota` value through a local var, tripping vet's lock-copy check; fixed by initializing the field directly in the struct literal.)
> - No `core/` changes — the throttle lives entirely in `platform/wecom/`. AGENTS.md notes rate limiting as a cross-cutting concern that "lives in core/", and indeed the **generic** engine-level outgoing limiter already does (`core/outgoing_ratelimit.go`, untouched here). This PR adds a **platform-local** sliding-window quota, not a new cross-cutting mechanism — mirroring the existing `weixin` `burst_limit`/`burst_window_secs` convention (`platform/weixin/weixin.go`). wecom's ~20-30 msgs/min cumulative cap is platform-specific behavior, so it belongs in `platform/wecom/` alongside the weixin precedent, not in the platform-agnostic core.
> - No new user-facing i18n strings — new error text appears only in `slog` logs and returned `error` values, not in user-visible UI.
> - `go test ./...`: wecom package passes (including `-race`). One unrelated failure exists on `main` independently of this PR: `core/TestCUJ_A5_FileReachesAgent` fails on a macOS TempDir cleanup race (`unlinkat ... directory not empty`) during session `.tmp` persistence — this PR does not touch `core/`.

## Related

- Issue:
- Related PR:

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1423,6 +1423,16 @@ app_secret = "your-feishu-app-secret"
 #             # 示例："http://你的VPS-IP:8888" — 将 VPS IP 加入可信 IP 列表
 # proxy_username = ""  # optional: proxy authentication username / 代理认证用户名
 # proxy_password = ""  # optional: proxy authentication password / 代理认证密码
+# burst_limit = 0           # Max outbound messages per sliding window; 0 = unlimited (default)
+#                            # 滑动窗口内最大出站消息数；0 = 不限（默认）
+#                            # WeChat Work throttles apps at ~20-30 msgs/min; set e.g. 20 to stay safe.
+#                            # 企业微信应用消息频率约 20-30 条/分钟；例如设 20 以留余量。
+# burst_window_secs = 60     # Sliding window length in seconds (default 60)
+#                            # 滑动窗口长度（秒，默认 60）
+#                            # Counts per API request: a long reply split into chunks consumes multiple slots.
+#                            # 按 API 请求计数：长回复被切分后每段占一个配额。阻塞等待，不丢消息。
+#                            # Stacks with [outgoing_rate_limit.platforms.wecom] (token bucket = per-second pacing).
+#                            # 与 [outgoing_rate_limit.platforms.wecom] 令牌桶叠加（令牌桶管每秒匀速）。
 
 # WeChat Work WebSocket / 企业微信长连接模式 (uncomment to enable / 取消注释以启用)
 # No public URL, no message encryption, no IP allowlist needed!
@@ -1442,6 +1452,14 @@ app_secret = "your-feishu-app-secret"
 # bot_id = "your-bot-id"
 # bot_secret = "your-bot-secret"
 # allow_from = "*"  # Allowed user IDs / 允许的用户 ID
+# burst_limit = 0           # Max outbound messages per sliding window; 0 = unlimited (default)
+#                            # 滑动窗口内最大出站消息数；0 = 不限（默认）
+#                            # WeChat Work throttles apps at ~20-30 msgs/min; set e.g. 20 to stay safe.
+#                            # 企业微信应用消息频率约 20-30 条/分钟；例如设 20 以留余量。
+# burst_window_secs = 60     # Sliding window length in seconds (default 60)
+#                            # 滑动窗口长度（秒，默认 60）
+#                            # Counts per API request: a long reply split into chunks consumes multiple slots.
+#                            # 按 API 请求计数：长回复被切分后每段占一个配额。阻塞等待，不丢消息。
 
 # Weixin personal (ilink bot) / 微信个人号（ilink 机器人网关）
 # Same HTTP API as OpenClaw @tencent-weixin/openclaw-weixin: long-poll getUpdates + sendMessage.
--- a/config.example.toml
+++ b/config.example.toml
@@ -1427,6 +1427,16 @@ app_secret = "your-feishu-app-secret"
 #             # 示例："http://你的VPS-IP:8888" — 将 VPS IP 加入可信 IP 列表
 # proxy_username = ""  # optional: proxy authentication username / 代理认证用户名
 # proxy_password = ""  # optional: proxy authentication password / 代理认证密码
+# burst_limit = 0           # Max outbound messages per sliding window; 0 = unlimited (default)
+#                            # 滑动窗口内最大出站消息数；0 = 不限（默认）
+#                            # WeChat Work throttles apps at ~20-30 msgs/min; set e.g. 20 to stay safe.
+#                            # 企业微信应用消息频率约 20-30 条/分钟；例如设 20 以留余量。
+# burst_window_secs = 60     # Sliding window length in seconds (default 60)
+#                            # 滑动窗口长度（秒，默认 60）
+#                            # Counts per API request: a long reply split into chunks consumes multiple slots.
+#                            # 按 API 请求计数：长回复被切分后每段占一个配额。阻塞等待，不丢消息。
+#                            # Stacks with [outgoing_rate_limit.platforms.wecom] (token bucket = per-second pacing).
+#                            # 与 [outgoing_rate_limit.platforms.wecom] 令牌桶叠加（令牌桶管每秒匀速）。
 
 # WeChat Work WebSocket / 企业微信长连接模式 (uncomment to enable / 取消注释以启用)
 # No public URL, no message encryption, no IP allowlist needed!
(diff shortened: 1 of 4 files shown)


chenhg5 (approved):
## QA Review: PR #1704

**Conclusion: Approve ✅**

Clean opt-in feature that mirrors how the `weixin` platform already handles outbound throttling, but with the right policy choice for `wecom` (block-until-window-slides, not fail-fast). Implementation is small, the sliding-window limiter is correct under concurrency, and the tests cover the invariants that matter: disabled-path is allocation-free, under-limit passes instantly, at-limit blocks then releases, ctx-cancel unblocks, and parsing is right for both HTTP and WebSocket modes.

### Diff review

`platform/wecom/wecom.go` (+112/-0)
- `sendQuota.wait()` is a textbook sliding-window limiter: drop out-of-window timestamps under the mutex, append `now` and return if there's room, otherwise compute `oldest + window - now` and `select` on a `time.NewTimer` vs `ctx.Done()`. Re-acquires the lock on the wake path and re-checks (no TOCTOU bug; a timer fire that races with another send taking the last slot is handled by the outer `for` loop). `waitFor <= 0` fast-path handles the case where the virtual clock has already advanced past the threshold before the goroutine wakes.
- `nowFunc = time.Now` package-level override is the standard Go testability hook and `pickInt` is a minimal TOML `int`/`int64`/`float64` coercion (silently returns 0 for string/nil — which is also what `TestPickInt` documents).
- Quota parsing correctly handles the "burst_limit > 0 with no window → default 60s" rule, and the negative-value clamps (`burstLimit < 0 → 0`, `burstWindow < 0 → 0`) are the right defensive defaults.
- `Reply()` puts `getAccessToken()` outside the chunk loop (pre-existing, but worth flagging — token is shared across chunks, so the quota on the *send* path doesn't double-count the token fetch as a slot, which would be the wrong behavior). Quota `wait()` is called per chunk after `splitByBytes`, which matches WeCom's "counts per API request" semantics.

`platform/wecom/websocket.go` (+33/-0)
- Same parsing path, same per-chunk `wait()` in `Send()` (chunked via `splitByBytes(content, 2000)`), single-shot `wait()` in `Reply()` because WS frames aren't pre-split here. Both error paths return wrapped `context.Canceled` and emit `slog.Error` with structured fields (`user`, `chunk`, `error`) — good operator ergonomics.

`config.example.toml` (+18/-0)
- Bilingual EN/ZH comment block in both the HTTP and WebSocket wecom sections. Explicitly notes the stacking with `[outgoing_rate_limit.platforms.wecom]` (token bucket = per-second pacing; this PR = per-minute cap). That's exactly the kind of footnote that prevents a future bug report from "the docs lied."

`platform/wecom/sendquota_test.go` (+261/-0, new file)
- `TestSendQuota_BlocksThenReleasesAfterWindow` is the standout: it starts a second `wait` in a goroutine, asserts the goroutine has *not* returned after 20ms (proving it actually blocks), advances the injected clock by 200ms, then asserts the goroutine *does* return within 1s. That's the right shape for a regression test on a blocking primitive — it would catch a future "fix" that accidentally turned `wait()` into a busy-loop or a no-op.
- `TestSendQuota_ContextCancel` is the matching pair: a blocked `wait` must surface `context.Canceled` so callers can stop on shutdown.
- `TestSendQuota_Disabled` does 100 calls and asserts no allocation/scheduling cost — good sanity for the fast path.
- `TestSendQuota_SlidingWindowEvicts` proves the window actually slides (send 2, advance 2s, send 2 more — all should pass). This is the test that would catch a "fix" that turned the limiter into a fixed-window or a leaky bucket.
- `TestNew_BurstLimitParsed` / `_DisabledByDefault` / `_CustomWindow` and the matching `TestNewWebSocket_*` cover both modes' parsing defaults and overrides.
- `withNow(t)` helper provides a per-test virtual clock with `t.Cleanup` to restore the original `nowFunc`. Internal `sync.Mutex` keeps `set` and the closure consistent. Clean.

### Risk

- Pure `platform/wecom/`, no public API or protocol change.
- No new dependencies (only `time` and `sync`, both already in the import set).
- `burst_limit=0` default means zero behavior change for existing deployments.
- Disabled-path is allocation-free (early return before the mutex), so no hot-path overhead for users who don't opt in.
- Per-chunk accounting is correct and matches WeCom's "counts API requests" model. A user splitting a long reply into 4 chunks at 20/min would consume 4 slots in 60s — matches what an operator would expect.

### Testing / Risk

- ✅ CI: 5/5 green on `5a597639`.
- ✅ Author local: `go build`, `go vet`, `go test`, `go test -race`, `gofmt -l` all clean on `platform/wecom/`.
- ✅ Concurrency model verified by reading: mutex protects `times` slice reads/append; sleep happens outside the lock; wake path re-acquires and re-checks.
- ✅ Token-fetch-outside-loop verified against `Platform.Reply` in head (line 563).
- ✅ Per-chunk `wait()` verified against `Platform.Reply` line 576 and `Platform.SendImage` line 613.

- The PR body doesn't call out that the quota is *platform-global* (not per-conversation or per-user). That's worth a footnote in the docs — a single noisy conversation can starve other users on the same wecom agent. Not blocking; just a note for future per-conversation quotas.
- A metrics/observability hook (current count, total waits, peak wait time) would make operator tuning much easier. Out of scope here, but cheap to add later via a `Stats() sendQuotaStats` accessor.
- `pickInt` silently returns 0 for string inputs. The test case documents this as intended, but a config-typo scenario like `burst_limit = "20"` (quoted string in TOML) would silently disable the quota. A one-line warning log on unexpected types would help operators notice.

### Next step

Maintainer: merge when convenient. A release-notes line ("wecom: new `burst_limit` / `burst_window_secs` options for sliding-window outbound throttling, disabled by default") would be helpful for operators who hit the WeCom 20-30 msgs/min limit. Post-merge, the cheapest confidence bump is a manual smoke: set `burst_limit = 2, burst_window_secs = 60`, send a 5-chunk reply, and confirm chunks 1-2 send immediately and chunks 3-5 wait out the window before going out.

PR #1706, opened 2026-08-18
fix(core): preserve queued messages after recall

AaronZ345:
## Summary

- preserve messages queued after an active message is recalled
- resume the oldest surviving message in a replacement agent session while retaining FIFO order
- avoid treating an intentionally closed agent event channel as a crash that drops the queue

## Root cause

The silent recall path cleared the entire `pendingMessages` queue when stopping the recalled turn. Messages sent after the recalled message were therefore accepted and acknowledged, then discarded without a response.

## Tests

- `go test -race ./core -run "TestCUJ_A8_RecallPreservesLaterQueuedMessage|TestHandleMessageRecallPreservesLaterQueueWhenAgentChannelCloses" -count=1`
- `go test ./core -count=1`
- `go test -tags no_web ./... -count=1`

## Maintenance

- Supersedes #1569 after the 30-day maintenance refresh.
- Carries the same patch on the latest `main`.

### DIFF
--- a/core/cuj_test.go
+++ b/core/cuj_test.go
@@ -1230,6 +1230,97 @@ func TestCUJ_A6_A7_CoveredByPlatformLayer(t *testing.T) {
 }
 
 // ===========================================================================
+// CUJ-A8 · Recalling one queued message must not discard messages sent after it
+// ===========================================================================
+
+func TestCUJ_A8_RecallPreservesLaterQueuedMessage(t *testing.T) {
+	env := newCUJEnv(t)
+	key := "test:recall-user"
+
+	// Keep the first turn busy long enough to queue two later messages. The
+	// first queued message will be recalled after it becomes the active turn;
+	// the second queued message must survive that cancellation boundary.
+	env.agent.setNextSessionEvents([]Event{{Type: EventResult, Content: "first reply", Done: true}}, 250)
+	env.engine.ReceiveMessage(plat(env.plat), &Message{
+		SessionKey: key,
+		Platform:   "test",
+		MessageID:  "msg-first",
+		UserID:     "recall-user",
+		UserName:   "recall-user",
+		Content:    "first long-running request",
+		ReplyCtx:   "ctx-first",
+	})
+
+	env.waitFor("first agent session starts", 2*time.Second, func() bool {
+		env.agent.mu.Lock()
+		defer env.agent.mu.Unlock()
+		return len(env.agent.sessions) == 1
+	})
+	env.agent.mu.Lock()
+	firstSession := env.agent.sessions[0]
+	env.agent.mu.Unlock()
+	firstSession.mu.Lock()
+	firstSession.delayMs = 5_000
+	firstSession.mu.Unlock()
+
+	// B and C both arrive while A is busy, so they are accepted in FIFO order.
+	env.engine.ReceiveMessage(plat(env.plat), &Message{
+		SessionKey: key,
+		Platform:   "test",
+		MessageID:  "msg-recalled",
+		UserID:     "recall-user",
+		UserName:   "recall-user",
+		Content:    "this queued request will be recalled",
+		ReplyCtx:   "ctx-recalled",
+	})
+	env.engine.ReceiveMessage(plat(env.plat), &Message{
+		SessionKey: key,
+		Platform:   "test",
+		MessageID:  "msg-after-recall",
+		UserID:     "recall-user",
+		UserName:   "recall-user",
+		Content:    "answer this message after the recall",
+		ReplyCtx:   "ctx-after-recall",
+	})
+
+	// Wait until B is actually active, then deliver the platform recall event.
+	// This pins the same path as the Feishu fallback probe in the live incident.
+	env.waitFor("recalled queued message becomes active", 2*time.Second, func() bool {
+		env.engine.interactiveMu.Lock()
+		state := env.engine.interactiveStates[key]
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
## QA Review: PR #1706

### Overall assessment
Well-designed bugfix for a real user-visible data-loss bug: recalling a message that was already queued in cc-connect silently dropped every later queued message in the same session. The fix removes two silent-wipe code paths, adds a `state.isStopped()` guard in `channelClosed:` so intentional stops aren't misclassified as crashes, and introduces a new helper `resumePendingAfterStoppedTurn` that explicitly transfers the preserved queue + lock ownership to a replacement agent session via a recursive re-entry. Test coverage is comprehensive at both unit and CUJ levels.

### Review scope
- `core/engine.go` (+81/-8), `core/engine_test.go` (+68/-0), `core/cuj_test.go` (+91/-0)
- +240/-8 in 3 files
- Focus: correctness of queue preservation across the recall boundary, lock-ownership transfer on re-entry, FIFO ordering, regression net quality, scope discipline

### ✅ What looks good
- **Three coordinated changes form one coherent fix**:
  1. New helper `resumePendingAfterStoppedTurn` snapshots the queue, clears it, merges in any messages that arrived during the transition window, and re-enters `processInteractiveMessageWith` with the oldest preserved message.
  2. The two `pendingMessages = nil` wipe sites in `stopInteractiveSessionWithOptions` (silent recall) and `normalCleanup` are deleted — those were the data-loss source.
  3. `processInteractiveEvents` `channelClosed:` now returns early when `state.isStopped()` so intentional stops aren't treated as crashes and drained twice.
  Each change alone would be incomplete; together they form the minimum surface area for the fix.
- **FIFO order preserved under transition**: `pending = append(pending, placeholder.pendingMessages...)` then re-setting `placeholder.pendingMessages = pending[1:]` keeps the snapshot + the new arrivals in arrival order. The state-detach window (where new messages can land on a placeholder) is handled defensively without dropping or reordering.
- **Lock ownership transfer is explicit**: the call site sets `unlocked = true` and returns, with a clear comment that the nested processor now owns the session lock. This is the right pattern for re-entry — the alternative (releasing then re-acquiring) would race with new arrivals.
- **Test design matches the production code path**:
  - `TestCUJ_A8_RecallPreservesLaterQueuedMessage` drives the public engine API (`ReceiveMessage`) end-to-end, exactly mimicking the Feishu recall flow that produced the original report.
  - `TestHandleMessageRecallPreservesLaterQueueWhenAgentChannelCloses` covers the agent-channel-close path that the `state.isStopped()` guard protects.
  - Both assert on `replacementSession.sentPrompts` to confirm only the *later* queued message reaches the replacement — guarding against accidentally re-feeding the recalled message.
- **Logging on resume**: `slog.Info("resuming queued messages after stopped turn", "session", ..., "next_msg_id", ..., "remaining_queue", ...)` is the right level for a state-machine transition that's notable but not an error. Operators can grep for this to validate recall behavior in production.

- **Document the semantic shift**: deleting `pendingMessages = nil` is a behavior change for *every* `stopInteractiveSessionWithOptions(notifyQueued=false)` and `normalCleanup` caller, not just recall. Current scan suggests only recall uses this path, but worth a one-line note in the PR description (or follow-up issue) confirming the audit, so future contributors don't add a silent-stop caller that expects the old wipe behavior.
- **`isStopped()` coverage check**: the new `state.isStopped()` early-return in `channelClosed:` only helps if `isStopped()` returns true for every *intentional* stop. Worth verifying that any future intentional-stop entry point also flips `state.stopped = true` *before* closing the agent channel, otherwise the bug returns through a different path. Worth a brief audit comment in the helper.

- **CHANGELOG / release notes**: `fix(core): preserve queued messages after recall (#1706, supersedes #1569)` is the right wording for the next release.
- **Recursion depth**: `resumePendingAfterStoppedTurn` re-enters `processInteractiveMessageWith`, which can re-enter itself on a chain of recalls. In practice depth is bounded by user behavior, but a defensive log warning when depth exceeds some threshold (say 5) would help future debugging. Optional.
- **CUJ catalog**: if `docs/cuj.md` exists, record CUJ-A8's name and trigger condition there for discoverability.

### 🚨 / 🔴 Must fix

### ❓ Questions
None.

### Testing / Risk
- **Author-validated**: `go test -race ./core -run "TestCUJ_A8_RecallPreservesLaterQueuedMessage|TestHandleMessageRecallPreservesLaterQueueWhenAgentChannelCloses"`, `go test ./core -count=1`, `go test -tags no_web ./... -count=1` — all per PR description
- **QA-verified**: diff inspected; the three changes are coherent; lock-ownership comments match the actual control flow; FIFO ordering invariant is enforced by the snapshot-then-rebuild pattern; the channel-closed guard is short-circuited before the drain logic that would have wiped state
- **Remaining risk**: low. The behavior change for non-recall `notifyQueued=false` callers (P2 above) is the only one worth a final scan; if audit finds another caller expecting wipe behavior, that caller would need to opt in explicitly (e.g., pass a new flag or call a separate helper). The fix itself does not introduce new attack surface or performance risk.

### Next step
Maintainer can merge when convenient. The P2 caller audit is the most valuable follow-up — should take 5 minutes with `grep` and either confirm scope is fine or spawn a follow-up to add an opt-out for any non-recall silent-stop caller.

AaronZ345:
Closing under the explicit age_30d maintenance policy after this PR crossed 30 days. Superseded by #1886, which carries the same approved patch on the latest main.

PR #1709, opened 2026-08-19
feat(cursor): deliver image attachments to the Cursor CLI via on-disk paths

Niko-Wu:
## Problem

`cursorSession.Send()` discarded every image:

```go
if len(images) > 0 {
    slog.Warn("cursorSession: images not yet supported in CLI mode, ignoring")
}
```

The drop is silent from the agent's point of view — no placeholder, nothing
appended to the prompt. A user who sends a screenshot plus "what's in here?"
gets an answer that quietly resolves the pronoun against unrelated
conversation history, which reads as a confident hallucination.

## Why it can work

The Cursor Agent CLI has no image flag, but it **can open image files from disk
and see their content**. Verified directly:

```
$ agent --print --workspace <dir> -- "read this image and quote the text: /tmp/shot.png"
```

returns an accurate transcription of the screenshot, watermarks included.

## Change

Route images through the same on-disk channel the file attachments already
use (`core.SaveFilesToDisk` + a path reference appended to the prompt).

Feishu image messages carry only bytes + MIME and **no filename**, so a name is
synthesised with an extension derived from the MIME type. Without this the
attachment lands as an extension-less blob and the agent has to guess the
format.

Two small helpers move into `core` so both backends share them:

- `core.ExtFromMime` — lifted out of `agent/claudecode` (was private, now shared;
  claudecode calls the shared one, behaviour unchanged, plus `image/bmp`)
- `core.AppendImageRefs` — like `AppendFileRefs`, but the wording tells the
  agent the paths are images the user attached

## Tests

- `core/message_image_test.go` — MIME→extension mapping, prompt wording, and
  that an image without a filename still lands with a usable extension
- `go test ./core/... ./agent/...` passes

## End-to-end

Running this build behind Feishu + cursor: sending a screenshot now logs the
attachment instead of the warning, and the agent reads the image content
correctly.

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -962,7 +962,7 @@ func (cs *claudeSession) Send(prompt string, messageID string, images []core.Ima
 
 	// Save and encode images
 	for i, img := range images {
-		ext := extFromMime(img.MimeType)
+		ext := core.ExtFromMime(img.MimeType)
 		fname := fmt.Sprintf("img_%d_%d%s", time.Now().UnixMilli(), i, ext)
 		fpath := filepath.Join(attachDir, fname)
 		if err := os.WriteFile(fpath, img.Data, 0o644); err != nil {
@@ -1008,19 +1008,6 @@ func (cs *claudeSession) Send(prompt string, messageID string, images []core.Ima
 		"type":    "user",
 		"message": map[string]any{"role": "user", "content": parts},
 	})
-}
-
-func extFromMime(mime string) string {
-	switch mime {
-	case "image/jpeg":
-		return ".jpg"
-	case "image/gif":
-		return ".gif"
-	case "image/webp":
-		return ".webp"
-	default:
-		return ".png"
-	}
 }
 
 // RespondPermission writes a control_response to the Claude process stdin.
--- a/agent/cursor/session.go
+++ b/agent/cursor/session.go
@@ -77,7 +77,25 @@ func newCursorSession(ctx context.Context, cmd string, extraArgs []string, workD
 
 func (cs *cursorSession) Send(prompt string, messageID string, images []core.ImageAttachment, files []core.FileAttachment) error {
 	if len(images) > 0 {
-		slog.Warn("cursorSession: images not yet supported in CLI mode, ignoring")
+		// The Cursor Agent CLI has no image flag, but it can open image files
+		// from disk and see their content. So route images through the same
+		// on-disk channel as files instead of dropping them.
+		imgFiles := make([]core.FileAttachment, 0, len(images))
+		for i, im := range images {
+			name := im.FileName
+			if name == "" {
+				// Feishu image messages carry no filename — synthesise one so
+				// the file lands with a correct extension.
+				name = fmt.Sprintf("img_%d_%d%s", time.Now().UnixMilli(), i, core.ExtFromMime(im.MimeType))
+			}
+			imgFiles = append(imgFiles, core.FileAttachment{
+				MimeType: im.MimeType,
+				Data:     im.Data,
+				FileName: name,
+			})
+		}
+		imagePaths := core.SaveFilesToDisk(cs.workDir, messageID, imgFiles)
+		prompt = core.AppendImageRefs(prompt, imagePaths)
 	}
 	if len(files) > 0 {
 		filePaths := core.SaveFilesToDisk(cs.workDir, messageID, files)
--- a/core/message.go
+++ b/core/message.go
@@ -341,6 +341,50 @@ func AppendFileRefs(prompt string, filePaths []string) string {
 	return prompt + "\n\n(Files saved locally, please read them: " + strings.Join(abs, ", ") + ")"
 }
 
+// ExtFromMime maps an image MIME type to a file extension. Attachments that
+// arrive without a filename (Feishu image messages carry only bytes + MIME)
+// must still land on disk with a usable extension, otherwise downstream agents
+// have to guess the format from the bytes.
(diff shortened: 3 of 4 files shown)


chenhg5 (approved):
## QA Review — PR #1709 (Cursor image attachments via on-disk paths)

**Conclusion:** Approve

**Overall assessment:** Small, well-scoped fix that turns a silent data-loss bug into a clean on-disk delivery, plus a tidy shared-helper refactor. CI is 5/5 SUCCESS, merge state CLEAN, head `2d66bae`. The diff is conservative: it doesn't change `cursorSession.Send`'s CLI invocation, doesn't widen the input surface, and lifts the MIME → extension helper into `core` so other backends can reuse it.

**Review scope:**
- `agent/cursor/session.go`: +19/-1 (image→file routing + path append)
- `agent/claudecode/session.go`: +1/-14 (use shared `core.ExtFromMime`)
- `core/message.go`: +44 (new `ExtFromMime`, `AppendImageRefs`)
- `core/message_image_test.go`: +56 (3 tests)
- Focused on: silent-loss bug fix, MIME handling, path correctness, test coverage.

**✅ What looks good:**

- **Root-cause fix is honest.** The pre-fix code was `slog.Warn("...images not yet supported in CLI mode, ignoring")` — silent to the user, who would see a confident answer that resolved the image reference against unrelated conversation history. The fix routes images through the same `core.SaveFilesToDisk` channel that file attachments already use, and appends a path reference via the new `core.AppendImageRefs` so the Cursor CLI sees the file on disk and reads it.
- **Verified that the CLI can do this.** Author tested directly with `agent --print --workspace <dir> -- "read this image and quote the text: /tmp/shot.png"` and got an accurate transcription. This is exactly the kind of "do the minimum viable thing the CLI already supports" approach I'd want for an image-delivery bug.
- **Feishu filename synthesis** is the right call: image messages carry bytes + MIME only, so a name like `img_<ms>_<i><ext>` is synthesised. `<ext>` comes from the shared `ExtFromMime` so the agent CLI doesn't have to sniff bytes for format.
- **Shared `core.ExtFromMime` refactor.** The previous private `extFromMime` in `agent/claudecode/session.go` was a textbook copy-paste candidate; promoting it to `core` with one extra case (`image/bmp`) is the lowest-friction way to give Cursor the same coverage. Behaviour for claudecode is unchanged except for the `.bmp` case, which is a strict expansion.
- **`AppendImageRefs` is distinct from `AppendFileRefs`** with image-specific wording ("the user attached image(s), saved locally — read the file(s) to view them"). Different prose is the right call: agents that handle both should not confuse image and non-image attachments.
- **Test coverage** covers the three real edge cases:
  1. `TestExtFromMime` pins every MIME mapping including the `"" → .png` fallback.
  2. `TestAppendImageRefs` pins empty paths (no spurious wording) and image-specific wording.
  3. `TestSaveFilesToDisk_ImageWithoutFileName` covers the Feishu-specific "no filename" path.

- **`AppendImageRefs` runs `filepath.Abs` per path even when `filepath.IsAbs` already passed** for absolute inputs — actually the diff correctly short-circuits via `if filepath.IsAbs(p)` first, so this is fine. Re-reading, no concern.
- The cursor session synthesises image filenames with `fmt.Sprintf("img_%d_%d%s", time.Now().UnixMilli(), i, core.ExtFromMime(im.MimeType))`. Within a single millisecond, two images at indices `i` and `i+1` are unique; across two `Send()` calls in the same millisecond, the filenames collide and `SaveFilesToDisk` will overwrite. Per the existing scope-bound behaviour of `SaveFilesToDisk` (which scopes by `messageID`), if `messageID` is non-empty this is fine; if `messageID` is empty the cursor path falls back to a global directory. Worth a quick check that Cursor's `Send` always passes a non-empty `messageID`, or note this as a follow-up. (Looking at the cursor session code, `messageID` is a parameter and `SaveFilesToDisk` requires it for unique names; the call site is `core.SaveFilesToDisk(cs.workDir, messageID, imgFiles)` so it's the caller's responsibility.)
- **Cursor CLI image format support is not enumerated in the PR body.** The PR assumes the CLI can read PNG/JPG/GIF/WebP/BMP from disk; if the user's `cursor-agent` version is older, MIME types like `image/avif` would land with `.png` extension and silently not render. Consider documenting the minimum supported Cursor CLI version in the PR body or in `agent/cursor/AGENTS.md` if one exists. Minor.

- A test that exercises the new path via `cursorSession.Send` end-to-end (mocking the CLI invocation) would lock in the fix. The existing `agent/cursor` test suite is currently absent from the diff; if there's no CI infrastructure for it, this is a separate problem.

**❓ Questions:** None blocking.

**Testing / Risk:**

- Verified evidence: CI 5/5 SUCCESS (`lint`, `unit-test`, `smoke-test`, `regression-test`, `performance-test`) on head `2d66bae`. Author confirms `go test ./core/... ./agent/...` green.
- End-to-end: author tested behind Feishu + cursor; screenshot now logs the attachment instead of the warning, agent reads the image correctly.
- Remaining risk:
  1. **Format coverage** — agents using older Cursor CLIs may not support all MIME types the helper maps; PNG is the safe fallback. Low risk.
  2. **Filename collision** — addressed by `messageID`-scoped save path. If a future caller passes an empty `messageID`, the collision window opens. Pre-existing behaviour, not introduced by this PR.
  3. **Disk footprint** — images now persist on disk rather than being dropped. The existing `cleanAttachments` lifecycle should clean them up; worth confirming in a follow-up if it doesn't already.

**Next step:** Recommend merge. The fix is small, root-cause clear, behaviour strictly improved (silent drop → readable image), and the shared-helper refactor prevents future drift.

PR #1710, opened 2026-08-19
fix(daemon): drop redundant CheckLinger stub that breaks all non-Linux builds

Niko-Wu:
## Problem

`main` does not compile on macOS (or Windows):

```
# github.com/chenhg5/cc-connect/daemon
daemon/launchd.go:32:6: CheckLinger redeclared in this block
	daemon/check_linger_other.go:7:6: other declaration of CheckLinger
```

`daemon/check_linger_other.go` declares `CheckLinger` under `//go:build !linux`,
but each platform already has its own definition:

| file | build tag |
| --- | --- |
| `daemon/systemd.go` | `linux` |
| `daemon/launchd.go` | `darwin` |
| `daemon/windows.go` | `windows` |
| `daemon/unsupported.go` | `!linux && !darwin && !windows` |

`!linux` overlaps darwin, windows **and** the catch-all case, so the package
only compiles on Linux. `v1.5.0` builds fine; the stub was added afterwards.

## Fix

The stub is fully redundant — `unsupported.go` already covers every platform
that lacks a native implementation. Removing the file restores the build.

## Verification

`go build -tags no_web ./cmd/cc-connect`

- darwin/arm64 — OK
- darwin/amd64 — OK
- linux/amd64 (`CGO_ENABLED=0`) — OK

Unrelated pre-existing issue spotted while checking windows/amd64, not touched
by this PR: `agent/pi/proc_windows.go:6:2: "os" imported and not used`.

## Why not just narrow the build tag?

Tested: retagging the stub to `!linux && !darwin && !windows` simply moves the collision to `daemon/unsupported.go`, which carries the same tag and already defines `CheckLinger`:

```
daemon/unsupported.go:15:6: CheckLinger redeclared in this block
	daemon/check_linger_other.go:7:6: other declaration of CheckLinger
```

Since the four files above partition every GOOS, there is **no possible build tag** under which this stub compiles anywhere — deletion is the only fix. (Semantics also favor `unsupported.go`: it returns `false` where the stub returns `true`.)

Related: #1738 fixes the remaining windows-only break in `agent/pi`; with both, darwin/arm64, darwin/amd64, windows/amd64, linux/amd64 and freebsd/amd64 all compile cleanly.

### DIFF
--- a/daemon/check_linger_other.go
+++ b/daemon/check_linger_other.go
@@ -1,9 +1,0 @@
-//go:build !linux
-
-package daemon
-
-// CheckLinger is a stub for non-Linux platforms where systemd linger does not
-// apply. Returns (true, "") so callers skip the linger warning.
-func CheckLinger() (enabled bool, user string) {
-	return true, ""
-}


chenhg5 (approved):
## Conclusion: Approve

### Overall assessment
A clean removal of a redundant stub that was breaking non-Linux builds. The deletion is small, the build-tag analysis is exactly right, and the fix is verified by local cross-platform builds across linux/darwin/windows.

### Review scope
- Reviewed `daemon/check_linger_other.go` (deletion, 9 lines).
- Verified the build-tag coverage map: `systemd.go` (linux), `launchd.go` (darwin), `windows.go` (windows), `unsupported.go` (`!linux && !darwin && !windows`).
- Cross-compiled `./daemon/...` and `./cmd/cc-connect` for `linux/amd64`, `darwin/amd64`, `windows/amd64` on the PR head.
- Cross-compiled `./daemon/...` on the PR head + current `origin/main` merge.

### Verification
- CI: 5/5 green — [run 32230897425](https://github.com/chenhg5/cc-connect/actions/runs/32230897425).
- Local cross-build (`GOOS={linux,darwin,windows} GOARCH=amd64 CGO_ENABLED=0 go build ./daemon/...`): all three pass clean.
- Local cross-vet (`go vet ./daemon/...`): clean on all three.
- `go test -race -count=1 ./daemon/` on PR head: ok (1.02s).
- `go test -race -count=1 ./daemon/` on `git merge origin/main` tree: ok (1.03s).
- Merge with current `origin/main` (`ab85b513`): auto-merges cleanly — no conflict.

### ✅ What looks good
- The PR message's build-tag audit is correct and verifiable: every platform is already covered by a dedicated file. `unsupported.go` (`!linux && !darwin && !windows`) provides the catch-all, so the `!linux` stub genuinely adds nothing.
- Minimal blast radius: a single file deletion. No collateral edits, no refactors.
- The deletion does not regress linux builds — `systemd.go` (linux-tagged) remains the authoritative `CheckLinger` for linux. Cross-build verified.

### 🚨/🔴 Must fix
- None. No blocker.

### 🟠 Should improve
- None material.

### 🔵 Optional
- P3 (unrelated to this PR, but worth noting in case the maintainer wants to address it): there is a **pre-existing Windows build break** in `cmd/cc-connect` caused by an unused `"os"` import in `agent/pi/proc_windows.go`. I confirmed the same error exists on `origin/main` before #1710 is applied, so this PR does not cause it. Fix is one line: remove `"os"` from the import block. PR #1738 in the QA queue does exactly this. Worth merging #1710 first (it's the simpler fix and unblocks the linux-only CI path), then #1738, to restore full cross-platform CI for the cmd binary.

### ❓ Questions
- None.

### Testing / Risk
- Verified evidence: CI 5/5 green, local cross-build + cross-vet + cross-test on linux/darwin/windows, auto-merge with current `main` clean.
- Remaining risk: very low. The deleted function returned `(true, "")` — i.e., it told callers "linger is fine, skip the warning." On darwin and windows, the actual platform-specific files (`launchd.go`, `windows.go`) provide their own `CheckLinger` implementations (or fall through to `unsupported.go`). For the "everything else" tier (`!linux && !darwin && !windows`, e.g. freebsd/openbsd/solaris), `unsupported.go` already covers it. No platform loses its `CheckLinger` after this deletion.

### Next step
- Merge when ready. This is a clean, low-risk fix.

PR #1714, opened 2026-08-20
fix(claudecode): bound session teardown — WaitDelay, kill retry, close/spawn race guard, fast graceful stop

Chendounai:
## Problem

On Windows, `/stop` on a Claude Code session frequently failed to actually stop the agent. In production (5 bridged projects, Feishu) this produced three user-visible symptoms:

1. The stopped session kept running for ~9 more minutes, writing to production data and sending messages, **while a new session was already live** — two identities talking at once.
2. Even when teardown eventually succeeded, `/stop` took **2m09s**, long enough that users concluded it was broken and pressed it again.
3. The failure was silent — nothing was reported back to the user.

Root-causing this took four rounds. The findings below are ordered by how much they actually mattered.

## Root cause 1 — `cmd.Wait()` never returns (the real one)

`agent/claudecode/session.go` attaches stderr to a `*bytes.Buffer` rather than an `*os.File`. `os/exec` therefore creates a pipe plus an `io.Copy` goroutine, and **`cmd.Wait()` blocks until that goroutine finishes**, which requires EOF on the pipe.

Claude Code spawns MCP servers as grandchildren. Any one of them inheriting the stderr write handle keeps the pipe open forever ⇒ `Wait()` never returns ⇒ `cs.done` never closes ⇒ `Close()` can only time out reporting "process tree still alive".

This is why `taskkill /T /F` reporting success and `Close()` reporting "still alive" **can both be true at the same time** — they are answering different questions. Two of our four debugging rounds were spent improving the kill path, which was never the problem.

Fix: `cmd.WaitDelay = 3 * time.Second`.

Worth noting: **`gemini`, `kimi`, `antigravity` and `hooks` all already set `WaitDelay`** in this repo (the comment in the gemini adapter even says it is there to stop the I/O goroutine blocking) — `claudecode` is the only adapter that was missing it. 3s rather than 1s because Claude Code can still be emitting a panic trace on exit, and the surrounding `Close()` backstop is 10s.

## Root cause 2 — close/spawn race lets two processes share one session

`stopInteractiveSessionWithOptions` deleted the state from the map **synchronously** while closing the process **asynchronously** (graceful window up to 2 minutes). A message arriving inside that window found an empty map and resumed the same `agent_session` id ⇒ two live processes appending to the same jsonl.

Note this does **not** require the kill to fail. A perfectly healthy teardown still exposes the window; a failed kill merely makes it permanent.

Fix:
- `closingSessions map[string]chan struct{}` registers in-flight teardowns. Registration happens **in the caller's goroutine**, otherwise there is still a gap between `/stop` returning and the goroutine being scheduled.
- `getOrCreateInteractiveStateWith` calls `awaitSessionClose(sessionKey)` **before taking `interactiveMu`** (waiting while holding the lock deadlocks the engine).
- `unsafeResume map[string]bool`: if Close errors, times out, or is abandoned, the next spawn is forced to start fresh instead of resuming. Losing history is strictly better than two agents fighting over one context.

## Root cause 3 — 120s spent waiting for a hook that does not exist

`gracefulStopTimeout` was hardcoded to 120s with a comment explaining it accommodates claude-mem's Stop hook. On a machine with no Stop hook configured at all, every `/stop` still paid the full 120s before escalating.

Fix: `defaultGracefulStopTimeout = 5 * time.Second`, with a comment warning that this must be revisited if Stop hook support is added.

## Root cause 4 — unbounded SIGKILL phase, and silent failure

`Close()`'s phase-3 was a single `taskkill` attempt followed by a bare `<-cs.done` with no timeout — the goroutine leaked and nobody was told.

Fix: retry the kill 3x at 2s intervals with a 10s final bound, return an error naming the PID, and surface teardown failures to the user via the existing `Platform.Send` path (`notifySessionCloseFailure`, new `MsgSessionCloseFailed` string in all five languages).

## Tests

`agent/claudecode/waitdelay_test.go`, 5 cases:

- `TestWithoutWaitDelayWaitHangsOnInheritedPipe` — control: reproduces the hang.
- `TestWaitDelayBoundsInheritedPipe` — treatment: bounded return.
- `TestProductionSessionSetsWaitDelay` — guards the production line itself; without it the first two stay green even if the fix is deleted.
- `TestGracefulStopTimeoutIsShortEnoughForUserStop`
- `TestIdleSessionExitsWithinGracefulWindow`

Reproduction note that cost us time: scaffolding the grandchild with `start /b` does **not** inherit handles, so `Wait()` returned in 0.03s and the hypothesis looked wrong. `Start-Process -NoNewWindow` inherits properly and reproduces immediately (>12s hang without the fix, 5s return with it).

## Verification

`go test ./agent/claudecode/ -run 'WaitDelay|GracefulStop|IdleSession'` — all 5 pass.

`go test ./core/` on Windows: this branch fails 4, pristine `main` fails 5, and the failing set rotates between runs (`TestCUJ_A3_ImageReachesAgent`, `TestCUJ_H2_TwoPlatformsConcurrentNoBleed`, `TestCUJ_A5_FileReachesAgent` swap places). `TestCompactReplyFooterPath_HomeRelativeDeepPathStaysFull` and both `TestAppendFileRefs_*` fail identically on pristine `main` — they look like path-semantics assumptions that do not hold on Windows. None of these are affected by this PR; verified by running the same suite in a clean `git worktree` of upstream `main`.

## Unrelated: `main` currently does not compile on Windows

While preparing this branch, `go build -tags no_web ./...` on **pristine `origin/main`** fails with:

```
daemon\windows.go:273:6: CheckLinger redeclared in this block
        daemon\check_linger_other.go:7:6: other declaration of CheckLinger
agent\pi\proc_windows.go:6:2: "os" imported and not used
```

`check_linger_other.go` appears to be missing a `//go:build !windows` constraint. Not touched here to keep this PR focused — happy to send a separate one if useful.

## Notes

These four rounds have been running in production since 2026-07-31 on v1.4.1, across 5 bridged projects; the symptoms have not recurred. This branch is that work rebased onto current `main` (clean, no conflicts).

The one behaviour change reviewers may want to weigh in on is `gracefulStopTimeout` 120s → 5s. If the claude-mem Stop hook case is important to preserve, making it configurable rather than a constant would be the better call and I'm glad to rework it that way.


chenhg5 (commented):
## QA Review — PR #1714 (claudecode /stop teardown)

**Conclusion:** Approve with one open question for the maintainer

**Overall assessment:** This is the most substantive bugfix in the current batch — a real production debugging writeup with 4 root causes, each with a measured symptom, a clearly-named fix, and explicit trade-off comments. CI is 5/5 SUCCESS on head `ae62d95`, merge state CLEAN. The diff is well-structured but touches a critical concurrency surface (`Engine` lifecycle maps), so the maintainer should ack the open question before merge.

**Review scope:**
- `agent/claudecode/session.go`: +90/-5 (WaitDelay + gracefulStopTimeout + SIGKILL retry loop)
- `agent/claudecode/waitdelay_test.go`: +131 (5 new tests including a meta-test guarding the production line itself)
- `core/engine.go`: +213/-15 (closingSessions / unsafeResume lifecycle + user-facing failure notification)
- `core/engine_test.go`: +19/-3
- `core/i18n.go`: +16 (MsgSessionCloseFailed, 5 languages)
- Focused on correctness of the wait delay fix, the close/spawn race guard, the kill retry logic, and the new sync surface in `Engine`.

**✅ What looks good:**

- **Root cause analysis is honest and reproducible.** The fact that `start /b` doesn't reproduce the hang while `Start-Process -NoNewWindow` does is exactly the kind of detail reviewers need; the waitdelay_test.go file documents it at the top and the `TestProductionSessionSetsWaitDelay` meta-test prevents the fix from being silently reverted (the author calls this out explicitly — same shape of bug as the "tool works but isn't wired up" pattern from amazon-monitor).
- **`cmd.WaitDelay = 3 * time.Second`** matches the convention of the other 4 adapters in the same repo (gemini/kimi/antigravity/hooks all use 1s; 3s is justified because Claude Code may still be emitting a panic trace on exit). Consistency with existing adapters is a strong correctness signal here — claudecode was the outlier, the fix brings it back in line.
- **Synchronisation order is correct.** `beginSessionClose` registers *before* the goroutine is scheduled, and `awaitSessionClose` is called *before* taking `interactiveMu`. The author's explicit warning ("spawning now would start a second process on the same agent session ID while the first one is still alive and writing to the same transcript") shows they thought through both the deadlock and the racy-registration failure modes.
- **`unsafeResume` + `MsgSessionResumeUnsafe`** is the right fallback — losing conversation history is strictly better than two agents fighting over the same jsonl transcript. Honest engineering.
- **Failure now surfaces to the user** via `notifySessionCloseFailure` and `MsgSessionCloseFailed`. The pre-fix "silent failure for 2m09s while the bot keeps running" was the most user-hostile behaviour in this batch; surfacing it is a real UX fix.
- **Production-validated.** 5 bridged projects, 2026-07-31 → today, symptoms not recurring. That's stronger evidence than any CI run.
- **The unrelated `origin/main` compile errors** (`CheckLinger` redeclared, `os` imported and not used in `agent/pi/proc_windows.go`) are correctly flagged but not touched in this PR. Keep them out of scope.

**🟠 Should improve (non-blocking but worth a maintainer ack):**

- **`gracefulStopTimeout` 120s → 5s** is a behaviour change, not just a bug fix. The author explicitly invites the question: *"If the claude-mem Stop hook case is important to preserve, making it configurable rather than a constant would be the better call and I'm glad to rework it that way."* On the maintainer's machine there is no Stop hook, so 5s is correct for this deployment. For other users (anyone running claude-mem or any future Stop hook), this will silently truncate their hook work. My recommendation: **land as-is, but file a follow-up to make the timeout configurable** (e.g. `[claudecode] graceful_stop_timeout = "5s"` in config.toml, defaulting to 5s). The constant doc comment already says it must be revisited if Stop hooks return — a config knob is the durable answer.
- **`closeTimeout = 150 * time.Second`** in `core/engine.go` (line ~80) is set well above `defaultGracefulStopTimeout` (5s) on purpose (covers potential custom overrides / future Stop-hook support). That coupling is correct but undocumented; a one-line comment next to `closeTimeout` explaining "this is the ceiling for the *outer* teardown; the inner graceful phase is `defaultGracefulStopTimeout`" would prevent a future maintainer from "fixing" one without touching the other.
- **`notifySessionCloseFailure` uses `e.waitOutgoing(platform)`** — if `waitOutgoing` is the one that holds up the close goroutine, this could chain stalls during a teardown storm. Worth checking that `waitOutgoing` returns promptly (or skips) when the platform is shutting down; otherwise the timeout-safety net becomes the next failure mode.

- The `awaitSessionClose` "loop again" path handles overlapping teardowns correctly (matching done channel comparison in `beginSessionClose`), but a quick comment at the top of that loop explaining *"an overlapping teardown may have been registered"* would save the next reader 5 minutes.
- `closingMu` and `interactiveMu` are held in different places; if any future code holds `interactiveMu` and tries to call into `closingMu` while a teardown is in flight, deadlock is possible. A doc comment on each lock ("may block up to closeTimeout; do not hold while awaiting teardown") would lock this in.

**❓ Questions (please ack before merge):**

1. **Configurable `gracefulStopTimeout`** — keep as constant for this PR and file a follow-up, or rework to a config knob now? My preference: ship as-is, file the follow-up. The behaviour change is well-justified by production data.
2. **The unrelated `origin/main` Windows compile errors** — would the maintainer like a separate PR fixing `check_linger_other.go` and `agent/pi/proc_windows.go`, or leave for the next contributor who hits it?

**Testing / Risk:**

- Verified evidence: CI 5/5 SUCCESS on head `ae62d95`. `go test ./agent/claudecode/ -run 'WaitDelay|GracefulStop|IdleSession'` passes per author.
- Honest disclosure: full `go test ./core/` on Windows has rotating flakes (`TestCUJ_A3_ImageReachesAgent`, `TestCUJ_H2_TwoPlatformsConcurrentNoBleed`, `TestCUJ_A5_FileReachesAgent`, `TestCompactReplyFooterPath_HomeRelativeDeepPathStaysFull`, both `TestAppendFileRefs_*`) — confirmed pre-existing on pristine `main`, not introduced by this PR.
- Remaining risk:
  1. **5s graceful timeout** will silently truncate any user-installed Stop hook. Low risk today, real risk once someone installs claude-mem. The doc comment + config-knob follow-up mitigates this.
  2. **Two new Engine maps** (`closingSessions`, `unsafeResume`) — both guarded by `closingMu`; no map grows without bound under normal use, but a session-storm scenario (many quick /stop → /new cycles) could leave stale entries. Not exercised in the new tests; the production data shows 5 bridged projects without symptoms, which is encouraging.
  3. **`closeTimeout = 150s` budget** for `awaitSessionClose` means a wedged close can block a new spawn for 3 minutes before the unsafe-resume fallback fires. That matches the existing teardown ceiling; just worth knowing.

**Next step:** Recommend merge once the maintainer acks the two questions above. The fix is well-evidenced, correctly synchronised, fails loudly instead of silently, and is already battle-tested in production.

chenhg5 (commented):
## QA Re-review — PR #1714 (claudecode /stop teardown)

**Conclusion:** Approve (with three P2/P3 nits — see below). No blocker remains.

### Re-review scope
Head SHA `ae62d95` is **unchanged** since my 2026-08-23 review, but main has gained 8 commits (including #1737 which also touches `agent/claudecode/session.go`). I re-validated the merge and re-checked for new findings introduced by the diff.

### Re-verification
- CI: 5/5 green — [run 32326216193](https://github.com/chenhg5/cc-connect/actions/runs/32326216193).
- Local merge: `git merge --no-commit --no-ff origin/main` auto-merges cleanly (no conflict in `agent/claudecode/session.go`, `core/engine.go`, or `core/i18n.go` — including the overlap with #1737's `innerArgs` block, which this PR does not touch).
- Local tests on the merged tree: `go test -race -count=1 ./agent/claudecode/ ./core/` → ok (14.4s + 48.1s).
- `go vet ./agent/claudecode ./core ./config` → clean.

### ✅ What still looks good (unchanged from prior review)
- Root-cause analysis is honest and reproducible. The `start /b` vs `Start-Process -NoNewWindow` distinction is the smoking gun, and `TestProductionSessionSetsWaitDelay` guards against silent reverts.
- `cmd.WaitDelay = 3 * time.Second` matches the convention of gemini/kimi/antigravity/hooks (1s); 3s is justified because Claude Code may still be emitting a panic trace on exit.
- Synchronisation order in `beginSessionClose` / `awaitSessionClose` is correct (register before scheduling the goroutine; await before taking `interactiveMu`).
- `unsafeResume` + `MsgSessionResumeUnsafe` is the right fallback. Two live agents fighting over one jsonl transcript is strictly worse than lost context.
- `notifySessionCloseFailure` + `MsgSessionCloseFailed` makes a previously silent failure visible to the user. Real UX win.
- Production-validated across 5 bridged projects, no recurring symptoms.

### 🔵 New minor finding (P3 nit, not in prior review)

**Two minor gofmt deviations introduced by this PR** (CI's `golangci-lint` set does not include `gofmt`, so the lint job missed them):

1. `agent/claudecode/session.go` line ~1203 — an indented code block in a doc comment needs to be wrapped in blank-comment lines:
   ```go
   //     17:45:20 /stop → 17:47:20 graceful 超时才发 SIGTERM → 17:47:29 真死，**用时 2 分 09 秒**。
   ```
   `gofmt` wants:
   ```go
   //
   //	17:45:20 /stop → 17:47:20 graceful 超时才发 SIGTERM → 17:47:29 真死，**用时 2 分 09 秒**。
   //
   ```

2. `agent/claudecode/waitdelay_test.go` line ~109 — double-space before `// 读 stdin，EOF 即退` should be a single space.

Both are pure whitespace; CI runs and tests are unaffected. Worth cleaning up for consistency but not blocking.

### 🟠 Should improve (unchanged from prior review)

- **`gracefulStopTimeout` 120s → 5s** is a behaviour change. The author explicitly invites the question of making it configurable. Recommendation: land as-is and file a follow-up for a `[claudecode] graceful_stop_timeout = "5s"` config knob (defaulting to 5s). The constant doc comment already flags the maintenance burden.
- **`closeTimeout = 150 * time.Second`** is set well above `defaultGracefulStopTimeout` (5s) on purpose, but that coupling is undocumented. One-line comment next to `closeTimeout` explaining the inner/outer split would prevent a future "fix one without touching the other" mistake.
- **`notifySessionCloseFailure` uses `e.waitOutgoing(platform)`** — confirm `waitOutgoing` returns promptly (or skips) when the platform is shutting down, so the close-failure notification doesn't get stuck behind another stall.

### 🟠 Coverage gap (still present, worth a maintainer ack)

The new Engine helpers (`beginSessionClose`, `awaitSessionClose`, `markUnsafeResume`, `consumeUnsafeResume`, `notifySessionCloseFailure`) have **no direct unit tests**. The `controllableAgentSession` struct got three new fields (`closeDelay`, `closeErr`, `closeFinished`) to support such coverage, but no test exercises them — only the existing `Close()` plumbing uses them. The behavior most likely to regress (close/spawn race guard, unsafe-resume fallback, failure-notify routing) is therefore only covered by the production-data argument.

A minimal regression test would be: spawn a session with `closeDelay: 100ms, closeErr: nil`; in parallel, send a new message to the same session key; assert `awaitSessionClose` blocks until the goroutine returns, then asserts the new spawn resumes. Author already shipped the harness — would be a ~50-line addition.

### 🔵 Optional (unchanged)
- Comment at the top of `awaitSessionClose`'s "loop again" path explaining *"an overlapping teardown may have been registered"*.
- Doc comment on `closingMu` and `interactiveMu` warning that neither should be held while awaiting teardown.

### ❓ Questions for maintainer (carry-over from prior review)

1. **Configurable `gracefulStopTimeout`** — ship as constant for this PR and file a follow-up config-knob, or rework to a config knob now? My preference: ship as-is, file the follow-up.
2. **The unrelated `origin/main` Windows compile errors** (`CheckLinger` redeclared, `os` imported and not used in `agent/pi/proc_windows.go`) — separate PR or leave for the next contributor? *Update 2026-08-26: `agent/pi/proc_windows.go` was fixed by #1738 (already merged). `check_linger_other.go` was fixed by #1710 (already in QA queue). So this question is moot now.*

### Testing / Risk (updated)

- Verified evidence: CI 5/5 green on head `ae62d95`, local merge with current `main` (8 commits newer than 2026-08-23, including #1737) auto-merges cleanly, local race tests on `agent/claudecode` + `core` pass.
- Remaining risk:
  1. 5s graceful timeout will silently truncate any user-installed Stop hook (low today, real once someone installs claude-mem).
  2. Two new Engine maps (`closingSessions`, `unsafeResume`) — guarded by `closingMu`, no unbounded growth under normal use; session-storm scenarios could leave stale entries (not exercised in tests).
  3. `closeTimeout = 150s` ceiling means a wedged close can block a new spawn for up to 3 minutes before the unsafe-resume fallback fires. Matches the existing teardown ceiling; just worth knowing.

### Next step
Recommend merge once the author runs `gofmt -w` on the two flagged files (or I can do it in this PR if maintainer prefers). The fix is well-evidenced, correctly synchronised, fails loudly instead of silently, and is battle-tested in production.

Chendounai:
Thanks for the re-review — both nits are fixed, and I took the coverage gap as well.

**gofmt (`c883e71`)**

Ran `gofmt -w` on the two flagged files only:

- `agent/claudecode/session.go` — the indented timeline block in the doc comment now has the blank comment lines gofmt wants (and a tab rather than spaces).
- `agent/claudecode/waitdelay_test.go` — single space before the trailing comment.

Deliberately *not* repo-wide: even with line endings corrected, `gofmt -l .` still lists 74 files across packages this PR never touches, so a blanket run would bury the actual diff. Happy to send that as a separate cleanup PR if you want it.

One note for anyone reproducing this on Windows: with `core.autocrlf=true` (the default on a fresh clone here) `gofmt -l` flags *every* file in the repo. `core.autocrlf=false` plus a fresh checkout is needed before the numbers mean anything — worth knowing before someone runs `gofmt -w .` and produces a 400-file diff.

**Coverage gap (`f84c491`)**

Added `core/session_teardown_test.go` — six cases, all on the existing `controllableAgentSession` harness:

- a spawn waits while a teardown is in flight
- nothing registered → no wait
- an overlapping teardown keeps the waiter blocked (the "loop again" path)
- the unsafe-resume flag is consumed exactly once
- engine shutdown unblocks a wedged teardown and reports the wait as unclean
- an empty session key never blocks

I checked that these actually guard something rather than merely passing: stubbing `awaitSessionClose` to `return true` immediately — i.e. the pre-fix behaviour — turns three of them red.

**Q1, configurable `gracefulStopTimeout`:** agreed with your preference — ship as-is, file the follow-up for `[claudecode] graceful_stop_timeout = "5s"` defaulting to 5s. Happy to open that PR once this lands.

**Q2:** noted as moot — thanks for chasing down which PRs fixed those.

**Still open from your list — tell me which you want inside this PR rather than as follow-ups:**

- the one-line comment next to `closeTimeout` explaining the inner/outer split
- doc comments on `closingMu` / `interactiveMu` warning that neither should be held while awaiting a teardown
- `waitOutgoing` promptness during platform shutdown — this one needs actual digging rather than a comment; I'd rather do it properly as its own issue unless you want it here

**Testing disclosure:**

- `go vet ./core ./agent/claudecode` — clean.
- `go test ./core/ -run 'AwaitSessionClose|OverlappingTeardown|UnsafeResume|SessionCloseGuard'` — 6/6 pass.
- `go test ./agent/claudecode/ -run 'WaitDelay|GracefulStop|IdleSession'` — pass.
- I could **not** run `-race` locally (no C toolchain on this machine), so I'm relying on CI and on your own race run for that.
- Full `./core/` still shows the pre-existing failures from your list. I re-ran them on pristine `ae62d95` to confirm they aren't mine: `TestCompactReplyFooterPath_HomeRelativeDeepPathStaysFull` and both `TestAppendFileRefs_*` fail there too; `TestCUJ_A5_FileReachesAgent` happened to pass that run, consistent with it being flaky rather than environmental.

Issue #145, opened 2026-03-15
Pi session corrupted by large file embeds — no way to recover without /new


quabug:
## Problem

Pi embeds file content (images, videos) inline in its session `.jsonl` files as `<file>` tags within user messages. When a turn fails — e.g., a video exceeds the model's context limit (`Range of input length should be [1, 983616]`) — the embedded content remains in the session file and **all subsequent messages on that session keep failing** with the same error.

Currently the only recovery is `/new`, which loses all conversation context.

### Reproduction

1. Start a pi session, have a few turns of conversation
2. Send a large video clip (e.g., a `.mov` file)
3. Pi embeds the raw video bytes inline in the session `.jsonl` as a `<file>` tag (~2-3MB of binary data in a text field)
4. The model API returns `400 InternalError.Algo.InvalidParameter: Range of input length should be [1, 983616]`
5. Every subsequent message on this session fails with the same error, because pi replays the full session history (including the embedded video) on each `--session` resume
6. Pi also auto-retries 3 times, making the user wait even longer

### Root Cause

Pi's session `.jsonl` format stores all content inline. Once a large file is written into the session history, there's no way to remove it — the session is permanently corrupted.

## Suggested Fix: `/revert` or `/undo` command

Add a command that removes the last turn (last user message + any subsequent assistant response lines) from the session `.jsonl` file. This is straightforward since the file is line-oriented JSONL:

1. Read the file, find the last `{"type":"message","message":{"role":"user",...}}` line
2. Truncate everything from that line onwards
3. Write back

### Design Consideration

This may only be relevant to agents that use file-based sessions (like pi). Rather than adding a global builtin command for all agents, consider supporting **agent-specific command registration** — agents could expose commands like `/revert` only when they implement the capability, similar to the existing optional interface pattern (`SessionDeleter`, `HistoryProvider`, etc.).

A possible approach:
- Define a `SessionReverter` optional interface in `core/`
- Agents that support it implement `RevertLastTurn(ctx, sessionID) error`
- Register `/revert` as a builtin that checks for the interface and falls back gracefully with "not supported"

happyTonakai:
/undo 命令我更倾向于cc的实现，即可以选择只回滚对话内容，还是回滚对话和代码。仅仅通过 jsonl 的截断无法实现代码回滚。这个功能我更倾向于直接使用 Pi Agent 的 Extension（官方example中已经有git-checkpoint），在cc-connect中只需要注册一个自定义命令即可。

PR #1724, opened 2026-08-23
fix(pi): pass attachments as @<path> argv refs instead of inlining bytes (#1723)

### LINKED ISSUE #1723: [Bug] Pi 智能体：收到的图片导致进程崩溃（fork/exec ... invalid argument），未能传给模型
## 问题概述

当项目的 agent 类型是 `pi` 时，用户发送**图片**给机器人，cc-connect 在拉起 `pi` 进程时直接失败，报
`start: fork/exec <pi路径>: invalid argument`。纯文本消息一切正常。

根因在 `agent/pi/session.go` 的 `piSession.Send`：入站图片先被保存到磁盘，但随后又**被读回来，把原始二进制字节直接内联进 prompt**，而这段 prompt 是作为 `pi` 的 `-p` 命令行参数传入的。图片字节里含 `0x00`，导致 `execve` 返回 `EINVAL`（`invalid argument`），`pi` 进程根本没有启动——图片从未到达模型。

这**不是"模型不支持视觉"**的问题：`pi` 用文件引用方式可以把图片正常传给模型，目标模型也确实支持视觉（下面有验证）。

## 环境

- cc-connect `1.5.0`（Homebrew）。**`main` 分支代码与此完全相同，同样存在该问题。**
- agent：`type = "pi"` → `pi` CLI（`@mariozechner/pi-coding-agent`，v0.84.2），以 Node shebang 脚本方式启动。
- 模型：`deepseek` provider / `deepseek-v4-flash-vision-exp`（`input: ["text","image"]`），已设为默认模型。
- 平台：飞书（Lark）走 WebSocket。（任何带图片的消息都会走同一代码路径。）
- 配置为默认的 json 一次性模式（未开启 `rpc`）。

## 复现步骤

1. 配置一个 `agent.type = "pi"` 且模型为视觉模型（如 `deepseek-v4-flash-vision-exp`）的项目。
2. 给机器人发任意一张图片（例如飞书截图）。
3. 观察：cc-connect 日志先记录收到入站图片，随后在尝试拉起 `pi` 时报错。

## 期望行为

图片被当作真实图片内容转发给模型（就像文本消息那样），助手正常回复。

## 实际行为

`pi` 从未被拉起，本次对话直接报错。日志：

```
2026-08-23 09:03:50  feishu: dispatched image batch ... image_count=1
2026-08-23 09:03:50  message received ... has_images=true, content_len=0
2026-08-23 09:03:51  ERROR  failed to send prompt  error="start: fork/exec /opt/homebrew/bin/pi: invalid argument"
```

同一个会话里的文本消息都能正常完成（`turn complete`），因为纯文本没有 NUL 字节。

## 根因

`agent/pi/session.go` 的 `piSession.Send`（约 322–361 行）：

```go
var atFiles []string
if len(images) > 0 {
    atFiles = append(atFiles, saveImagesToDisk(attachDir, images)...)
}
if len(files) > 0 {
    atFiles = append(atFiles, saveFilesToDisk(attachDir, files)...)
}

// 把附件内容内联进消息
var promptText strings.Builder
promptText.WriteString(msg)
for _, f := range atFiles {
    data, err := os.ReadFile(f)
    if err != nil { ... continue }
    promptText.WriteString("\n\n--- " + filepath.Base(f) + " ---\n")
    promptText.Write(data)          // <-- 把图片原始字节内联进 prompt
}
...
return s.sendJSON(promptText.String())
```

`sendJSON`（约 365 行）随后执行：

```go
args := append(append([]string{}, s.extraArgs...), "--mode", "json", "-p", prompt)
...
cmd := exec.CommandContext(s.ctx, s.cmd, args...)
...
if err := cmd.Start(); err != nil { return fmt.Errorf("start: %w", err) }
```

整段 prompt（含原始字节）作为单个 `-p` 参数传入；内部含 `0x00`，`execve` 返回 `EINVAL` →
`start: fork/exec ... invalid argument`。

注：RPC 模式下同样发送这段 prompt，但走的是 stdin（二进制安全），所以 `rpc = true` 只能避开崩溃，
它仍然把字节当文本内联，模型依旧拿不到真正的图片内容。

## 验证：pi + 模型其实支持视觉

直接用 `pi` 文件引用方式调用，模型能正确看图：

```
pi --provider deepseek --model deepseek-v4-flash-vision-exp \
   -p "图里有什么形状/颜色，哪个在左边？" @/tmp/img.png
```

（图片是：左边红色方块、右边蓝色矩形）结果：

```
The image has two squares on a white background: a red square on the left and a blue square on the right.
The color on the LEFT side is red.   ✅
```

所以问题完全在 cc-connect 如何传附件，而不是模型能力。

## 建议修复

在 `piSession.Send` 中，把已保存的附件路径用 pi 原生的 `@file` 文件引用方式传给 `pi`（已被证明可行），
而不是把文件读回并把原始字节内联进 prompt。具体就是在 `sendJSON` 和 `startRPC` 两条路径里都把这些路径
以 `@<path>` 追加到 `pi` 调用参数，`msg` 保持为文本 prompt，例如：

```go
for _, f := range atFiles {
    args = append(args, "@"+f)   // pi 将 @path 视为文件/附件引用
}
```

这样图片能作为附件真正到达 pi，再由 pi 当作多模态内容传给模型，而不是让 `execve` 收到一个含 NUL 的 argv。

---

*（附带相关小问题：`~/.pi/agent/models-store.json` 仍只列出纯文本模型，而 `models.json` 已包含视觉模型；
把 store 同步一下，可避免任何读取 store 的组件误判模型为不支持视觉。）*

---
**相关 / related:** #145（Pi session corrupted by large file embeds）——同样是 pi 把附件内容内联进 prompt 引发的问题，可一并修复。

chenhg5:
Resolves #1723. Also fixes the related #145 (Pi session corrupted by large file embeds) since large embedded bytes were the same root cause.

## Root cause (user RCA, adopted as-is)

`agent/pi/session.go`\ Send() (lines 322-361 in the pre-fix code) read saved attachment files back and appended their raw bytes to the prompt text. The prompt then went into `-p <argv>` for json mode. Image bytes contain NUL, so execve returned EINVAL and the pi process never started — the model never saw the image.

- **rpc = false** (default): direct crash, model never receives the image.
- **rpc = true**: stdin doesn't blow up, but the model still gets image bytes as text rather than a real visual input.

## Fix

Stop inlining bytes; let pi's @<path> mechanism handle attachments:

- **sendJSON (json mode)**: append `@<path>` as argv entries after the `-p` argument.
- **sendRPC (rpc mode)**: embed `@<path>` references into the message text under an `Attachments:` header.
- **Send()**: no longer reads file bytes back; just hands the `atFiles` list to whichever transport is active.

## Refactor (pure-function extraction, no behavior change outside the attachment fix)

- `buildJSONArgs(extraArgs, prompt, sessionID, model, thinking, atFiles)` — assembles argv; `sendJSON` calls it instead of inlining arg-building.
- `composeRPCPrompt(prompt, atFiles)` — assembles rpc message text; `sendRPC` calls it.

Extracting these makes the @-path refactor unit-testable without spawning a real pi process, which is the whole point of the regression tests below.

## Tests (`agent/pi/session_test.go`, 17 new assertions)

**buildJSONArgs** (12 cases):
- basic shape (`--mode json -p <prompt>` plus optional flags in order)
- empty `atFiles` produces zero `@`-prefixed args (bug guard)
- attachments become `@<path>` argv entries in the same order
- paths with spaces / dashes / non-ASCII round-trip cleanly
- prompt text untouched by attachments (path fragments never leak into `-p`)
- prompt with NUL bytes survives arg construction (regression for #1723)
- attachment bytes never leak into any argv element (regression for #1723 root cause)
- `@`-args come after all flag args, not interleaved
- optional flags (`--session-id`, `--model`, `--thinking`) only emitted when non-empty
- `extraArgs` are prepended before `--mode json -p`
- `extraArgs` slice not mutated by the helper (defends against repeat-call accumulation)
- each call returns a fresh slice (no aliasing across calls)

**composeRPCPrompt** (5 cases):
- empty `atFiles` returns the prompt unchanged (no spurious `Attachments:` header)
- non-empty `atFiles` get `@<path>` references appended
- attachment order preserved (pi loads in this order; multi-image messages depend on it)
- attachment bytes never inlined (regression for #1723)
- NUL bytes in the original prompt preserved (regression for #1723)

## Validation

- `go build ./agent/pi/...` clean
- `go vet ./agent/... ./core/...` clean
- `go test -race -count=1 ./agent/pi/` green (full pi package, 17 new + existing tests)
- `go build -tags no_web ./cmd/cc-connect/...` clean (the `web/dist` embed is a pre-existing gitignore issue unrelated to this change)

## Out of scope (deferred per PM triage)

- ~/.pi/agent/models-store.json vs models.json vision-model-list sync (PM noted as separate follow-up)
- Visual-model end-to-end verification with deepseek-v4-flash-vision-exp + Feishu: requires real pi + real model access in CI, not automatable here. The regression tests pin the argv shape; manual verification at PR time is sufficient.

## Risks

- The `@<path>` mechanism is pi-specific; if pi's CLI parsing of `@<path>` references changes in a future release, the same paths might stop being picked up. Low risk — pi has supported this syntax across all 0.x versions we have visibility into, and the args layer is the canonical place to handle this.
- We do not validate that `atFiles` paths actually exist before passing them to pi; `saveImagesToDisk` / `saveFilesToDisk` already log+skip on write failure, but a partially-failed save could still leave pi with a stale path. This matches the pre-fix behavior — the bug we are fixing is the execve crash, not the save-error surface.

### DIFF
--- a/agent/pi/session.go
+++ b/agent/pi/session.go
@@ -341,38 +341,30 @@ func (s *piSession) Send(msg string, messageID string, images []core.ImageAttach
 		atFiles = append(atFiles, saveFilesToDisk(attachDir, files)...)
 	}
 
-	// Build the message with attachment contents embedded
-	var promptText strings.Builder
-	promptText.WriteString(msg)
-	for _, f := range atFiles {
-		data, err := os.ReadFile(f)
-		if err != nil {
-			slog.Warn("piSession: cannot read attachment", "file", f, "error", err)
-			continue
-		}
-		promptText.WriteString("\n\n--- " + filepath.Base(f) + " ---\n")
-		promptText.Write(data)
-	}
-
+	// Issue #1723: do NOT inline attachment bytes back into the prompt.
+	// json mode passes the prompt via -p <argv>, so embedding raw image
+	// bytes triggers execve EINVAL on NUL bytes and the pi process never
+	// starts (model never sees the image). rpc mode goes through stdin
+	// so it doesn't crash, but the model still gets image bytes as text
+	// rather than a real visual input. Both paths must hand pi the saved
+	// paths via pi's @<path> mechanism — sendJSON appends them as argv
+	// entries; sendRPC embeds them in the message text, which pi parses
+	// the same way.
 	if s.rpc {
-		return s.sendRPC(promptText.String())
+		return s.sendRPC(msg, atFiles)
 	}
-	return s.sendJSON(promptText.String())
+	return s.sendJSON(msg, atFiles)
 }
 
 // sendJSON spawns `pi --mode json -p <prompt>` as a one-shot process,
 // reads all output events, and sends them to the events channel.
-func (s *piSession) sendJSON(prompt string) error {
-	args := append(append([]string{}, s.extraArgs...), "--mode", "json", "-p", prompt)
-	if sid := s.CurrentSessionID(); sid != "" {
-		args = append(args, "--session-id", sid)
-	}
-	if s.model != "" {
-		args = append(args, "--model", s.model)
-	}
-	if s.thinking != "" {
-		args = append(args, "--thinking", s.thinking)
-	}
+//
+// Issue #1723: attachment paths are passed as @<path> argv entries (Pi's
+// standard mechanism). Embedding the file bytes into -p would crash
+// execve on NUL bytes, and would also break pi's vision pipeline which
+// needs to know the file is on disk rather than be handed raw bytes.
+func (s *piSession) sendJSON(prompt string, atFiles []string) error {
+	args := buildJSONArgs(s.extraArgs, prompt, s.CurrentSessionID(), s.model, s.thinking, atFiles)
 
 	slog.Debug("piSession: spawning json mode", "cmd", s.cmd, "sessionID", s.CurrentSessionID())
 
@@ -465,13 +457,51 @@ func (s *piSession) writeRPCCommand(cmd map[string]any) error {
 // sendRPC writes a JSON "prompt" command to the persistent RPC process stdin.
 // Events are read asynchronously by readLoopRPC, including agent_end which
 // triggers EventResult.
-func (s *piSession) sendRPC(prompt string) error {
+//
+// Issue #1723: attachment paths are embedded into the message text as
+// @<path> references (pi's standard mechanism, parsed the same way as in
+// json mode). The rpc stdin pipe doesn't crash on NUL bytes, but the
+// model still needs to load the files from disk — embedding raw bytes
+// in the message field would give the model text-shaped garbage instead
+// of a real visual/file input.
+func (s *piSession) sendRPC(prompt string, atFiles []string) error {
 	cmd := map[string]any{
 		"type":    "prompt",
-		"message": prompt,
+		"message": composeRPCPrompt(prompt, atFiles),
 	}
 	slog.Debug("piSession: sending RPC prompt", "bytes", len(prompt))
 	return s.writeRPCCommand(cmd)
+}
+
+// composeRPCPrompt builds the rpc-mode message text, appending
+// "@<path>" references for any attachments. Pulled out of sendRPC so
+// the Issue #1723 attachment refactor is unit-testable without a live
+// rpc process.
+//
+// The format is stable on purpose:
+//   - a blank line separates the user's text from the attachment list
+//   - "Attachments:" header mirrors pi's CLI help so users see
+//     consistent wording whether they invoke the binary directly or
+//     via cc-connect
+//   - each @<path> is on its own line in the same order as atFiles
+//
+// When atFiles is empty the original prompt is returned unchanged —
+// we do NOT emit an empty "Attachments:" header, because an
+// assistant model could interpret that as a request to attach
+// something.
+func composeRPCPrompt(prompt string, atFiles []string) string {
+	if len(atFiles) == 0 {
+		return prompt
+	}
+	var b strings.Builder
+	b.WriteString(prompt)
+	b.WriteString("\n\nAttachments:\n")
+	for _, f := range atFiles {
+		b.WriteString("@")
+		b.WriteString(f)
+		b.WriteByte('\n')
+	}
+	return b.String()
 }
 
 // ── Event Handling (shared by both modes) ───────────────────
@@ -1183,6 +1213,31 @@ func (s *piSession) GetContextUsage() *core.ContextUsage {
 	}
 	u := *s.lastUsage
 	return &u
+}
+
+// buildJSONArgs assembles the argv slice for the one-shot `pi --mode json`
+// process. Pulled out of sendJSON so the @<path> attachment refactor
+// (Issue #1723) can be unit-tested without spawning a process.
+//
+// Returns a fresh slice on every call — never aliased to the caller's
+// extraArgs, since sendJSON mutates it.
+func buildJSONArgs(extraArgs []string, prompt, sessionID, model, thinking string, atFiles []string) []string {
+	args := append(append([]string{}, extraArgs...), "--mode", "json", "-p", prompt)
+	if sessionID != "" {
+		args = append(args, "--session-id", sessionID)
+	}
+	if model != "" {
+		args = append(args, "--model", model)
+	}
+	if thinking != "" {
+		args = append(args, "--thinking", thinking)
+	}
+	// Issue #1723: append attachment paths as @<path> argv entries (NOT
+	// inlined into the prompt text). pi treats them as inputs to load.
+	for _, f := range atFiles {
+		args = append(args, "@"+f)
+	}
+	return args
 }
 
 // ── Attachment helpers ───────────────────────────────────────
--- a/agent/pi/session_test.go
+++ b/agent/pi/session_test.go
@@ -1,0 +1,341 @@
+package pi
+
+import (
+	"bytes"
+	"os"
+	"strings"
+	"testing"
+)
+
+// ── buildJSONArgs (Issue #1723) ────────────────────────────────
+//
+// Issue #1723 regression coverage: the old code inlined saved attachment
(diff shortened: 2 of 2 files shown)


chenhg5 (commented):
## QA review: Approve (GitHub: comment only — author=self, cannot self-approve)

Approving this PR through the Multigent QA gate. GitHub does not allow the PR author to self-approve, so I'm posting this as a comment instead of a formal approve event; the QA decision (Multigent `review_decision`) is still `approve`.

### What looks good

- **Root cause matches user RCA exactly.** `Send()` no longer reads attachment bytes back; `sendJSON` and `sendRPC` both pass saved paths via pi's native `@<path>` mechanism. Both transports are fixed at the same layer — the right place, since both old behaviors shared the same `promptText.String()` -> `-p <argv>` (or stdin message) sink.
- **Pure-function refactor is the right move.** Extracting `buildJSONArgs` and `composeRPCPrompt` makes the @-path change unit-testable without spawning a real pi process. The 17 new assertions pin argv shape, byte-leak prevention, NUL-byte safety, and slice-aliasing — exactly what a regression test needs.
- **`Attachments:` header is conditional.** `composeRPCPrompt` returns the prompt unchanged when `atFiles` is empty, avoiding a spurious header that could be misread by the assistant as "please attach something." Subtle but correct.
- **`extraArgs` is defensively copied** via `append([]string{}, extraArgs...)`, with the helper returning a fresh slice. `TestBuildJSONArgs_ReturnsFreshSlice` and `TestBuildJSONArgs_ExtraArgsNotMutated` together pin this.
- **`@-args` come after all flag args.** Matches pi CLI semantics where trailing positionals are inputs. Verified by `TestBuildJSONArgs_AttachmentsAfterCoreFlags`.
- **#145 fixed for free.** Same root cause; same fix surface.

### Must fix

None.

### Should improve (non-blocking)

- 🟠 **P2 — `sendRPC` log underreports bytes.** `slog.Debug("bytes", len(prompt))` still logs the *original* prompt length. After the fix, the actual bytes written to stdin is `len(composeRPCPrompt(prompt, atFiles))`. Cosmetic — doesn't affect correctness — but worth aligning for debuggability.

### Optional

- 🔵 **P3 — missing trailing newline** in `session_test.go`. CI lint passes so this is project-style-only.

### Testing / Risk

- **CI**: lint / unit-test / smoke-test / regression-test / performance-test all SUCCESS on run 32642115862. `mergeStateStatus=CLEAN`.
- **E2E visual-model verification**: out of scope for CI. User already confirmed `@<path>` works with `deepseek-v4-flash-vision-exp` in the original issue. Recommend a quick manual smoke at release time; not a merge blocker.
- **Rollback**: low risk, public API unchanged, single-PR revert.
- **Compatibility**: this is a fix, not a breaking change — downstream goes from "pi never starts / model gets bytes as text" to "pi processes real visual/file input."

### Next step

Maintainer: please merge at your convenience via the GitHub UI. QA will not auto-merge. Multigent `review_decision=approve`, `github_review_type=comment` (forced by self-author restriction).

Thanks for the clean PR and for bundling #145 in.

chenhg5 (commented):
## QA Review — PR #1724 (Fixes #1723)

**QA Conclusion:** Approve (posted as comment because this is the author's own PR; GitHub blocks self-approval).

**Overall assessment:** Clean, well-scoped bugfix from the maintainer. CI is 5/5 SUCCESS, merge state CLEAN, current head `03f73e9` has no prior formal QA review — this review covers the full diff.

**Review scope:**
- `agent/pi/session.go`: +83/-28 (Send / sendJSON / sendRPC + new helpers)
- `agent/pi/session_test.go`: +346 new (17 test cases)
- Focused on correctness, regression coverage for the EINVAL/NUL root cause, refactor cleanliness, and risk surface.

**✅ What looks good:**
- **Pure-function extraction** of `buildJSONArgs` and `composeRPCPrompt` is the right call. It makes the @<path> refactor unit-testable without spawning a real pi process, which is the only way to lock down this regression.
- **17 dedicated test cases** pin both the new behaviour and the specific failure mode of the old code: NUL bytes in prompt, attachment bytes inlined into argv / message text, ordering, flag/positional separation, extra-args mutation safety, fresh-slice aliasing. This is the regression coverage I'd want for an execve-level bug.
- **Behavior change is minimal and reversible** if pi changes its CLI parsing — the helpers wrap a single decision point.
- **Root cause comment** at the top of `Send()` is well-written and will save the next maintainer a debug session.
- **Both transports covered** (json via argv `-p`, rpc via stdin message) — the asymmetry between the two fixes is correctly documented.

- `agent/pi/session.go` `sendRPC`: the debug log still uses `len(prompt)` (the user-supplied message), but the bytes actually sent are `len(composeRPCPrompt(prompt, atFiles))`. With non-empty `atFiles` the log will under-report by exactly the size of the attachments header — the opposite of what's useful when debugging attachment throughput. One-line fix.

- `composeRPCPrompt` format (`\n\nAttachments:\n@<path>\n...`) is described as "stable on purpose", which is good. A short note in `AGENTS.md` or a package-level comment saying "Pi rpc prompt format is consumed by the pi binary's @<path> parser; do not silently change `Attachments:` wording" would protect future contributors who might be tempted to "clean up" the header.

**❓ Questions:** None blocking.

**Testing / Risk:**
- Verified evidence: CI 5/5 SUCCESS (`lint`, `unit-test`, `smoke-test`, `regression-test`, `performance-test`) on head `03f73e9`.
- Remaining risk: (1) `@<path>` is pi-specific — low risk, documented. (2) `atFiles` paths not validated before passing to pi — matches pre-fix behaviour, no regression. (3) No vision-model E2E in CI — correctly noted as not automatable.

**Next step:** Recommend merge. The change is small, root cause is clear, tests pin behaviour, and CI is fully green. The slog nit can be fixed in this PR or follow-up.

PR #1730, opened 2026-08-24
fix(codex): propagate failed app-server turns

AaronZ345:
Supersedes #1599.

This is a same-patch replacement rebased onto the latest `main`.

## Summary

- inspect `turn/completed` status and error data from Codex app-server
- emit `EventError` instead of a successful empty `EventResult` when the turn failed
- clear failed-turn state so the later idle notification cannot emit a duplicate completion

## Why

Codex app-server can report a failed turn through `turn/completed` with `turn.status = "failed"` and an error object. The adapter currently ignores both fields and always completes successfully. Scheduled jobs then render `(empty response)` and are recorded as completed, hiding the upstream error.

## Tests

- `go test -race ./agent/codex -run 'TestAppServerSession_FailedTurnEmitsError$' -count=1`
- `git diff --check`

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -1126,7 +1126,18 @@ func (s *appServerSession) handleNotification(method string, paramsRaw json.RawM
 	case "turn/completed":
 		var notif turnNotification
 		if err := json.Unmarshal(paramsRaw, &notif); err == nil {
-			s.completeTurn()
+			if strings.EqualFold(strings.TrimSpace(notif.Turn.Status), "failed") || notif.Turn.Error != nil {
+				errMsg := ""
+				if notif.Turn.Error != nil {
+					errMsg = strings.TrimSpace(notif.Turn.Error.Message)
+				}
+				if errMsg == "" {
+					errMsg = "turn failed (no details)"
+				}
+				s.failTurn(fmt.Errorf("%s", errMsg))
+			} else {
+				s.completeTurn()
+			}
 		}
 
 	case "thread/status/changed":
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## QA review for PR #1730 — approve

### Conclusion
**Approve.** Small, focused, well-tested fix for a real Codex app-server visibility bug. CI 5/5 green, all package tests pass under race locally.

### Verification
- CI: 5/5 green on run [32740364687](https://github.com/chenhg5/cc-connect/actions/runs/32740364687) (lint, unit-test, smoke-test, regression-test, performance-test). Note: `mergeStateStatus: UNKNOWN` (likely a transient GitHub classifier; `git merge` against `origin/main` was clean on #1735 — will spot-check here if needed).
- Local on head `d7f435d`: `go vet ./agent/codex/...` clean, `gofmt -l` clean, `go test -race -count=1 -run TestAppServerSession_FailedTurnEmitsError ./agent/codex/` → ok (1.0s), full `go test -race -count=1 ./agent/codex/` → ok (2.7s, no regressions).

### What looks good
- **Correct trigger logic.** Branches on `turn.status == "failed"` (case-insensitive, trimmed) *or* `turn.error != nil` — covers both server-reported failures and explicit error payloads even if status string is missing.
- **Defensive fallthrough.** If status is "failed" but no error message arrived, the user still gets `"turn failed (no details)"` rather than a silent empty response — exactly the symptom this PR is fixing.
- **State cleanup is correct and minimal.** The new `failTurn` clears `currentTurn` and `pendingMsgs` under the existing `stateMu`, mirroring the cleanup path used elsewhere in the session. Without this, the subsequent idle notification would race and possibly emit a duplicate completion — and the test specifically asserts against that.
- **Reuses existing primitives.** `s.emitError(err)` is the same path used for other error emissions, so any downstream handler that already handles `EventError` (operator UI, audit log, scheduled-job rendering) keeps working without changes.
- **Test covers all three claims** in the PR body: (1) error event with the right message, (2) state cleared, (3) no duplicate event from the subsequent idle notification. The third assertion is the one most likely to regress in future refactors, and it locks the behavior in.
- **Clean rebase.** The PR body notes this supersedes #1599 rebased onto current `main` — same patch, fresh head.

### Risk
- **Low.** The change is additive (new branch + new method) inside a single notification handler. The existing happy-path branch (`completeTurn`) is preserved verbatim. No new dependencies, no config schema change, no platform-handler change.
- Worst-case regression if `failTurn` ever deadlocks: `currentTurn` stays set and the session will refuse new turns until restarted. The test exercises the lock acquisition explicitly.
- Rollback: revert single commit; no data migration.

- Watch for scheduled Codex jobs in production logs that previously rendered "(empty response)" — those should now show a real error message and a non-success completion record.
- If a turn fails while `pendingMsgs` had buffered user input, that input is intentionally discarded; operators should be aware of this if they expected queued messages to retry on the next turn.

Merge: ready when maintainer is — no blockers from QA.

PR #1733, opened 2026-08-24
fix(weixin): make inbound message dedup configurable (#1667)

### LINKED ISSUE #1667: Bug: WeChat message duplication causes duplicate agent responses
## Problem
WeChat server sometimes retransmits the same message (same msg_id) to cc-connect, causing the agent (Claude Code) to process the same message twice and send two different responses to the user.

## Evidence
From cc-connect logs on a production NAS deployment:
```
msg_id=7492913259736648968 received at 20:02:13.179
msg_id=7492913259736648968 received at 20:02:17.827  (same msg_id, 4 seconds later)
```

Both invocations generated independent Claude Code responses (different text answering the same question), confusing the user.

## Root Cause
The message processing pipeline in cc-connect does not deduplicate by msg_id before forwarding to the agent. When WeChat retransmits, cc-connect processes it as a new message.

## Suggested Fix
Add an LRU cache (e.g., `sync.Map` or `go-cache`) in the message dispatch layer that:
1. Checks incoming msg_id against recently processed IDs (configurable window, default 30s)
2. If a duplicate is found within the window, silently drop it (or respond with a cached reply)
3. Clean cache entries older than the window

This should be configurable per-platform (WeChat is the main offender) with an option like:
```toml
[projects.platforms.options]
  dedup_window_seconds = 10
  dedup_enabled = true
```

## Environment
- cc-connect v1.4.1 (commit 5d4c96dd)
- Platform: WeChat (ilink)
- OS: Linux (Debian 12)

## Impact
Every WeChat user experiences this issue intermittently (maybe 5-10% of messages). It makes the bot appear confused/flaky, answering the same question twice with different wording.

chenhg5:
Fixes #1667.

## Problem
WeChat (ilink) occasionally retransmits the same `message_id` on ACK delay or network jitter. The original report (cc-connect v1.4.1, commit 5d4c96dd) showed two identical `msg_id=7492913259736648968` entries 4 seconds apart, both generating independent agent responses.

## Root cause
On `main`, dispatchInbound already had an inline dedup map (5-minute window, keyed on `from|message_id|seq|create_time_ms|client_id`) — that landed in commit 760079b after the issue was filed. But the cache was hardcoded: no per-deployment override, no opt-out, no shared helper. Different platforms (feishu, telegram, discord, etc.) re-implement their own dedup or none at all.

## Fix
1. **`core/dedup`** — `MessageDedup` now accepts a configurable TTL via `NewMessageDedup(ttl)`; zero-value still uses `DefaultDedupTTL` (60s) for back-compat. Adds nil-receiver safety so callers embedding an uninitialised cache never panic.
2. **`platform/weixin`** — replace the inline map with `core.MessageDedup`, parse two new options:
   - `dedup_enabled` (default `true`)
   - `dedup_window_seconds` (default `300`)

   Drop path emits a DEBUG-level log line.
3. **Tests**
   - `core/dedup_test.go`: nil-receiver, configurable TTL expiry, default-TTL fallback for `ttl<=0`, race-detector stress.
   - `platform/weixin/dedup_test.go` (new): `pickBool` parser, default-on factory, opt-out factory, custom-window factory, dispatch drops duplicate `message_id` and accepts duplicates when `dedup_enabled=false`.
4. **`config.example.toml`** — documents the new keys (zh + en) in the weixin platform block.

No behaviour change for deployments that do not touch the new options: weixin continues to dedup with the same 5-minute window.

## Verification
- `go vet ./core/ ./platform/weixin/ ./cmd/...` — clean
- `go build ./...` — clean (the `web/embed.go` "all:dist" warning is a pre-existing web-asset issue, unrelated)
- `go test -race -count=1 ./core/ ./platform/weixin/` — both packages pass
- `go build -tags 'no_discord no_dingtalk no_qq no_qqbot no_line no_weibo no_webex no_matrix no_tuitui no_max no_wecom no_wps_xiezuo no_wps_agentspace no_cloud_web no_googlechat no_yuanbao' ./cmd/cc-connect` — clean

## Risk
Low. The new options default to the pre-existing behaviour (enabled, 5-minute window). All other platforms that already use `core.MessageDedup` (wecom, feishu, matrix, qq, qqbot, wps-agentspace, cloud-web) are unaffected because zero-value TTL still falls back to `DefaultDedupTTL`.

## Follow-ups (not in this PR)
- Extend `dedup_enabled` / `dedup_window_seconds` to feishu/telegram per the user's spec; default off, opt-in. Easy follow-up because the core helper already supports it.
- If weixin still occasionally leaks duplicates after this lands, the next step is to log the dedup key shape so we can spot a key collision (e.g. when `MessageID == 0`).

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1484,6 +1484,18 @@ app_secret = "your-feishu-app-secret"
 # 使其低于触发阈值。设为 0 关闭配额。
 # burst_limit = 4           # max separate messages per window / 每窗口最多独立消息数
 # burst_window_secs = 86400 # window length (default 24h) / 窗口时长（秒，默认24小时）
+#
+# Inbound message dedup (issue #1667):
+# ilink can retransmit the same message_id on ACK delay or jitter, which
+# otherwise causes the bot to reply twice with different answers. The dedup
+# cache is keyed on from + message_id + seq + create_time_ms + client_id.
+# Default: enabled with a 5-minute window. Set dedup_enabled=false to disable.
+# 入站消息去重（issue #1667）：ilink 在 ACK 延迟或网络抖动时会重传相同
+# message_id，导致 bot 回复两次不同的答案。缓存键由 from + message_id + seq
+# + create_time_ms + client_id 组合。默认开启，窗口 5 分钟。设为
+# dedup_enabled=false 可关闭。
+# dedup_enabled = true
+# dedup_window_seconds = 300
 
 # -----------------------------------------------------------------------------
 # Tencent Yuanbao (yuanbao.tencent.com bot platform)
--- a/core/dedup.go
+++ b/core/dedup.go
@@ -5,7 +5,11 @@ import (
 	"time"
 )
 
-const dedupTTL = 60 * time.Second
+// DefaultDedupTTL is the default dedup window used when a platform does not
+// configure one explicitly. It is intentionally generous — long enough to
+// absorb a network jitter / ACK-delay retry, short enough to keep the map
+// bounded under steady traffic.
+const DefaultDedupTTL = 60 * time.Second
 
 // StartTime is set once at process startup.
 // Platforms use it to discard messages created before the current process started,
@@ -14,15 +18,40 @@ var StartTime = time.Now()
 
 // MessageDedup tracks recently seen message IDs to prevent duplicate processing.
 // Safe for concurrent use.
+//
+// A zero-value MessageDedup uses DefaultDedupTTL; platforms can override via
+// NewMessageDedup(ttl). The struct intentionally stays a value type so existing
+// platform code that embeds `dedup core.MessageDedup` keeps working unchanged.
 type MessageDedup struct {
 	mu   sync.Mutex
 	seen map[string]time.Time
+	ttl  time.Duration // 0 means "use DefaultDedupTTL"
+}
+
+// NewMessageDedup returns a MessageDedup with a custom TTL. ttl <= 0 falls back
+// to DefaultDedupTTL.
+func NewMessageDedup(ttl time.Duration) *MessageDedup {
+	if ttl <= 0 {
+		ttl = DefaultDedupTTL
+	}
+	return &MessageDedup{ttl: ttl}
+}
+
+// TTL returns the dedup window in effect (handy for diagnostics and tests).
+func (d *MessageDedup) TTL() time.Duration {
+	if d == nil {
+		return 0
+	}
+	if d.ttl <= 0 {
+		return DefaultDedupTTL
+	}
+	return d.ttl
 }
 
 // IsDuplicate returns true if msgID was already seen within the TTL window.
 // Empty msgID is never considered a duplicate.
 func (d *MessageDedup) IsDuplicate(msgID string) bool {
-	if msgID == "" {
+	if d == nil || msgID == "" {
 		return false
 	}
 	d.mu.Lock()
@@ -31,8 +60,9 @@ func (d *MessageDedup) IsDuplicate(msgID string) bool {
 		d.seen = make(map[string]time.Time)
 	}
 	now := time.Now()
+	ttl := d.TTL()
 	for k, t := range d.seen {
-		if now.Sub(t) > dedupTTL {
+		if now.Sub(t) > ttl {
 			delete(d.seen, k)
 		}
 	}
--- a/core/dedup_test.go
+++ b/core/dedup_test.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"sync"
 	"testing"
 	"time"
 )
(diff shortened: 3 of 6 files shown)


chenhg5 (commented):
Conclusion: Comment (QA decision: approve — see Multigent QA report doc-20260824-d7j3ve)

> Note: posting as `--comment` because the PR author is the maintainer (`chenhg5`) and GitHub blocks self-approval. The QA decision is **approve**; this PR is **ready-to-merge**. The Human Merge Gate will see the same conclusion in the decision report.

## Overall assessment
Clean, well-scoped fix for issue #1667. The 5-minute inline dedup map that landed in commit `760079b` is now a reusable `core.MessageDedup` with a configurable TTL, and the weixin platform opts into it via two new config keys. Default behaviour is identical to the pre-existing inline map, so no upgrade risk. Tests pin both the regression path and the new opt-out semantics.

## What looks good
- **Default behaviour preserved.** `dedup_enabled = true` and `dedup_window_seconds = 300` in the factory reproduce the inline 5-minute map exactly — zero upgrade risk for existing deployments.
- **`MessageDedup` stays a value type.** Other platforms (wecom/feishu/matrix/qq/qqbot/wps-agentspace/cloud-web) embed `core.MessageDedup` directly; this change keeps that pattern working because zero-value TTL falls back to `DefaultDedupTTL`.
- **Nil-receiver safety** on both `IsDuplicate` and `TTL()`. This is the kind of small defensive change that prevents a future factory-init ordering bug from becoming a panic in production.
- **`pickBool` correctly handles the "key absent" case.** The factory uses an explicit `if _, ok := opts["dedup_enabled"]; !ok { dedupEnabled = true }` rather than relying on `pickBool(nil) == false` to be the default. That's the right call: without it, an unset key would silently disable dedup.
- **Two existing tests updated in lockstep** (`TestPollLoop_*` use `core.NewMessageDedup(5 * time.Minute)` instead of `make(map[string]time.Time)`). The PR description flagged this as a known boundary and the change landed correctly.
- **`TestDispatchInbound_DropsDuplicateMessageID`** is the right regression guard — it sends the exact same composite key twice and asserts only the first reaches the handler.

## Verified locally
- `go build ./core/... ./platform/weixin/...` clean
- `go vet ./core/ ./platform/weixin/ ./cmd/...` clean (only pre-existing `web/embed.go` "no matching files" warning, unrelated)
- `go test -race -count=1 -v` on the new tests — **all PASS**:
  - core: `TestMessageDedup_NilReceiver`, `TestMessageDedup_ConfigurableTTL`, `TestMessageDedup_DefaultTTLForZero`, `TestMessageDedup_ConcurrentConfigurable`
  - weixin: `TestPickBool` (14 sub-cases), `TestNew_DedupDefaultsToEnabled`, `TestNew_DedupCanBeDisabled`, `TestNew_DedupWindowRespected`, `TestDispatchInbound_DropsDuplicateMessageID`, `TestDispatchInbound_DedupDisabledAcceptsDuplicates`
  - existing: `TestPollLoop_NotifiesReadyForPollAfterFirstSuccessfulGetUpdates`, `TestPollLoop_DoesNotNotifyReadyForPollWhileGetUpdatesFails`
- `go test -race -count=1 -short ./core/` and `./platform/weixin/` — both `ok`
- `go build -tags 'no_discord no_dingtalk no_qq no_qqbot no_line no_weibo no_webex no_matrix no_tuitui no_max no_wecom no_wps_xiezuo no_wps_agentspace no_cloud_web no_googlechat no_yuanbao' ./cmd/cc-connect` clean

CI run 32750281595: lint / unit-test / smoke-test / regression-test / performance-test all SUCCESS, `mergeStateStatus=CLEAN`.

## Optional follow-ups (non-blocking, P3)
- **Extend the same two keys to feishu/telegram** (default off, opt-in). The core helper already supports it; this PR lays the groundwork but stops at weixin as planned. Worth doing as a single follow-up PR.
- **`pickBool` is currently weixin-local.** If more platforms add bool options in the near future, consider moving `pickBool` to `core` (mirroring `pickInt` would be a good pattern).
- **DEBUG-only log on drop.** Reasonable default. If ops turn `dedup_enabled = false`, there's no operator-facing signal that dedup is off — a small observability gap, P3.
- **Dedup-key shape logging on collision.** The PR description's "if weixin still occasionally leaks duplicates" follow-up (log the `from|message_id|seq|create_time_ms|client_id` shape when `MessageID == 0`) is a sensible next debugging hook — not needed for this PR.

## Testing / Risk
- Verified evidence: CI 5/5 green, full local race tests green, new tests cover nil-safety, configurable TTL expiry, default-TTL fallback for `ttl<=0`, race-detector stress, factory defaults, opt-out, custom window, regression drop path, and opt-out bypass.
- Remaining risk: minimal. Default behaviour unchanged; other platforms unaffected; new config keys are opt-in. The only behaviour change is `pickBool`-shaped (no platform embeds it yet).

## Next step
Ready to merge. Recommend merging and adding a one-line note to the next release announcement: "Weixin platform inbound dedup is now configurable: `dedup_enabled` (default true) and `dedup_window_seconds` (default 300). Defaults preserve existing 5-minute dedup behaviour."

PR #1725, opened 2026-08-24
fix(feishu): fail closed when bot open_id discovery fails at startup (#1618)

chenhg5:
## 关联 Issue
- https://github.com/chenhg5/cc-connect/issues/1618
- 决策报告: doc-20260824-prcefv (Lark/Feishu 启动 fail-open)

## 根因
`platform/feishu/feishu.go` 的 `Platform.Start()` 调用 `/open-apis/bot/v3/info` 解析 bot 自身 open_id。如果失败(本例:本地 HTTP 代理未就绪),之前的行为是 `slog.Warn` 一行后让 `botOpenID == ""`,而下游组聊过滤器(feishu.go:1386)`if chatType == "group" && !p.groupReplyAll && p.getBotOpenID() != ""` 直接把整个 `if` 块短路掉 — bot 在进程剩余生命周期内回复所有组消息(包括没 @ 它的),3h10m 静默失能窗口 + 难以察觉。

## 修复方案(采纳用户建议 A+B)
1. **Fail closed**:open_id 解析失败 → 组聊静默(DM 不受影响,因为 DM 不需要 mention 检查),而非接收全员消息
2. **Retry with backoff**:初始 fetch 包了 `withTransientRetry`(3 次重试,指数退避 500ms-5s + jitter),大部分瞬时失败自愈,根本到不了 degraded 状态
3. **重试预算耗尽后启动后台 supervisor**:每 5 分钟重试一次,直到成功或进程退出 — 透明代理/VPN/DNS 抖动自愈,无需 restart
4. **日志升级**:失败从 `WARN` 提到 `ERROR`;周期性的 supervisor 重试也是 ERROR
5. **可观测性**:新增可选 `core.PlatformHealth` 接口,feishu 实现 `PlatformHealth()` 暴露 degraded 状态;`/status` 卡片在 platforms 行加 `⚠️` 标记并附 `**⚠️ Degraded**` 块列出原因;management API `/status` 加 `degraded_platforms` 字段,`/projects/{name}` 把硬编码的 `"connected": true` 改成真实状态
6. **顺手修 webhook 模式**:Lark 私有部署走 webhook 时同样 `botOpenID == ""`,过滤器本来就 fail-open。本次修复后,若 webhook 模式后续也想要 bot open_id,可直接复用 supervisor

## 顺手检查 Feishu 国内版
[...]

PR #1735, opened 2026-08-25
fix(feishu): retry bot open_id discovery instead of permanently disabling group chat filtering

Niko-Wu:
# fix(feishu): retry bot open_id discovery instead of permanently disabling group chat filtering

> **Baseline:** branched from `main` (`580df5ea`), a single commit touching only
> `platform/feishu/`. No unrelated changes are bundled in.

## Symptom

When `cc-connect` starts in WebSocket mode and the very first Feishu API call fails,
the bot's own `open_id` is never learned — for the entire lifetime of the process.
The only trace is a single `WARN` at startup:

```
level=WARN msg="feishu: failed to get bot open_id, group chat filtering disabled"
```

The process then keeps running, looks healthy, serves DMs correctly, and gives no
further indication that anything is wrong. In the incident that prompted this PR the
daemon ran degraded for **43 minutes** until it was manually restarted.

## Reproduction conditions

Any transient failure of the *first* `/open-apis/bot/v3/info` call reproduces it:

- the process is launched at boot / by `launchd` / by systemd **before** the network
  stack or DNS resolver is ready;
- a laptop wakes from sleep and the supervisor restarts the service before Wi-Fi
  reassociates (this is the observed case — macOS `launchd` + wake-from-sleep);
- a brief DNS outage, captive portal, or proxy hiccup at the moment of startup.

The window is small, but the consequence is permanent, so over a long-running
deployment it is hit reliably.

## Real-world log

```
time=2026-08-25T13:52:32.693+08:00 level=WARN
  msg="feishu: failed to get bot open_id, group chat filtering disabled"
  error="api call: Post \"https://open.feishu.cn/open-apis/auth/v3/tenant_access_token/internal\": dial tcp: lookup ..."
```

## Root cause

`Platform.Start` (`platform/feishu/feishu.go`) treats bot identity discovery as a
one-shot, best-effort operation:

```go
if !p.shouldUseWebhookMode() {
    if openID, err := p.fetchBotOpenID(); err != nil {
        slog.Warn(p.platformName+": failed to get bot open_id, group chat filtering disabled", "error", err)
    } else {
        p.mu.Lock()
        p.botOpenID = openID
        p.mu.Unlock()
        slog.Info(p.platformName+": bot identified", "open_id", openID)
    }
}
```

There is no retry. A transient error is converted into a permanent capability loss —
a classic startup race, where the recovery path (the network coming up a few seconds
later) exists but is never taken.

## Security impact — this fails **open**, not closed

`p.botOpenID` is not merely used to *match* mentions; its emptiness disables the
mention gate altogether. From `onMessage`:

```go
if chatType == "group" && !p.groupReplyAll && p.getBotOpenID() != "" {
    if !isBotMentioned(msg.Mentions, p.getBotOpenID()) {
        // ... drop the message
    }
}
```

When `getBotOpenID()` returns `""` the entire condition is skipped, so **every message
in every group chat is dispatched to the agent**, including messages that never
mentioned the bot. For a deployment that relies on `@bot` as the activation gate, this
means unrelated group chatter silently triggers agent runs — burning tokens, and
exposing conversation content the operator did not intend to route to an agent.

This PR does not change that fallback behaviour (a separate decision about whether an
unknown bot identity should fail closed), but it does shrink the exposure window from
"until someone notices and restarts" to "until the network comes back".

Two other consumers of `getBotOpenID()` degrade more benignly and are likewise
untouched: `filterQuotedFilesForUser` fails closed (drops quoted files), and
`stripMentions` merely leaves the `@bot` prefix in the forwarded text.

## Fix

When the eager fetch fails, start a background goroutine that retries with
exponential backoff until it succeeds or the platform shuts down:

- backoff starts at **2s**, doubles, and is capped at **60s**; retries are unbounded,
  because the condition being waited on (network availability) has no deadline;
- on success the value is written under the existing `p.mu` and an `INFO` line records
  the recovery, so operators can tell degraded-then-recovered from degraded-forever;
- the goroutine owns a `context.CancelFunc` stored on the `Platform` and cancelled
  from `Stop()`, so it cannot leak and an in-flight HTTP request is cancelled too;
- `startBotOpenIDRetry` is idempotent — a second call while a loop is running is a
  no-op rather than a duplicate goroutine;
- `fetchBotOpenID` now takes a `context.Context` instead of hard-coding
  `context.Background()`, which is what makes cancellation actually reach the request.

**Webhook mode is deliberately unchanged.** The existing comment explains why startup
there must not depend on a bot-info call, and that reasoning still holds — the retry
only applies to the WebSocket path that already performed eager discovery.

## Scope of change

| File | Change |
| --- | --- |
| `platform/feishu/feishu.go` | +101 / −4 — three new `Platform` fields, two new methods, `fetchBotOpenID` takes a ctx, `Start`/`Stop` wiring |
| `platform/feishu/bot_openid_retry_test.go` | new — 4 tests |

No behaviour changes on the success path: when the first fetch works, the code runs
exactly as before and no goroutine is started.

## Testing

**Unit tests** — `go test -count=1 -race ./platform/feishu/...` is green on a clean
`main` checkout (34.7s), as are `gofmt -l` and `go vet ./platform/feishu/`:

- `TestStartBotOpenIDRetry_RecoversAfterFailure` — `httptest` server fails the first
  two `/open-apis/bot/v3/info` calls, then succeeds; asserts `botOpenID` is populated
  and that it took ≥3 attempts.
- `TestStartBotOpenIDRetry_OnlyOneGoroutine` — two `start` calls, one `stop`; asserts
  API traffic ceases, proving no duplicate goroutine.
- `TestStopBotOpenIDRetry_StopsGoroutine` — always-failing server; asserts the attempt
  counter freezes after `stop`, and that `stop` is idempotent.
- `TestStop_CancelsBotOpenIDRetry` — asserts `Platform.Stop()` cancels the loop.

Both new code paths were mutation-checked: neutering the goroutine launch fails two
tests, and neutering the `cancel()` call fails two others — the tests are not vacuous.

**End-to-end against the live Feishu API.** The built binary was started behind a
`CONNECT` proxy that was *not yet listening*, reproducing the original failure, then
the proxy was brought up mid-backoff:

```
15:08:53 WARN  feishu: failed to get bot open_id, group chat filtering disabled until retry succeeds
               error="... proxyconnect tcp: dial tcp 127.0.0.1:18080: connect: connection refused"
15:08:55 WARN  feishu: bot open_id retry failed  attempt=1 next_delay=4s
15:08:59 WARN  feishu: bot open_id retry failed  attempt=2 next_delay=8s
15:09:07 WARN  feishu: bot open_id retry failed  attempt=3 next_delay=16s
        <-- proxy started here -->
15:09:23 INFO  feishu: bot identified after retry, group chat filtering restored
               open_id=ou_b3b9... attempt=4
```

Backoff progression (2s → 4s → 8s → 16s) and recovery both behave as designed. On the
previous binary this scenario stays degraded indefinitely.

**Happy path unaffected:** the new binary was installed over the running service and
restarted via `launchctl`; startup logs the unchanged `feishu: bot identified` line
and no retry goroutine is created.

## Note on scope of verification

This PR does not touch `daemon/`. On non-Linux hosts a whole-repo `go build ./...`
currently trips over a duplicate `CheckLinger` declaration in that package, which
predates and is unrelated to this change, so verification here is reported at package
level: `go test -count=1 -race ./platform/feishu/...` passes on a clean `main`
baseline.

### DIFF
--- a/platform/feishu/bot_openid_retry_test.go
+++ b/platform/feishu/bot_openid_retry_test.go
@@ -1,0 +1,150 @@
+package feishu
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	lark "github.com/larksuite/oapi-sdk-go/v3"
+)
+
+// newBotInfoTestServer serves the tenant-access-token endpoint plus the bot
+// info endpoint. botInfoStatus decides, per attempt, whether the bot info call
+// fails; attempts counts how many times bot info was hit.
+func newBotInfoTestServer(t *testing.T, openID string, failFirst int64) (*httptest.Server, *atomic.Int64) {
+	t.Helper()
+	var attempts atomic.Int64
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		if r.URL.Path == "/open-apis/auth/v3/tenant_access_token/internal" {
+			_, _ = w.Write([]byte(`{"code":0,"msg":"ok","tenant_access_token":"t-test","expire":7200}`))
+			return
+		}
+		if r.URL.Path == "/open-apis/bot/v3/info" {
+			n := attempts.Add(1)
+			if failFirst < 0 || n <= failFirst {
+				w.WriteHeader(http.StatusInternalServerError)
+				_, _ = w.Write([]byte(`{"code":500,"msg":"internal error"}`))
+				return
+			}
+			_, _ = w.Write([]byte(`{"code":0,"msg":"ok","bot":{"open_id":"` + openID + `"}}`))
+			return
+		}
+		w.WriteHeader(http.StatusNotFound)
+	}))
+	t.Cleanup(srv.Close)
+	return srv, &attempts
+}
+
+func newBotOpenIDRetryPlatform(srv *httptest.Server, initial, maxDelay time.Duration) *Platform {
+	return &Platform{
+		platformName: "feishu",
+		client: lark.NewClient("app-id", "app-secret",
+			lark.WithOpenBaseUrl(srv.URL),
+			lark.WithHttpClient(srv.Client()),
+			lark.WithEnableTokenCache(false),
+		),
+		botOpenIDRetryInitialDelay: initial,
+		botOpenIDRetryMaxDelay:     maxDelay,
+	}
+}
+
+// TestStartBotOpenIDRetry_RecoversAfterFailure covers the startup race that
+// motivated this retry loop: the first fetch fails (network not ready yet) and
+// a later attempt succeeds, restoring group chat filtering.
+func TestStartBotOpenIDRetry_RecoversAfterFailure(t *testing.T) {
+	srv, attempts := newBotInfoTestServer(t, "ou_bot_recovered", 2)
+	p := newBotOpenIDRetryPlatform(srv, 10*time.Millisecond, 40*time.Millisecond)
+	defer p.stopBotOpenIDRetry()
+
+	p.startBotOpenIDRetry()
+
+	deadline := time.Now().Add(5 * time.Second)
+	for p.getBotOpenID() == "" && time.Now().Before(deadline) {
+		time.Sleep(5 * time.Millisecond)
+	}
+	if got := p.getBotOpenID(); got != "ou_bot_recovered" {
+		t.Fatalf("botOpenID = %q after %d attempts, want %q", got, attempts.Load(), "ou_bot_recovered")
+	}
+	if n := attempts.Load(); n < 3 {
+		t.Fatalf("attempts = %d, want at least 3 (two failures then success)", n)
+	}
+}
+
+// TestStartBotOpenIDRetry_OnlyOneGoroutine ensures a second call while a retry
+// loop is already running does not spawn a duplicate goroutine: a single stop
+// must then silence all retry traffic.
+func TestStartBotOpenIDRetry_OnlyOneGoroutine(t *testing.T) {
+	srv, attempts := newBotInfoTestServer(t, "ou_bot", -1) // always fails
+	p := newBotOpenIDRetryPlatform(srv, 10*time.Millisecond, 20*time.Millisecond)
+	defer p.stopBotOpenIDRetry()
+
+	p.startBotOpenIDRetry()
+	p.startBotOpenIDRetry()
+
+	deadline := time.Now().Add(5 * time.Second)
+	for attempts.Load() == 0 && time.Now().Before(deadline) {
+		time.Sleep(5 * time.Millisecond)
+	}
+	p.stopBotOpenIDRetry()
+
+	settled := attempts.Load()
+	time.Sleep(200 * time.Millisecond)
+	if after := attempts.Load(); after > settled+1 {
+		t.Fatalf("a second retry goroutine survived the single stop: attempts %d -> %d", settled, after)
+	}
+}
+
+// TestStopBotOpenIDRetry_StopsGoroutine verifies the retry goroutine exits on
+// shutdown instead of leaking and hammering the API forever.
+func TestStopBotOpenIDRetry_StopsGoroutine(t *testing.T) {
+	srv, attempts := newBotInfoTestServer(t, "ou_bot", -1) // always fails
+	p := newBotOpenIDRetryPlatform(srv, 10*time.Millisecond, 20*time.Millisecond)
+
+	p.startBotOpenIDRetry()
+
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## QA review for PR #1735 — approve

### Conclusion
**Approve.** Solid fix for a real, well-characterized production failure. CI 5/5 green, clean auto-merge with `main`, race-safe tests pass locally.

### Verification
- CI: 5/5 green on run [32821509154](https://github.com/chenhg5/cc-connect/actions/runs/32821509154) (lint, unit-test, smoke-test, regression-test, performance-test). Note: GitHub shows `mergeStateStatus: UNKNOWN`, but `git merge --no-commit` against `origin/main` succeeds with no conflicts — appears to be a transient GitHub classifier, not a real blocker.
- Local: `go vet ./platform/feishu/...` clean, `gofmt -l` clean, `go test -race -count=1 -run 'TestStartBotOpenIDRetry|TestStop' ./platform/feishu/` → ok (1.6s, including 4 new tests), full `go test -race -count=1 ./platform/feishu/` → ok (33s).
- New tests:
  - `TestStartBotOpenIDRetry_RecoversAfterFailure` — httptest fails 2x then succeeds; asserts `botOpenID` populated and ≥3 attempts.
  - `TestStartBotOpenIDRetry_OnlyOneGoroutine` — idempotency guard.
  - `TestStopBotOpenIDRetry_StopsGoroutine` — counter freezes after stop; idempotent.
  - `TestStop_CancelsBotOpenIDRetry` — `Platform.Stop()` cancels the loop.
- Author reports mutation-checked coverage; the goroutine launch and the `cancel()` call each have a dedicated test that breaks if either is neutered.

### What looks good
- **Real incident, real RCA.** The PR body documents a 43-minute degraded window caused by a single failed DNS lookup at boot, with the actual log line. The fix matches the failure mode exactly.
- **Defensive but tight scope.** Webhook mode is intentionally left alone (per the existing comment about why it must not depend on bot-info). The change touches only the WebSocket path that already performed eager discovery.
- **Idempotent retry goroutine.** The `if p.botOpenIDRetryCancel != nil { cancel(); return }` guard means concurrent `Start` calls can't multiply the loop. The mutex is held only for the assignment, not for the sleep or HTTP request.
- **Context actually reaches the request.** This is the one change in `fetchBotOpenID`'s signature, and it's the bit that makes `Stop()` actually able to cancel an in-flight HTTP request rather than waiting for it to time out.
- **Backoff bounds with documentation.** 2s → 60s exponential cap, unbounded attempts (justified in the PR: the condition being waited on has no deadline). Tests use the override fields (`botOpenIDRetryInitialDelay` / `MaxDelay`) to keep wall-clock time low.
- **Test/E2E honesty.** Author documents the live proxy-not-listening repro and the recovery log, plus flags the pre-existing `daemon` `CheckLinger` duplicate (separate from this change) as the reason verification is at package level rather than `./...`.

### 🟠 P2 (worth knowing, not blocking)
- The PR explicitly punts on **fail-closed vs fail-open** for unknown bot identity: today the `getBotOpenID() == ""` path in `onMessage` skips the entire gate, so a transient failure (this PR's case) *or* a permanent one silently fans every group message to the agent. The retry goroutine shrinks the window dramatically but does not close it. **#1725** (already QA-approved) appears to be addressing the permanent-failure path; once both are merged the security posture is meaningfully tighter. Worth keeping these two in mind together when cutting a release.

### 🔵 P3 (nit, optional)
- The two override fields (`botOpenIDRetryInitialDelay`, `botOpenIDRetryMaxDelay`) live as exported-effectively struct fields. Future tests will likely want shorter bounds; a small helper (`func (p *Platform) withBotOpenIDRetryTiming(...) *Platform`) would keep that pattern readable. Not a blocker.

### Risk
- **Low.** Single platform file change + new test file. No public API change, no config schema change, no handler signature changes outside the package. Behaviour on the success path is unchanged.
- Worst-case regression: if `startBotOpenIDRetry` ever panics during launch, the goroutine never starts and we fall back to today's behaviour (silent degraded mode). The test suite covers this path.
- Rollback: revert the single commit; no data migration needed.

- Watch for the new `feishu: bot identified after retry, group chat filtering restored` log line in production deployments — its presence confirms the fix is active, its absence combined with the older `group chat filtering disabled until retry succeeds` line tells operators the retry is still in flight.
- Pair with #1725 deployment to ensure the fail-closed path covers the case where retries themselves never succeed.

Merge: ready when the maintainer is — no blockers from QA's side.

Niko-Wu:
Closing as superseded by #1725, which landed the same startup retry plus a 5-minute background supervisor — and with safer fail-closed semantics for the group mention filter (this PR kept the filter fail-open during the outage window). Nothing left here that #1725 doesn't cover. Thanks!

PR #1338, opened 2026-06-13
Add per-session idle timeout for live agent processes

hl1221hl:
## Summary

Adds a per-project `agent_session_idle_timeout_mins` setting that closes an idle live agent process after a clean interactive turn while preserving the cc-connect session and saved agent session ID.

When the next message arrives in the same chat/thread, cc-connect starts a new agent process and resumes the same underlying conversation. This gives deployments a way to reclaim idle live CLI processes without rotating the user-facing session, and differs from `workspace_idle_timeout_mins`, which reaps an entire multi-workspace workspace.

## Behavior

- `agent_session_idle_timeout_mins` under `[[projects]]` is optional and disabled when unset or set to `0`.
- Idle close timers are scheduled only after a clean turn when the live agent session is still healthy.
- A new foreground turn cancels the pending idle close timer.
- Pending permission requests, queued messages, stopped sessions, abnormal event state, or dead agent sessions prevent idle close.
- Reloading config with the setting removed or disabled cancels already scheduled idle close timers.
- The saved `agent_session_id` remains intact so the next turn can resume.

## Validation

- `go test ./core -run 'TestAgentSessionIdleTimeout|TestProcessInteractiveMessage_SchedulesAgentSessionIdleCloseAfterCleanTurn' -count=1`
- `go test -race ./core -run 'TestAgentSessionIdleTimeout|TestProcessInteractiveMessage_SchedulesAgentSessionIdleCloseAfterCleanTurn' -count=1`
- `go test ./config -run 'TestLoad_.*AgentSessionIdleTimeout' -count=1`
- `go test ./core -count=1`
- `go test ./config -count=1`

`go test ./... -count=1` was also attempted. The packages touched by this PR passed, but the full run is currently blocked in this checkout by existing environment/upstream issues unrelated to this change: missing `web/dist` embed assets, duplicate macOS daemon `CheckLinger` declarations, and a claudecode process-kill test returning `operation not permitted`.

PR #1737, opened 2026-08-26
fix(claudecode): drop --replay-user-messages so /compact and friends stay reachable (#1736)

### LINKED ISSUE #1736: [Bug] /compact fails with "No active session" because Claude Code process exits after each turn
## Description

`/compact` consistently fails with "没有活跃的会话可以压缩。请先发送一条消息。" (No active session to compress) because the Claude Code process exits immediately after processing each message, leaving no active process to receive the compress command.

## Steps to Reproduce

1. Send any message to the bot (Feishu p2p chat)
2. Wait for the bot to reply
3. Send `/compact`
4. Bot responds: "没有活跃的会话可以压缩。请先发送一条消息。"

## Expected Behavior

`/compact` should successfully compress the session context.

## Actual Behavior

`/compact` fails because the Claude Code process has already exited after processing the previous message.

## Root Cause

cc-connect spawns Claude Code with `--replay-user-messages`, which causes Claude Code to exit after processing queued messages instead of staying alive in idle state waiting for input.

Log evidence:
```
00:06:07  /compact → audit: command_executed command=compress → no active process → fails silently
00:06:17  normal message → claudeSession: starting (new process spawned)
00:31:05  agent process exited ← process exits after processing
00:33:52  normal message → claudeSession: starting (new process spawned again)
00:36:03  agent process exited ← exits again
```

The `/compact` command can only succeed if sent during the brief window while Claude Code is still processing a message — which is impractical for users.

## Environment

- cc-connect v1.4.1 (built 2026-07-25)
- Claude Code v2.1.207
- macOS, Feishu p2p chat
- Agent type: claudecode
- Permission mode: bypassPermissions

## Workaround

Configured `auto_compress` as a workaround:
```toml
[projects.auto_compress]
max_tokens = 650000
min_gap_mins = 5
```

However, `auto_compress` has its own issues (#1115 — idle reset never fires).

## Suggestion

Consider one of:
1. Remove `--replay-user-messages` from Claude Code spawn args and use stdin for message delivery only
2. Keep Claude Code process alive between turns (idle mode) instead of exiting
3. Implement `/compact` at the cc-connect level (compress JSONL file directly) without needing an active Claude Code process
4. Add a config option to control whether `--replay-user-messages` is used

chenhg5:
Fixes #1736.

## Problem

cc-connect spawned Claude Code with `--replay-user-messages`. Per Claude Code's protocol, that flag tells the CLI to drain queued stdin messages and exit after the current turn. The next in-session slash command (`/compact`, `/clear`, `/resume`, ...) therefore had no live process to receive it; the following user message spawned a brand-new process and lost all prior context — making `/compact` and friends effectively unreachable.

## Fix

Drop `--replay-user-messages` from the inner spawn args in `agent/claudecode/session.go`. Without it, Claude Code keeps reading stdin until either:

- the engine calls `Close()`, or
- the v1.5.0 `agent_session_idle_timeout_mins` reaper (#1338) reaps the idle process.

Both paths are already wired in cc-connect, so removing the flag is safe. The flag only governed echo-of-user-message behaviour, which cc-connect does not consume.

## Test

Adds `TestNewClaudeSession_NoReplayFlagKeepsProcessAlive` in `agent/claudecode/session_test.go`:

- spawns the CLI substituted by a helper that asserts `--replay-user-messages` is not in argv (exits non-zero if it is),
- drives 100 regular user messages plus interleaved /compact, /clear, /resume turns (103 total) through the live process,
- asserts `cs.Alive()` stays true after every send,
- drains the result stream and verifies exact ordering of the echoed turns (envelope unwrapped, so echoes carry the human text, not the JSON envelope).

The helper reuses `TestHelperProcess` with a new `claude-stdin-echo` mode token. The argv walk now detects the mode token after `--`, with backward-compat fallback to `os.Args[len(os.Args)-1]` for older invocations.

## Validation

- `go test -race ./agent/claudecode/` — pass
- `go test ./...` — pass (pre-existing `cmd/cc-connect` and `web` failures from missing `dist/` build artifact are unrelated)
- `go vet ./agent/claudecode/` — clean
- `gofmt -l` on changed files — clean

## Risk

Low. The flag was redundant — removing it changes process lifecycle from "exit after each turn" to "stay alive until Close()/idle reaper". The engine and #1338 idle timeout already handle both ends.

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -231,11 +231,18 @@ func newClaudeSession(ctx context.Context, workDir, cliBin string, cliExtraArgs 
 	// innerArgs are Claude Code CLI flags — when a wrapper is used with
 	// cmdArgsFlag these get bundled into a single passthrough string.
 	// outerArgs are flags the wrapper itself understands (e.g. --model).
+	//
+	// We intentionally do NOT pass `--replay-user-messages`: that flag tells
+	// Claude Code to drain queued stdin messages and exit, which breaks any
+	// subsequent in-session slash command such as `/compact`, `/clear`, or
+	// `/resume` — issue #1736. Without it the CLI keeps reading stdin until
+	// either the user closes the session or `agent_session_idle_timeout_mins`
+	// (#1338) reaps the idle process; both paths are already handled by the
+	// engine and Close() below.
 	innerArgs := []string{
 		"--output-format", "stream-json",
 		"--input-format", "stream-json",
 		"--permission-prompt-tool", "stdio",
-		"--replay-user-messages",
 	}
 	if !disableVerbose {
 		innerArgs = append(innerArgs, "--verbose")
--- a/agent/claudecode/session_test.go
+++ b/agent/claudecode/session_test.go
@@ -1,8 +1,10 @@
 package claudecode
 
 import (
+	"bufio"
 	"bytes"
 	"context"
+	"encoding/json"
 	"io"
 	"os"
 	"os/exec"
@@ -631,9 +633,9 @@ func makeFiller(n int) string {
 // assistant text reaches the user.
 //
 // Cases covered:
-//  - string content (plain text result)
-//  - array content (Anthropic SDK multi-block: [{type:"text", text:"..."}])
-//  - is_error=true (exit code 1, success=false)
+//   - string content (plain text result)
+//   - array content (Anthropic SDK multi-block: [{type:"text", text:"..."}])
+//   - is_error=true (exit code 1, success=false)
 func TestHandleUserEmitsToolResult(t *testing.T) {
 	cases := []struct {
 		name        string
@@ -649,10 +651,10 @@ func TestHandleUserEmitsToolResult(t *testing.T) {
 				"message": map[string]any{
 					"content": []any{
 						map[string]any{
-							"type":          "tool_result",
-							"tool_use_id":   "toolu_abc",
-							"is_error":      false,
-							"content":       "command output here",
+							"type":        "tool_result",
+							"tool_use_id": "toolu_abc",
+							"is_error":    false,
+							"content":     "command output here",
 						},
 					},
 				},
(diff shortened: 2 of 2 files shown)


chenhg5 (commented):
## QA review for PR #1737 — approve

### Conclusion
**Comment (de-facto approve, blocked by self-author on `--approve`).** Head `809aadcc` is merge-ready. Review decision = **approve**, gating only on maintainer's GitHub merge.

### Verification
- CI: 5/5 green on run [32914322777](https://github.com/chenhg5/cc-connect/actions/runs/32914322777) (lint, unit-test, smoke-test, regression-test, performance-test). `mergeStateStatus` = CLEAN.
- Local: `go vet ./agent/claudecode/...` clean; `gofmt -l` clean; `go build ./agent/claudecode/...` clean; `go test -race -count=1 ./agent/claudecode/` → ok (14.4s, includes the new 103-message regression test).

### What looks good
- The new regression test (`TestNewClaudeSession_NoReplayFlagKeepsProcessAlive`) is exactly the right shape: it doesn't just assert that the helper didn't see the flag — it drives 100 regular messages + `/compact` + `/clear` + `/resume` through the real `newClaudeSession` path and asserts `cs.Alive()` after each send and exact echo ordering. That catches both halves of the issue (flag gone, slash commands reachable).
- Helper's startup check (`os.Exit(3)` if `--replay-user-messages` reappears in argv) acts as a future-proof guard against accidentally re-introducing the flag.
- The expanded `TestHelperProcess` argv walk keeps backward compatibility for the older helper invocations (`TestHelperProcess_StdinEofExit` and friends).
- The new innerArgs comment makes the rationale + dependency on #1338 explicit, which is great for future maintainers.

### 🟠 P2 (suggest, optional in this PR)
- The `claude-stdin-echo` case re-declares `mode := "unknown"` and re-walks `os.Args` a second time (session_test.go ~lines 813-820). The outer `switch mode` already routed into this case using the same logic, so the inner re-walk is dead-code-by-construction. Just use the outer `mode` variable.

### 🔵 P3 (nit, no action needed)
- Local `itoa(int) string` helper works, but `strconv.Itoa` would be the more idiomatic choice if more regression tests land later.
- The gofmt cleanup of a few comment-block alignments (`session_test.go` lines 633-637, 651-655) is fine.

### Risk
- **Low.** The change is a single literal removal in `innerArgs`; without the flag, Claude Code's default behaviour is to keep reading stdin until EOF — which is what `Close()` and the #1338 idle reaper already terminate on. No public API, no config schema, no platform handler changes. Rollback = add the literal back.
- v1.5.0's `agent_session_idle_timeout_mins` is the safety net for long-idle sessions; the fix relies on that being enabled (it is by default per #1338). Worth a quick sanity check on your deployments before merge.

- After rollout, watch `/compact`, `/clear`, `/resume` success rates across Feishu / Telegram / Discord — should jump significantly.
- Verify long-idle memory profile matches v1.5.x pre-PR (it should, since the CLI was always designed to be long-running).

Internal QA decision doc: `doc-20260826-oeby6j` (Multigent docs). Merging is your call — I won't auto-merge.

PR #1641, opened 2026-08-05
fix(codex): retry server overloaded turns

gpYang:
## Summary
- classify Codex server_overloaded/capacity failures as retriable
- retry overloaded/rate-limit turns instead of surfacing raw failures
- render retry notices in rich/progress cards instead of standalone messages
- bypass progress-card edit throttling for retry notices and finalize old retry cards before replay/exhaustion

## Validation
- go build -o /tmp/cc-connect-retry.candidate ./cmd/cc-connect
- go test ./core -run 'TestProcessInteractiveTurnWithRetry_(ReplaysPromptAfterOverloaded|RichCardNoticeUpdatesCard|ProgressCardNoticeUpdatesCard|ProgressCardRetryExhaustionFinalizesFailed|ProgressCardRetryNoticeBypassesThrottle|StopCancelsRetryDelay|ReplaysQueuedPromptAfterOverloaded)|TestProcessInteractiveEvents_ReturnsRetriableErrorWithoutSendingRawError|TestCompactProgressWriter|TestBuildAndParseProgressCardPayload|TestParseProgressCardPayload' -count=3
- go test ./core -run 'TestProcessInteractiveTurnWithRetry_(ProgressCardRetryExhaustionFinalizesFailed|ProgressCardRetryNoticeBypassesThrottle|RichCardNoticeUpdatesCard|ProgressCardNoticeUpdatesCard|StopCancelsRetryDelay)' -count=1 -race
- go test ./agent/codex -run 'TestSend_TurnFailedServerOverloadedIsRetriable|TestSend_WithImages_PassesImageArgsAndDefaultPrompt|TestSend_ResumeWithImages_PlacesSessionBeforeImageFlags|TestSend_HandlesLargeJSONLines|TestGetModelAndReasoningEffort_FromRuntimeConfigWhenUnset' -count=3 -p 1 -parallel=1
- go test ./platform/feishu ./platform/discord ./platform/telegram -count=1
- go test ./core -run TestCUJ_A3_ImageReachesAgent -count=3

Note: built candidate only; did not replace or restart the local cc-connect service.

PR #1740, opened 2026-08-26
fix(core,feishu): suppress fallback spam on card update failure

gpYang:
从原 #1641 分支拆出的独立改动（407d9632）。

compact/card 进度失败时，不再把每个后续工具事件退化为聊天消息（仅 legacy 模式回退），最终回复保持权威。
同时保留上游 tool_max_len 对 progress_style=card 的截断逻辑。

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -5231,8 +5231,13 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 				// the rich-card path. event.ToolInput itself is left untouched.
 				cardToolInput := truncateIf(toolInput, e.display.ToolMaxLen)
 				if !cp.AppendEvent(ProgressEntryToolUse, cardToolInput, event.ToolName, toolMsg) {
-					for _, chunk := range SplitMessageCodeFenceAware(toolMsg, maxPlatformMessageLen) {
-						sendWorkspace(p, replyCtx, chunk)
+					// Compact/card progress is best-effort. If its editable card
+					// fails, do not turn every subsequent tool event into a new
+					// chat message; the final answer remains authoritative.
+					if progressStyleForTarget(p, replyCtx) == progressStyleLegacy {
+						for _, chunk := range SplitMessageCodeFenceAware(toolMsg, maxPlatformMessageLen) {
+							sendWorkspace(p, replyCtx, chunk)
+						}
 					}
 				}
 			}
@@ -5277,7 +5282,7 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 						Success:  event.ToolSuccess,
 					}
 					if !cp.AppendStructured(entry, resultMsg) {
-						if !SuppressStandaloneToolResultEvent(p) {
+						if progressStyleForTarget(p, replyCtx) == progressStyleLegacy && !SuppressStandaloneToolResultEvent(p) {
 							e.sendRaw(p, replyCtx, resultMsg)
 						}
 					}
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -352,6 +352,7 @@ type stubCompactProgressPlatform struct {
 	stubPlatformEngine
 	style          string
 	supportPayload bool
+	updateErr      error
 	previewMu      sync.Mutex
 	previewStarts  []string
 	previewEdits   []string
@@ -379,7 +380,7 @@ func (p *stubCompactProgressPlatform) UpdateMessage(_ context.Context, _ any, co
 	p.previewMu.Lock()
 	p.previewEdits = append(p.previewEdits, content)
 	p.previewMu.Unlock()
-	return nil
+	return p.updateErr
 }
 
 func (p *stubCompactProgressPlatform) BuildRichCard(status CardStatus, title string, steps []ToolStep, markdown string, streaming bool, statusFooter string) string {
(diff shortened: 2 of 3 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Split-out companion to #1641 (already merged): fixes two related but distinct issues — fallback-spam on transient card-update failures, and Feishu having no throttling on burst card edits. +55/-4 across 3 files, 1 commit, 5/5 CI green. The change is small, focused, and clearly anchored in observable user-facing pain (Feishu PATCH rate limits + transient timeouts turning into noisy fallback chat messages). Branch is a month old but that reflects work-saturation, not PR defect.

Review scope:
- `core/engine.go` (+8/-3): two `if progressStyleForTarget(p, replyCtx) == progressStyleLegacy` gates added at the fallback sites in `processInteractiveEvents`
- `core/engine_test.go` (+43/-1): new `TestProcessInteractiveEvents_CardProgressDoesNotSpamFallbackWhenUpdateFails` (with two subtests for `compact` and `card` styles); `stubCompactProgressPlatform.UpdateMessage` now returns `updateErr` instead of always nil
- `platform/feishu/feishu.go` (+4/-0): `ProgressUpdateInterval()` returning 1500ms to opt into the existing `ProgressUpdateThrottler` interface
- Cross-checked `core.ProgressUpdateThrottler` interface exists and is consumed by `progress_compact.go` (`if throttler, ok := p.(ProgressUpdateThrottler); ok { w.minUpdateInterval = throttler.ProgressUpdateInterval() }`) — the Feishu side just plugs into an already-wired capability.
- Cross-checked `progressStyleForTarget` returns the per-reply style (via `progressStyleHintProvider`), defaulting to `progressStyleForPlatform(p)`. The fix correctly uses per-target resolution, so replies routed to different platforms don't accidentally cross-talk.

✅ What looks good:
- **Correct, narrow fix**: the two fallback sites now check `progressStyleForTarget(p, replyCtx) == progressStyleLegacy` before emitting chunked chat messages. For compact/card styles, a transient `cp.AppendEvent` failure is now silent — the final EventResult remains the authoritative reply, exactly as the PR description states.
- **Test is well-structured**: `updateErr` is wired through `stubCompactProgressPlatform.UpdateMessage`, and the test pumps Thinking → ToolUse → ToolResult → Text → Result events through `processInteractiveEvents` with the simulated failure. Asserts `len(sent) == 1 && sent[0] == "done"` — exactly the right contract ("only the final reply reaches the chat"). The two subtests over `compact` and `card` styles pin both modes.
- **Throttler opt-in is one-liner and reversible**: `ProgressUpdateInterval() time.Duration { return 1500 * time.Millisecond }` plus a 2-line doc comment. Discord already implements this interface (per `search_code`), so Feishu is joining an existing pattern, not inventing one.
- **No regression risk for legacy platforms**: the gate is `== progressStyleLegacy`, so platforms that don't set `progress_style` (default = legacy) keep their existing chunked-fallback behavior byte-for-byte.
- **Stack discipline**: this PR is explicitly a split-out from #1641 (per body: "从原 #1641 分支拆出的独立改动"). Smaller surface = easier review = easier rollback.

- **P2 — `slog` trace for the swallowed failure**: when `cp.AppendEvent` fails under compact/card, the fix silently drops the tool event. There is no log line at all, so operators debugging "why is my progress card not updating?" will have nothing to grep for. Consider adding a `slog.Debug` with the tool name and the underlying error before the swallow, e.g. `slog.Debug("compact/card progress append failed, final reply will remain authoritative", "tool", event.ToolName, "error", err)`. The existing `e.sendRaw` path also has no Debug here — same gap.
- **P2 — `ProgressUpdateInterval` value 1500ms is hardcoded**: same pattern as the suggested stallTimeout exposure in #1869 review. If Feishu's PATCH limits ever change (or operators want to tune for high-throughput bots), this requires a rebuild. Consider threading it through the Feishu `Platform` config struct (or `config.Config`) so operators can override without code changes. Discord's value should be checked for parity.
- **P2 — `stubCompactProgressPlatform.UpdateMessage` now returns `p.updateErr` instead of nil**: this is a test-fixture change that *should* be fine for every other test using this stub (none of them assert that `UpdateMessage` returns nil), but worth a quick `grep` to confirm. If any test silently relied on a nil return, they would now break in surprising ways.

🔵 Optional:
- P3: the PR title says "fix(core,feishu)" but the throttler opt-in is one line — the conventional split would be two PRs (one core-only, one feishu-only). Both are tiny and the merged PR is more coherent, so this is just a style note.
- P3: the `progressStyleForTarget` call is computed twice in `processInteractiveEvents` (once for the ToolUse fallback, once for the ToolResult fallback). At current code density the cost is negligible, but a single up-front computation would be marginally cleaner if this function ever lands on a hot path.

❓ Questions:
- None.

Testing / Risk:
- Verified: GitHub CI 5/5 green on head `646f0b74aa` (lint, unit-test, smoke-test, regression-test, performance-test all success).
- New test: `TestProcessInteractiveEvents_CardProgressDoesNotSpamFallbackWhenUpdateFails` with two subtests covering `compact` and `card` styles. Asserts only the final EventResult reaches the chat when progress append fails.
- Existing `TestProcessInteractiveEvents_CardProgressUsesCardTemplate` is adjacent in the file (line 1751) and still passes (per `updateErr` default = nil, so it exercises the happy path).
- Remaining risk: minimal. The two fallback sites are the only ones that produced chat-message spam on transient card failures; both are gated. Legacy platforms are unaffected.
- Latent concern: the `slog` silence on swallowed failures is a debuggability gap. Not a blocker; suggest a follow-up to add `slog.Debug` lines.

Next step:
- Merge. PR is small, isolated, behaviorally narrow, and well-tested. The companion split-out of #1641 was the right call.

Issue #1742, opened 2026-08-26
[Bug] weixin: send budget (burst_limit=4/24h) is applied to the reply path and silently blocks normal conversations after v1.5.0


tian-sandwichlab:
## Summary

#1643 added a per-account weixin send budget (`burst_limit = 4`, `burst_window_secs = 86400`)
to avoid tripping ilink's `ret=-2` throttle. The quota is applied to **both** the proactive-push
path and the **reply** path — the PR describes this as "keeping the push and reply paths
consistent".

On an interactive deployment this means the bot **stops replying after 4 messages per 24h**,
even though — on our account — the gateway does not throttle replies at all. After upgrading
v1.4.1 → v1.5.0 our three weixin bots became effectively unusable within a few hours, and the
failure is invisible to the person chatting: they simply get no answer.

## Evidence: the gateway throttles *pushes*, not *replies*

Two contrasting days from the same bot account, taken from `daemon.log` **before** the upgrade
(v1.4.1, i.e. no local quota — so what we see is the gateway's own behaviour):

**2026-08-15 — zero inbound messages all day, only bot-initiated sends:**

```
09:09:27  ret=-2 ×6  → send
18:11:08  ret=-2 ×6  → send
19:13:54  ret=-2 ×6  → send
22:01:16  send
```

Every proactive send tripped `ret=-2`. (Six lines per episode is v1.4.1's `3×500ms` retry loop
plus the "incomplete delivery notice" — exactly the escalation #1643 fixed.)

**2026-08-17 — 54 inbound messages, 36 outbound replies, same account:**

```
09:50:36  IN
09:51:48  OUT
09:55:54  IN
09:56:17  OUT
...
```

Every single OUT is preceded by an IN. All 36 were delivered. There was exactly one throttle
episode (16:59–17:00) and it recovered immediately on the next send.

**36 replies in one day is 9× the default budget**, and this pattern ran for weeks without
the account being throttled.

We are not disputing the 5–6/24h figure in #1643 — we reproduce it exactly for *proactive*
sends. The claim here is narrower: that measurement does not appear to extend to replies, and
the PR's own opening line scopes the problem to pushes:

> Fixes weixin (ilink) `sendMessage ret=-2 "prepare failed"` failures that blocked
> **proactive pushes (cron/timer)**.

## Impact after upgrading to v1.5.0

```
12:20:04  INFO  message received  platform=weixin ...
12:23:45  ERROR platform send failed  platform=weixin
[...]

Issue #1712, opened 2026-08-19
[Bug] 连接个人微信，手机端发送四条消息后就就不再回复


Ahanobody:
### cc-connect Version

v1.5.0

### Operating System

Windows (WSL)

### Agent Type

Codex (OpenAI)

### Platform

N/A

### Installation Method

npm (npm install -g cc-connect)

### Bug Description

连接个人微信后，通过微信端向agent发送4条消息并得到回复后，再次发送信息便不产生回复
日志显示
msg="platform send failed" platform=weixin error="weixin: send budget exhausted (4 messages in the last 24h0m0s); ilink throttles the bot after roughly 5-6 sends per window — reduce messages or re-login later" content_len=179

### Steps to Reproduce

通过AI提示词进行部署，连接个人微信后可正常交互。codex通过ccswitch使用mimo模型

### Expected Behavior

-

### Actual Behavior

发送四次消息后无回复

### Configuration (config.toml)

```toml

```

### Logs / Error Output

```shell
time=2026-08-19T18:58:17.935+08:00 level=INFO msg="message received" platform=weixin msg_id=[REDACTED_MSG] session=weixin:dm@[REDACTED_INSTANCE] user=[REDACTED_USER] content_len=6 has_images=false has_audio=false has_files=false
time=2026-08-19T18:58:17.936+08:00 level=INFO msg="processing message" platform=weixin user=[REDACTED_USER] session=[REDACTED_SESSION]
time=2026-08-19T18:58:24.036+08:00 level=INFO msg="turn complete" session=[REDACTED_SESSION] agent_session=[REDACTED_AGENT] msg_id=[REDACTED_MSG] tools=0 response_len=81 turn_duration=6.10043s input_tokens=0 output_tokens=0 silent=false
time=2026-08-19T18:58:24.036+08:00 level=ERROR msg="platform send failed" platform=weixin error="weixin: send budget exhausted (4 messages in the last 24h0m0s); ilink throttles the bot after roughly 5-6 sends per window — reduce messages or re-login later" content_len=179
```

chenhg5:
Thanks for reporting — this is the same v1.5.0 weixin quota regression as #1716. Marking as duplicate and tracking the fix there. Please follow #1716 for progress and the upcoming v1.5.x release notes.

PR #1744, opened 2026-08-26
fix(weixin): let burst_limit = 0 actually disable the send quota (#1716)

### LINKED ISSUE #1716: [Bug] weixin quota: default window (24h) caps the bot at 4 messages/day, and burst_limit=0 does not disable it
### cc-connect Version

v1.5.0 (Homebrew, macOS)

### Operating System

macOS (Darwin 24.6.0, Apple Silicon)

### Agent Type

Claude Code

### Platform

WeChat / ilink

### Installation Method

Homebrew

### Bug Description

Two problems with the weixin send quota added in v1.5.0.

**1. The default window is wrong.** Defaults are `burst_limit = 4` + `burst_window_secs = 86400`, i.e. **4 outbound messages per 24 hours**. But cc-connect's own config comment describes the upstream rule as a *short-window* burst throttle:

> the gateway throttles the bot after roughly 5-6 separate messages within a **short window** (multi-chunk sends are fine)

Implementing a burst guard with a 24h window turns "don't send 5-6 back to back" into a hard daily cap of 4. The bot goes silent for the rest of the day after four replies.

**2. `burst_limit = 0` does not disable the quota.** The config comment says `0 disables the quota`. Setting it to `0` and restarting still produces `send budget exhausted (4 messages in the last 24h0m0s)` — the value appears to be treated as unset and falls back to the default. Only an explicit large value works.

Also worth noting: the message is **dropped, not queued**. Nothing is visible on the messaging side; the only trace is `level=ERROR` in the local log. Conversations die mid-thread and neither side can tell why.

### Steps to Reproduce

1. v1.5.0 with a weixin platform, no quota keys set.
2. Send 4 outbound messages; the 5th is dropped with `send budget exhausted`.
3. Set `burst_limit = 0`, restart — still dropped with the same error.

### Expected Behavior

- Default `burst_window_secs` should match ilink's actual short window (~600-1800s), not 86400.
- `burst_limit = 0` should disable the quota as documented.
- Ideally queue and pace instead of dropping, so no message is silently lost.

### Actual Behavior

4 messages per 24h, silently discarded afterwards, and the documented escape hatch (`0`) has no effect.

```toml
[projects.platforms.options]
allow_from = "*"
token = "${WEIXIN_BOT_TOKEN}"
base_url = "https://ilinkai.weixin.qq.com"
```

### Logs / Error Output

```shell
level=ERROR msg="platform send failed" platform=weixin error="weixin: send budget exhausted (4 messages in the last 24h0m0s); ilink throttles the bot after roughly 5-6 sends per window — reduce messages or re-login later"
```

Note: in three weeks of use before v1.5.0, ilink's real `ret=-2` throttle was never triggered at this traffic level — the guard fires long before the thing it guards against.

### Additional Context

Working workaround (`0` does not work):

```toml
burst_limit = 100
burst_window_secs = 600
```

Related: #1712 — same symptom on Windows WSL + Codex, so it is not OS- or agent-specific.

meikocho1:
Fixes #1716. Also unblocks #1712 and #1742.

## The bug

`New()` reads both quota options through `pickInt()`, which returns `0` for an
absent key **and** for an explicitly configured `0`. Forty lines later:

```go
if burstLimit <= 0 {
	burstLimit = defaultBurstLimit
}
if burstWindow <= 0 {
	burstWindow = defaultBurstWindowSecs
}
```

So `burst_limit = 0` was rewritten back to `4`, and the quota stayed on.

That matters because `0` is the only documented way out. `config.example.toml`
says, in both languages:

> `0 disables the quota.`
> `设为 0 关闭配额。`

With the shipped defaults (`burst_limit = 4`, `burst_window_secs = 86400`) a
bot stops sending after four outbound messages per 24h. The send is dropped,
not queued — the only trace is a local `level=ERROR` line, so the person
chatting just stops getting answers. Three reports so far (#1712, #1716,
#1742), and each one hit the same wall: setting `burst_limit = 0` and
restarting changed nothing.

## The fix

Resolve both options by key presence, matching the `dedup_enabled` idiom
already used a few lines below:

- key absent → default (unchanged behaviour)
- explicit `0` → quota disabled, as documented
- negative → clamped to disabled (unchanged)
- explicit positive → honoured (unchanged)

The `<= 0` fallback block is removed since the defaults are now applied at
the point of parsing.

Second commit hunk corrects the `config.example.toml` prose: it described the
gateway throttle as firing on "5-6 separate messages within a **short**
window", while the default window is 86400s. As written the two halves
contradict each other, which is what led #1716 to read the daily cap as a
bug in itself. The note now states plainly that the shipped defaults are a
cap of 4 outbound messages per day.

## Not changed here — needs your call

#1742 argues the quota should apply to the proactive-push path only, not to
replies, with log evidence from a pre-v1.5.0 account: 36 replies delivered in
one day with a single throttle episode, versus every proactive send tripping
`ret=-2` on a day with zero inbound messages. #1643 applied it to both paths
deliberately ("keeping the push and reply paths consistent"), and the
`~5-6/window` figure is your measurement, so I did not touch either the
push/reply split or the `86400` default. Happy to follow up in a separate PR
if you want the reply path exempted or the default window shortened.

## Tests

`TestNew_BurstQuotaOptions` covers all five resolution cases and asserts the
quota is genuinely off (via `checkSendQuota`) whenever it is configured off.

Verified against `golang:1.25`:

- before the fix, the new test fails on `explicit zero limit`, `explicit zero
  window` and `negative values` — e.g. `sendQuotaLimit = 4, want 0`
- after the fix, `go vet ./platform/weixin/` is clean and
  `go test ./platform/weixin/` passes

(`go build ./...` fails on `web/embed.go: pattern all:dist: no matching files
found` on a fresh checkout without a frontend build — pre-existing and
unrelated to this change.)

---

## 中文摘要

`pickInt()` 无法区分「未配置」与「显式配置为 0」，随后的 `<= 0` 回退把显式的
`0` 改回默认值 4，导致 `config.example.toml` 中承诺的「设为 0 关闭配额」完全
失效。默认配置下机器人每 24 小时只能外发 4 条消息，超出后直接丢弃（仅本地
`ERROR` 日志），用户侧毫无提示 —— 已有 #1712、#1716、#1742 三份报告。

本 PR 改为按 key 是否存在来解析（与下方 `dedup_enabled` 写法一致）：未配置走
默认值，显式 `0` 或负数关闭配额，正数按配置生效。同时修正
`config.example.toml` 中「短窗口」与 86400 秒默认值自相矛盾的说明。

#1742 提出的「配额应只作用于主动推送、不作用于回复」涉及 #1643 的既定设计
选择，本 PR 未改动，留待您决定。

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1475,13 +1475,17 @@ app_secret = "your-feishu-app-secret"
 # proxy_password = ""
 #
 # Send-volume quota vs ilink's burst throttle (sendMessage ret=-2 "prepare failed"):
-# the gateway throttles the bot after roughly 5-6 separate messages within a short
-# window (multi-chunk sends are fine), and the penalty escalates with every attempt
-# made while active. The quota paces separate messages so the bot stays below the
-# trigger. 0 disables the quota.
-# 发送量配额（防 ilink 突发节流 ret=-2）：网关在短窗口内约 5-6 条独立消息即节流
+# the gateway throttles the bot after roughly 5-6 separate messages per window
+# (multi-chunk sends are fine), and the penalty escalates with every attempt made
+# while active. The quota paces separate messages so the bot stays below the
+# trigger. Note the default window is 24h, so the defaults below cap the bot at 4
+# outbound messages per day — raise burst_limit / shorten burst_window_secs for
+# interactive deployments, or set burst_limit = 0 to disable the quota entirely.
+# 发送量配额（防 ilink 突发节流 ret=-2）：网关在单个窗口内约 5-6 条独立消息即节流
 # （多切块发送不计数），且节流期内每次尝试都会升级惩罚。配额对独立消息限速，
-# 使其低于触发阈值。设为 0 关闭配额。
+# 使其低于触发阈值。注意默认窗口为 24 小时，即下方默认值会将机器人限制为
+# 每天 4 条外发消息；交互式部署请调大 burst_limit 或调小 burst_window_secs，
+# 或设 burst_limit = 0 完全关闭配额。
 # burst_limit = 4           # max separate messages per window / 每窗口最多独立消息数
 # burst_window_secs = 86400 # window length (default 24h) / 窗口时长（秒，默认24小时）
 #
--- a/platform/weixin/weixin.go
+++ b/platform/weixin/weixin.go
@@ -151,12 +151,20 @@ func New(opts map[string]any) (core.Platform, error) {
 	}
 	lp := pickInt(opts["long_poll_timeout_ms"])
 
-	// Send-volume quota (see defaultBurstLimit constants). 0 disables the quota.
-	burstLimit := pickInt(opts["burst_limit"])
+	// Send-volume quota (see defaultBurstLimit constants). An explicitly
+	// configured 0 disables the quota as documented in config.example.toml;
+	// only an absent key falls back to the default (issue #1716).
+	burstLimit := defaultBurstLimit
+	if v, ok := opts["burst_limit"]; ok {
+		burstLimit = pickInt(v)
+	}
 	if burstLimit < 0 {
 		burstLimit = 0
 	}
-	burstWindow := pickInt(opts["burst_window_secs"])
+	burstWindow := defaultBurstWindowSecs
+	if v, ok := opts["burst_window_secs"]; ok {
+		burstWindow = pickInt(v)
+	}
 	if burstWindow < 0 {
 		burstWindow = 0
 	}
@@ -191,13 +199,6 @@ func New(opts map[string]any) (core.Platform, error) {
 	cdnHttpClient := &http.Client{
 		Timeout:   60 * time.Second,
 		Transport: &http.Transport{Proxy: nil},
-	}
-
-	if burstLimit <= 0 {
-		burstLimit = defaultBurstLimit
-	}
-	if burstWindow <= 0 {
-		burstWindow = defaultBurstWindowSecs
 	}
 
 	// dedup_enabled (default true) and dedup_window_seconds (default 300s, the
--- a/platform/weixin/weixin_test.go
+++ b/platform/weixin/weixin_test.go
@@ -571,6 +571,77 @@ func TestCheckSendQuota_DisabledWhenZero(t *testing.T) {
 	}
 }
 
+// TestNew_BurstQuotaOptions verifies how burst_limit / burst_window_secs are
+// resolved from platform options (issue #1716): an absent key falls back to the
+// documented default, while an explicit 0 disables the quota as config.example.toml
+// promises. Before the fix an explicit 0 was indistinguishable from "unset" and
+// silently became the default, capping bots at 4 messages/24h with no escape hatch.
+func TestNew_BurstQuotaOptions(t *testing.T) {
+	cases := []struct {
+		name       string
+		opts       map[string]any
+		wantLimit  int
+		wantWindow time.Duration
+	}{
+		{
+			name:       "unset falls back to defaults",
+			opts:       map[string]any{"token": "tok"},
+			wantLimit:  defaultBurstLimit,
+			wantWindow: time.Duration(defaultBurstWindowSecs) * time.Second,
+		},
+		{
+			name:       "explicit zero limit disables the quota",
+			opts:       map[string]any{"token": "tok", "burst_limit": 0},
+			wantLimit:  0,
+			wantWindow: time.Duration(defaultBurstWindowSecs) * time.Second,
+		},
+		{
+			name:       "explicit zero window disables the quota",
+			opts:       map[string]any{"token": "tok", "burst_window_secs": 0},
+			wantLimit:  defaultBurstLimit,
+			wantWindow: 0,
+		},
+		{
+			name:       "negative values are clamped to disabled",
+			opts:       map[string]any{"token": "tok", "burst_limit": -1, "burst_window_secs": -5},
+			wantLimit:  0,
+			wantWindow: 0,
+		},
+		{
+			name:       "explicit values are honored",
+			opts:       map[string]any{"token": "tok", "burst_limit": 30, "burst_window_secs": 600},
+			wantLimit:  30,
+			wantWindow: 600 * time.Second,
+		},
+	}
+	for _, tc := range cases {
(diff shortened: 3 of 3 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- High-quality bugfix that closes the long-standing "burst_limit=0 does nothing" defect documented in `config.example.toml`. The fix is minimal (two `pickInt` calls swapped for `if v, ok := opts[key]; ok` blocks; removal of the `<= 0` fallback), the regression test directly covers the reported failure modes, and the CI is 5/5 green on commit `245213869381439218735fff29197ea5f5d29b38`, mergeable, base = main, no conflicts. The accompanying `config.example.toml` prose correction is the right fix-up to ship alongside, because the original wording was what primed #1716 to misread the daily cap as a bug in itself.

Review scope:
- Reviewed `platform/weixin/weixin.go` (`New()` quota option parsing), `platform/weixin/weixin_test.go` (new `TestNew_BurstQuotaOptions` + the existing `TestCheckSendQuota_DisabledWhenZero` baseline), and `config.example.toml` (the prose correction around the weixin burst-quota block).
- Confirmed the `pickInt` semantics (returns `0` for `nil`, missing key, `string`, `bool`, etc.) and the `sendQuotaLimit` (`int`) / `sendQuotaWindow` (`time.Duration`) field types.
- Confirmed the existing `dedup_enabled` idiom the PR points to (`if v, ok := opts["dedup_enabled"]; ok { ... }`) actually exists a few lines below the changed block, so the fix genuinely matches an in-repo pattern rather than inventing a new one.
- Focused on correctness, key-presence semantics, the clamp-to-disabled branch for negative values, and the doc-prose change.

✅ What looks good:
- The root cause is correctly diagnosed: `pickInt` collapses "absent" and "explicit 0" into the same value (`0`), and the post-parse `<= 0` fallback then promotes that `0` back to the default, so the documented "0 disables the quota" escape hatch was unreachable. The fix preserves the documented contract by switching to key-presence checks at parse time, which is the only correct way to disambiguate absent vs. zero in a `map[string]any` settings bag.
- The fix mirrors the `dedup_enabled` idiom already in the file (`if v, ok := opts["dedup_enabled"]; ok { ... }`), so a future reader sees consistency rather than a one-off trick. The two quota keys are handled symmetrically — the same parse-and-default pattern, the same `< 0 → 0` (disabled) clamp — which is the right factoring.
- Removing the `<= 0` fallback block is correct: defaults are now applied at parse time, and a `0` that survives to the platform struct unambiguously means "disabled". This makes `sendQuotaLimit == 0 || sendQuotaWindow == 0` a safe predicate for "quota off" anywhere downstream, including `checkSendQuota` which the new test exercises directly.
- The negative-value clamp (`if burstLimit < 0 { burstLimit = 0 }`) is preserved. That matches the contract "negative → clamp to disabled", which is what `TestNew_BurstQuotaOptions` asserts for the `negative values are clamped to disabled` case. No silent negative-vs-default semantics — both produce a disabled quota, which is what a defensive platform config should do.
- The new test covers all five resolution cases the PR describes (unset / explicit-0-limit / explicit-0-window / negative / explicit-positive) and goes a step further by directly invoking `checkSendQuota` `defaultBurstLimit+5` times in the "disabled" cases. That is exactly the failure mode reported in #1716/#1712/#1742 — the bot sending several messages and being silently throttled — so this assertion is meaningful, not cosmetic.
- The doc prose change is the right scope: it no longer claims a "short window" that contradicted the `86400s` default, and it tells the operator explicitly that the shipped defaults cap the bot at 4 outbound messages per day, plus the three knobs (`burst_limit`, `burst_window_secs`, `burst_limit = 0`). This is what the user needed to read in the first place to understand the cap.
- The PR body explicitly carves out #1742 (push-vs-reply split, default window length) as a separate decision for the maintainer, with a clear "I did not touch X because of Y, happy to follow up" framing. That is the right way to ship a scoped bugfix without quietly reopening a previously-debated design choice.

🚨/🔴 Must fix:

🟠 Should improve:
- None worth blocking on.

🔵 Optional:
- `if burstLimit < 0 { burstLimit = 0 }` and the equivalent for `burstWindow` are now sitting between two `pickInt` calls rather than next to the `<= 0` block they replaced. Consider hoisting both clamps into a single helper (e.g. `clampQuotaInt(v, defaultVal) int`) so the "negative → disabled" contract is documented once instead of twice. Cosmetic.
- `TestCheckSendQuota_DisabledWhenZero` still exists and is now somewhat redundant with the new `TestNew_BurstQuotaOptions` "disabled" sub-cases. Not worth deleting — it tests `checkSendQuota` directly, whereas the new test goes through `New()` — but worth a one-line comment on the older test pointing to the new one as the parse-path coverage, so the next refactor doesn't accidentally drop either layer.
- The PR's prose-correction commit is correct, but `config.example.toml` still leaves the `burst_limit = 4 / burst_window_secs = 86400` defaults in place, which the doc now says will throttle the bot at 4 outbound/day. If 4/day is genuinely wrong as a default for any non-trivial deployment (which #1742's logs suggest), a follow-up PR that raises `defaultBurstLimit` to something more humane (e.g. 30/24h or 100/1h) would be a more user-visible improvement than the prose change. Strictly out of scope here — flagging only.

❓ Questions:
- None.

Testing / Risk:
- `go vet ./platform/weixin/` clean per the PR body's local verification.
- The new test fails on the pre-fix code on exactly the three cases the bug report describes (explicit zero limit, explicit zero window, negative values), which is the strongest possible regression coverage — the test would have caught the bug at PR-#1716 time if it had existed.
- Risk scope is bounded to the `weixin` platform's `New()` parser. Other platforms with quota / rate-limit options (`feishu`, `telegram`, `slack`, etc.) have their own opt-parsing code; this PR does not touch them, so any similar bug there remains to be found and fixed separately. Worth a sweep in a future PR.
- Backward compatibility: an existing user who relied on the buggy "explicit 0 silently becomes 4" behavior to *enforce* a 4/day cap will see a behavior change (their explicit `0` now genuinely disables the cap). That is exactly the documented behavior and the right fix, but it is a behavior change for a narrow edge case. Worth a release-notes line.

Next step:
- Recommend maintainer merge. No follow-up PR required from this PR; the optional follow-ups (helper extraction, default-limit bump, sweep other platforms for the same idiom) can land separately.

PR #1745, opened 2026-08-26
feat(core): add message.finalized hook event for post-reply automation

gpYang:
## 背景

实现 Agent 完成一轮代码改动后的自动代码审核（AI review）需要"最终回复发送完成"这一生命周期事件。当前只有 `message.sent`（发送前），无法在回复完成后触发外部脚本。

## 改动

- **core/hooks.go**：新增 `message.finalized` 事件类型；`HookEvent` 增加 `turn_id` / `workspace` / `source` / `internal` / `reply_kind` 字段；`eventToEnv` 导出对应 `CC_HOOK_*` 环境变量
- **core/engine.go**：
  - 前台最终回复（rich card / stream preview / 普通分片 / suppress-duplicate / fallback 路径）完整发送成功后 Emit，`source=agent.final_reply`
  - 后台回复（unsolicited reader）发送成功后 Emit，`source=agent.background_reply`
  - 静默回复（NO_REPLY）、空回复、发送失败不触发
  - `interactiveState` 持久化 `ccSessionKey` 供 hook 上下文使用
- **core/hooks_test.go / core/engine_test.go**：事件类型、env 变量、发送后触发顺序、后台回复触发、发送失败不触发等测试
- **daemon/check_linger_other.go**：build tag 从 `!linux` 改为 `!linux && !darwin`，修复 macOS 上与 `launchd.go` 的 `CheckLinger` 重复声明导致的本地构建失败

## 测试

- `go build ./...` 通过
- 新增测试全部通过（`TestMessageFinalizedEventType`、`TestEventToEnv`、`TestProcessInteractiveEvents_FinalizedHookRunsAfterReplySend`、`TestUnsolicitedReader_EmitsFinalizedHookAfterReplySend`、`TestUnsolicitedReader_DoesNotEmitFinalizedHookWhenReplySendFails`）
- hooks 相关测试全部通过

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -412,7 +412,8 @@ level = "info" # debug, info, warn, error
 #
 # Supported events / 支持的事件:
 #   message.received   — user message arrives / 用户消息到达
-#   message.sent       — agent reply sent / Agent 回复发送完成
+#   message.sent       — legacy pre-delivery agent reply hook / 兼容旧版发送前事件
+#   message.finalized  — final reply send completed / 最终回复发送完成
 #   session.started    — agent session spawned / Agent 会话启动
 #   session.ended      — agent session stopped / Agent 会话结束
 #   cron.triggered     — cron job fires / 定时任务触发
--- a/config.example.toml
+++ b/config.example.toml
@@ -414,6 +414,16 @@ level = "info" # debug, info, warn, error
 #   message.received   — user message arrives / 用户消息到达
 #   message.sent       — legacy pre-delivery agent reply hook / 兼容旧版发送前事件
 #   message.finalized  — final reply send completed / 最终回复发送完成
+#                       (CC_HOOK_CHANGED=true only when the turn actually
+#                       modified the git working tree. Semantics: false does NOT
+#                       mean "no change" — non-git workspaces, fresh repos
+#                       without commits, or git failures always report false.
+#                       Fingerprint windows are serialized per workspace, so
+#                       concurrent turns cannot pollute each other's flag.)
+#                       (CC_HOOK_CHANGED 仅在该轮确实改动 git 工作树时为 true。
+#                       注意：false 不等于"无改动"——非 git 工作区、无提交的
+#                       新仓库或 git 失败时恒为 false。指纹窗口按工作区互斥，
+#                       并发轮次不会互相污染该标志。)
 #   session.started    — agent session spawned / Agent 会话启动
 #   session.ended      — agent session stopped / Agent 会话结束
 #   cron.triggered     — cron job fires / 定时任务触发
--- a/core/engine.go
+++ b/core/engine.go
@@ -455,10 +455,15 @@ type Engine struct {
 	workspaceInitAllowLocalPaths bool
 	workspaceBindings            *WorkspaceBindingManager
 	workspacePool                *workspacePool
-	initFlows                    map[string]*workspaceInitFlow // workspace channel key → init state
-	initFlowsMu                  sync.Mutex
-	sendWorkDirMu                sync.RWMutex
-	sendWorkDirs                 map[string]string // sessionKey → work_dir assigned by send --cwd
+	// workspaceFpMu/workspaceFpLocks serialize the git fingerprint window per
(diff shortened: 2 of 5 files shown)


chenhg5 (approved):
## Conclusion: Approve

Solid addition. The new `message.finalized` hook event unlocks a long-awaited automation surface (auto code review, metrics, post-send tests) without breaking existing `message.sent` consumers. Tests are the standout: the marker-file ordering test proves hook runs **strictly after** `platform.send`, the queued-message re-baseline test catches a subtle correctness bug, and the per-workspace fingerprint lock test defends `Changed` against concurrent-turn pollution.

### What looks good

- **Opt-in / backward compatible**: new event type only fires for users who configure it; `message.sent` semantics unchanged. Zero migration risk for existing hook consumers.
- **Strict ordering invariant**: `finalized && !isSilent && !hadEmptyReply` gate is tested with marker files, not just "count == 1" — this catches race regressions a future refactor would otherwise reintroduce.
- **Per-workspace fingerprint lock** (`workspaceFpMu` + `workspaceFpLocks`) serializes the baseline-vs-current git diff window so two concurrent turns in the same workspace don't corrupt each other's `Changed` flag. Lazy map init inside lock is correct.
- **5s timeout on `workspaceGitFingerprint`**: prevents a hung git/network FS from stalling the event loop. The baseline + emit scan adds at most ~2× git status per turn; on small repos this is negligible.
- **Queued-message re-baseline** (`TestProcessInteractiveEvents_QueuedMessageGetsFreshBaseline`): catches the subtle case where a follow-up queued message would otherwise inherit the previous turn's dirty baseline and report `Changed=true` even though nothing in *this* turn touched files.
- **Bilingual docs** in `config.example.toml` for both EN and ZH readers.
- **Bonus**: the `daemon/check_linger_other.go` build tag fix (`!linux` → `!linux && !darwin`) unblocks local macOS daemon builds that were previously broken by the duplicate `CheckLinger` declaration with `launchd.go`. Nice catch.

### 🟠 Should fix (not blocking)

- **P2: `Changed` semantics doc vs. implementation drift.** In `core/engine.go`, the emit uses `Changed: turnStartGitFingerprint != "" && turnStartGitFingerprint != workspaceGitFingerprint(ctx, hookWorkspace)`. The config comment claims "git failures always report false", but the actual behavior is:
  - First fingerprint fails (`""`), second fails → `Changed = false` ✅ matches doc
  - First succeeds (`X`), second **fails** (transient git error → returns `""`) → `Changed = true` ❌ docs say false
  
  The transient-failure case (second call flakes) will be reported as "workspace changed" and may trigger false auto-reviews. Either:
  - **A** (preferred): `Changed: turnStartGitFingerprint != "" && turnStartGitFingerprint == workspaceGitFingerprint(...)` (only "both succeed and equal" → false, anything else → false; conservative)
  - **B**: keep current behavior, update the doc comment to "Changed is `true` only when both fingerprints succeed and differ"
  
  Pick one, but make doc and code consistent.

### 🔵 Optional

- **P3**: `workspaceGitFingerprint` swallows all git errors. Consider emitting a structured log (e.g. `slog.Warn` with the stderr) when git fails, so ops can diagnose transient failures instead of being invisible.
- **P3**: Per-workspace fingerprint lock serializes across sessions on the same workspace. This is by design but worth a comment in the code calling it out for future maintainers.
- **P3**: Future enhancement: short-window git fingerprint cache (invalidate on next hook event) to amortize the cost for high-frequency turn workloads. Not for this PR.

### ❓ Questions

- Is there a docs page (e.g. `docs/hooks.md`) that should also be updated, or is `config.example.toml` the single source of truth for hook event documentation? If the former exists, please add a brief mention there too.

### Testing / Risk

- **Verified**: CI 5/5 green on `4b5ab052`; 7 new tests in `core/engine_test.go` cover the critical invariants. The marker-file ordering test in particular is a high-signal design.
- **Remaining risk**: P2 above is the only correctness gap. Performance overhead is bounded by the 5s timeout.

### Next step

Please address the P2 docs/implementation drift (either A or B). After that, this is merge-ready. Great contribution overall — the marker-file ordering test is exactly the kind of high-fidelity test we want for lifecycle hooks.

PR #1588, opened 2026-07-22
feat(feishu): on-demand download of quoted files when user mentions bot (#1560)

chenhg5:
## Summary

Fixes #1560 — when a Feishu group member uploads a file and then replies/quotes it while @-mentioning the bot, the bot could not see the file because cc-connect does not persist message bodies.

This PR introduces on-demand quoted-file retrieval that runs **only** when the reply explicitly @-mentions the bot AND the original file's Feishu sender matches the user who triggered the mention.

## Acceptance criteria mapping

| PM criterion | Where it is enforced |
| --- | --- |
| Detect reply/quote metadata (parent_id / file_key) | `fetchSingleMessage` now extracts metadata from `file` and `media` chain items into a new `quotedFileMeta` struct |
| Only fetch when triggering message quotes a file AND mentions the bot | `filterQuotedFilesForUser` gates on `isBotMentioned`. If the gate fails, zero calls hit `/open-apis/im/v1/messages/:id/resources/:file_key` |
| Same IM user / same session only (no cross-user) | `filterQuotedFilesForUser` also drops any meta whose `senderID` differs from the triggering `userID` |
| Transient in-session state — dropped on offline | The new `quotedFileMeta` lives only inside the per-invocation chain slice; nothing is written into `~/.cc-connect/sessions/*.json` |
| Route through existing attachment processing | Once filtered + downloaded, files are attached as `core.FileAttachment` on `core.Message.Files`, the same path `SaveFilesToDisk` consumes (issue #1552 alignment) |
| Tests for mention+quote fetch / quote-only no fetch / ordinary no fetch | `platform/feishu/quote_file_test.go` (new) — see `TestDispatchMessageQuotedFileAcceptance` |
| Validate required Feishu scopes and rate-limit behavior | Reads require `im:message` (or `im:message:readonly`) + `im:resource`. Feishu returns explicit scope errors which are logged and the file is skipped — bounded by `withTransientRetry`. Rate-limit pressure is bounded *structurally* because no fetch happens unless the gate passes |

## Non-goals

- No full message-history persistence (the reporter's chosen option `c`).
- No cross-user file retrieval (privacy default safe even though the reporter only asked for same-user).
- No unrelated attachment-collision fixes — that lives in PR #1557 (issue #1552).

## Files touched

[...]

PR #1743, opened 2026-08-26
fix(weixin): split send budget by path — reply bypasses, push retains (Issue #1742)

chenhg5:
Fixes #1742.

## Problem

v1.5.0's #1643 burst_limit=4/24h was applied to **both** the proactive-push path and the **reply** path. On interactive weixin deployments ilink does not throttle replies, so the local quota silently bricked bots at 4 replies per 24h — a user would send a message, the agent would answer, and the local quota would drop the response with no chat-side notification. The failure is invisible to the user.

## Fix

Scope the burst budget to where the gateway actually throttles us (push + media) and let replies through unconditionally:

- New `sendPath` enum (`sendPathReply` | `sendPathPush`) threaded through `checkSendQuota` / `sendChunks` / `sendSingleItem`. `Reply()` routes to the reply path; `Send()` and outbound media route to the push path.
- `checkSendQuota` early-returns nil on `sendPathReply` — replies never consume bucket space, never increment the over-budget counter, and cannot be blocked regardless of how many the user already exchanged.
- On push-path exhaustion, log a structured ERROR `weixin: push_path_budget_exceeded` (fields `path`/`used`/`limit`/`window`/`hint`) and increment `PushBudgetExceededTotal()` so operators can see push-budget pressure in `cc-connect doctor` / telemetry. Never a silent drop.
- Error message updated to say "push messages" instead of "messages" so the cause is obvious from `daemon.log` alone.

## Tests

- `TestCheckSendQuota_ReplyBypassesQuota` — 50 reply calls under `burst_limit=1` all pass; counter stays at 0; `sendQuotaTimes` stays empty.
- `TestCheckSendQuota_PushStillEnforced` — push path still trips on the 5th send at limit=4; counter goes to 1.
- `TestCheckSendQuota_PushBlockedIncrementsCounter` — repeated over-budget attempts increment monotonically.
- `TestSendChunks_ReplyIgnoresBudget` — end-to-end: 50 reply `sendChunks` all hit the API at `burst_limit=1`.
- `TestSendChunks_PushRespectsBudget` — push path through `sendChunks` blocks the 2nd attempt; over-budget send never reaches the API.
- `TestSendSingleItem_PushRespectsBudget` — media path consults the push budget like the text path.

## Validation

- `go test -race ./platform/weixin/` — pass (3.97s)
- `go vet ./platform/weixin/` — clean
- `gofmt -l` on changed files — clean

## Risk

[...]

PR #1746, opened 2026-08-26
fix(feishu): download large resources via HTTP Range to bypass code=234037 (#1741)

### LINKED ISSUE #1741: [Bug] Feishu: large file download fails with code=234037 although the resource API supports HTTP Range
## Summary

Downloading a file attachment from Feishu fails for large files with
`code=234037 Downloaded file size exceeds limit.`

The Feishu resource API **does** serve the file when the request carries an HTTP `Range`
header, but cc-connect issues a plain whole-file `GET`. As a result any file above Feishu's
whole-file cap is completely unreachable, even though partial download is permitted.

Both the direct-download path and the new quoted-file path added in #1588 are affected.

## Environment

| | |
| --- | --- |
| cc-connect | `v1.4.1` (commit `5d4c96dd`) and `v1.5.0` (commit `17c61062`) — **both affected** |
| Platform | Feishu (`open.feishu.cn`), group chat, long connection |
| OS | macOS (arm64), launchd daemon |
| Failing file | `142,735,026` bytes (~136 MB) `.zip` |
| Working files | 3.2 MB and 13 MB `.zip` (same chat, same bot) |

## Observed behaviour

**Path 1 — user sends the file directly** (`platform/feishu.(*Platform).downloadResource`):

```
level=INFO  msg="feishu: file received" user=ou_REDACTED file_key=file_v3_REDACTED file_name=large-archive.zip
level=ERROR msg="feishu: download file failed" error="feishu: resource API code=234037 msg=Downloaded file size exceeds limit."
```

The whole inbound message is then dropped — there is no subsequent `message received` /
`processing message` / `turn complete`. The agent never learns that a file was sent.

**Path 2 — user quotes the file and @-mentions the bot** (the on-demand retrieval from #1588,
v1.5.0):

```
level=WARN msg="feishu: download quoted file failed; skipping this entry" error="feishu: resource API code=234037 msg=Downloaded file size exceeds limit."
```

Same API error, only downgraded to `WARN` and the entry is skipped.

Round-trip is ~0.7 s in both cases, i.e. the request is rejected up front — this is not a
timeout and not a transfer failure.

## Root cause — the API accepts `Range`, cc-connect does not use it

Same `file_key`, same token, two requests. Only the `Range` header differs:

```sh
URL="https://open.feishu.cn/open-apis/im/v1/messages/$MSG_ID/resources/$FILE_KEY?type=file"

# 1) whole-file GET — what cc-connect does today
curl -sD- -o /dev/null "$URL" -H "Authorization: Bearer $TENANT_ACCESS_TOKEN"
# HTTP/1.1 400 Bad Request
# {"code":234037,"msg":"Downloaded file size exceeds limit.", ...}

# 2) identical URL, with Range
curl -sD- -o chunk.bin "$URL" -H "Authorization: Bearer $TENANT_ACCESS_TOKEN" \
                              -H "Range: bytes=0-1048575"
# HTTP/1.1 206 Partial Content
# Content-Type: application/zip
# Content-Range: bytes 0-1048575/142735026
# Inner_file_data_meta: {"Mime":"application/zip","FileSize":142735026,"IsShardingFile":false}
```

Request 2 returns 1 MiB of valid ZIP data. The file is intact on Feishu's side and the total
size is advertised in `Content-Range`. **Only the whole-file fetch is capped.**

I did not determine the exact threshold; it lies somewhere between 13 MB (works) and 136 MB
(rejected).

## Secondary problem: this can be reported to the user as success

Because path 1 drops the entire message and path 2 skips the entry, the agent has no way to
know that an attachment existed and failed. In our setup the bot is prompted to upload any
`.zip` it receives, and to re-run "the most recent zip" when the user types `upload`.

What the user experienced:

1. Sent a 136 MB zip → silently dropped (only visible in `daemon.log`).
2. Quoted it and typed `@bot upload` → quoted download skipped; the agent saw a bare `upload`.
3. The agent fell back to the most recent zip path in its history — **a different, older
   package from the previous day** — re-ran it, and reported *"upload finished, 0 new,
   20 deduplicated"*.

So a failed transfer surfaced as a successful upload of unrelated content. Surfacing the
failure to the agent (or at least to the reply) would prevent this class of confusion.

## Suggested fix

1. In `downloadResource`, fall back to (or default to) a `Range`-based chunked download:
   issue `Range: bytes=0-<chunk-1>`, read `Content-Range: bytes .../<total>`, then loop until
   `total` bytes are fetched. This is a self-contained change and needs no new configuration.
2. Reuse the same helper for the quoted-file path from #1588 so both routes benefit.
3. Optionally bound it with the existing size configuration and, when the bound is exceeded,
   emit a user-visible message instead of dropping the inbound message silently.

Happy to test a patch against the same 136 MB file if that helps.

---

# 中文版

## 问题

从飞书下载大文件附件时失败，报 `code=234037 Downloaded file size exceeds limit.`

飞书的资源下载接口**在请求带 `Range` 头时是可以正常返回文件内容的**，但 cc-connect 发的是
普通的整文件 `GET`，因此只要文件超过飞书对"整文件拉取"的上限就完全下载不到——尽管分片下载是被允许的。

直接下载路径，以及 #1588 在 v1.5.0 中新增的引用文件按需下载路径，**两条都受影响**。

## 环境

| | |
| --- | --- |
| cc-connect | `v1.4.1`（commit `5d4c96dd`）与 `v1.5.0`（commit `17c61062`）**均复现** |
| 平台 | 飞书（`open.feishu.cn`），群聊，长连接 |
| 系统 | macOS (arm64)，launchd 守护进程 |
| 失败文件 | `142,735,026` 字节（约 136 MB）的 `.zip` |
| 成功文件 | 同一个群、同一个机器人下的 3.2 MB 与 13 MB `.zip` |

## 现象

**路径一 —— 用户直接发送文件**（`platform/feishu.(*Platform).downloadResource`）：

```
level=INFO  msg="feishu: file received" file_name=large-archive.zip
level=ERROR msg="feishu: download file failed" error="feishu: resource API code=234037 msg=Downloaded file size exceeds limit."
```

之后**整条入站消息被丢弃**，不再有 `message received` / `processing message` / `turn complete`。
agent 完全不知道有人发过文件。

**路径二 —— 用户引用该文件并 @ 机器人**（#1588 在 v1.5.0 引入的按需拉取）：

```
level=WARN msg="feishu: download quoted file failed; skipping this entry" error="feishu: resource API code=234037 msg=Downloaded file size exceeds limit."
```

同样的 API 错误，只是降级为 `WARN` 并跳过该条目。

两条路径的请求往返都只有约 0.7 秒，即请求是被直接拒绝的——**不是超时，也不是传输中断**。

## 根因 —— 接口支持 `Range`，cc-connect 没有使用

同一个 `file_key`、同一个 token，两次请求只差一个 `Range` 头：

```sh
URL="https://open.feishu.cn/open-apis/im/v1/messages/$MSG_ID/resources/$FILE_KEY?type=file"

# 1) 整文件 GET —— cc-connect 当前的做法
curl -sD- -o /dev/null "$URL" -H "Authorization: Bearer $TENANT_ACCESS_TOKEN"
# HTTP/1.1 400 Bad Request
# {"code":234037,"msg":"Downloaded file size exceeds limit.", ...}

# 2) 完全相同的 URL，加上 Range
curl -sD- -o chunk.bin "$URL" -H "Authorization: Bearer $TENANT_ACCESS_TOKEN" \
                              -H "Range: bytes=0-1048575"
# HTTP/1.1 206 Partial Content
# Content-Type: application/zip
# Content-Range: bytes 0-1048575/142735026
# Inner_file_data_meta: {"Mime":"application/zip","FileSize":142735026,"IsShardingFile":false}
```

第 2 次请求返回了 1 MiB 有效的 ZIP 数据，`Content-Range` 里还带着文件总长度。
说明**文件在飞书侧完好无损，被限制的只有整文件拉取这一种取法**。

具体阈值我没有测定，只知道介于 13 MB（可用）与 136 MB（被拒）之间。

## 次生问题：失败会被当成成功汇报给用户

由于路径一丢弃整条消息、路径二跳过该条目，agent 无从知道"曾经有一个附件且下载失败了"。

我们的机器人被设定为：收到 `.zip` 就自动上传；用户发 `upload` 时则重跑"最近的那个 zip"。
于是实际发生的是：

1. 发送 136 MB 的 zip → 被静默丢弃（只有 `daemon.log` 里看得到）。
2. 引用它并 `@机器人 upload` → 引用下载被跳过，agent 只看到一个光秃秃的 `upload`。
3. agent 从历史里取"最近的 zip 路径"——那是**前一天另一个不相关的旧包**——重跑一遍，
   并回复"上传完成，新增 0 个，去重 20 个"。

一次失败的传输，最终呈现为一次针对无关内容的成功上传。
只要把失败传达给 agent（哪怕只体现在回复里），就能避免这一类误导。

## 建议的修复

1. 在 `downloadResource` 中改用（或在失败时回退到）基于 `Range` 的分片下载：
   先发 `Range: bytes=0-<chunk-1>`，从 `Content-Range: bytes .../<total>` 读出总长度，
   再循环取满。改动自包含，不需要新增配置项。
2. 让 #1588 的引用文件路径复用同一个下载函数，两条路径一起受益。
3. 可选：用现有的大小配置设个上限，超限时给用户一条可见的提示，而不是静默丢弃整条消息。

如果需要，我可以用同一个 136 MB 的文件验证补丁。

chenhg5:
## Summary

The larkim SDK's `GetMessageResource` builder does not expose the HTTP `Range` header, so any resource over Feishu's ~2 MiB streaming cap is truncated with code=234037 ("download interrupted"). This affects both generic file/audio downloads (4 call sites in `platform/feishu/feishu.go`) and the #1588 quoted-file path, on v1.4.1 and v1.5.0.

## Fix

Added `platform/feishu/resource_download.go` with a new `downloadResourceChunked` helper that bypasses the larkim SDK, sends a raw HTTP GET with `Range: bytes=0-0` first, and on 206 (Partial Content) loops Range GETs in default 8 MiB chunks (configurable via `resource_chunk_size_bytes`, clamped to [1 MiB, 64 MiB]). Servers that don't honour Range return 200 with the full body and we short-circuit.

`downloadResource` (used by audio body, file body, merge_forward file, and #1588 quoted file) and `downloadImage` now both route through the new helper, so all five existing call sites benefit automatically without changing the call signature.

## Defensive bounds

- **`resourceMaxBytes`** (default 512 MiB) caps total download size against adversarial servers that report an unboundedly large Content-Length.
- **Transient network errors** on any single chunk are retried with the existing exponential-backoff policy (3 attempts, 500ms→5s).
- **Context cancellation** aborts the loop cleanly so user cancellation doesn't leave orphan goroutines.
- **Auth failures** propagate with feishu context, not raw transport errors.
- **Empty `messageID` / `fileKey`** rejected before any network call.

## Tests

Added `platform/feishu/resource_download_test.go` with 9 new tests covering:

- `TestParseContentRangeTotal` — parser correctness (8 cases incl. malformed headers)
- `TestDownloadResourceChunked_SmallFileUsesSingleGet` — exactly one outbound GET for small files (matches pre-#1741 behaviour for files that fit Feishu's cap)
- `TestDownloadResourceChunked_LargeFileUsesRangeLoop` — non-multiple-of-chunk-size total, verifies byte order and Content-Range round-trip
- `TestDownloadResourceChunked_TotalSizeMismatchFails` — server lies about total mid-loop, download rejected
- `TestDownloadResourceChunked_ProbeFailureFallsBackToSingleGet` — first-chunk 5xx falls back to plain GET
- `TestDownloadResourceChunked_RespectsMaxBytes` — oversize resource rejected, no allocation
- `TestDownloadResourceChunked_RetriesTransientChunk` — connection reset on second chunk is retried with backoff
- `TestDownloadResourceChunked_ContextCancelStopsDownload` — context cancellation aborts
- `TestDownloadResourceChunked_RejectsEmptyIDs` / `TestDownloadResourceChunked_AuthFailurePropagates` — input validation + auth error wrapping

Full `platform/feishu` test suite green (32.5s). Race detector clean. Build clean.

## Risk / compatibility

- **No new dependencies** — uses only `net/http` from stdlib + existing `lark` SDK for token minting.
- **No changes to public API** — `downloadResource` and `downloadImage` keep their signatures; only the implementation changes.
- **No platform-name references in `core/`** — all changes scoped to `platform/feishu/`.
- **Configurable chunk size** — operators with edge-case deployments can tune `resource_chunk_size_bytes`; default 8 MiB matches Feishu's documented guidance.
- **Behaviour for small files unchanged** — files under Feishu's streaming cap still cost exactly one outbound GET (same as before).

Fixes #1741

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -197,6 +197,29 @@ type Platform struct {
 	imageBatchMu     sync.Mutex
 	imageBatch       map[string]*imageBatchEntry
 	imageBatchWindow time.Duration // quiet period before flushing a batch; 0 means use defaultImageBatchWindow
+
+	// resourceDownloadHTTP is the bare HTTP client used to download message
+	// resources directly from Feishu with HTTP Range requests. The larkim SDK's
+	// GetMessageResource does not expose a Range header (#1741), so for files
+	// larger than ~2MB the SDK issues a plain GET and Feishu rejects the
+	// response with code=234037. Bypassing the SDK with our own client and
+	// Range header is the supported workaround.
+	resourceDownloadHTTP *http.Client
+	// resourceChunkSize is the byte size of each Range request issued during
+	// chunked downloads. 8 MiB matches Feishu's documented guidance and keeps
+	// memory bounded. Operators can override via resource_chunk_size_bytes in
+	// config; values are clamped to [1 MiB, 64 MiB].
+	resourceChunkSize int64
+	// resourceMaxBytes caps the total bytes a single resource download may
+	// consume, guarding against adversarial servers that report an
+	// unboundedly large Content-Length. 512 MiB matches cc-connect's own
+	// inbound attachment cap and is large enough for any plausible bot user
+	// attachment on Feishu/Lark today.
+	resourceMaxBytes int64
+	// fetchResourceToken returns the bearer token used for resource downloads.
+	// When nil, defaults to fetchFreshTenantAccessToken. Indirected so unit
+	// tests can inject a stub without spinning up the full lark SDK.
+	fetchResourceToken func(ctx context.Context) (string, error)
 }
 
 // defaultImageBatchWindow is the quiet period after the last image in a
@@ -207,6 +230,17 @@ type Platform struct {
 // image sends. Operators that need a longer or shorter window can override
 // it via the platform option `image_batch_window_ms`.
 const defaultImageBatchWindow = 500 * time.Millisecond
+
+// defaultResourceMaxBytes caps the total bytes a single Feishu resource
+// download may consume. Feishu's message-resource endpoint does not validate
+// caller-side size limits beyond the per-app upload cap (typically 1 GiB for
+// files; 60 MiB for images), and a misconfigured server can advertise a
+// Content-Length orders of magnitude larger than the actual resource. 512 MiB
+// matches the inbound attachment cap cc-connect applies everywhere else and is
+// large enough to cover any realistic bot user attachment on Feishu/Lark.
+// Operators that need to download larger files can raise this via
+// resource_max_bytes; the value is clamped to a sane minimum of 1 MiB.
+const defaultResourceMaxBytes int64 = 512 * 1024 * 1024
 
 // batchWindow returns the effective image-batch coalesce window for this
 // Platform. Tests and zero-initialised Platforms fall back to the default so
@@ -382,6 +416,27 @@ func newPlatform(name, domain string, opts map[string]any) (core.Platform, error
 		imageBatchWindow = time.Duration(ms) * time.Millisecond
 	}
 
+	// resource_chunk_size_bytes: byte size for each Range request when chunked-
+	// downloading Feishu message resources (issue #1741). The larkim SDK does
+	// not expose Range headers, so for resources above ~2 MiB a plain GET
+	// returns code=234037. Default 8 MiB; clamped to [1 MiB, 64 MiB].
+	resourceChunkSize := int64(8 * 1024 * 1024)
+	if raw, ok := opts["resource_chunk_size_bytes"]; ok {
+		n, err := coerceMilliseconds(raw)
+		if err != nil {
+			return nil, fmt.Errorf("%s: invalid resource_chunk_size_bytes %v: %w", name, raw, err)
+		}
+		if n > 0 {
+			resourceChunkSize = n
+		}
+	}
+	if resourceChunkSize < 1*1024*1024 {
+		resourceChunkSize = 1 * 1024 * 1024
+	}
+	if resourceChunkSize > 64*1024*1024 {
+		resourceChunkSize = 64 * 1024 * 1024
+	}
+
 	// Webhook mode configuration (for Lark international version)
 	port, _ := opts["port"].(string)
 	if port == "" {
@@ -426,6 +481,9 @@ func newPlatform(name, domain string, opts map[string]any) (core.Platform, error
 		mentionMap:                 mentionMap,
 		imageBatch:                 make(map[string]*imageBatchEntry),
 		imageBatchWindow:           imageBatchWindow,
+		resourceDownloadHTTP:       &http.Client{Timeout: 60 * time.Second},
+		resourceChunkSize:          resourceChunkSize,
+		resourceMaxBytes:           defaultResourceMaxBytes,
 	}
 	if !useInteractiveCard {
 		base.self = base
@@ -3065,24 +3123,14 @@ func buildFeishuFileMessageContent(msgType, fileKey string) (string, error) {
 }
 
 func (p *Platform) downloadImage(messageID, imageKey string) ([]byte, string, error) {
-	resp, err := p.client.Im.MessageResource.Get(context.Background(),
-		larkim.NewGetMessageResourceReqBuilder().
-			MessageId(messageID).
-			FileKey(imageKey).
-			Type("image").
-			Build())
+	// Issue #1741: large image bodies suffer the same code=234037 truncation
+	// as files when fetched through the larkim SDK (which cannot set Range
+	// headers). Route image downloads through the same chunked helper used
+	// for files so a 20-MiB screenshot lands whole instead of being capped
+	// at the SDK's ~2 MiB streaming ceiling.
+	data, err := p.downloadResourceChunked(context.Background(), messageID, imageKey, "image")
 	if err != nil {
-		return nil, "", fmt.Errorf("%s: image API: %w", p.tag(), err)
-	}
-	if !resp.Success() {
-		return nil, "", fmt.Errorf("%s: image API code=%d msg=%s", p.tag(), resp.Code, resp.Msg)
-	}
-	if resp.File == nil {
-		return nil, "", fmt.Errorf("%s: image API returned nil file body", p.tag())
-	}
-	data, err := io.ReadAll(resp.File)
-	if err != nil {
-		return nil, "", fmt.Errorf("%s: read image: %w", p.tag(), err)
+		return nil, "", fmt.Errorf("%s: image download: %w", p.tag(), err)
 	}
 
 	mimeType := detectMimeType(data)
@@ -3091,24 +3139,14 @@ func (p *Platform) downloadImage(messageID, imageKey string) ([]byte, string, er
 }
 
 func (p *Platform) downloadResource(messageID, fileKey, resType string) ([]byte, error) {
-	resp, err := p.client.Im.MessageResource.Get(context.Background(),
-		larkim.NewGetMessageResourceReqBuilder().
-			MessageId(messageID).
-			FileKey(fileKey).
-			Type(resType).
-			Build())
+	// Issue #1741: the larkim SDK issues a plain GET that Feishu truncates
+	// with code=234037 for resources above ~2 MiB. downloadResourceChunked
+	// bypasses the SDK, sends Range headers, and reassembles the bytes
+	// client-side. All four existing call sites (audio body, file body,
+	// merge_forward file, #1588 quoted file) flow through here unchanged.
+	data, err := p.downloadResourceChunked(context.Background(), messageID, fileKey, resType)
 	if err != nil {
-		return nil, fmt.Errorf("%s: resource API: %w", p.tag(), err)
-	}
-	if !resp.Success() {
-		return nil, fmt.Errorf("%s: resource API code=%d msg=%s", p.tag(), resp.Code, resp.Msg)
-	}
-	if resp.File == nil {
-		return nil, fmt.Errorf("%s: resource API returned nil file body", p.tag())
-	}
-	data, err := io.ReadAll(resp.File)
-	if err != nil {
-		return nil, fmt.Errorf("%s: read resource: %w", p.tag(), err)
+		return nil, err
 	}
 	slog.Debug(p.tag()+": downloaded resource", "key", fileKey, "type", resType, "size", len(data))
 	return data, nil
@@ -3815,8 +3853,10 @@ func isTenantAccessTokenInvalid(err error) bool {
 	return strings.Contains(msg, "99991663") || strings.Contains(msg, "invalid access token")
 }
 
-// Transient retry constants for network-level failures.
-const (
+// Transient retry settings for network-level failures. Declared as var (not
+// const) so tests can shrink the retry window; production callers never
+// touch them after init.
+var (
 	maxTransientRetries    = 3
 	transientRetryInitial  = 500 * time.Millisecond
 	transientRetryMaxDelay = 5 * time.Second
--- a/platform/feishu/resource_download.go
+++ b/platform/feishu/resource_download.go
@@ -149,7 +149,7 @@ func (p *Platform) resourceFetchFirstChunk(ctx context.Context, token, messageID
 	if err != nil {
 		return nil, 0, fmt.Errorf("first-chunk request: %w", err)
 	}
-	defer resp.Body.Close()
+	defer func() { _, _ = io.Copy(io.Discard, resp.Body); _ = resp.Body.Close() }()
 
 	switch resp.StatusCode {
 	case http.StatusPartialContent:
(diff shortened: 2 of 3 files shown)


chenhg5 (commented):
## Conclusion: Approve

(Self-author block: `gh pr review --approve` is rejected because the PR author `chenhg5` is the maintainer running the QA session. GitHub review is therefore posted as `comment`, but `review_decision` per Multigent workflow is `approve` — the QA conclusion is that this PR should enter the human merge gate.)

### Overall assessment
Clean, well-scoped hotfix for #1741. CI is fully green, the diff auto-merges with current `main`, the 9 new tests (with 8 parser sub-cases) directly cover the regression and edge cases, and the design — bypass the SDK with raw HTTP, HEAD-probe then Range-loop — is exactly what the Feishu docs imply. No new dependencies, no public API change, behaviour for small files preserved. Low-risk.

### Review scope
- Reviewed `platform/feishu/feishu.go` (+112/-36), `platform/feishu/resource_download.go` (new, +398), `platform/feishu/resource_download_test.go` (new, +487). Total +961/-36.
- Head `de1ef969`, base `1e4c4b2`, PR sits exactly 2 commits ahead of current `main` — no merge conflict.
- Focused on: HTTP Range semantics, byte-order safety, retry/backoff alignment with existing policy, max-bytes cap against adversarial servers, test coverage of edge cases (mismatch, oversize, cancel, auth, empty), and whether the change breaks small-file behaviour.

### Verification
- CI: 5/5 green — [run 32950475162](https://github.com/chenhg5/cc-connect/actions/runs/32950475162) (lint 2m24s, unit-test 4m45s, smoke-test 26s, regression-test 33s, performance-test 54s).
- Local merge: `git merge --no-commit --no-ff origin/main` → Already up to date (PR is 2 commits ahead of main, both from this branch).
- Local tests on PR head: `go test -race -count=1 ./platform/feishu/` → ok (34.07s, in line with author's 32.7s).
- Local targeted run `go test -v -run 'TestParseContentRangeTotal|TestDownloadResourceChunked_'` → 10/10 pass in 1.048s; all 8 parser sub-cases pass.
- `go vet ./platform/feishu/` → clean.
- `gofmt -l` on changed files → only `feishu.go` is flagged; verified the drift is **pre-existing on `origin/main`** (field-alignment around `groupFilterDegraded*` block, unrelated to this PR). PR introduces no new gofmt drift.

### ✅ What looks good
- **HEAD-then-Range design is right.** Probing with `Range: bytes=0-0` lets the helper decide between (a) full-body 200 → return immediately and (b) 206 + Content-Range total → loop the remaining chunks. Small files cost exactly one GET (matches pre-#1741 behaviour); large files succeed where the SDK path failed.
- **The 5 call sites are correctly re-routed.** `downloadResource` (4 sites: audio body, file body, merge_forward file, #1588 quoted file) and `downloadImage` (1 site) both go through `downloadResourceChunked`. The `larkim` SDK's `GetMessageResource` calls are removed cleanly; `larkim` import is preserved for event handlers (`P2MessageReceiveV1`, etc.) which still need it.
- **Defensive bounds are layered.** `resourceMaxBytes=512 MiB` caps total bytes (matches cc-connect's inbound attachment cap, sized for any realistic bot user file on Feishu/Lark today). `resourceRangeMaxRangeHeaderBytes=64 MiB` caps any single Range request as defence in depth. `LimitReader(body, cap+1)` on both probe and fallback paths catches servers that lie about Content-Length. `io.Copy(io.Discard, resp.Body)` before close ensures connection reuse.
- **Retry policy reuses existing primitives.** `transientRetryInitial`, `maxTransientRetries`, `transientRetryMaxDelay`, `isTransientError` are imported from the existing package — not a new retry ladder. Same policy as the SDK path used to apply.
- **Probe-fail fallback is honest.** When the probe fails (4xx/5xx), the helper logs a warning and falls back to a plain GET. This catches servers that reject Range entirely with 4xx (rather than silently ignoring it). For the failing-server case, the existing failure mode is preserved (and may still hit the ~2 MiB cap) — at least the failure mode is no worse than today.
- **Content-Range verification is thorough.** Every chunk request checks the response's `Content-Range` total matches the probe's advertised total. Mismatch → fail-fast with a clear error (`Content-Range total mismatch: got N want M`). Defends against servers that change their mind mid-stream.
- **Input validation happens before network.** Empty `messageID` / `fileKey` rejected before any outbound HTTP. Auth failure (`fetchResourceToken` returns error) wraps with feishu context prefix.
- **Context cancellation propagates cleanly.** Verified by `TestDownloadResourceChunked_ContextCancelStopsDownload` — first chunk request returns `context.Canceled`, helper bails out, no orphan goroutines (the chunk loop is synchronous).
- **Test coverage is comprehensive.** 10 test functions, 17 sub-cases total when including the parser table. Uses `httptest.NewServer` for realistic HTTP semantics. Covers: chunked loop with non-multiple-of-chunk-size total, single-GET fallback, Content-Range mismatch, max-byte cap, transient retry, context cancel, auth failure, empty-input rejection, and 8 malformed Content-Range parser inputs (including the edge cases `*`, empty, negative, no slash, garbage).
- **Configurable chunk size matches Feishu's documented guidance.** Default 8 MiB per Feishu's API docs; operator-overridable via `resource_chunk_size_bytes`; clamped to `[1 MiB, 64 MiB]` to keep memory bounded.
- **PR body and commit messages explain the why.** Both reference #1741, mention `code=234037`, and explain the SDK limitation that motivates the bypass. The second commit ("satisfy golangci-lint") cleanly fixes the errcheck warnings from the first commit.

### 🚨/🔴 Must fix
- None. No blocker.

### 🟠 Should improve
- None material.

### 🔵 Optional (non-blocking)
- **P2**: `resource_max_bytes` is documented in the PR body as "Operators that need to download larger files can raise this via `resource_max_bytes`", but `newPlatform` does **not** actually parse this config key — `resourceMaxBytes` is hard-coded to `defaultResourceMaxBytes = 512 MiB`. Either add the option parser (matching the `resource_chunk_size_bytes` pattern) or update the comment to make the hard-cap nature explicit.
- **P2**: The full resource is buffered in memory (`bytes.NewBuffer(make([]byte, 0, total))`). For a 100 MiB file, this allocates ~100 MiB of process RSS; at 512 MiB cap the process could spike +512 MiB on a single download. For deployments with tight container memory limits, a disk-backed implementation (write chunks to temp file, rename on success — the path the PM originally proposed) would be safer. Out of scope for this PR; worth a follow-up if memory pressure shows up in the field.
- **P3**: `coerceMilliseconds` (defined at `feishu.go:258`) is reused as a byte-size parser at `feishu.go:425`. Function name is misleading (it's just a generic numeric TOML coercion). Worth extracting a `coerceInt64(v any) (int64, error)` helper and routing both call sites through it.
- **P3**: A `slog.Info` fires on every chunked download (`feishu: resource chunked download complete file_key=… chunks=… chunk_size=…`). On a high-volume deployment this could spam the log. Either downgrade to `Debug` or aggregate into a counter (`feishu_download_total`, `feishu_download_bytes_total`) matching the metrics the PM originally proposed.
- **P3**: `isTransientError` (imported from existing package) is the existing transient-error predicate — verify it correctly classifies the connection-reset-on-second-chunk case the test exercises (`TestDownloadResourceChunked_RetriesTransientChunk`). Out of scope for this PR, but worth confirming in a follow-up that the same predicate is what #1640 used.

### ❓ Questions
- None blocking.

### Testing / Risk
- Verified evidence: CI 5/5 green (run 32950475162), local race tests pass (34.07s for full `./platform/feishu/`, 1.048s for the new tests), `go vet` clean, `gofmt -l` clean (no new drift), auto-merge with `origin/main` clean (PR is exactly 2 commits ahead).
- Remaining risk: very low. The new behaviour is what the user explicitly asked for (Range chunks to bypass code=234037), the implementation uses stdlib only, retry policy aligns with existing policy, and the 512 MiB cap + LimitReader pattern defends against adversarial servers.

### Next step
- Maintainer can merge when ready. As called out in PM triage for #1741, this should ship alongside the v1.5.1 hotfix that includes #1743 (weixin reply path). Combined, both PRs restore v1.4.1-equivalent behaviour on the two main interaction paths (weixin reply, feishu large files) that v1.5.0 silently broke.

Refs #1741, #1640, #1588.

PR #1748, opened 2026-08-26
fix(core): normalize workspace path separators for cross-platform key consistency

6moran:
### Problem

`normalizeWorkspacePath` returns whatever `filepath.Clean` / `filepath.EvalSymlinks` produce. On Windows that is backslash-separated, on Linux/macOS forward-slash. The result is used as a **map key** in several places: workspace bindings, the agent/workspace state cache, and interactive session keys. Because callers feed the same logical path from different origins (config, filesystem, platform messages), separators can differ, so the same workspace can produce two different keys on Windows (e.g. `C:/a/b` vs `C://a//b`) -> duplicate workspace states / agent instances and missed bindings. The repo's own footer code already normalizes to forward slashes at display time (`compactReplyFooterPath` calls `filepath.ToSlash`), confirming forward-slash is the intended canonical form.

### Reproduction (on Windows)

```
go test ./core/ -run TestNormalizeWorkspacePath -v
```

Fails:

```
normalizeWorkspacePath("/nonexistent/path/./foo/../bar") = "\nonexistent\path\bar", want "/nonexistent/path/bar"
```

### Fix

Return `filepath.ToSlash(...)` from `normalizeWorkspacePath` in both branches. On Linux/macOS `ToSlash` is a no-op, so behavior there is unchanged (CI stays green); on Windows the key becomes separator-consistent. The test expectation is normalized with `filepath.ToSlash` as well so it is portable across platforms.

### Verification

- `go test ./core/ -run 'TestNormalizeWorkspacePath|TestNormalizeBeforePoolProducesSameKey|TestInteractiveKeyForSessionKey_NormalizesWorkspace|TestWorkspacePool' -v` -> `nonexistent uses Clean only` now PASS; other workspace-key tests PASS.
- Full `go test ./core/` on Windows: no new failures vs. baseline (remaining failures are pre-existing platform differences, e.g. symlink evaluation needing elevated privileges, unrelated to this change).
- On Linux/macOS the change is a no-op, so CI results are unaffected.

### Files changed

- `core/workspace_state.go` - `normalizeWorkspacePath` now returns forward-slash-normalized paths in both the symlink-resolved and Clean-only branches.
- `core/workspace_state_test.go` - expected value in `TestNormalizeWorkspacePath` normalized to forward slashes for portability.

> Note: this replaces the earlier #1747, whose head branch was rebuilt onto the latest `main`; GitHub blocks reopening a PR whose branch was deleted/recreated, so it is superseded by this one.

### DIFF
--- a/core/workspace_state.go
+++ b/core/workspace_state.go
@@ -11,16 +11,22 @@ import (
 // mismatches caused by trailing slashes, symlinks, or relative segments.
 // If the path cannot be resolved (e.g. doesn't exist yet), falls back to
 // filepath.Clean only.
+//
+// The result is always returned with forward-slash separators so that it can
+// be used as a stable, cross-platform key (workspace bindings, agent cache,
+// interactive session keys, etc.). On Windows, filepath.Clean/EvalSymlinks emit
+// backslashes, which would otherwise make the same logical path collide with
+// itself depending on which caller supplied the separators.
 func normalizeWorkspacePath(path string) string {
 	cleaned := filepath.Clean(path)
 	resolved, err := filepath.EvalSymlinks(cleaned)
 	if err != nil {
-		return cleaned
+		return filepath.ToSlash(cleaned)
 	}
 	if resolved != path {
 		slog.Debug("workspace path normalized", "original", path, "normalized", resolved)
 	}
-	return resolved
+	return filepath.ToSlash(resolved)
 }
 
 // workspaceState holds the runtime state for a single workspace.
--- a/core/workspace_state_test.go
+++ b/core/workspace_state_test.go
@@ -88,11 +88,14 @@ func TestNormalizeWorkspacePath(t *testing.T) {
 	}
 
 	// Resolve the expected path through EvalSymlinks so that the test works
-	// on macOS where /var is a symlink to /private/var.
-	resolvedRealDir, err := filepath.EvalSymlinks(realDir)
+	// on macOS where /var is a symlink to /private/var. Normalize to forward
+	// slashes to match normalizeWorkspacePath's separator-consistent output,
+	// keeping the expectation portable across Windows/Linux/macOS.
+	resolved, err := filepath.EvalSymlinks(realDir)
 	if err != nil {
 		t.Fatal(err)
 	}
+	resolvedRealDir := filepath.ToSlash(resolved)
 
 	tests := []struct {
 		name  string


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Two-line core fix that makes `normalizeWorkspacePath` always return forward-slash paths so it can be used as a stable cross-platform key. CI is 5/5 green on commit `a89d5541a3df569c53d6286589a88a2b7464149c`, mergeable, base = main, no conflicts. On Linux/macOS `filepath.ToSlash` is a no-op, so CI behavior is unchanged and the only platform where semantics shift is Windows, which is exactly the platform that needs it.

Review scope:
- Reviewed `core/workspace_state.go` (`normalizeWorkspacePath`) and `core/workspace_state_test.go` (test expectation normalization), plus confirmed callers via search: `core/workspace_state.go` (itself + `workspacePool` consumers), `core/workspace_state_test.go`, `core/multi_workspace_test.go`, `core/cuj_test.go`, `core/relay_test.go`, and design docs.
- Confirmed that the existing display code at `core/claude_status_footer_test.go` (`compactReplyFooterPath`) already canonicalizes via `filepath.ToSlash`, which is the right prior art — the change brings the key-side normalizer in line with the display-side normalizer.
- Focused on correctness, portability of the test fix, and call-site safety.

✅ What looks good:
- The fix is minimally scoped: both branches of the existing function now end with `filepath.ToSlash(...)`. There is no restructuring of the symlink evaluation logic, which keeps the diff trivial to audit and trivial to revert.
- `filepath.ToSlash` is documented to be a no-op on non-Windows, so Linux/macOS CI results are unchanged. The fix only changes behavior on the platform that has the bug, which is exactly what you want.
- The function's doc comment now spells out the contract ("The result is always returned with forward-slash separators so that it can be used as a stable, cross-platform key") and explicitly calls out the failure mode the fix closes. Future callers can rely on the invariant instead of having to re-normalize on their end.
- Test-side normalization (`resolvedRealDir := filepath.ToSlash(resolved)`) keeps `TestNormalizeWorkspacePath` portable across Windows/Linux/macOS. The expectation that the function returns forward slashes is encoded in the test in a way that doesn't depend on the host platform's separator.
- The PR description's reproduction (`normalizeWorkspacePath("/nonexistent/path/./foo/../bar") = "\nonexistent\path\bar", want "/nonexistent/path/bar"`) lines up exactly with the production-code change: that input hits the `EvalSymlinks` error path and the cleaned value is now `ToSlash`'d before return.
- All callers I traced (`workspacePool` and downstream in `core/multi_workspace_test.go`, `core/cuj_test.go`, `core/relay_test.go`) treat the result as a map key — i.e. separator-stability is what they want, and the change strictly improves that.

🚨/🔴 Must fix:
- None. No correctness, security, or compatibility issues introduced on the platforms the existing CI matrix covers.

🟠 Should improve:
- The fix addresses separators but not case. On Windows the same logical path can appear as `C:\Foo\bar` and `c:\foo\BAR`, which would still produce two distinct map keys after this fix. Not introduced by this PR (the previous code had the same property), so not blocking — but worth a one-line doc comment noting the scope so the next reader doesn't assume the function is case-insensitive too. Optional.

🔵 Optional:
- Consider adding a focused cross-platform test case that demonstrates the same logical path with different separators (e.g. `C:\a\b` vs `C:/a/b` vs `C:\\a\\b`) collapsing to the same key. CI on this matrix is Linux-only, so the test won't fail on Linux either way, but it would document the invariant for future Windows CI.
- `compactReplyFooterPath` doing `filepath.ToSlash` is now redundant for any path that originated from `normalizeWorkspacePath`. Not worth changing here, but a `ToSlash` at the boundary of map-key vs display-key could be a follow-up if more callers appear.

❓ Questions:
- None.

Testing / Risk:
- The existing `TestNormalizeWorkspacePath` covers both branches of `normalizeWorkspacePath` (the EvalSymlinks-error path via `/nonexistent/...` and the success path via `realDir`); the test's expectations are now platform-portable thanks to the `filepath.ToSlash(resolved)` normalization.
- Risk scope: any caller that today does something like `if strings.Contains(path, "\\")` after going through `normalizeWorkspacePath` would silently change behavior. None of the call sites I traced do that — they all treat the result as a map key. The design docs referenced in the search results (`session-resilience-plan.md` / `-design.md`) talk about workspace paths as keys, not as display strings, which is consistent with the new contract.
- Backward compatibility: a Linux/macOS user who has existing on-disk session files keyed by an un-normalized path is unaffected, because the function is a no-op on those platforms. A Windows user who has been affected by the bug will start seeing correct keys on next launch — there is no migration concern, since the prior state was the wrong map key anyway.

Next step:

PR #1752, opened 2026-08-26
fix(slack): inject thread context into agent prompt (#1749)

### LINKED ISSUE #1749: Slack: bot @-mentioned in an existing thread gets no thread context

chenhg5:
Fixes #1749.

When the bot is @-mentioned inside an existing Slack thread, fetch the thread history via `conversations.replies` and inject it into `core.Message.ExtraContent` so the agent sees what humans said earlier in the thread. Without this fix, the agent only sees the most recent message and loses the conversation context.

## Changes

- **`.platform/slack/thread_context.go`** (new): `fetchThreadContext`, `formatThreadContext`, and a soft-degrading `maybeFetchThreadContext` wrapper (10s timeout, log + metric on failure, never blocks dispatch). Filters out bot/app messages so the agent only sees the human side.
- **`platform/slack/slack.go`**: add config knobs `thread_context_enabled` (default `true`) and `thread_context_max_messages` (default `20`, hard cap `100`), wire `maybeFetchThreadContext` into both `AppMentionEvent` and `MessageEvent` branches when `ThreadTimeStamp` is set, add atomic metrics: `slack_thread_context_fetch_total`, `slack_thread_context_fetch_failed_total`, `slack_thread_context_messages_total`.
- **`platform/slack/thread_context_test.go`** (new): 14 tests covering fetch, format, bot filtering, pagination cap, only-bots short-circuit, nil-client guard, empty inputs, soft degradation, options parsing, and display-name resolution.
- **`docs/slack.md`** + **`config.example.toml`**: document the new options.

## Design

Mirrors the Feishu `fetchReplyChain` / `formatReplyChain` pattern from `platform/feishu/feishu.go` for cross-platform consistency.

The fetch is best-effort: any Slack API failure (network, missing scope, rate limit) is logged at WARN and the agent still receives the new message without context. The main message flow is never blocked.

## Tests

```
go test ./platform/slack/ -count=1
go test -race ./platform/slack/ -count=1
```

All platform/slack tests pass (existing + 14 new). The full `core/ ./config/ ./agent/...` test suite is also green.

## Metrics

```
slack_thread_context_fetch_total          # every fetch attempt
slack_thread_context_fetch_failed_total   # failures (logged + swallowed)
slack_thread_context_messages_total       # total non-bot messages fetched
```

### DIFF
--- a/platform/slack/slack.go
+++ b/platform/slack/slack.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/chenhg5/cc-connect/core"
@@ -19,6 +20,17 @@ import (
 	"github.com/slack-go/slack/slackevents"
 	"github.com/slack-go/slack/socketmode"
 )
+
+// defaultSlackThreadContextMaxMessages is the fallback cap for the number of
+// non-bot thread replies we include in core.Message.ExtraContent when the
+// operator has not explicitly configured slack.thread_context_max_messages.
+// Chosen to fit comfortably in a Claude prompt without overwhelming context.
+const defaultSlackThreadContextMaxMessages = 20
+
+// slackThreadContextMaxMessagesHardCap is the upper bound enforced regardless
+// of user configuration; protects against accidental misconfiguration that
+// would balloon prompts and burn Slack rate-limit quota.
+const slackThreadContextMaxMessagesHardCap = 100
 
 func init() {
 	core.RegisterPlatform("slack", New)
@@ -41,6 +53,21 @@ type Platform struct {
 	channelNameCache map[string]string
 	channelCacheMu   sync.RWMutex
 	userNameCache    sync.Map // userID -> display name
+
+	// Thread context injection (issue #1749). When threadContextEnabled is
+	// true and an inbound event arrives inside an existing Slack thread
+	// (ThreadTimeStamp != ""), the platform calls conversations.replies to
+	// build a human-readable transcript and injects it as core.Message.
+	// ExtraContent. Failures are soft-degraded: the main message is still
+	// dispatched, just without the transcript.
+	threadContextEnabled     bool
+	threadContextMaxMessages int
+	// Metric counters — exposed via the slog-debug "slack: thread context
+	// summary" line and any future /metrics hook. Atomic so concurrent event
+	// handlers can increment without locking.
+	metricThreadContextFetchTotal        atomic.Int64
+	metricThreadContextFetchFailedTotal  atomic.Int64
+	metricThreadContextMessagesTotal     atomic.Int64
 }
 
 func New(opts map[string]any) (core.Platform, error) {
@@ -58,12 +85,35 @@ func New(opts map[string]any) (core.Platform, error) {
 			"if your agent runtime is tmux, also set window_per_session=true — " +
 			"without it, concurrent threads share a single pane and their output will interleave")
 	}
+	threadEnabled := true
+	if v, ok := opts["thread_context_enabled"].(bool); ok {
+		threadEnabled = v
+	}
+	threadMax := defaultSlackThreadContextMaxMessages
+	if v, ok := opts["thread_context_max_messages"]; ok {
+		switch n := v.(type) {
+		case int:
+			threadMax = n
+		case int64:
+			threadMax = int(n)
+		case float64:
+			threadMax = int(n)
+		}
+	}
+	if threadMax < 1 {
+		threadMax = 1
+	}
+	if threadMax > slackThreadContextMaxMessagesHardCap {
+		threadMax = slackThreadContextMaxMessagesHardCap
+	}
 	return &Platform{
-		botToken:         botToken,
-		appToken:         appToken,
-		allowFrom:        allowFrom,
-		sessionScope:     scope,
-		channelNameCache: make(map[string]string),
+		botToken:                 botToken,
+		appToken:                 appToken,
+		allowFrom:                allowFrom,
+		sessionScope:             scope,
+		channelNameCache:         make(map[string]string),
+		threadContextEnabled:     threadEnabled,
+		threadContextMaxMessages: threadMax,
 	}, nil
 }
 
@@ -220,6 +270,9 @@ func (p *Platform) handleEvent(evt socketmode.Event) {
 					MessageID: ev.TimeStamp,
 					ReplyCtx:  replyContext{channel: ev.Channel, timestamp: threadTS},
 				}
+				if extra := p.maybeFetchThreadContext(ev.Channel, ev.ThreadTimeStamp); extra != "" {
+					msg.ExtraContent = extra
+				}
 				p.handler(p, msg)
 
 			case *slackevents.AssistantThreadStartedEvent:
@@ -280,6 +333,9 @@ func (p *Platform) handleEvent(evt socketmode.Event) {
 					Content:  ev.Text, Images: images, Files: docFiles, Audio: audio,
 					MessageID: ts,
 					ReplyCtx:  replyContext{channel: ev.Channel, timestamp: threadTS},
+				}
+				if extra := p.maybeFetchThreadContext(ev.Channel, ev.ThreadTimeStamp); extra != "" {
+					msg.ExtraContent = extra
 				}
 				p.handler(p, msg)
 			}
--- a/platform/slack/thread_context.go
+++ b/platform/slack/thread_context.go
@@ -1,0 +1,203 @@
+package slack
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strings"
+	"time"
+
+	"github.com/slack-go/slack"
+)
+
+// fetchThreadContext fetches the Slack thread that starts at threadTS and
+// returns a formatted, chronological summary suitable for injection into
+// core.Message.ExtraContent. The summary only includes messages the bot did
+// NOT post so the agent sees the human-readable transcript.
+//
+// Behaviour:
+//   - maxMessages caps the number of *non-bot* messages we keep; Slack's API
+//     hard limit per request is 1000, we still cap lower for prompt safety.
+//   - API failures (network, auth, scope) are returned as errors. Callers
+//     log WARN + continue without context; we never block the main message.
+//   - Bot messages are skipped at the formatter stage, not at the API stage,
+//     so the count of API messages is still reported in metrics.
+//
+// Returns ("", nil) when the thread has zero non-bot messages so callers can
+// distinguish "no context needed" from "fetch failed".
+func (p *Platform) fetchThreadContext(ctx context.Context, channelID, threadTS string, maxMessages int) (string, error) {
+	if channelID == "" || threadTS == "" {
+		return "", errors.New("slack: fetchThreadContext: empty channel or thread ts")
+	}
+	if maxMessages <= 0 {
+		maxMessages = defaultSlackThreadContextMaxMessages
+	}
+	if p.client == nil {
+		return "", errors.New("slack: fetchThreadContext: client not initialised")
+	}
+
+	p.metricThreadContextFetchTotal.Add(1)
+
+	// Slack paginates thread replies by cursor. We request Inclusive so the
+	// thread root is included; the formatter will still skip it if it is a
+	// bot message. Limit stays at the default (~50) — Slack returns at most
+	// 50 per page; we follow cursor if we need more, but cap by maxMessages.
+	const pageSize = 50
+	collected := make([]slack.Message, 0, pageSize)
+	cursor := ""
+	for {
+		params := &slack.GetConversationRepliesParameters{
+			ChannelID: channelID,
+			Timestamp: threadTS,
+			Limit:     pageSize,
+			Inclusive: true,
+			Cursor:    cursor,
+		}
+		msgs, _, nextCursor, err := p.client.GetConversationRepliesContext(ctx, params)
+		if err != nil {
+			// Failure metric is incremented by the caller's logThreadContextFailure
+			// path so direct callers (who handle errors themselves) and the
+			// soft-degrading maybeFetchThreadContext wrapper don't double-count.
+			return "", fmt.Errorf("conversations.replies: %w", err)
+		}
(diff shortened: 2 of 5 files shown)


chenhg5 (commented):
## Conclusion: Approve

**Overall assessment:** This PR brings Slack platform up to parity with Feishu's `fetchReplyChain` behaviour — when a user @-mentions the bot inside an existing thread, the agent now sees the human side of the conversation that came before. CI is fully green, the implementation is well-tested (13 tests + 5 sub-cases, all pass under `-race`), the API loop self-caps at `maxMessages*2` to protect rate limit, and the feature soft-degrades (WARN log + failure counter + main dispatch never blocked). Auto-merge against `origin/main` is clean.

**Review scope:**
- Reviewed `platform/slack/thread_context.go` (new, 203 lines), `platform/slack/slack.go` (+66/-5), `platform/slack/thread_context_test.go` (new, +533), `docs/slack.md` (+35), `config.example.toml` (+5). Total +837/-5.
- Focused on correctness, security, soft-degrade behaviour, metric accuracy, config parsing, scope implications, and cross-platform consistency with Feishu's `fetchReplyChain` / `formatReplyChain`.

**✅ What looks good:**
- **Soft-degrade design** (`maybeFetchThreadContext`): wraps the fetch in `context.WithTimeout(..., 10*time.Second)`, all errors are logged + metric'd then swallowed, main dispatch is never blocked. This is exactly the right shape for a context-fetch that's a "nice to have" — agents still receive the triggering message.
- **Self-cap before cursor exhaustion**: the loop breaks when `len(collected) >= maxMessages*2` or `nextCursor == ""`, which protects Slack rate limit on bot-heavy threads (e.g. CI notification bots).
- **Metric discipline**: failure counter only increments in `logThreadContextFailure` (single source of truth), so callers that handle errors themselves don't double-count.
- **Config clamps**: `thread_context_max_messages` is parsed as `int/int64/float64`, clamped to `[1, 100]`; disabled flag short-circuits before any API call.
- **Cross-platform parity**: `formatThreadContext` deliberately mirrors Feishu's `formatReplyChain` shape (numbered chain for multi-message, single-message format for short threads) — operators moving between platforms get consistent agent behaviour.
- **Test coverage is comprehensive**: fetch success/failure, pagination, bot filtering, nil client, empty inputs, disabled flag, soft-degrade, options parsing (defaults, explicit disable, smaller cap, over-cap clamp, zero clamp), display name fallback.

**🟠 Should improve (not blocking, but worth doing in a follow-up):**
- **P2 — Scope visibility**: `conversations.replies` is a new required scope for Slack apps; deployment docs should highlight this prominently so operators don't discover it via a `WARN` log on the first production thread @. Currently the WARN includes the API name which is helpful, but the doc change in this PR doesn't call out the scope addition explicitly.
- **P3 — Metric naming precision**: `metricThreadContextMessagesTotal` adds `int64(len(collected))` (total API messages, including bot). Doc comment says "non-bot messages fetched". For now these are consistent at the API layer, but if anyone later adds bot-filtering at the API stage the metric would silently include more. Consider renaming to `slack_thread_context_api_messages_total` or splitting into two counters.

**🔵 Optional:**
- **P3 — gofmt drift in all 3 changed files** (`slack.go`, `thread_context.go`, `thread_context_test.go`): field-alignment padding and a missing trailing newline at the end of `thread_context.go`. Pure whitespace, CI lint doesn't run `gofmt -l`, so it slips through. `gofmt -w` on these three files before merge is recommended but not blocking — same pattern we saw on #1714.
- **P3 — Timeout configurability**: `10*time.Second` in `maybeFetchThreadContext` is hardcoded. If operators ever want to tune it (e.g. rate-limit-aware deployments), exposing `thread_context_timeout_secs` would be a one-liner addition; default behaviour unchanged.

**❓ Questions:**
- None blocking. (If `thread_context_max_messages` is set to a value > 100 in some other config file in the wild, this PR will silently clamp it. That's the right behaviour but is worth a release-note line.)

**Testing / Risk:**
- **Verified locally:**
  - `go test -race -count=1 ./platform/slack/` → ok (2.579s)
  - `go test -v -run 'TestFetchThreadContext_|TestFormatThreadContext|TestMaybeFetchThreadContext_|TestSlackOptionsForThreadContext' ./platform/slack/` → 13 tests + 5 sub-cases pass
  - `go vet ./platform/slack/` → clean
  - `git merge --no-commit --no-ff origin/main` → Already up to date
- **CI:** 5/5 green (run 32974898798): lint 4m46s, unit-test 4m57s, smoke-test 28s, regression-test 33s, performance-test 55s.
- **Remaining risk:** Behaviour change is enabled by default. If your Slack workspace has unusual rate-limit tier or very chatty threads, monitoring `slack_thread_context_fetch_failed_total` is the right signal. The feature is reversible via `thread_context_enabled = false` per-platform in `config.toml`.

**Next step:** Approve & merge. If maintainer prefers zero nit noise on the main branch, suggest author run `gofmt -w` on the 3 changed files in the same PR — but it's not required to ship this.

🤖 Posted by cc-connect QA workflow (chenhg5/cc-connect). Decision: **approve**. GitHub refused `--approve` due to self-author block on `chenhg5` author, so review is submitted as `--comment` per Multigent workflow rule; the structured `review_decision` is still `approve`.

platform/slack/thread_context.go
```diff
@@ -0,0 +1,203 @@
...
+			// path so direct callers (who handle errors themselves) and the
+			// soft-degrading maybeFetchThreadContext wrapper don't double-count.
+			return "", fmt.Errorf("conversations.replies: %w", err)
+		}
+		collected = append(collected, msgs...)
+		// Cap early — once we have enough candidates the formatter will trim
+		// bot messages down to maxMessages, but we stop the API loop here so
+		// we don't burn rate-limit quota on threads with thousands of bot
+		// replies.
+		if len(collected) >= maxMessages*2 || nextCursor == "" {
```
zhoukuncheng, line 69: `conversations.replies` returns oldest-first, so this break makes long threads return the *beginning* of the thread.

With the defaults `pageSize` is 50 and `maxMessages*2` is 40, so the first page already satisfies `len(collected) >= 40` and the loop exits — the cursor path below is unreachable. Line 112 then keeps "the most recent 20" out of the *oldest 50*. On a 300-message thread the agent gets messages ~31-50 and never sees what was just said, which inverts the point of the feature. It only happens to be correct for threads of ≤50 messages.

Either page to the end of the thread (following `nextCursor` while discarding all but the last `maxMessages`), or fetch from the newest end and reverse.

platform/slack/thread_context.go
```diff
@@ -0,0 +1,203 @@
...
+			break
+		}
+		cursor = nextCursor
+	}
+
+	kept := formatThreadContext(collected, p, maxMessages)
+	if kept == "" {
+		return "", nil
+	}
+	p.metricThreadContextMessagesTotal.Add(int64(len(collected)))
```
zhoukuncheng, line 79: This counts every message the API returned, bot messages included, while the PR description and docs call it "total non-bot messages fetched" — `len(humanMessages)` is the number that matches the name.

It is also skipped entirely by the `kept == ""` early return above, and on the `client == nil` / empty-args paths `logThreadContextFailure` increments the failure counter while `metricThreadContextFetchTotal` was never incremented, so failures can exceed attempts.

platform/slack/thread_context.go
```diff
@@ -0,0 +1,203 @@
...
+// usually < 500 ms, but a slow network must not stall message dispatch.
+func (p *Platform) maybeFetchThreadContext(channelID, threadTS string) string {
+	if !p.threadContextEnabledFor() {
+		return ""
+	}
+	if threadTS == "" {
+		// Top-level message or DM with no thread — nothing to fetch.
+		return ""
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
zhoukuncheng, line 195: This 10s ctx only bounds `GetConversationRepliesContext`. The formatting pass that runs after it calls `displayNameFor` per message (lines 117 and 123) → `resolveUserName` → `p.client.GetUserInfo` (`slack.go:669`), which takes no context and is issued once per uncached user.

`handleEvent` runs synchronously on the single socket-mode consumer goroutine (`slack.go:196`), so a cold cache with 20 distinct posters means 20 serial `users.info` calls (Tier-3, rate limited) outside the stated bound, with every other channel's events queued behind them. The request is already acked at `slack.go:222` so Slack will not redeliver — the cost is latency for everyone else, not duplicates — but "The main message flow is never blocked" in the docs does not hold.

Passing `ctx` through to the name resolution, or resolving names concurrently under the same deadline, would make the bound real.

platform/slack/slack.go
```diff
@@ -220,6 +270,9 @@ func (p *Platform) handleEvent(evt socketmode.Event) {
 					MessageID: ev.TimeStamp,
 					ReplyCtx:  replyContext{channel: ev.Channel, timestamp: threadTS},
 				}
+				if extra := p.maybeFetchThreadContext(ev.Channel, ev.ThreadTimeStamp); extra != "" {
```
zhoukuncheng, line 273: The message that triggered this dispatch is not excluded, so the user's text reaches the agent twice: once as `Content`, and again as the last entry of `ExtraContent`.

`fetchThreadContext` requests the thread with `Inclusive: true` and gets no trigger timestamp, and the mention itself is already posted by the time the event arrives, so it comes back in the replies and passes the non-bot / non-empty filter. The comment at `thread_context.go:89-94` notes it cannot compare timestamps "without passing the trigger ts" — `ev.TimeStamp` is available right here (and `ts` in the `MessageEvent` branch).

While there: that same comment says it takes "the oldest maxMessages non-bot messages", but line 112 takes the newest.

zhoukuncheng (commented):
Reviewed at `05ecfffa` — `go test ./platform/slack/` is green and so is CI, but three of the findings below are behavioural rather than style, and the trigger semantics differ from what #1749 asked for.

**Gaps against #1749**

- **Fetches once per message, not once per thread.** Both branches (`slack.go:273`, `slack.go:337`) call `maybeFetchThreadContext` whenever `ThreadTimeStamp` is set, so every follow-up turn re-injects the whole transcript. That is the #764 "the transcript buries the user's actual text" concern the issue calls out, and it costs one `conversations.replies` per turn indefinitely.
- **No skip for threads the bot itself rooted.** Those messages already reached the agent as ordinary turns, so re-reading them is pure duplication.
- **Other bots' messages are dropped, not labelled** (`isSlackBotMessage`). In a multi-bot thread (#1610) "what did the other bot just say" is the main reason to read a thread you were pulled into; it is also how the agent loses its own earlier answers.
- **`session_scope` is still absent from `docs/slack.md`** (`grep session_scope docs/slack.md` → no hits). That was the issue's secondary ask — the default is `user`, so per-thread sessions stay undiscoverable for anyone setting Slack up from the docs.
- The description says this mirrors Feishu's `fetchReplyChain`/`formatReplyChain`. It mirrors the output format; the trigger semantics are different. `feishu.go:2282` walks a quoted-reply parent chain and only when a quote exists — it never re-reads a whole thread per message.

**Relationship to #1750**

#1750 targets the same issue and implements exactly these four: a per-thread claim map so a thread is bootstrapped once, exclusion of the trigger message and anything after it, `assistant` labelling for bot messages (so the #1610 case works), and a 5s bound chosen because the socket-mode loop dispatches serially. It also documents `session_scope`. Worth deciding which branch to land before either goes further — happy either way, but landing both would leave two thread-context code paths in the adapter.

**Minor:** `slack_thread_context_fetch_total` / `_failed_total` / `_messages_total` are documented in `docs/slack.md` as exposed metrics, but `grep -rn slack_thread_context --include='*.go'` returns nothing — they are unexported atomics that nothing reads. And `gofmt -l ./platform/slack/` flags all three files (missing trailing newline in `thread_context.go`, struct field alignment in `slack.go`). CI misses it because `.golangci.yml` is v2 format with no `formatters:` section.

PR #1755, opened 2026-08-27
fix(claudecode): add claude-fable-5 to the /model fallback list (#1642)

### LINKED ISSUE #1642: fix(claudecode): /model picker fallback list is stale — Claude 5 family (claude-fable-5) not selectable for subscription-auth users
## Summary

For `claudecode` users authenticated via **Claude subscription OAuth** (no `ANTHROPIC_API_KEY`), the `/model` picker can never show the Claude 5 family (`claude-fable-5`). The picker falls back to a hardcoded 4-entry list that predates the Claude 5 launch, so the current flagship model is simply not selectable from the UI. This is the claudecode twin of the codex issue fixed in #1546 (stale static allowlist → GPT-5.6 not selectable).

## Environment

- cc-connect v1.4.1 (also reproduced by reading `main` — code unchanged)
- Agent: `claudecode`, auth = Claude subscription (OAuth), **no** `ANTHROPIC_API_KEY` set
- Platform: Feishu (but the picker is platform-independent)

## Root cause

`agent/claudecode/claudecode.go` → `AvailableModels()` has three tiers:

1. `configuredModels()` — only populated when the user sets `models = [...]` in config (most don't);
2. `fetchModelsFromAPI()` — **returns `nil` immediately when `ANTHROPIC_API_KEY` is empty**, which is the normal state for subscription-auth users (Claude Code OAuth has no x-api-key to use here);
3. hardcoded fallback:

```go
return []core.ModelOption{
    {Name: "sonnet", Desc: "Claude Sonnet (balanced)"},
    {Name: "opus", Desc: "Claude Opus (most capable)"},
    {Name: "opus[1m]", Desc: "Claude Opus (1M context)"},
    {Name: "haiku", Desc: "Claude Haiku (fastest)"},
}
```

Subscription users therefore *always* land on tier 3, and tier 3 has no `fable` / `claude-fable-5` entry. The `/model` card/buttons in `core/engine.go` (`cmdModel`) are built exclusively from this list, so the model cannot be picked.

## What already works (workaround)

Free-text switching is fine: `parseModelSwitchArgs` + `resolveModelSwitchTarget` fall through to `return input` without validating against the list, so

```
/model switch claude-fable-5
```

passes the name straight to `--model` and persists it via `modelSaveFunc`. Confirmed working end-to-end on Feishu. So this is purely a **discoverability** bug — the escape hatch exists but nothing in the UI hints at it, and users reasonably conclude the model is unsupported (I did, several times).

## Suggested fix

Prefer the same philosophy as #1546 — stop trusting a static snapshot:

1. **Minimal**: refresh the fallback list for the current lineup and include the `fable` family (e.g. `fable` / `claude-fable-5`, plus current opus/sonnet/haiku aliases). Cheap, but it will go stale again at the next model launch.
2. **Better**: when tier 2 is unavailable (no API key), treat the fallback as *suggestions*, and render a hint in the picker card: "or reply `/model switch <model-id>` for any model". One i18n string, permanent fix for future launches.
3. **Optional**: accept `claude-*` prefixed ids as always-valid (prefix pattern matching, like the codex fix), so numbered lists and typed ids never diverge in behavior.

(2)+(3) together would make claudecode `/model` launch-proof.

## Related

- #1546 — codex `/model` gpt-5.x visibility (same class of bug, fixed via prefix matching in v1.5.0-beta.2)
- #1505 — 1M context reporting for Fable/Mythos-class models

Niko-Wu:
Fixes #1642.

The hardcoded fallback list in `AvailableModels()` predates the Claude 5 family, so subscription-auth users (where the dynamic listing is unavailable) cannot pick `claude-fable-5` from `/model`.

One-line fix: add `{Name: "claude-fable-5", Desc: "Claude Fable 5 (frontier)"}` to the fallback list, between `opus[1m]` and `haiku` to keep the existing capability ordering.

Deliberately minimal:
- The issue and triage only name `fable` as missing; `opus`/`sonnet`/`haiku` aliases already roll forward to the latest generation in the CLI, so no redundant `claude-opus-5`/`claude-sonnet-5` entries.
- Full id `claude-fable-5` rather than a short alias — the reporter verified end-to-end that this id passes through `--model` correctly.
- No entries removed.

Verification: `go build ./agent/...` and `go test ./agent/claudecode/...` pass.

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -393,6 +393,7 @@ func (a *Agent) AvailableModels(ctx context.Context) []core.ModelOption {
 		{Name: "sonnet[1m]", Desc: "Claude Sonnet (1M context)"},
 		{Name: "opus", Desc: "Claude Opus (most capable)"},
 		{Name: "opus[1m]", Desc: "Claude Opus (1M context)"},
+		{Name: "claude-fable-5", Desc: "Claude Fable 5 (frontier)"},
 		{Name: "haiku", Desc: "Claude Haiku (fastest)"},
 	}
 }


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- One-line addition to the `AvailableModels()` hardcoded fallback in `agent/claudecode/claudecode.go` adds `claude-fable-5` between `opus[1m]` and `haiku`. CI is 5/5 green on commit b112e028c1b8ef9569a7ac9f25f4c07c6816af91, mergeable, base = main, no conflicts. Risk is minimal.

Review scope:
- Reviewed `agent/claudecode/claudecode.go` and the related issue #1642.
- Focused on correctness, naming, ordering, and whether the fallback path is still load-bearing after this change.

✅ What looks good:
- The model id `claude-fable-5` matches the model id used by the Claude CLI (consistent with the Anthropic `/v1/models` listing on the dynamic path), so subscription-OAuth users falling through to the static list will be able to select the model the CLI itself advertises.
- Placement is correct: `opus[1m]` is the current frontier in the static list, so inserting `claude-fable-5` between `opus[1m]` and `haiku` preserves the existing capability ordering (frontier → fast) that users navigating `/model` rely on.
- The diff deliberately scopes the change to the fallback list (not to `configuredModels()`, providers, or the dynamic `/v1/models` fetch path), so this PR is a pure bugfix that does not alter how API-key users pick models.
- The companion issue #1642 is closed, and the fix matches the title — exact, scoped, and minimal.

🚨/🔴 Must fix:
- None. No correctness, security, or compatibility issues introduced by this one-line addition.

🟠 Should improve:
- None worth blocking on.

🔵 Optional:
- The static fallback list and the dynamic API list will continue to drift independently over time; consider adding a small smoke test that exercises the static-fallback branch (e.g. when `configuredModels()` is empty and `fetchModelsFromAPI` returns `nil`) and asserts `claude-fable-5` is present, so future model launches cannot silently regress the subscription-OAuth path again. Not blocking.
- Description string `"Claude Fable 5 (frontier)"` is fine, but if you want it to line up with what the dynamic fetch returns you could derive the display name from a shared constant. Cosmetic.

❓ Questions:
- None.

Testing / Risk:
- `agent/claudecode/claudecode_model_test.go` covers `configuredModels()` boundary conditions and `GetModel()` precedence but does not directly assert the static fallback list — see optional suggestion above.
- Risk surface is the `/model` picker UI for subscription-OAuth users when `ANTHROPIC_API_KEY` is unset AND no per-provider models are configured. Before this PR that list lacked `claude-fable-5` entirely; after this PR it matches the dynamic list and the issue title.

Next step:
- Recommend maintainer merge. The remaining "drift" risk is small and is exactly what the existing fallback + dynamic-fetch layered design already accepts; no follow-up PR required.

PR #1757, opened 2026-08-27
fix(core): trailing NO_REPLY must not suppress text already delivered before a tool call (#1614)

### LINKED ISSUE #1614: Claude Code: trailing `NO_REPLY` suppresses substantive text emitted before a tool call
## Summary

With the Claude Code agent, a turn can emit substantive user-facing text and a
tool call in the same turn, then end the trailing continuation with `NO_REPLY`
because the text was already emitted. cc-connect treats the final `NO_REPLY` as
applying to the whole turn: the substantive text is not finalized as the
response, and the persisted assistant history collapses to `NO_REPLY`.

This is not an agent crash — the work completes successfully, but the final
delivery is suppressed and the body is lost from session history, so subsequent
turns have no record of the answer.

## Environment

- cc-connect: `v1.4.1`
- Agent: `claudecode`
- Platform: Feishu group chat
- Display: default `full` mode (`tool_messages` defaults to `true`), `card_mode = legacy`

The classification happens in shared `processInteractiveEvents` before normal
final delivery, so the bug is platform-independent for platforms on the legacy
streaming-preview path.

## Original user scenario

The user asked Claude to perform a multi-step code review, ask a second local
coding agent for an independent review, consolidate both, and return the
prioritized findings to the group. Claude used task-tracking tools
(`TaskCreate`/`TaskUpdate`) while doing this work, then ended the trailing
continuation with `NO_REPLY`.

## Reproduction

Sanitized Claude Code event sequence for the affected turn (timestamps UTC+8,
identities redacted):

```text
10:57:58 user message
10:58:06 assistant text (short preamble)
10:58:07 assistant tool_use  -> user tool_result
10:58:10 assistant tool_use  -> user tool_result
10:58:56 assistant text (substantive final report, ~4.4k chars)
10:58:56 assistant tool_use  -> user tool_result   # report + task update, same turn
10:59:01 assistant text "NO_REPLY"                  # trailing continuation
10:59:01 result event = "NO_REPLY"
```

cc-connect log for the same turn:

```text
turn complete tools=3 response_len=8 output_tokens=2368 silent=true
turn response response=NO_REPLY
silent reply suppressed
```

The task tracker was marked completed and the report had already been
generated, but the persisted turn response was only `NO_REPLY`.

No application IDs, user/open IDs, chat IDs, message IDs, API endpoints,
credentials, model/provider names, or message contents are included.

## Expected behavior

Text already emitted as committed user-facing assistant output before a tool
call should be delivered exactly once (finalized), even if the trailing
continuation is `NO_REPLY`; and the persisted assistant history should retain
that body, not collapse to `NO_REPLY`.

`NO_REPLY` must remain silent when the turn contains no committed user-facing
text — e.g. a group-chat gate decision, or a turn where a tool already delivered
the intended attachment/message.

## Actual behavior

The final `NO_REPLY` marks the whole turn silent. In `processInteractiveEvents`
(`core/engine.go`, the `EventResult` handler):

1. `fullResponse := event.Content` takes only the last assistant segment
   (`"NO_REPLY"`). The two `textParts` fallback branches do **not** trigger
   under default `full` mode (`tool_messages = true`, `segmentStart > 0`).
2. `isSilent := isSilentReply(baseResponse)` is `true`, so the turn enters the
   `else if isSilent` branch, which calls `sp.discard()` and skips all send paths.
3. `session.AddHistory("assistant", baseResponse)` persists `"NO_REPLY"`,
   losing the body from history.

The body had been surfaced as a streaming preview (frozen + detached at the
tool call, leaving it visible in-place), but it is never finalized as the turn
response and is lost from history.

## Relation to #549

#549 fixed the quiet-mode case where `event.Content` overwrote the accumulated
preview (`fullResponse := event.Content` ignoring `textParts`). This report is
the full-mode / streaming-preview counterpart: the same `event.Content` bias
loses the body when the trailing continuation is a bare `NO_REPLY`. Always
preserving preview text would regress #549, because provisional progress text
is not necessarily a final response — so the fix must distinguish provisional
preview from committed user-facing text.

## Relation to #1585 (not covered)

PR #1585 ("fix: prevent NO_REPLY markers from leaking to users", open) targets
a related but distinct set of cases:

- a standalone trailing `NO_REPLY` text segment after rich-card text;
- a tool-boundary flush leaking a bare `NO_REPLY`;
- unsolicited/background turns resolving to `NO_REPLY`.

I verified #1585 against this scenario by checking out its `core/engine.go`
(`head = d548504`, branch `codex/fix-no-reply-leak`) and running the regression
below: the body is still lost (turn marked `silent=true`, history `NO_REPLY`),
identical to `main`. #1585's `joinTextSegments` helper only runs inside the two
`textParts` fallback branches; under default `full` mode neither branch runs
(`event.Content` is non-empty and `tool_messages = true`), and #1585 does not
touch the `else if isSilent { sp.discard() }` branch that drops the body. So
this report complements #1585 rather than duplicating it.

## Suggested fix

Track `committedText` — the user-facing assistant text emitted before a tool
call — separately from provisional preview text. On the streaming-preview path
each tool call freezes and detaches the preview (leaving the text visible
in-place) and advances `segmentStart`, so `textParts[:segmentStart]` is exactly
the text already surfaced to the user. When the result resolves to a bare
`NO_REPLY` but `committedText` is non-empty, deliver the committed body instead
of suppressing the turn.

Output selection as a pure function (validated):

```go
func resolveTurnOutput(resultText, committedText string) (text string, silent bool) {
	// Bare NO_REPLY marker: fall back to committed body when present.
	if isSilentReply(resultText) {
		committedText = strings.TrimSpace(committedText)
		if committedText != "" {
			return committedText, false
		}
		return "", true
	}
	// Trailing NO_REPLY marker preceded by substantive text: strip the marker.
	if stripped, ok := stripTrailingSilent(resultText); ok {
		if strings.TrimSpace(stripped) == "" {
			return "", true
		}
		return stripped, false
	}
	return resultText, false
}
```

`committedText` population:

- include user-visible text blocks emitted before a tool call
  (`textParts[:segmentStart]`, i.e. text already detached/surfaced via preview);
- exclude thinking blocks, tool progress, and synthetic status previews
  (they never enter `textParts`);
- when `committedText != ""` and the result is bare `NO_REPLY`, persist and
  deliver `committedText` instead of the marker;
- in the `toolCount > 0 && segmentStart > 0` finalize branch, strip a trailing
  bare `NO_REPLY` from the unsent remainder so the marker is never delivered.

This preserves `NO_REPLY` semantics for genuinely silent turns (no committed
text) while preventing a trailing continuation from erasing text Claude Code
already presented as user-visible output.

## Regression tests

Validated against `main` (bug reproduces) and with the fix applied (all pass)
in a temporary clone. New file `core/no_reply_legacy_preview_test.go`:

1. `TestProcessInteractiveEvents_LegacyPreview_BodyThenToolThenNoReply_PreservesBody`
   — `text + tool_use -> tool_result -> NO_REPLY -> result NO_REPLY` on the
   legacy preview path: body delivered once, `NO_REPLY` not leaked, history
   retains the body. Fails on `main` (history collapses to `NO_REPLY`).
2. `TestProcessInteractiveEvents_LegacyPreview_ProgressOnlyNoReply_StaysSilent`
   — provisional preview only -> `NO_REPLY`: stays silent (preserves #549).
3. `TestProcessInteractiveEvents_LegacyPreview_NormalResult_Unchanged`
   — ordinary non-silent result: unchanged.
4. `TestProcessInteractiveEvents_LegacyPreview_MultiToolMultiBody_NoDuplicate`
   — multiple tool calls with multiple text segments: bodies preserved in
   order, not duplicated, `NO_REPLY` not delivered.
5. `TestProcessInteractiveEvents_LegacyPreview_NoReplyCaseInsensitiveWhitespace`
   — `NO_REPLY` / `no_reply` / whitespace-padded variants: stay silent.

## Validation

Tested on a clone of `chenhg5/cc-connect` `main` with Go 1.25.

```bash
go test ./core/ -run 'TestProcessInteractiveEvents_LegacyPreview' -count=1 -v
go test ./core/ -run 'NoReply|Silent|SilentReply|Preview|RichCard|Unsolicited|TextThenNoReply|DoesNotFlush|LegacyPreview' -count=1
```

Results:

- On `main`: test 1 fails (`persisted assistant history is only NO_REPLY; body
  lost from history: "NO_REPLY"`), tests 2–5 pass.
- With the suggested fix applied: all five pass.
- With #1585's `core/engine.go` checked out instead: test 1 still fails
  (identical to `main`), confirming #1585 does not cover this scenario.
- Existing `NoReply|Silent|Preview|RichCard|Unsolicited` tests remain green —
  no regression. The full `go test ./core/` has one unrelated flaky failure
  (`TestCUJ_A3_ImageReachesAgent`, a macOS temp-dir cleanup race that passes on
  retry and is documented in #1585's PR notes), not caused by this change.

Niko-Wu:
Fixes #1614.

## Root cause

In `processInteractiveEvents`, `EventResult.Content` only carries the **last** assistant segment — `"NO_REPLY"` in the issue's scenario. `isSilentReply` then silences the whole turn and `AddHistory` records the bare marker, even though every tool/thinking/permission boundary had already flushed `textParts[:segmentStart]` to the user as committed output. The marker was being applied retroactively to the entire turn instead of to its own trailing segment — exactly the granularity bug identified in the triage comment.

## Fix

- Before the silent check, reconstruct the committed body (`textParts[:segmentStart]`, marker-stripped). If the final segment is a bare marker **and** committed text exists, the marker only closes the trailing continuation: the committed body becomes the turn response and history keeps it instead of `"NO_REPLY"`.
- Guarded with `segmentStart > 0 && !hasRichCard` — rich-card/streaming-card paths keep `segmentStart == 0` and are untouched.
- A turn consisting of nothing but `NO_REPLY` (with or without tool calls) stays fully silent, locked by a regression test.
- The `toolCount > 0` remainder-send branch now also strips a trailing marker so it can never leak to the platform.

## Tests

- `TestProcessInteractiveEvents_TrailingNoReplyPreservesTextBeforeTool` — reproduces the issue sequence (text → tool → body → tool → `NO_REPLY`); fails before the fix with the exact symptoms from the report (`tools=2 response_len=8 silent=true`, history collapsed to the marker), passes after.
- `TestProcessInteractiveEvents_BareNoReplyWithToolsStaysSilent` — regression guard for the intentional silent path.

`go test ./core/ -count=1` passes (full package, no regressions); `go vet ./core/` clean.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -5598,17 +5598,47 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 				}
 			}
 
+			// Text already surfaced to the user in earlier segments: every tool /
+			// thinking / permission boundary flushes textParts[segmentStart:] (as a
+			// detached frozen preview or as sent chunks) and advances segmentStart,
+			// so textParts[:segmentStart] is committed user-facing output. On the
+			// rich-card / streaming-card paths segmentStart stays 0 and the card
+			// keeps the body itself, so they are left untouched.
+			committedText := ""
+			if segmentStart > 0 && segmentStart <= len(textParts) && !hasRichCard {
+				committed := strings.Join(textParts[:segmentStart], "")
+				if stripped, ok := stripTrailingSilent(committed); ok {
+					committed = stripped
+				}
+				committed = strings.TrimRight(committed, " \t\r\n")
+				if strings.TrimSpace(committed) != "" {
+					committedText = committed
+				}
+			}
+
 			// Detect NO_REPLY marker on the base response (before indicators/footer are appended).
-			// Three cases:
-			//   1. bare marker (isSilentReply)               → fully silent
-			//   2. trailing marker with non-empty reasoning  → strip marker, deliver reasoning
-			//   3. trailing marker with empty strip result   → fully silent
+			// Four cases:
+			//   1. bare marker with committed text           → marker only closes the trailing
+			//                                                  continuation; the committed body
+			//                                                  is the turn response
+			//   2. bare marker, nothing committed            → fully silent
+			//   3. trailing marker with non-empty reasoning  → strip marker, deliver reasoning
+			//   4. trailing marker with empty strip result   → fully silent
 			// History records the ORIGINAL baseResponse so the agent retains context of its own
 			// decision; only the outbound platform text gets rewritten/suppressed.
+			isSilent := isSilentReply(baseResponse)
+			if isSilent && committedText != "" {
+				// Case 1: event.Content only carries the last assistant segment
+				// ("NO_REPLY"). Suppressing the turn here would drop the body the
+				// user already saw and collapse history to the marker.
+				baseResponse = committedText
+				cleanResponse = committedText
+				isSilent = false
+			}
+
 			session.AddHistory("assistant", baseResponse)
 			sessions.Save()
 
-			isSilent := isSilentReply(baseResponse)
 			if !isSilent {
 				if stripped, ok := stripTrailingSilent(baseResponse); ok {
 					if strings.TrimSpace(stripped) == "" {
@@ -5816,6 +5846,12 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 				sp.discard()
 				if segmentStart < len(textParts) {
 					unsent := strings.Join(textParts[segmentStart:], "")
+					// The remainder is raw agent text, so it can still carry the
+					// trailing NO_REPLY marker (baseResponse was stripped above but
+					// is not what gets sent here). Never leak the marker.
+					if stripped, ok := stripTrailingSilent(unsent); ok {
+						unsent = strings.TrimRight(stripped, " \t\r\n")
+					}
 					if unsent != "" {
 						if !sendChunksWithStatusFooter(e.ctx, p, replyCtx, unsent, statusFooter, sendWorkspaceWithError) {
 							return
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -1373,6 +1373,105 @@ func TestProcessInteractiveEvents_ToolSegmentsKeepFinalFooter(t *testing.T) {
 	}
 }
 
+// TestProcessInteractiveEvents_TrailingNoReplyPreservesTextBeforeTool covers
+// the claudecode shape where the agent emits substantive user-facing text,
+// issues a tool call in the same turn, and then closes the trailing
+// continuation with a bare NO_REPLY (EventResult.Content = "NO_REPLY").
+// The marker only ends that continuation: the text already committed to the
+// user before the tool call must stay the turn response and must survive in
+// session history instead of collapsing to the marker.
+func TestProcessInteractiveEvents_TrailingNoReplyPreservesTextBeforeTool(t *testing.T) {
+	p := &stubPlatformEngine{n: "telegram"}
+	e := NewEngine("test", &stubAgent{}, []Platform{p}, "", LangEnglish)
+	e.SetDisplayConfig(DisplayCfg{ThinkingMessages: true, ThinkingMaxLen: 300, ToolMaxLen: 500, ToolMessages: true})
+
+	sessionKey := "telegram:user-trailing-noreply"
+	session := e.sessions.GetOrCreateActive(sessionKey)
+	agentSession := newControllableSession("s-trailing-noreply")
+	state := &interactiveState{
+		agentSession: agentSession,
+		platform:     p,
+		replyCtx:     "ctx-trailing-noreply",
+	}
+	e.interactiveStates[sessionKey] = state
+
+	const report = "结论：三处风险，已按优先级排序。"
+	agentSession.events <- Event{Type: EventText, Content: "先检查一下。"}
+	agentSession.events <- Event{Type: EventToolUse, ToolName: "Bash", ToolInput: "pwd"}
+	agentSession.events <- Event{Type: EventText, Content: report}
+	agentSession.events <- Event{Type: EventToolUse, ToolName: "TaskUpdate", ToolInput: "completed"}
+	agentSession.events <- Event{Type: EventText, Content: "NO_REPLY"}
+	agentSession.events <- Event{Type: EventResult, Content: "NO_REPLY", Done: true}
+	e.processInteractiveEvents(state, session, e.sessions, sessionKey, "m-trailing-noreply", time.Now(), nil, nil, state.replyCtx)
+
+	history := session.GetHistory(0)
+	if len(history) == 0 {
+		t.Fatal("history is empty, want the assistant turn recorded")
+	}
+	last := history[len(history)-1]
+	if last.Role != "assistant" {
+		t.Fatalf("last history role = %q, want assistant", last.Role)
+	}
+	if !strings.Contains(last.Content, report) {
+		t.Fatalf("history lost the committed body: %q", last.Content)
+	}
+	if strings.Contains(last.Content, "NO_REPLY") {
+		t.Fatalf("history collapsed to the NO_REPLY marker: %q", last.Content)
+	}
+
+	sent := p.getSent()
+	delivered := false
+	for _, msg := range sent {
+		if strings.Contains(msg, "NO_REPLY") {
+			t.Fatalf("NO_REPLY marker leaked to the platform: %q (all sent = %#v)", msg, sent)
+		}
+		if strings.Contains(msg, report) {
+			delivered = true
+		}
+	}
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Targeted fix to `processInteractiveEvents` that re-establishes segment-level granularity when a turn closes its trailing continuation with a bare `NO_REPLY`. CI is 5/5 green on commit `8f6d2ff9d8b43b301f27ecf0631d9fd66fa1d566`, mergeable, base = main, no conflicts. Both new regression tests directly reproduce the issue shape from #1614 and the complementary fully-silent case.

Review scope:
- Reviewed `core/engine.go` (the fix in `processInteractiveEvents` + the `toolCount > 0` remainder-send branch) and `core/engine_test.go` (two new regression tests + the existing `TestProcessInteractiveEvents_ToolSegmentsKeepFinalFooter` baseline).
- Confirmed that `silentReplyTrailingRe`, `silentReplyRe`, `isSilentReply`, and `stripTrailingSilent` were left untouched on the head commit — this PR only re-orders how they are used, which keeps the regex semantics single-sourced.
- Focused on correctness, history preservation, marker leak to platform, and rich-card isolation.

✅ What looks good:
- The root cause is correctly diagnosed: `EventResult.Content` carries only the last assistant segment, so a turn that ends with `NO_REPLY` looks like a fully-silent turn even though earlier segments were already committed. Reconstructing `committedText` from `textParts[:segmentStart]` and consulting it first matches the actual data flow (every tool / thinking / permission boundary flushes + advances `segmentStart`).
- The new case table ("bare marker with committed text → marker only closes the trailing continuation; committed body is the turn response") is consistent with how streaming-preview previews already treat earlier segments.
- The `segmentStart > 0 && !hasRichCard` guard is the right isolation: rich-card / streaming-card paths keep `segmentStart == 0` and the card body replaces the text flow, so they are explicitly left untouched. This is the same granularity the rest of the function already relies on.
- Both `baseResponse` and `cleanResponse` are overwritten in the case-1 branch, and the history call uses `baseResponse` after the rewrite — so `session.AddHistory("assistant", committedText)` records the committed body, which is what subsequent turns / resume / `--resume` rely on. This matches the issue's "history collapses to NO_REPLY" symptom and fixes it directly.
- The companion patch in the `toolCount > 0` remainder-send branch (`unsent = strings.TrimRight(stripped, " \t\r\n")`) closes the leak where a trailing marker could still be sent to the platform after the suppression was applied earlier. The exact strip behavior matches what `stripTrailingSilent` already returns.
- `TestProcessInteractiveEvents_TrailingNoReplyPreservesTextBeforeTool` reproduces the user's event sequence verbatim (text → tool → body → tool → `NO_REPLY`) and asserts both the platform delivery and the history content. `TestProcessInteractiveEvents_BareNoReplyWithToolsStaysSilent` is the regression guard for the case where the turn has no committed body and must stay silent.
- Edge handling in the case-1 path: `strings.TrimSpace(committed) != ""` check ensures the marker isn't overwritten onto whitespace-only committed text. `committed = strings.TrimRight(committed, " \t\r\n")` keeps formatting intact.

🚨/🔴 Must fix:

🟠 Should improve:
- None worth blocking on.

🔵 Optional:
- `stripTrailingSilent` is now called in three places in this function (committed-body strip, case-1 → case-3/4 strip, and the `toolCount > 0` remainder strip). Consider a one-line comment near the regex / helper definition noting that this is the central point of truth for "trailing marker suppression" so future refactors don't accidentally split the behavior. Cosmetic.
- The case-1 branch only fires when `isSilent && committedText != ""`. If `isSilent && committedText == ""` (committed body was whitespace-only and got trimmed to empty), the marker path remains. That is the desired behavior, but a direct test case for "trailing marker with whitespace-only committed body" would lock the invariant. Not blocking.

❓ Questions:
- None.

Testing / Risk:
- New tests are at the same `processInteractiveEvents` boundary as the existing segment tests, so they exercise the real flow end-to-end without needing a full live agent harness.
- Risk surface: the change is bounded to the silent-reply handling in `processInteractiveEvents`. The case-1 rewrite replaces `baseResponse` and `cleanResponse` only when both `isSilent` and `committedText != ""`, and only when `segmentStart > 0 && !hasRichCard` — three independent conditions all of which must hold. The rich-card path, the streaming-card path, the bare-silent path, and the trailing-marker-with-reasoning path are explicitly untouched.
- Failure mode worth knowing about: if any future change moves the `session.AddHistory("assistant", baseResponse)` call above the case-1 rewrite, history will start collapsing to the marker again. Worth a one-line inline comment at the AddHistory call to flag the ordering, but not required.

Next step:

Issue #1115, opened 2026-05-25
auto_compress estimateTokens ignores system prompt / tools overhead + idle reset never fires due to Unlock() updating UpdatedAt


kevinsutianxing:
### Description

The `auto_compress` feature triggers when estimated token usage exceeds `max_tokens`. However, `estimateTokensWithPendingAssistant` (in `core/engine.go`, line 572) only sums the character length of history entries and any pending assistant response, then divides by 4 as a rough heuristic:

```go
func estimateTokensWithPendingAssistant(entries []HistoryEntry, pendingAssistant string) int {
    count := 0
    for _, h := range entries {
        count += len([]rune(h.Content))
    }
    if pendingAssistant != "" {
        count += len([]rune(pendingAssistant))
    }
    if count == 0 {
        return 0
    }
    return (count + 3) / 4
}
```

The problem: this estimate ignores the large fixed token cost of system prompts, tool definitions (tool_use), skills, memory injection, and other per-turn overhead that the upstream agent (Claude Code, Codex, etc.) injects into each API call. In practice, these can consume ~60K tokens per turn.

With `max_tokens = 15000`, auto-compress only triggers when history content alone reaches 15K tokens. But by that point, the actual total context (history + system prompt + tools + skills + memory) is already ~75K tokens, well past the model's useful context window. The compress fires too late.

### Steps to Reproduce

1. Configure `auto_compress` with `max_tokens = 15000`.
2. Use a cc-connect session with Claude Code (which injects ~60K tokens of system prompt / tools / skills / memory per turn).
3. Have a conversation until history accumulates enough characters that `estimateTokens` returns >= 15000.
4. Observe: by the time auto-compress triggers, the actual context window is far beyond the intended threshold, and the model has already been degraded for many turns.

### Expected Behavior

Auto-compress should trigger when the *total* estimated context (history + fixed overhead) reaches `max_tokens`. The `max_tokens` threshold should reflect actual token consumption, not just the history portion.

### Suggested Fix

Add a configurable `system_overhead_tokens` field (or auto-detect it from the first API call's usage response) that represents the fixed per-turn token cost:

```go
func estimateTokensWithPendingAssistant(entries []HistoryEntry, pendingAssistant string, systemOverhead int) int {
    count := systemOverhead  // ~60K for Claude Code with tools/skills/memory
    for _, h := range entries {
        count += len([]rune(h.Content))
    }
    if pendingAssistant != "" {
        count += len([]rune(pendingAssistant))
[...]

PR #1761, opened 2026-08-28
fix(core): use real API-reported input tokens for auto-compress trigger

### LINKED ISSUE #1764: [Bug] auto_compress trigger uses text-only heuristic that misses 70-85% of context (tool_use/tool_result blocks + system prompt overhead)
## Summary

`auto_compress.max_tokens` triggers `/compress` when the engine estimates
context usage exceeds the threshold. Today the estimate is a text-only
heuristic:

```go
func estimateTokensWithPendingAssistant(entries []HistoryEntry, pendingAssistant string) int {
    count := 0
    for _, h := range entries {
        count += len([]rune(h.Content))
    }
    // ... + pendingAssistant runes ...
    return (count + 3) / 4
}
```

This systematically undercounts. `HistoryEntry.Content` only contains the
textual portion of the conversation — it does NOT include:

- `tool_use` blocks in assistant messages (1-5% of context)
- **`tool_result` blocks` in user messages (70-85% of context — the bulk of any conversation that read files, ran grep, or executed commands)
- System prompt (10-30K tokens per Claude Code session)
- Tool definitions (5-20K tokens)
- Skills / MCP servers / attachments (1-10K tokens)
- Last turn's `thinking` blocks (0-30K tokens, counted as input next turn)

A session that reads several files and runs shell commands will have a
trigger that sees only a tiny fraction of the actual prompt, and never
fires until the conversation grows far past the model's context window.

## Reproduction

1. cc-connect with a claudecode project, `auto_compress.max_tokens = 15000`
2. Conversation that reads files / runs tools (`cat`, `grep`, `bash`)
3. Watch the chat footer: the "ctx N%" indicator (driven by Claude Code's
   own API-reported usage) climbs past 50% while auto-compress never triggers
4. Eventually conversation hits context overflow / model degradation before auto-compress ever fires

## Root cause (already diagnosed)

The Claude Code adapter already tracks the real, API-reported prompt token
count per turn via `claudeSession.lastUsage` (populated in
`agent/claudecode/session.go::handleAssistant` from the last assistant
event's `usage` block). The same number drives the `ctx N%` footer.

The trigger code, however, uses the heuristic — the same data is right
there but the trigger never reads it.

Per Anthropic's usage semantics, three disjoint subsets sum to the
complete prompt: `input_tokens` + `cache_creation_input_tokens` +
`cache_read_input_tokens`. They are not overlapping — see Anthropic's
pricing structure where `cache_creation` is billed at 1.25× and
`cache_read` at 0.1× of `input_tokens`, which would be incoherent under
a subset interpretation.

## Related

- **#1115** (OPEN) — `auto_compress estimateTokens ignores system prompt /
  tools overhead + idle reset never fires due to Unlock() updating
  UpdatedAt`. This issue covers BOTH bugs; my fix addresses Bug 1
  (the real-usage half). The idle-reset / UpdatedAt half (Bug 2) is
  orthogonal and not addressed by this PR.

## Proposed solution

Prefer the agent's actual API-reported input tokens when the session
implements `ContextUsageReporter` — fallback to the existing heuristic
when the interface is not implemented or returns nil.

```go
estimate := contextEstimate // existing heuristic
if usage := replyFooterSessionContextUsage(state.agentSession); usage != nil && usage.UsedTokens > 0 {
    estimate = usage.UsedTokens
}
// existing threshold comparison using `estimate`
```

## Fallback / backward compatibility

When `ContextUsageReporter` is not implemented (e.g. agents other than
claudecode and copilot), or `GetContextUsage()` returns nil (before first
API response, after restart with no usage retained), the trigger falls
back to the existing text heuristic. Same config schema; no breaking change.

YongmaoLuo:
## Summary

Switch the auto-compress trigger from the text-only heuristic to the agent's actual API-reported input tokens when the session implements `ContextUsageReporter`. The Claude Code and Copilot adapters already populate this field for the reply footer's "ctx N%" indicator — this PR reuses the same signal for the trigger.

## Problem

`estimateTokensWithPendingAssistant` (the current trigger) only counts the rune length of user/assistant history text and divides by 4. It systematically misses the bulk of context — see issue #1764 for the full breakdown. Short version: `tool_result` blocks alone are 70-85% of any conversation that reads files, runs grep, or executes commands, and they are not in `HistoryEntry.Content`. Plus system prompt (10-30K), tool definitions (5-20K), skills/MCP/attachments (1-10K) — none in `HistoryEntry` at all.

The fix reuses `claudeSession.lastUsage.UsedTokens` (already computed for the footer) via the existing `ContextUsageReporter` interface. Falls back to the heuristic when the interface isn't implemented or returns nil.

## Anthropic usage semantics

`input_tokens + cache_creation_input_tokens + cache_read_input_tokens` are three **disjoint subsets** that sum to the complete prompt — not overlapping. See Anthropic's pricing (`cache_creation` 1.25×, `cache_read` 0.1× of `input_tokens`): incoherent under a subset interpretation. So:

```
claudeSession.lastUsage.UsedTokens == full prompt size of the next inference call
```

This already includes system prompt, tool definitions, skills, MCP servers, every `tool_use`/`tool_result` block, prior `thinking` blocks, attachments, and cache prefixes.

## Backward compatibility

- `ContextUsageReporter` not implemented → existing heuristic (no change)
- `GetContextUsage()` returns nil (before first API response, or after restart) → heuristic (no change)
- Same config schema, no breaking change
- The Copilot adapter also implements `ContextUsageReporter`, so the same fix applies automatically

## Tests

Two new tests in `core/engine_test.go`:

- **`TestAutoCompress_UsesRealUsageWhenAvailable`** — sets `contextUsage.UsedTokens = 50_000` while the heuristic threshold is 4 tokens; asserts `/compact` is sent (would have failed without the fix).
- **`TestAutoCompress_FallsBackToHeuristicWhenNoUsage`** — leaves `contextUsage` nil; asserts the existing heuristic path still triggers compression.

All 9 auto-compress/compress tests pass deterministically. `go vet ./...` clean, `gofmt -l` clean.

## Diff

```diff
@@ -5778,10 +5778,21 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
            contextEstimate := estimateTokensWithPendingAssistant(session.GetHistory(0), baseResponse)

-           // Evaluate auto-compress trigger (token estimate on user+assistant text,
-           // including this turn's assistant reply before it is appended to history).
+           // Evaluate auto-compress trigger.
+           //
+           // Prefer the agent's actual API-reported input tokens when the session
+           // implements ContextUsageReporter — the text-only heuristic misses
+           // tool_use/tool_result blocks (~70-85% of context) and the fixed overhead
+           // of system prompt + tools + skills (issue #1115). The real number is the
+           // size of the last assistant event's prompt: input + cache_creation +
+           // cache_read (the three are disjoint subsets that sum to the full prompt,
+           // per Anthropic's usage semantics), which already includes everything
+           // the model will see on the next inference call.
            if e.autoCompress.max_tokens > 0 {
                estimate := contextEstimate
+               if usage := replyFooterSessionContextUsage(state.agentSession); usage != nil && usage.UsedTokens > 0 {
+                   estimate = usage.UsedTokens
+               }
                now := time.Now()
                state.mu.Lock()
                last := state.lastAutoCompressAt
```

15 lines added in `core/engine.go` (12 of them comment), 118 lines of tests added.

## Related

- Closes #1764 — the new umbrella issue for this bug class
- Partial fix for #1115 — addresses Bug 1 (real-usage half of the trigger); Bug 2 (idle reset / UpdatedAt) is orthogonal and not in scope
- The fix is purely additive and composes with any future `system_overhead_tokens` config knob that #1115 might suggest

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -5778,10 +5778,21 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 
 			contextEstimate := estimateTokensWithPendingAssistant(session.GetHistory(0), baseResponse)
 
-			// Evaluate auto-compress trigger (token estimate on user+assistant text,
-			// including this turn's assistant reply before it is appended to history).
+			// Evaluate auto-compress trigger.
+			//
+			// Prefer the agent's actual API-reported input tokens when the session
+			// implements ContextUsageReporter — the text-only heuristic misses
+			// tool_use/tool_result blocks (~70-85% of context) and the fixed overhead
+			// of system prompt + tools + skills (issue #1115). The real number is the
+			// size of the last assistant event's prompt: input + cache_creation +
+			// cache_read (the three are disjoint subsets that sum to the full prompt,
+			// per Anthropic's usage semantics), which already includes everything
+			// the model will see on the next inference call.
 			if e.autoCompressEnabled && e.autoCompressMaxTokens > 0 {
 				estimate := contextEstimate
+				if usage := replyFooterSessionContextUsage(state.agentSession); usage != nil && usage.UsedTokens > 0 {
+					estimate = usage.UsedTokens
+				}
 				now := time.Now()
 				state.mu.Lock()
 				last := state.lastAutoCompressAt
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -9925,6 +9925,124 @@ func TestAutoCompress_TriggerAfterResult(t *testing.T) {
 	}
 }
 
+// TestAutoCompress_UsesRealUsageWhenAvailable verifies that when the agent session
+// implements ContextUsageReporter and reports a real API token count, the auto-compress
+// trigger uses that real number instead of the text-only heuristic — even when the
+// heuristic alone is below threshold. This guards against the bug where sessions with
+// large tool_use/tool_result blocks (the bulk of real context) never trigger compress.
+func TestAutoCompress_UsesRealUsageWhenAvailable(t *testing.T) {
+	p := &stubPlatformEngine{n: "test"}
+	sess := newQueuingSession("auto-compress-real-usage")
+	agent := &stubCompressorAgent{cmd: "/compact"}
+	e := NewEngine("test", agent, []Platform{p}, "", LangEnglish)
+	// Heuristic threshold of 4 tokens would NOT trigger on a short message,
+	// but the real usage of 50_000 tokens (representing tool_use/tool_result overhead)
+	// MUST trigger.
+	e.SetAutoCompressConfig(true, 4, 0)
+
+	key := "test:user1"
+	state := &interactiveState{
+		agentSession: sess,
+		platform:     p,
+		replyCtx:     "ctx",
+	}
+	e.interactiveMu.Lock()
+	e.interactiveStates[key] = state
+	e.interactiveMu.Unlock()
+
+	// Inject a real usage snapshot: 50k tokens used of a 200k window.
+	sess.contextUsage = &ContextUsage{
+		UsedTokens:    50_000,
+		ContextWindow: 200_000,
+	}
+
+	session := e.sessions.GetOrCreateActive(key)
+	session.AddHistory("user", "hi")
+
+	go e.processInteractiveEvents(state, session, e.sessions, key, "msg1", time.Now(), func() {}, nil, nil)
+	sess.events <- Event{Type: EventResult, Content: "response", Done: true}
+
+	deadline := time.After(2 * time.Second)
+	for {
+		sess.sendMu.Lock()
+		n := len(sess.sendCalls)
+		sess.sendMu.Unlock()
+		if n > 0 {
+			break
+		}
+		select {
+		case <-deadline:
+			t.Fatal("timed out waiting for auto-compress send — real usage (50k) should have triggered despite short text heuristic")
+		default:
+			time.Sleep(10 * time.Millisecond)
+		}
+	}
+
+	sess.sendMu.Lock()
+	last := sess.sendCalls[len(sess.sendCalls)-1]
+	sess.sendMu.Unlock()
+	if last != "/compact" {
+		t.Fatalf("expected /compact auto-compress driven by real usage, got %q", last)
+	}
+}
+
+// TestAutoCompress_FallsBackToHeuristicWhenNoUsage verifies that when the agent
+// session's GetContextUsage returns nil (no usage data yet), the trigger falls
+// back to the text-only heuristic — backward compatible behavior.
+func TestAutoCompress_FallsBackToHeuristicWhenNoUsage(t *testing.T) {
+	p := &stubPlatformEngine{n: "test"}
+	// Use a queuing session whose contextUsage is nil — exercises the
+	// "real usage unavailable, fall back to text heuristic" path.
+	sess := newQueuingSession("no-usage")
+	agent := &stubCompressorAgent{cmd: "/compact"}
+	e := NewEngine("test", agent, []Platform{p}, "", LangEnglish)
+	e.SetAutoCompressConfig(true, 4, 0)
+
+	key := "test:user1"
+	state := &interactiveState{
+		agentSession: sess,
+		platform:     p,
+		replyCtx:     "ctx",
+	}
+	e.interactiveMu.Lock()
+	e.interactiveStates[key] = state
+	e.interactiveMu.Unlock()
+
+	session := e.sessions.GetOrCreateActive(key)
+	// Long enough text to cross the heuristic threshold of 4 tokens.
+	session.AddHistory("user", strings.Repeat("a", 64))
+
+	go e.processInteractiveEvents(state, session, e.sessions, key, "msg1", time.Now(), func() {}, nil, nil)
+	sess.events <- Event{Type: EventResult, Content: "response", Done: true}
+
+	deadline := time.After(2 * time.Second)
+	for {
+		sess.sendMu.Lock()
+		n := len(sess.sendCalls)
+		sess.sendMu.Unlock()
+		if n > 0 {
+			break
+		}
+		select {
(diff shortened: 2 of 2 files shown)


chenhg5 (approved):
## Conclusion: Approve

**Overall assessment:** Excellent targeted change that fixes the long-standing `auto_compress` regression by switching the trigger from a rune/4 text heuristic (which silently misses 70-85% of real context) to the real, API-reported prompt token count that's already populated for the `ctx N%` footer. Implementation is minimal (4 lines of production code + 118 lines of focused tests), backward-compatible (preserves the existing path as fallback for sessions that don't implement `ContextUsageReporter`), and the test coverage directly proves both the fix and the fallback. CI is 5/5 green, `go vet` and `gofmt -l` are clean, full `core` package race-enabled tests pass in ~48 s, and the PR auto-merges against `origin/main` cleanly.

**Review scope:**
- Reviewed `core/engine.go` (+13/-2) and `core/engine_test.go` (+118/-0); total +131/-2 across 2 files.
- Traced the change through the existing `replyFooterSessionContextUsage` helper (`core/engine.go:7665`) and `ContextUsageReporter` interface (`core/interfaces.go:496`).
- Verified the three adapters that implement the interface: `agent/claudecode/session.go:1180` (`claudeSession.GetContextUsage()`), `agent/copilot/session.go:813` (with `var _ core.ContextUsageReporter = (*copilotSession)(nil)`), `agent/pi/session.go:1208`.
- Focused on correctness, backward compatibility, semantic accuracy of the `usage.UsedTokens` semantics, test coverage of both fix and fallback paths, and Anthropic's disjoint-subsets claim.

**✅ What looks good:**
- **Surgical change with maximum leverage**: 4 lines of production logic that reuse a data path the system already maintains. `claudeSession.lastUsage` was already populated for the reply footer's `ctx N%` indicator — this PR just reuses it for the trigger. No new data plumbing, no new dependencies.
- **Strict-superset semantics**: When `GetContextUsage()` returns nil (before first API response, after restart, or for agents that don't implement the interface), the existing heuristic path is preserved byte-identical. The fallback test (`TestAutoCompress_FallsBackToHeuristicWhenNoUsage`) explicitly proves this.
- **Anthropic usage semantics are correct**: The PR's claim that `input_tokens + cache_creation_input_tokens + cache_read_input_tokens` are three disjoint subsets summing to the complete prompt is consistent with Anthropic's pricing model (1.25× for cache_creation, 0.1× for cache_read — incoherent under overlapping interpretation). `UsedTokens` therefore equals the full next-inference-call prompt, including system prompt, tool definitions, every tool_use/tool_result block, thinking blocks, skills, MCP, and cache prefixes — exactly what the heuristic was missing.
- **Direct regression coverage**: `TestAutoCompress_UsesRealUsageWhenAvailable` constructs the exact failure scenario (50 k tokens of real usage vs. a 4-token heuristic threshold) and asserts `/compact` is sent. Without the fix, this test would time out. This is the right way to encode the regression.
- **Honest scope boundary**: PR description explicitly notes it addresses Bug 1 of #1115 (real-usage half) and Bug 2 (idle-reset / UpdatedAt) is orthogonal. Avoids scope creep.
- **Code hygiene**: 12 lines of the 13-line diff in `engine.go` are an explanatory comment referencing both issue numbers and the disjoint-subsets reasoning. Future maintainers will thank you.

**🟠 Should improve (not blocking, but worth a follow-up):**
- **P3 — Documentation**: The user-facing doc for `auto_compress.max_tokens` could add one line: "Estimate uses the agent adapter's API-reported prompt token count when available (claudecode/copilot/pi); falls back to a text-only heuristic otherwise." Operators reading the config schema won't otherwise know the trigger accuracy differs by agent.
- **P3 — Adapter template enforcement**: Three of the four agent adapters implement `ContextUsageReporter`. Any future adapter that forgets to implement it will silently fall back to the heuristic. A short comment in `core/interfaces.go` near `ContextUsageReporter`, or a startup-time INFO log when an adapter's session doesn't implement it, would catch this drift early.

**🔵 Optional:**
- **P3 — Idle-reset follow-up**: The orthogonal Bug 2 of #1115 (idle reset / `UpdatedAt`) deserves its own issue if maintainer wants to track it. The PR correctly stays out of scope.

**❓ Questions:**
- None.

**Testing / Risk:**
- **Verified locally:**
  - `CGO_ENABLED=1 go test -race -count=1 ./core/` → ok 48.337 s (full core package, no regressions)
  - `CGO_ENABLED=1 go test -race -count=1 -v -run 'TestAutoCompress_UsesRealUsageWhenAvailable|TestAutoCompress_FallsBackToHeuristicWhenNoUsage' ./core/` → 2/2 pass in 1.112 s
  - `go vet ./core/ ./agent/claudecode/` → clean
  - `gofmt -l core/engine.go core/engine_test.go` → clean
  - `git fetch origin main && git merge --no-commit --no-ff origin/main` → Already up to date
- **CI:** 5/5 green (run 33134956659): lint 2m13s, unit-test 4m58s, smoke-test 30s, regression-test 35s, performance-test 49s. `mergeStateStatus: CLEAN`.
- **Remaining risk:** Behaviour change in the trigger. If an operator's `auto_compress.max_tokens` was tuned for v1.5.0's undercounting heuristic, the real usage will reach threshold earlier. In most cases this is the desired behaviour (the trigger should fire more, not less), but operators should review their thresholds once. The fallback path keeps the old behaviour intact for any agent that doesn't implement the interface.

**Next step:** Approve & merge. Ship in v1.5.1 alongside #1743 / #1746 / #1752. A short release-note line: `auto_compress` trigger now uses the agent's API-reported prompt token count when available (claudecode / copilot / pi), falling back to the text heuristic otherwise.

🤖 Posted by cc-connect QA workflow (`chenhg5/cc-connect`). Decision: **approve**. External contribution review (`author != chenhg5`), so this is a proper `--approve` review.

PR #1765, opened 2026-08-28
fix(claudecode): implement graceful CancelTurn via control_request interrupt so /stop keeps the session alive

### LINKED ISSUE #1763: [Bug] /stop kills the claudecode subprocess; subsequent /compress, /clear, /resume fail with "No active session"
## Summary

After `/stop` on the claudecode adapter, the Claude Code subprocess is killed
and the engine's `interactiveStates` entry is deleted. The very next
`/compress` (or `/clear`, `/resume`) message replies with "❌ 没有活跃的会话可以压缩"
because `cmdCompress` (`core/engine.go:10449`) requires
`state.agentSession.Alive()` to be true.

This is inconsistent with the design established by #1737 (turn-to-turn keeps
the subprocess idle) and with native Claude Code TUI behavior (Esc interrupts
the current turn without ending the session).

## Reproduction

1. cc-connect v1.5.1-beta.1 with a claudecode project (latest beta, includes PR #1737 — i.e. the between-turn keep-idle fix is in).
2. Send a user message; let the agent start processing.
3. Send `/stop` — engine replies "Execution stopped".
4. Send `/compress` (or `/clear`, `/resume`).
5. Engine replies "❌ 没有活跃的会话可以压缩。请先发送一条消息。"

The claudecode subprocess has been killed by `/stop`, so there is no live CLI
to receive the `/compact` slash command via stdin.

## Root cause

`core/engine.go:10335 stopInteractiveSessionWithOptions` has two paths:

- **Canceller path** (10354-10397) — calls `CancelTurn()` on the agent
  session, keeps the subprocess and `interactiveStates` entry alive.
  Currently only ACP implements this interface (since #1275).
- **normalCleanup path** (10399-10421) — kills the subprocess, deletes
  the state, fires `HookEventSessionEnded`. This is what claudecode hits.

The claudecode adapter does NOT implement `AgentSessionCanceller`, so
`/stop` falls through to `normalCleanup` and the session is destroyed.

## Why this is the right fix

The `AgentSessionCanceller` interface (`core/interfaces.go:526`) and the
canceller-success code path in `stopInteractiveSessionWithOptions` were
added specifically for this scenario — they are unused infrastructure
today because only ACP implements it. Making claudecode implement it
brings the claudecode adapter in line with the existing pattern.

Empirically verified (against Claude Code v2.x in `--input-format stream-json`
mode) that the CLI accepts a stream-json `control_request` of subtype
`interrupt` over stdin: it replies with `control_response`, emits a
`result` event with `subtype=error_during_execution`, and returns to its
idle-waiting state with the same Claude session ID. The subprocess stays
alive, stdin stays open, the next user message resumes via stdin. This
matches TUI Esc semantics without using OS signals.

## Related history

- **#1736 / #1737** (CLOSED via PR #1737) — `/compact fails because CLI exits after each turn`. Fixed by dropping `--replay-user-messages`.
  Same symptom class (subprocess not available for `/compact`), different
  mechanism (CLI auto-exit vs explicit `/stop`). #1737 keeps the CLI alive
  between turns; this issue is about keeping it alive between stops.
- **#628** (OPEN, xukp20, 2026-04-15) — adds a separate `/interrupt`
  slash command that uses `control_request subtype=interrupt` for Claude
  Code (matching the mechanism I propose) and `turn/interrupt` for Codex
  app-server. The author's PR was marked "superseded by #957" by chenhg5's
  QA triage note on 2026-06-19; xukp20 has not closed it. The mechanism is
  the same; the design choice that differs is: PR #628 adds a NEW `/interrupt`
  command, my proposal hooks into the EXISTING `/stop` and makes it graceful
  (so users don't need to learn a second command). If maintainers prefer
  a separate command, the underlying fix is reusable.
- **#957** (MERGED 2026-06-07) — `feat(core): add /cancel command to
  interrupt and reset session`. Implements the engine-level
  `AgentSessionCanceller` dispatch path. My proposal is the claudecode-side
  concrete implementation of the interface #957 added — the chenhg5 QA note
  on #628 explicitly called this out as "follow-up to #957, not a replacement".
- **#1714** (MERGED 2026-08-27, yesterday) — `fix(claudecode): bound session
  teardown`. Adds `closingSessions` race-guard for the close/spawn window.
  Complementary to this proposal (their fix is about the kill-and-respawn
  path; mine is about avoiding the kill in the first place when graceful).

## Proposed solution

Make `*claudeSession` implement `AgentSessionCanceller` by writing a
`control_request` JSON message to stdin. Concretely:

```go
func (cs *claudeSession) CancelTurn() error {
    // ... liveness / stdin checks ...
    return cs.writeJSON(map[string]any{
        "type":       "control_request",
        "request_id": fmt.Sprintf("interrupt-%d", time.Now().UnixNano()),
        "request":    map[string]any{"subtype": "interrupt"},
    })
}
```

`cmdCancel` must follow up with an explicit close to prevent a subprocess
leak when the canceller branch keeps the CLI alive (cmdCancel's intent is
to discard everything, unlike `/stop` which preserves the session). Fix
shape: capture `state.agentSession` BEFORE calling `stopInteractiveSession`
and invoke `closeAgentSessionAsync` unconditionally after.

Add `HookEventTurnCancelled` to distinguish from `HookEventSessionEnded`:
the canceller-success path keeps the session alive, so emitting
`session.ended` would mislead listeners.

## Scope

claudecode only. Copilot and codex appserver would each need their own
transport-specific implementation (deferred). Class B adapters
(cursor/kimi/qoder/antigravity/gemini/codex exec) do not apply — they
spawn per turn and have no cross-turn idle state to preserve.

## Fallback

If `CancelTurn` returns an error, the engine's existing fallback path
(`engine.go:10381-10386`) already handles it correctly: fall through to
`normalCleanup` and kill the subprocess. This is what ACP uses today.

YongmaoLuo:
## Summary

Make `*claudeSession` implement `AgentSessionCanceller` by sending a `control_request` JSON message of subtype `interrupt` to the CLI's stdin. The CLI responds with `control_response`, emits `result/subtype=error_during_execution`, and returns to its idle-waiting state with the same Claude session ID. The subprocess stays alive and stdin stays open — the next user message resumes via stdin without a re-spawn.

Closes #1763.

## Problem

`/stop` on the claudecode adapter takes the `normalCleanup` path (`core/engine.go:10399-10421`): kills the Claude Code subprocess and deletes the engine's `interactiveStates` entry. The very next `/compress`, `/clear`, or `/resume` message replies with "❌ 没有活跃的会话可以压缩" because `cmdCompress` requires `state.agentSession.Alive()` (`engine.go:10449`).

This is inconsistent with the design philosophy established by #1737 (turn-to-turn keeps the subprocess idle) and with native Claude Code TUI behavior (Esc interrupts the current turn without ending the session).

The `AgentSessionCanceller` interface (`core/interfaces.go:526`) and the canceller-success code path in `stopInteractiveSessionWithOptions` were added specifically for this scenario. They are unused infrastructure today because only ACP implements the interface.

## Solution

```go
func (cs *claudeSession) CancelTurn() error {
    cs.stdinMu.Lock()
    stdinClosed := cs.stdin == nil
    cs.stdinMu.Unlock()
    if stdinClosed  { return fmt.Errorf("claudecode: stdin closed") }
    if !cs.alive.Load() { return fmt.Errorf("claudecode: process not alive") }
    return cs.writeJSON(map[string]any{
        "type":       "control_request",
        "request_id": fmt.Sprintf("interrupt-%d", time.Now().UnixNano()),
        "request":    map[string]any{"subtype": "interrupt"},
    })
}
```

Empirically verified against Claude Code v2.x in `--input-format stream-json` mode (test harness `/tmp/cc-cancel-verify/v7` in the local fork): the CLI accepts this stdin message, the subprocess stays alive, and the next user message after the interrupt completes a normal turn successfully.

## Why `control_request` and not SIGINT / control_response

- **Not SIGINT**: SIGINT in stream-json mode causes the CLI to emit `result/error_during_execution` AND close stdin — the next user message gets broken pipe (verified, v2 in the local harness). The CLI's SIGINT handler is not the same as TUI Esc.
- **Not bare stdin slash command** like `/stop`: slash commands are processed as regular user-message text — they don't interrupt the current turn (verified, v5 in the local harness).
- **Not `control_response`**: that's the response to a CLI-issued `control_request`. The wire is symmetric — we can issue a `control_request` to interrupt a turn.

## Three small follow-ups

1. **cmdCancel leak fix** (`core/engine.go`): cmdCancel's intent is to discard everything. When the canceller branch keeps the subprocess alive (newly reachable after this PR), cmdCancel would leak. Fix: capture `state.agentSession` BEFORE calling `stopInteractiveSession`, then call `closeAgentSessionAsync` unconditionally after. Idempotent with normalCleanup's already-invoked `Close()`.

2. **New `HookEventTurnCancelled`** (`core/hooks.go`): the canceller-success path emits `turn.cancelled` instead of `session.ended`. The session is NOT ended (subprocess alive, reusable). `session.ended` stays on `normalCleanup`. Listeners filtering on `session.ended` should add a second filter on `turn.cancelled`.

3. **Latent double-Unlock bug** (`stopInteractiveSessionWithOptions`): the previous code unlocked `e.interactiveMu` at the start of the canceller branch then unlocked it again at `normalCleanup` on the cancel-fails-fallback path. No agent previously implemented `AgentSessionCanceller` AND returned an error, so the bug was dormant. Restructured to call `CancelTurn` while holding the lock; only unlock on the success path. CancelTurn implementations must not acquire `interactiveMu` themselves.

## Tests

- `TestStopInteractiveSession_CancellerPath_KeepsStateAlive` — canceller path: state preserved, subprocess `Alive()==true`, Close() not called
- `TestCmdCancel_ForcesCloseAfterGracefulCancel` — leak-fix regression test
- `TestCmdStop_GracefulAllowsCompress` — end-to-end: `/stop` then `/compress` does NOT reply with `MsgCompressNoSession`
- `controllableAgentSession.CancelTurn` fixture (default returns error so existing tests stay on `normalCleanup`; opt-in via `cancelTurnFn`)
- `stubCancellerCompressorAgent` fixture (combines canceller + `CompressCommand`)

`go test ./core/ ./agent/claudecode/ -count=1` passes. `go vet ./...` clean, `gofmt -l` clean on touched files.

## Related

- #1736 / #1737 — same symptom class (subprocess not available for `/compact`), different cause (CLI auto-exit). #1737 keeps the CLI alive between turns; this PR extends the same principle to between stops.
- #628 (OPEN, xukp20) — adds a separate `/interrupt` slash command using the same `control_request subtype=interrupt` mechanism for Claude Code. The mechanism overlaps with this PR; the design choice differs: #628 adds a NEW command, this PR makes the EXISTING `/stop` graceful. If maintainers prefer a separate command, the underlying mechanism is reusable across both approaches.
- #957 — `feat(core): add /cancel command to interrupt and reset session`. Implements the engine-level `AgentSessionCanceller` dispatch path. This PR is the claudecode-side concrete implementation; the chenhg5 QA note on #628 called this out as "follow-up to #957, not a replacement".
- #1714 (MERGED 2026-08-27) — `fix(claudecode): bound session teardown`. Adds `closingSessions` race-guard for the close/spawn window. Complementary to this PR (their fix is about the kill-and-respawn path; mine is about avoiding the kill in the first place when graceful works).

## Scope

claudecode only. Copilot and codex appserver would each need their own transport-specific `CancelTurn` implementation (deferred). Class B adapters (cursor/kimi/qoder/antigravity/gemini/codex exec) do not apply — they spawn per turn and have no cross-turn idle state to preserve.

### DIFF
--- a/agent/claudecode/session.go
+++ b/agent/claudecode/session.go
@@ -1191,6 +1191,42 @@ func (cs *claudeSession) Alive() bool {
 	return cs.alive.Load()
 }
 
+// CancelTurn implements core.AgentSessionCanceller. Asks Claude Code
+// to interrupt the current generation over the existing stream-json
+// stdin channel — `{"type":"control_request","request":{"subtype":
+// "interrupt"}, "request_id": "..."}`. The CLI replies with a
+// `control_response`, emits a `result` event with
+// subtype="error_during_execution", and returns to its idle-waiting
+// state with the same Claude session ID. The subprocess stays alive
+// and stdin stays open so the next user message can be delivered
+// without a re-spawn — matching TUI Esc/Ctrl+C semantics through the
+// stream-json wire protocol instead of an in-process AbortController
+// (TUI-only) or an OS signal (which closes stdin in stream-json mode
+// and is unreliable on Windows).
+//
+// Verified empirically against Claude Code v2.x in --input-format
+// stream-json mode; see /tmp/cc-cancel-verify/v7 in the local fork.
+func (cs *claudeSession) CancelTurn() error {
+	cs.stdinMu.Lock()
+	stdinClosed := cs.stdin == nil
+	cs.stdinMu.Unlock()
+	if stdinClosed {
+		return fmt.Errorf("claudecode: stdin closed, session torn down")
+	}
+	if !cs.alive.Load() {
+		return fmt.Errorf("claudecode: process not alive")
+	}
+	requestID := fmt.Sprintf("interrupt-%d", time.Now().UnixNano())
+	if err := cs.writeJSON(map[string]any{
+		"type":       "control_request",
+		"request_id": requestID,
+		"request":    map[string]any{"subtype": "interrupt"},
+	}); err != nil {
+		return fmt.Errorf("claudecode: write interrupt: %w", err)
+	}
+	return nil
+}
+
 // defaultGracefulStopTimeout 是 Close() Phase 1「关掉 stdin、等它自己干净退出」
 // 的等待上限。
 //
--- a/core/engine.go
+++ b/core/engine.go
@@ -5778,10 +5778,21 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 
 			contextEstimate := estimateTokensWithPendingAssistant(session.GetHistory(0), baseResponse)
 
-			// Evaluate auto-compress trigger (token estimate on user+assistant text,
-			// including this turn's assistant reply before it is appended to history).
+			// Evaluate auto-compress trigger.
+			//
+			// Prefer the agent's actual API-reported input tokens when the session
+			// implements ContextUsageReporter — the text-only heuristic misses
+			// tool_use/tool_result blocks (~70-85% of context) and the fixed overhead
+			// of system prompt + tools + skills (issue #1115). The real number is the
+			// size of the last assistant event's prompt: input + cache_creation +
+			// cache_read (the three are disjoint subsets that sum to the full prompt,
+			// per Anthropic's usage semantics), which already includes everything
+			// the model will see on the next inference call.
 			if e.autoCompressEnabled && e.autoCompressMaxTokens > 0 {
 				estimate := contextEstimate
+				if usage := replyFooterSessionContextUsage(state.agentSession); usage != nil && usage.UsedTokens > 0 {
+					estimate = usage.UsedTokens
+				}
 				now := time.Now()
 				state.mu.Lock()
 				last := state.lastAutoCompressAt
@@ -10294,11 +10305,26 @@ func (e *Engine) cmdCancel(p Platform, msg *Message) {
 
 	slog.Info("cmdCancel: stopping execution and creating new session", "session_key", msg.SessionKey)
 
+	// Capture the live agent session BEFORE stopInteractiveSession so we can
+	// force-close it regardless of which path stopInteractiveSession takes.
+	// When the agent implements AgentSessionCanceller (claudecode after
+	// PR-style fix, ACP, etc.) the canceller branch keeps the subprocess
+	// alive — without this follow-up close, cmdCancel would leak the
+	// subprocess. closeAgentSessionAsync is idempotent with the
+	// normalCleanup path, so the double-close is harmless.
+	savedAgent := e.interactiveStateForCancel(interactiveKey)
+
 	// Stop the current execution (like /stop)
 	stopped := e.stopInteractiveSession(interactiveKey, p, msg.ReplyCtx)
 	if !stopped {
 		// No execution in progress, but still create a new session
 		slog.Debug("cmdCancel: no execution to stop, proceeding with new session", "session_key", msg.SessionKey)
+	}
+
+	// Force-close the captured subprocess in case the canceller path kept
+	// it alive. Idempotent with normalCleanup, which already invoked Close.
+	if savedAgent != nil {
+		e.closeAgentSessionAsync(interactiveKey, savedAgent, p, msg.ReplyCtx)
 	}
 
 	// Clear old session's agent session ID so it cannot be resumed
@@ -10340,13 +10366,26 @@ func (e *Engine) stopInteractiveSessionWithOptions(sessionKey string, notifyQueu
 	closeReplyCtx := state.replyCtx
 	state.mu.Unlock()
 
-	// If the agent session supports graceful turn cancellation (e.g. ACP),
-	// send a cancel notification and keep the session alive for the next
-	// user message, rather than killing the process and destroying state.
+	// If the agent session supports graceful turn cancellation (e.g. ACP,
+	// claudecode via stream-json control_request interrupt), call CancelTurn
+	// and keep the session alive for the next user message rather than
+	// killing the process and destroying state.
+	//
+	// We hold e.interactiveMu across the CancelTurn call so that on failure
+	// (cancelErr != nil) we can fall through to normalCleanup below without
+	// a double-unlock. CancelTurn implementations must NOT acquire
+	// e.interactiveMu themselves — they operate on the agent subprocess's
+	// own state (stdin write, etc.) and don't touch engine bookkeeping.
 	if canceller, ok := agentSession.(AgentSessionCanceller); ok && agentSession != nil {
-		// Keep the state in the map so the next message reuses this session.
-		// Don't markStopped — the session is still usable.
-		// Don't delete from interactiveStates — keep it alive.
+		cancelErr := canceller.CancelTurn()
+		if cancelErr != nil {
+			slog.Warn("agent session CancelTurn failed, falling back to Close",
+				"session_key", sessionKey, "error", cancelErr)
+			// Fall through to normalCleanup below — keep lock held.
+			goto normalCleanup
+		}
+
+		// Canceller succeeded. Now release the lock and finalize state.
 		e.interactiveMu.Unlock()
 
 		if pending != nil {
@@ -10366,19 +10405,11 @@ func (e *Engine) stopInteractiveSessionWithOptions(sessionKey string, notifyQueu
 		state.eventsNeedResync = true
 		state.mu.Unlock()
 
-		cancelErr := canceller.CancelTurn()
-		if cancelErr != nil {
-			slog.Warn("agent session CancelTurn failed, falling back to Close",
-				"session_key", sessionKey, "error", cancelErr)
-			// Fall through to normal cleanup below.
-			goto normalCleanup
-		}
-
 		slog.Info("agent session turn cancelled, session kept alive",
 			"session_key", sessionKey)
 
 		e.hooks.Emit(HookEvent{
-			Event:      HookEventSessionEnded,
+			Event:      HookEventTurnCancelled,
 			SessionKey: sessionKey,
 		})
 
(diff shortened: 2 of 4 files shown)


chenhg5 (approved):
## Conclusion: Approve

**Overall assessment:** Clean, well-scoped change that finally activates the previously-dormant `AgentSessionCanceller` infrastructure for the claudecode adapter. Implementation matches the public stream-json `control_request interrupt` protocol (verified by the author against Claude Code CLI v2.x), preserves full backward compatibility (failure path falls back to `normalCleanup`; non-claudecode adapters untouched), and proactively fixes three collateral issues that would have been latent bugs once the canceller became reachable: the cmdCancel subprocess leak, a double-`Unlock` race in `stopInteractiveSessionWithOptions`, and the wrong hook event (`session.ended` → `turn.cancelled`). Test coverage is targeted at exactly the new behavior. CI is 5/5 green and auto-merge against `origin/main` is clean.

**Review scope:**
- Reviewed all 4 changed files: `agent/claudecode/session.go` (+36), `core/engine.go` (+65/-17), `core/hooks.go` (+20/-19, mostly gofmt whitespace), `core/engine_test.go` (+312). Total +433/-36.
- Traced the change through `core.AgentSessionCanceller` (interface at `core/interfaces.go:526`) and the new `core.HookEventTurnCancelled` constant.
- Verified the `CancelTurn` protocol against the stream-json wire format the author documented.
- Focused on correctness, the cmdCancel leak fix, the double-Unlock race fix, hook event semantics, fallback safety, and cross-adapter impact.

**✅ What looks good:**
- **Strict-superset semantics**: claudecode now implements an existing interface; the canceller path is gated on the type assertion. Every other adapter (ACP, pi, copilot, codex, etc.) takes the same pre-existing path as before. Zero risk for non-claudecode users.
- **Defensive double-call guards**: `CancelTurn` checks `stdin == nil` and `!cs.alive.Load()` before sending — both correctly return errors that fall through to `normalCleanup`. `closeAgentSessionAsync` is documented as idempotent with the normalCleanup path's `Close()` so the cmdCancel follow-up is safe even on overlap.
- **Latent double-Unlock fix is real**: pre-PR, `stopInteractiveSessionWithOptions` would have `e.interactiveMu.Unlock()` then `goto normalCleanup` where the same lock would have been unlocked again. No agent previously implemented `AgentSessionCanceller` AND returned error, so the bug was dormant — but the new claudecode `CancelTurn` could trip it. Reordering to "hold lock across `CancelTurn`, unlock only on success" is the right call.
- **Hook event semantics**: `session.ended` previously fired on the canceller-success path too, which was misleading — the session is *not* ended (subprocess alive, reusable). The new `turn.cancelled` cleanly separates the two states. Existing `session.ended` listeners are unaffected.
- **Test design**: `controllableAgentSession` defaults its `CancelTurn` to return error so existing tests stay on the `normalCleanup` path — opt-in via `cancelTurnFn`. This avoids regression in the broader test suite, which I confirmed by running `./core/` + `./agent/claudecode/` under `-race` (both pass).
- **End-to-end test (`TestCmdStop_GracefulAllowsCompress`)**: directly encodes the issue's reproduction — `/stop` then `/compress` — and asserts success. This is the right level of coverage.

**🟠 Should improve (not blocking, but worth flagging):**
- **P2 — Coordination with PR #1761**: this PR's single commit (`9296243`) **also contains PR #1761's auto-compress real-usage change** (lines 5781-5795 in `core/engine.go`). The same author wrote both, and the changes were merged into one commit. PR #1761 is approved but not yet merged. Maintainer must decide:
  - **Option A** (recommended): merge #1761 first, then rebase #1765 to drop the duplicate.
  - **Option B**: merge #1765 directly (gets both fixes), then close #1761 as superseded.
  - Either preserves both fixes; the QA approval on #1761 stays valid in either case. **Not a blocker**, but worth surfacing before merge so the maintainer isn't surprised by the overlap.
- **P3 — CLI version compatibility**: `CancelTurn` sends `control_request subtype=interrupt`, which is a Claude Code CLI v2.x feature. Older CLI versions will silently accept the write but ignore it — `/stop` would feel like it didn't work (turn continues to natural completion). No protocol handshake to detect this. Possible mitigation: log a one-time WARN when `claudeVersion` (if exposed) is below a known minimum, but this is a follow-up.

**🔵 Optional:**
- **P3 — Test coverage gap**: the 3 new tests cover happy paths; the "CancelTurn returns error → fallback to normalCleanup → subprocess killed + interactiveStates cleared" path is exercised by every pre-existing test that uses the default `controllableAgentSession` but isn't asserted explicitly. A dedicated negative test would harden this regression vector.
- **P3 — Docs for `HookEventTurnCancelled`**: the hook reference doc (if any) should call out the new event and the semantic difference from `session.ended`. Operators writing cross-platform hooks will want to know which to subscribe to.
- **P3 — Copilot / Codex appserver Canceller**: author explicitly deferred; worth a tracking issue so it doesn't get lost.

**❓ Questions:**
- None.

**Testing / Risk:**
- **Verified locally:**
  - `CGO_ENABLED=1 go test -race -count=1 ./core/ ./agent/claudecode/` → ok 48.4 s + 14.4 s (no regressions)
  - `CGO_ENABLED=1 go test -race -count=1 -v -run 'TestStopInteractiveSession_CancellerPath_KeepsStateAlive|TestCmdCancel_ForcesCloseAfterGracefulCancel|TestCmdStop_GracefulAllowsCompress' ./core/` → 3/3 pass in 1.15 s
  - `CGO_ENABLED=1 go test -race -count=1 -run 'TestAutoCompress_UsesRealUsageWhenAvailable|TestAutoCompress_FallsBackToHeuristicWhenNoUsage' ./core/` → 2/2 pass (confirms PR #1761 tests work on PR #1765's head)
  - `go vet ./core/ ./agent/claudecode/` → clean
  - `gofmt -l` on all 4 changed files → clean
  - `git fetch origin main && git merge --no-commit --no-ff origin/main` → Already up to date
- **CI:** 5/5 green (run 33146810078): lint 2m27s, unit-test 4m46s, smoke-test 26s, regression-test 32s, performance-test 49s. `mergeStateStatus: CLEAN`.
- **Remaining risk:** Behaviour change for claudecode `/stop` users — session stays alive, `/compress` / `/clear` / `/resume` work afterwards. This is the desired fix. Hook listeners subscribed only to `session.ended` will miss the new `turn.cancelled` events; they need to add a second subscription if they care about turn-level cancellation.

**Next step:** Approve & merge. **Maintainer decision required before merge**: confirm Option A (rebase after #1761 lands) or Option B (merge #1765, close #1761). Either is fine for shipping in v1.5.1 — the v1.5.1 hotfix candidate set (#1743 + #1746 + #1752 + #1761 + #1765) becomes 4 PRs either way.

🤖 Posted by cc-connect QA workflow (`chenhg5/cc-connect`). Decision: **approve**. External contribution review (`author != chenhg5`), so this is a proper `--approve` review with explicit P2 note about the #1761 overlap.

PR #1768, opened 2026-08-29
fix(pi): reference non-image attachments by plain path, not @<path> (#1767)

### LINKED ISSUE #1767: [Bug] pi attachments passed as @<path> get inlined by pi, blowing model context for large files (inconsistent with claudecode path references)
## Summary

Sending a **non-image** file (e.g. a 5.8MB PDF or a video) to a **pi** agent via cc-connect results in a `400` from the model: the file content is inlined into the user message, pushing the request over the model's context limit.

This is caused by a **design inconsistency in cc-connect**: pi attachments are passed as `@<path>` argv references (which pi then **inlines**), whereas claudecode attachments are saved to disk and passed as **plain path references** in the prompt (no inlining).

## Reproduction

1. Send a file (>~1MB, non-image, e.g. a PDF) to a pi agent through cc-connect (web/weixin).
2. cc-connect 1.5.1-beta.1 passes it as `@<path>` (Issue #1723, commit `acd22356`).
3. pi's `processFileArguments` (`dist/cli/file-processor.js`) reads the whole file as UTF-8 text and inlines it into the user message (`<file name="...">...content...</file>`), with **no size limit**.
4. The model API returns:
   ```
   400: This model's maximum context length is 1048576 tokens. However, you requested 3965411 tokens ...
   ```

Observed with a **5.8MB PDF** → the inlined text was ~6.15M chars → ~3.9M tokens.

## Root Cause

cc-connect Issue #1723 intentionally stopped cc-connect from inlining attachment bytes (which previously caused a `bufio.Scanner: token too long` crash in the pi subprocess), and switched to passing `@<path>` argv refs so **pi** loads them.

However, **pi's `processFileArguments` only treats *images* as attachments** (`images.push(...)`, `processImage`). Every other file type (PDF, video, audio, ...) falls into the "text file" branch and is inlined in full:

```js
// non-image branch
const content = stripBom(await readFile(absolutePath, "utf-8"));
text += `<file name="${absolutePath}">\n${content}\n</file>\n`;
```

There is **no size cap** and binary content is dumped as text (useless to the model and huge).

## Comparison with claudecode

`agent/claudecode/session.go` (Send) does NOT inline:

```go
filePaths := core.SaveFilesToDisk(cs.workDir, messageID, files)
textPart += "\n\n(Files saved locally, please read them: " + strings.Join(filePaths, ", ") + ")"
```

So claudecode hands the model a **path reference** and lets it read via tools; pi gets the **whole file inlined**.

## Impact

- Any large non-image attachment sent to a pi agent blows the model context.
- Small non-image files "work" but are dumped as raw bytes (not useful as text).
- This compounds with #145 (session becomes permanently corrupted once a large file is embedded inline).

## Suggested Fix — remove inlining entirely

The correct fix is to **completely stop inlining file content into the prompt/messages for pi**. All files should be passed as **path references**, never as embedded bytes. This also makes pi behave consistently with claudecode, and avoids the "binaries dumped as text" problem for small files.

1. **cc-connect (primary)**: pass pi attachments as **path references in the prompt** — mirror claudecode's `SaveFilesToDisk` + `(Files saved locally, please read them: <paths>)`. Do **not** use `@<path>` argv for file attachments. The model then reads via tools (pi's read tool truncates to 50KB, so context stays safe).
   - Images may still go as proper image attachments (pi's `processImage`), which is not "inlining" — that's fine.
   - Non-image / binary / any content: **reference only, never embed**.

2. **pi (secondary / defense in depth)**: in `processFileArguments`, **do not inline the file content for `@<path>` args at all**. Emit only the reference (`<file name="..."></file>`), leaving it to the model to decide whether to read via tools. This also removes the session-corruption vector from #145.

## Environment

- cc-connect `1.5.1-beta.1`
- `@earendil-works/pi-coding-agent@0.84.3`
- Node `v22.23.2`

## Related

- #145 — "Pi session corrupted by large file embeds — no way to recover without /new" (same symptom family; this issue is about the root cause + design inconsistency and the fix).

chenhg5:
## Summary

Fixes #1767: pi's `processFileArguments` inlines the full UTF-8 contents of every `@<path>` attachment into the prompt the model sees. For a >~1MB non-image attachment this triggers a 400 from the model provider. The `agent/pi` adapter was passing both images AND non-image files through the same single `atFiles` slice, so every non-image file hit the inliner.

Mirrors the `claudecode` behaviour:

- **Images** continue to use pi's `@<path>` argv / message-text mechanism. `pi`'s `processImage` handles them correctly and the bytes never enter our prompt text.
- **Non-image files** are saved to disk (unchanged) but referenced by **plain path** in a trailer appended to the prompt: `(Files saved locally, please read them: <paths>)`. Pi's own file-reading tools then load only the bytes the model actually needs.

The `agent/pi.Send()` pipeline now separates the two streams:

```
imageAtFiles  → buildJSONArgs / composeRPCPrompt → @<path> argv or message-text refs (Issue #1723 path)
filePaths     → promptWithFileRefs(prompt, filePaths) → plain trailer appended to prompt (Issue #1767 path)
```

Both `sendJSON` and `sendRPC` route through the new `promptWithFileRefs` helper so the trailer wording stays identical across modes.

## Root cause (Issue #1767)

Before this PR the `atFiles` slice in `Send()` aggregated both `saveImagesToDisk(...)` and `saveFilesToDisk(...)` into one bucket:

```go
var atFiles []string
if len(images) > 0 {
    atFiles = append(atFiles, saveImagesToDisk(attachDir, images)...)
}
if len(files) > 0 {
    atFiles = append(atFiles, saveFilesToDisk(attachDir, files)...)
}
```

Both `sendJSON` (via `buildJSONArgs`) and `sendRPC` (via `composeRPCPrompt`) then treated every entry as a `@<path>` reference. For images this is correct; for non-image files pi's `processFileArguments` reads the file contents and inlines them as UTF-8 text into the prompt.

## Changes

- `agent/pi/session.go`
  - `Send()`: split the single `atFiles` slice into `imageAtFiles` (only images) and `filePaths` (only non-image files). Comments now document the two distinct mechanisms.
  - `sendJSON(prompt, imageAtFiles, filePaths)`: pass `imageAtFiles` to `buildJSONArgs` and prepend the trailer produced by `promptWithFileRefs(prompt, filePaths)` to the prompt before it goes into `-p`. The trailer wording mirrors the `claudecode` agent so users see consistent phrasing across runtimes.
  - `sendRPC(prompt, imageAtFiles, filePaths)`: same pattern via `composeRPCPrompt`.
  - New `promptWithFileRefs(prompt, filePaths) string` helper: returns the original prompt when `filePaths` is empty (no spurious "Files saved locally" trailer), otherwise appends `"\n\n(Files saved locally, please read them: <paths>)"`.

- `agent/pi/session_test.go` (+273 lines)
  - 9 new tests pinning the new contract:
    - `TestPromptWithFileRefs_NoFilesKeepsPrompt` — empty `filePaths` → no trailer (matches the `composeRPCPrompt` invariant on the image side).
    - `TestPromptWithFileRefs_AppendsPlainPathRefs` — paths appear in trailer, **never** `@`-prefixed (would re-trigger #1767).
    - `TestPromptWithFileRefs_PreservesOrder` — multi-file reviews keep upload order.
    - `TestPromptWithFileRefs_DoesNotInlineFileBytes` — recognisable marker bytes written to disk must NOT appear in the returned prompt.
    - `TestPromptWithFileRefs_LargeNonImageFileDoesNotInline` — **spec-grade regression**: 1.5MB file with a recognisable marker scattered through it; assert the marker never appears in the prompt and the path is referenced as plain text.
    - `TestBuildJSONArgs_OnlyImageAtFilesBecomeAtPath` — JSON-mode invariant: `@`-args contain only image paths, never non-image paths.
    - `TestComposeRPCPrompt_OnlyImagesBecomeAtPathRefs` — RPC-mode twin.
    - `TestSaveFilesToDisk_LargeFileRoundTrip` — `saveFilesToDisk` must round-trip a 1.2MB file without truncation.
    - `TestSaveImagesToDisk_DoesNotConsumeFileSlots` — defensive guard against future refactors that mix the two producers.

## Tests

```
go test ./agent/pi/ -count=1
go test -race ./agent/pi/ -count=1
golangci-lint run ./agent/pi/...
```

All existing `agent/pi` tests still pass (the only `errcheck` items lint reports on the touched files are the pre-existing ones in `pi.go` / `pi_test.go`, untouched by this PR). 9 new tests added.

## Risk / compatibility

- Behaviour change: image attachments still work exactly as before (Issue #1723 path). Non-image attachments now arrive at the model as path references rather than inlined bytes — the model will need to call its file-reading tool to load them, which is the same behaviour as the `claudecode` agent. This is the intended fix.
- File bytes still land on disk exactly as the platform sent them (`saveFilesToDisk` unchanged); the only change is how those bytes reach the model.
- No public API changes; `Send()` signature is unchanged.

## Related

- Issue #1767
- Issue #1723 (predecessor: pass attachments as `@<path>` argv instead of inlining bytes — preserved for images)

### DIFF
--- a/agent/pi/session.go
+++ b/agent/pi/session.go
@@ -333,38 +333,48 @@ func (s *piSession) Send(msg string, messageID string, images []core.ImageAttach
 	}
 	cleanAttachments(attachDir)
 
-	var atFiles []string
+	// Issue #1723: images are passed via pi's @<path> argv / message-text
+	// mechanism. pi's processImage loads them as visual inputs and the
+	// bytes never enter our prompt text.
+	//
+	// Issue #1767: non-image files are NOT passed via @<path>. pi's
+	// processFileArguments reads every @<path> file's full UTF-8 contents
+	// and inlines them into the prompt the model sees. For a >~1MB text
+	// attachment this can blow the model's context and trigger a 400 from
+	// the provider. Mirror the claudecode behaviour: save non-image files
+	// to disk and tell the model where they live via a plain path
+	// reference; pi's own file-reading tools will load only the bytes the
+	// model actually needs.
+	var imageAtFiles []string
 	if len(images) > 0 {
-		atFiles = append(atFiles, saveImagesToDisk(attachDir, images)...)
+		imageAtFiles = append(imageAtFiles, saveImagesToDisk(attachDir, images)...)
 	}
+	var filePaths []string
 	if len(files) > 0 {
-		atFiles = append(atFiles, saveFilesToDisk(attachDir, files)...)
+		filePaths = append(filePaths, saveFilesToDisk(attachDir, files)...)
 	}
 
-	// Issue #1723: do NOT inline attachment bytes back into the prompt.
-	// json mode passes the prompt via -p <argv>, so embedding raw image
-	// bytes triggers execve EINVAL on NUL bytes and the pi process never
-	// starts (model never sees the image). rpc mode goes through stdin
-	// so it doesn't crash, but the model still gets image bytes as text
-	// rather than a real visual input. Both paths must hand pi the saved
-	// paths via pi's @<path> mechanism — sendJSON appends them as argv
-	// entries; sendRPC embeds them in the message text, which pi parses
-	// the same way.
 	if s.rpc {
-		return s.sendRPC(msg, atFiles)
+		return s.sendRPC(msg, imageAtFiles, filePaths)
 	}
-	return s.sendJSON(msg, atFiles)
+	return s.sendJSON(msg, imageAtFiles, filePaths)
 }
 
 // sendJSON spawns `pi --mode json -p <prompt>` as a one-shot process,
 // reads all output events, and sends them to the events channel.
 //
-// Issue #1723: attachment paths are passed as @<path> argv entries (Pi's
-// standard mechanism). Embedding the file bytes into -p would crash
+// Issue #1723: image paths are passed as @<path> argv entries (Pi's
+// standard mechanism). Embedding the image bytes into -p would crash
 // execve on NUL bytes, and would also break pi's vision pipeline which
 // needs to know the file is on disk rather than be handed raw bytes.
-func (s *piSession) sendJSON(prompt string, atFiles []string) error {
-	args := buildJSONArgs(s.extraArgs, prompt, s.CurrentSessionID(), s.model, s.thinking, atFiles)
+//
+// Issue #1767: filePaths (non-image attachments) are NOT passed as
+// @<path>. We append a plain "Files saved locally, please read them: …"
+// trailer to the prompt instead — pi's processFileArguments would
+// otherwise inline the entire UTF-8 contents into the prompt the model
+// sees, which triggers a 400 for files larger than the model's context.
+func (s *piSession) sendJSON(prompt string, imageAtFiles []string, filePaths []string) error {
+	args := buildJSONArgs(s.extraArgs, promptWithFileRefs(prompt, filePaths), s.CurrentSessionID(), s.model, s.thinking, imageAtFiles)
 
 	slog.Debug("piSession: spawning json mode", "cmd", s.cmd, "sessionID", s.CurrentSessionID())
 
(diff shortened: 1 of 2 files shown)


chenhg5 (commented):
Conclusion: Approve (commented because chenhg5 cannot self-approve)

Overall assessment:
- PR cleanly fixes Issue #1767 by mirroring the claudecode behavior for pi non-image attachments. Image attachments keep the `@<path>` argv path from #1723, so neither regression is reintroduced. CI is 5/5 green and all new tests pass with `-race` locally.

Review scope:
- Reviewed `agent/pi/session.go` (+61/-25) and `agent/pi/session_test.go` (+273/-0).
- Focused on correctness (no inlining, no image regression), test coverage, and consistency with the claudecode attachment contract.

✅ What looks good:
- The split between `imageAtFiles` (passed as `@<path>` argv) and `filePaths` (plain path trailer) is exactly the right factoring — it makes the #1723 vs #1767 distinction explicit at the call site (`Send`) and lets both `sendJSON` and `sendRPC` share the same composition layer via `promptWithFileRefs`.
- `promptWithFileRefs` returns the original prompt unchanged when `filePaths` is empty, with an explicit comment about why an empty "Files saved locally" trailer would confuse the model. Same defensive pattern as `composeRPCPrompt` already had for `atFiles`.
- The trailer wording `"(Files saved locally, please read them: …)"` matches the claudecode wording exactly, so users see consistent phrasing across runtimes.
- Regression tests are well-targeted:
  - `TestPromptWithFileRefs_DoesNotInlineFileBytes` and `TestPromptWithFileRefs_LargeNonImageFileDoesNotInline` pin the core invariant (no byte leakage) using recognizable markers.
  - `TestBuildJSONArgs_OnlyImageAtFilesBecomeAtPath` and `TestComposeRPCPrompt_OnlyImagesBecomeAtPathRefs` confirm the split is observable end-to-end without spawning pi.
  - `TestSaveFilesToDisk_LargeFileRoundTrip` guards against a future regression where the saved file is truncated (the path is the contract — the file must be the full payload).
  - `TestSaveImagesToDisk_DoesNotConsumeFileSlots` is a nice defensive guard against a future refactor accidentally routing FileAttachment bytes through `saveImagesToDisk`.
- Code comments on both `Send`/`sendJSON`/`sendRPC` explicitly cite #1723 and #1767, so future maintainers immediately understand why the two paths differ.

🔵 Optional (nit, non-blocking):
- `agent/pi/session_test.go` is missing a trailing newline (`gofmt -l` reports it). Trivial follow-up; one-line fix.

Testing / Risk:
- Verified evidence:
  - GitHub Actions: lint, unit-test, smoke-test, regression-test, performance-test — all SUCCESS (run 33222544862).
  - Local: `gofmt -l` (only the EOF nit), `go vet ./agent/pi/` clean, `go build ./agent/pi/` clean, `CGO_ENABLED=1 go test -race -count=1 ./agent/pi/` passes (7.079s).
  - Auto-merge with `origin/main`: already up to date, no conflicts.
- Remaining risk:
  - Prompt text gains a small trailer proportional to path-list size — negligible for typical use; consistent with the existing claudecode wording.
  - If a future pi version changes `processFileArguments` to stop inlining `@<path>` text, this PR's split remains correct (images still benefit from `@<path>` visual handling; non-images still avoid inlining by design).

Next step:
- Merge. The remaining nit is not worth a separate round trip. If the author wants to land the EOF fix, a one-line follow-up commit is fine; otherwise it can go in any future cleanup PR.

Issue #1244, opened 2026-06-07
Canonical: new agent/platform integrations should use ACP first


chenhg5:
## Goal

Use this issue as the canonical backlog item for new agent/platform integration requests that are not already tracked by a concrete implementation issue.

## ACP-first policy

For new AI agent integrations, the preferred path is:

1. Try the generic ACP adapter first: `type = "acp"` with the agent's documented ACP command/args.
2. If the agent works through ACP, do not create a dedicated adapter issue unless there is a concrete compatibility gap.
3. If the agent does not expose ACP or needs first-class behavior that ACP cannot provide, add details here:
   - official docs link
   - headless/stdio/API entrypoint
   - session/resume behavior
   - streaming output format
   - permission/tool-call behavior
   - concrete user workflow

## Currently consolidated examples

- just-every/code
- CodeBuddy
- Zoom platform
- Linear Agent
- Grok / model-agent request
- Codex Desktop backend
- Other future agent/platform requests without a concrete PR

## Non-goal

This issue does not replace concrete bug reports. If an ACP-based integration fails with logs, open a focused bug issue with config, version, and logs.

chenhg5:
Thanks @chenhg5 — this is a useful canonical tracker. We've added it to the PM backlog as the default entry point for ACP-first integration requests. Concrete bug reports and specific integration issues should still be opened separately (per the non-goal). When a new request comes in that fits the consolidation list, we'll link it back here. Closing the visibility gap (no maintainer ack on this thread).

PR #1770, opened 2026-08-29
fix(copilot): send enableConfigDiscovery so skills load (#1769)

### LINKED ISSUE #1769: copilot adapter: skills under ~/.agents/skills never load (session.create omits enableConfigDiscovery)
## Summary

When cc-connect drives GitHub Copilot CLI via `type = "copilot"`, **all environment-discovered skills are silently dropped**. Only the 2 builtin skills survive. Skills under `~/.agents/skills`, `~/.copilot/skills`, `.github/skills`, `.claude/skills` etc. are never visible to the agent, even though the same skills load fine when Copilot CLI is used directly.

Root cause: the `copilot` adapter never sends `enableConfigDiscovery` in its `session.create` params. Copilot CLI's session-config builder defaults that field to `false`, and `enableSkills` is derived from it — so every cc-connect session starts with skill discovery disabled.

## Environment

| | |
|---|---|
| cc-connect | v1.5.0 (`17c61062`), bug also present at HEAD `80713167` / v1.5.1-beta.1 |
| GitHub Copilot CLI | 1.0.81 |
| OS | Linux 6.17.0-1022-azure x86_64 |
| Agent type | `type = "copilot"` |

## Reproduction

Config:

```toml
[[projects]]
  name = "copilot"
  mode = "multi-workspace"
  base_dir = "/home/user/copilot"

  [projects.agent]
    type = "copilot"

    [projects.agent.options]
      mode = "bypassPermissions"
      model = "gpt-5.6-sol"
```

With 33 skills installed under `~/.agents/skills/`, ask the agent in chat to list its available skills. It reports only `customize-cloud-agent` and `github-pr-media`.

Verified directly against Copilot CLI's own prompt: parsing the `<available_skills>` block out of `~/.copilot/logs/process-*.log` shows

```
skills=2   locations: ['builtin', 'builtin']
```

whereas `copilot skill list` and `copilot -p "..."` in the same directory both see all 35.

## Root cause

`agent/copilot/session.go:66` — `copilotSessionConfig` has no field for `enableConfigDiscovery`, so it can never appear in the `session.create` payload:

```go
type copilotSessionConfig struct {
	SessionID                      string                     `json:"sessionId,omitempty"`
	ClientName                     string                     `json:"clientName,omitempty"`
	Model                          string                     `json:"model,omitempty"`
	Provider                       *copilotWireProviderConfig `json:"provider,omitempty"`
	RequestPermission              *bool                      `json:"requestPermission,omitempty"`
	WorkingDirectory               string                     `json:"workingDirectory,omitempty"`
	Streaming                      *bool                      `json:"streaming,omitempty"`
	IncludeSubAgentStreamingEvents *bool                      `json:"includeSubAgentStreamingEvents,omitempty"`
	EnvValueMode                   string                     `json:"envValueMode,omitempty"`
}
```

On the Copilot CLI side (1.0.81 `app.js`), the session-config builder defaults the flag to `false` and ties skills to it:

```js
enableConfigDiscovery: t.enableConfigDiscovery ?? !1,
// ...
enableSkills: t.enableSkills ?? t.enableConfigDiscovery,
```

Note the asymmetry that makes this easy to miss: the lower-level discovery helpers default to `?? !0` (**true**), so `copilot` used interactively or via `-p` discovers skills normally. Only the programmatic `session.create` path defaults to `false`. Any client that omits the field therefore gets skill discovery silently turned off.

## Evidence (A/B against the same binary, same cwd, same skills)

Two `--headless --stdio` sessions differing **only** by that one param, skill count parsed from Copilot's own `<available_skills>` prompt block:

| `session.create` params | skills injected |
|---|---|
| `{clientName, workingDirectory, model}` — what cc-connect sends today | **2** (both `builtin`) |
| same + `"enableConfigDiscovery": true` | **35** (33 `project` + 2 `builtin`) |

Reproduced twice in independent process pairs, attributed by matching each log's own `session.create` blob rather than by mtime.

## Workarounds that do NOT work

Worth recording, since several look plausible:

| Attempt | Result |
|---|---|
| `enableSkills: true` alone | 2 — upstream derives it from `enableConfigDiscovery`, which is still false |
| `trustWorkingDirectory: "session"` | 2 — not a trust/permission issue |
| `--add-dir ~/.agents` via extra args | 2 — that flag only affects `.github/skills` |
| `COPILOT_SKILLS_DIRS` env var | 2 — read inside the gated code path |
| `enableConfigDiscovery: true` | **35** ✅ |

`cmd`/extra-args (`core.ParseCmdOpts`, `agent/copilot/copilot.go:47`) can't fix this either — it's a JSON-RPC parameter, and `copilot --help` exposes no equivalent CLI flag.

## Suggested fix

Add the field and set it to `true` (or make it configurable, defaulting to true) in `session.create`:

```go
 	EnvValueMode                   string                     `json:"envValueMode,omitempty"`
+	EnableConfigDiscovery          *bool                      `json:"enableConfigDiscovery,omitempty"`
 }
```

Since cc-connect already documents skill support for other agents, enabling discovery by default seems right; a `skills = false` opt-out could be added if some users want the leaner prompt.

## Note on the ACP workaround

Switching the project to `type = "acp"` with `cmd = "copilot"` / `args = ["--acp", "--stdio"]` does load all 35 skills (verified: skill actually invoked, `Using skill: find-skills`), and per #1244's ACP-first policy that's a reasonable path. But it costs the `copilot` adapter's optional interfaces — `/model` (`ModelSwitcher`), `/compress` (`CompressCommand`), `GetSessionHistory`, `DeleteSession`, provider switching — since `agent/acp` doesn't implement them. So ACP isn't a full substitute for fixing the native adapter.

Happy to send a PR if the one-line direction looks right.

lprincewhn:
## Summary

With `type = "copilot"`, every environment-discovered skill is silently dropped — the agent sees only the 2 builtin skills, while `copilot skill list` in the same directory lists all of them. `copilotSessionConfig` had no field for `enableConfigDiscovery`, so `session.create` / `session.resume` never sent it, and Copilot CLI defaults it to `false` on that path while deriving `enableSkills` from it. This adds the field and sets it to `true`.

## Type of change

- [ ] Documentation only

## Root cause

Copilot CLI 1.0.81's session-config builder:

```js
enableConfigDiscovery: t.enableConfigDiscovery ?? !1,
enableSkills: t.enableSkills ?? t.enableConfigDiscovery,
```

The asymmetry that made this easy to miss: Copilot's lower-level discovery helpers default the same flag to `?? !0` (**true**), so interactive `copilot` and `copilot -p` load skills normally. Only the programmatic `session.create` path defaults it off, so any client omitting the field gets skill discovery silently disabled. Affects `.github/skills`, `.agents/skills`, `.claude/skills`, `~/.copilot/skills`, `~/.agents/skills`.

Since `sessionConfig()` backs both create and resume, one change covers resumed sessions too.

Alternatives that do **not** work, for the record: `enableSkills: true` alone (derived from this flag), `trustWorkingDirectory: "session"`, `--add-dir` (only covers `.github/skills`), `COPILOT_SKILLS_DIRS` (read inside the gated path). `cmd`/extra-args can't help either — it's a JSON-RPC param with no CLI-flag equivalent.

## Testing

Verified against Copilot CLI 1.0.81 — two `--headless --stdio` sessions differing **only** by this param, skill count parsed from Copilot's own `<available_skills>` prompt block in `~/.copilot/logs/process-*.log`:

| `session.create` params | skills injected |
|---|---|
| `{clientName, workingDirectory, model}` — pre-fix | **2** (both `builtin`) |
| same + `"enableConfigDiscovery": true` | **35** (33 `project` + 2 `builtin`) |

Reproduced twice in independent process pairs, attributed by matching each log's own `session.create` blob rather than by mtime.

- `TestSessionConfig_SendsEnableConfigDiscoveryForSkills` in `agent/copilot/session_test.go` — asserts `sessionConfig()` sets `EnableConfigDiscovery` to `true`, and that the marshaled `session.create` payload actually carries `"enableConfigDiscovery": true` (`omitempty` on a `*bool` only elides a nil pointer, so the wire shape is pinned explicitly).

- Regression test name: `TestSessionConfig_SendsEnableConfigDiscoveryForSkills`
  - [x] Reverted the fix locally; the regression test failed as expected.

Verified it fails for the right reason rather than on a compile error — with the struct field kept but the assignment removed (the true pre-fix behavior), it fails on the assertion:

```
--- FAIL: TestSessionConfig_SendsEnableConfigDiscoveryForSkills
    session_test.go:440: EnableConfigDiscovery = nil (field omitted from session.create);
    Copilot CLI then defaults it to false and disables all skill discovery
```

- [x] No CUJ touched (change is confined to `agent/copilot`; no files under `core/` are modified)

Ran it anyway for safety: `go test ./core/ -run TestCUJ` → `ok ... 4.917s`.

Skills placed in any Copilot-discovered directory now actually load in cc-connect sessions. In my setup the agent went from 2 skills to 35.

Note this fixes Copilot's *own runtime* skill loading. The `copilot` adapter separately doesn't implement cc-connect's `SkillProvider` (`SkillDirs()`), so the `/skills` command surface still won't list them — a distinct gap, not addressed here.

- [x] `go build ./...` passes — clean with `-tags no_web`. Untagged, it fails on `web/embed.go:13:12: pattern all:dist: no matching files found`, which is the gitignored generated frontend `dist/` missing from a fresh clone; reproduced identically on unmodified `origin/main`.
- [x] `go test ./...` passes — `./agent/copilot/` green. Two pre-existing failures unrelated to this change, both reproduced on unmodified `origin/main`:
  - `cmd/cc-connect` and `web`: the same missing `dist/` above.
  - Flaky tests in `core` under full-suite parallel load — a *different* test failed on each run (`TestCUJ_G3_PlatformReconnectReinitializesAndDelivers`, then `TestCUJ_G5_ToolFailureSurfacesToUser` + `TestCmdCronExec_TriggersJob`), mostly `TempDir RemoveAll cleanup: directory not empty`. All pass in isolation, and a clean `go test ./core/` on unmodified `origin/main` fails the same way. `core` does not import `agent/copilot` (`go list -deps ./core/` → no match), so this change cannot reach it. CI's `unit-test` job is green, consistent with a local-environment race.
- [x] No new hardcoded platform/agent names in `core/` — no `core/` files touched
- [x] i18n strings have all-language translations — no new user-facing strings
- [x] `gofmt` / `go vet` clean on both changed files (`gofmt -l` also flags `agent/copilot/copilot.go` and `copilot_test.go`, but those are pre-existing on `origin/main` and left untouched)

## Related

- Issue: Closes #1769
- Related PR: —
- Related policy discussion: #1244 (ACP-first). Switching to `type = "acp"` with `args = ["--acp", "--stdio"]` also loads all 35 skills and is a reasonable workaround, but it costs the native adapter's optional interfaces — `/model` (`ModelSwitcher`), `/compress` (`CompressCommand`), `GetSessionHistory`, `DeleteSession`, provider switching — which `agent/acp` doesn't implement. So ACP isn't a full substitute for fixing this.

Per your routing note in #1769: the fix follows the three points you listed (add the field mirroring `RequestPermission`/`Streaming`, set it in `cs.sessionConfig()`, regression test asserting the `session.create` params). I left the `skills = false` opt-out out of scope as you suggested. On your "ideally also" fourth item — a fixture-driven test parsing `<available_skills>` out of `~/.copilot/logs/process-*.log` — I did use exactly that as the verification method above, but didn't add it to the suite since it needs a real authenticated Copilot CLI process and reads the user's home directory. Happy to add it behind a build tag or as an opt-in integration test if you'd prefer it committed.

### DIFF
--- a/agent/copilot/session.go
+++ b/agent/copilot/session.go
@@ -73,6 +73,12 @@ type copilotSessionConfig struct {
 	Streaming                      *bool                      `json:"streaming,omitempty"`
 	IncludeSubAgentStreamingEvents *bool                      `json:"includeSubAgentStreamingEvents,omitempty"`
 	EnvValueMode                   string                     `json:"envValueMode,omitempty"`
+	// EnableConfigDiscovery must be sent explicitly: Copilot CLI's
+	// session.create/session.resume path defaults it to false, and derives
+	// enableSkills from it. Omitting it disables every environment-discovered
+	// skill source (.github/skills, .agents/skills, .claude/skills,
+	// ~/.copilot/skills, ~/.agents/skills), leaving only builtin skills.
+	EnableConfigDiscovery *bool `json:"enableConfigDiscovery,omitempty"`
 }
 
 type copilotPermissionResult struct {
@@ -214,6 +220,7 @@ func (cs *copilotSession) sessionConfig(sessionID string) copilotSessionConfig {
 	requestPermission := true
 	streaming := true
 	includeSubAgentStreaming := true
+	enableConfigDiscovery := true
 	return copilotSessionConfig{
 		SessionID:                      sessionID,
 		ClientName:                     "cc-connect",
@@ -224,6 +231,7 @@ func (cs *copilotSession) sessionConfig(sessionID string) copilotSessionConfig {
 		Streaming:                      &streaming,
 		IncludeSubAgentStreamingEvents: &includeSubAgentStreaming,
 		EnvValueMode:                   "direct",
+		EnableConfigDiscovery:          &enableConfigDiscovery,
 	}
 }
 
--- a/agent/copilot/session_test.go
+++ b/agent/copilot/session_test.go
@@ -419,6 +419,51 @@ func TestSessionConfig_MatchesCopilotCreateResumeShape(t *testing.T) {
 	}
 }
 
+// TestSessionConfig_SendsEnableConfigDiscoveryForSkills is a regression test
+// for skills under ~/.agents/skills (and every other discovered source) never
+// loading in cc-connect sessions.
+//
+// Copilot CLI's session.create/session.resume config builder defaults
+// enableConfigDiscovery to false and derives enableSkills from it
+// (`enableSkills: t.enableSkills ?? t.enableConfigDiscovery`). Note the
+// asymmetry that made this easy to miss: the lower-level skill discovery
+// helpers default the same flag to true, so interactive `copilot` and
+// `copilot -p` load skills fine — only the programmatic session.create path
+// defaults it off. Omitting the field therefore left cc-connect sessions with
+// builtin skills only, while `copilot skill list` in the same directory
+// listed all of them.
+func TestSessionConfig_SendsEnableConfigDiscoveryForSkills(t *testing.T) {
+	cs := &copilotSession{model: "gpt-5.2", workDir: "/work/project"}
+	cfg := cs.sessionConfig("sess-1")
+
+	if cfg.EnableConfigDiscovery == nil {
+		t.Fatal("EnableConfigDiscovery = nil (field omitted from session.create); " +
+			"Copilot CLI then defaults it to false and disables all skill discovery")
+	}
+	if !*cfg.EnableConfigDiscovery {
+		t.Fatalf("EnableConfigDiscovery = %v, want true", *cfg.EnableConfigDiscovery)
+	}
+
+	// The wire payload must actually carry the flag — `omitempty` on a *bool
+	// only elides a nil pointer, but assert it to pin the serialized shape
+	// Copilot CLI reads.
+	raw, err := json.Marshal(cfg)
+	if err != nil {
+		t.Fatalf("marshal session config: %v", err)
+	}
+	var wire map[string]any
+	if err := json.Unmarshal(raw, &wire); err != nil {
+		t.Fatalf("unmarshal session config: %v", err)
+	}
+	got, ok := wire["enableConfigDiscovery"]
+	if !ok {
+		t.Fatalf("session.create payload has no enableConfigDiscovery key: %s", raw)
+	}
+	if got != true {
+		t.Fatalf("enableConfigDiscovery = %v, want true, payload: %s", got, raw)
+	}
+}
+
 func TestRespondPermission_RPCUsesCopilotResultShape(t *testing.T) {
 	var buf bytes.Buffer
 	cs := &copilotSession{


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- This is a textbook bugfix: a tiny struct-field addition plus a default-true assignment, with a regression test that pins both the Go field and the JSON-RPC wire shape (`json.Marshal` + key assertion). CI is 5/5 green and all local tests pass with `-race`. Approving.

Review scope:
- Reviewed `agent/copilot/session.go` (+8/-0) and `agent/copilot/session_test.go` (+45/-0).
- Focused on correctness (no behavior change beyond sending the missing flag), wire-format stability, and test coverage of the `omitempty + *bool` edge case.

✅ What looks good:
- The struct field mirrors the existing `RequestPermission` / `Streaming` `*bool` pattern, so it lands cleanly in the wire payload with the same JSON tag style.
- The doc comment above the new field explains *why* it must be sent explicitly (Copilot CLI's asymmetric defaults — discovery helpers default `true`, but `session.create` defaults `false`). That asymmetry is exactly the kind of thing a future contributor might "simplify away", so the comment is load-bearing.
- The regression test is well-targeted:
  - Checks `cfg.EnableConfigDiscovery != nil` (catches the original bug class — field omitted entirely).
  - Checks `*cfg.EnableConfigDiscovery == true` (catches an accidental default-false regression).
  - `json.Marshal` + `map[string]any` unmarshal + key existence check pins the *wire* shape. This is the right call: `omitempty` on a `*bool` only elides a nil pointer, so a `true` should serialize, but a future contributor flipping the tag or adding a custom `MarshalJSON` could break that invariant silently. Pinning the wire shape prevents that.
  - The failure message in the test doubles as a diagnostic — it explicitly names the upstream default behavior, which is invaluable for whoever debugs a future regression.
- PR body is exceptional: A/B matrix with the same binary / cwd / skill set, exhaustive "workarounds that don't work" table (saves maintainer time when someone suggests the same things again), and a clear "Note on ACP" explaining why the ACP workaround isn't a substitute. `Closes #1769` is wired correctly.

🔵 Optional (nit, non-blocking):
- None material. `gofmt -l` reports nothing on the changed files; `go vet ./agent/copilot/` is clean; `go build` is clean; `-race -count=1 ./agent/copilot/` passes (1.044s).

Testing / Risk:
- Verified evidence:
  - GitHub Actions: lint, unit-test, smoke-test, regression-test, performance-test — all SUCCESS (run 33239304482).
  - Local: `gofmt -l` clean on changed files, `go vet ./agent/copilot/` clean, `go build ./agent/copilot/` clean, `CGO_ENABLED=1 go test -race -count=1 ./agent/copilot/` passes (1.044s), including the new `TestSessionConfig_SendsEnableConfigDiscoveryForSkills` and the adjacent `TestSessionConfig_MatchesCopilotCreateResumeShape`.
  - Auto-merge with `origin/main`: already up to date, no conflicts.
  - The pre-fix failure mode is documented in the PR body (field present but assignment removed → test fails on the assertion, not on a compile error), which is the right way to validate a regression test.
- Remaining risk:
  - Behavior change is bounded: enabling skill discovery is the documented intent. Users without skills see no observable change; users with skills now see them, which is the bugfix.
  - If a user later reports "too many skills inflating the prompt", an opt-out (`[[projects.agent.options]] skills = false`) can land in a follow-up. Not blocking for this PR.
  - `core` package does not import `agent/copilot` (`go list -deps ./core/` returns no match), so the pre-existing flaky tests under `go test ./...` parallel load cannot be reached by this change.

Next step:
- Merge. The fix is minimal, the regression test is tight, the documentation is exemplary, and the PR body answers every maintainer question in advance. Recommended for first-time contributor — would be a good reference PR for future external contributors.

Welcome to cc-connect, @lprincewhn — great first PR. 🎉

PR #1771, opened 2026-08-30
fix(weixin): realign quota defaults to short window and honor burst_limit=0

### LINKED ISSUE #1716: [Bug] weixin quota: default window (24h) caps the bot at 4 messages/day, and burst_limit=0 does not disable it

lyjx29:
## Summary

Fixes #1716. Aligns the weixin send-volume quota with the documented short-window burst guard and makes `burst_limit = 0` truly disable the quota. Both defects were confirmed real by the maintainer (v1.5.0 regression from #1643).

## Changes

1. **Default window → short window**: `defaultBurstWindowSecs` `86400` (24h) → `1800` (30m), matching the config comment ("short-window burst guard"). The quota now paces separate messages within a short window instead of capping the bot at 4 messages per day.
2. **`burst_limit = 0` now disables the quota**: the init code `if burstLimit <= 0 { burstLimit = defaultBurstLimit }` treated an explicit `0` as "unset" and fell back to the default. Now only an explicitly provided `burst_limit` / `burst_window_secs` overrides the default (including `0`); unset values use the defaults.
3. **Tests**: `TestNew_BurstLimitZeroDisablesQuota`, `TestNew_DefaultQuota`, `TestNew_ExplicitBurstLimit`, `TestNew_DefaultQuotaEnforced`. The 0-disable test fails on the old logic (limit falls back to 4) and passes after the fix.
4. **Docs**: updated `config.example.toml` comments (default now 1800s, short-window semantics) and noted the behavior change in `CHANGELOG.md` — configs relying on the previous 24h default must set `burst_window_secs` explicitly.

## Behavior change

Configs that relied on the default 24h window (4 messages/day) will now see a 30-minute window instead. Set `burst_window_secs` explicitly to keep the old behavior.

## Verification

- `go test ./platform/weixin/ -count=1` passes (all existing + new tests)
- `go vet ./platform/weixin/` clean
- `gofmt` clean


chenhg5 (changes requested):
## CI Gate — review deferred

@lyjx29 thanks for the PR! Before I can do a full diff/QA review, I need CI to be green on this branch.

### What's missing
- **No checks reported on `fix-weixin-quota-defaults`.** Fork PRs don't auto-trigger CI in this repo unless a maintainer pushes to the branch (which I can see hasn't happened — last activity 2026-08-30).
- The PR body doesn't mention local test results, and the diff is non-trivial (6 files, +377/−133 in `platform/weixin/...`). I want CI as a baseline before I trust manual results.

### Suggested fix (either is fine)
1. **Easiest:** allow maintainer edits and ask me (or @chenhg5) to push a no-op to trigger CI, OR
2. Push any new commit (even a rebase onto current `main`) — that retriggers the `pull_request` workflow.

Once CI is green, I'll do the full QA review: correctness of the new quota math, the `burst_limit=0` semantics, behavior change vs. previous defaults, and test coverage of edge cases (e.g. concurrent sends, config-missing fallback, burst=0 disables throttling entirely vs. keeps only rate-limit).

### Quick sanity check while you wait
Could you confirm one thing in the PR description: when `burst_limit=0`, is the intent "no burst cap at all" (drop the burst check) or "treat 0 as default and apply the configured `burst_limit`"? The current code path matters here — affects whether heavy senders can hit the rate-limiter immediately vs. being throttled even on first send.

PR #1777, opened 2026-09-01
fix(core): never write temp files in target directory (fixes t.TempDir cleanup)

TongLing916:
## Problem

`AtomicWriteFile` 在目标目录中创建临时文件（如 `.tmp-...`)。当 session manager 的 `saveLocked()` 通过后台 goroutine（`ReceiveMessage` → `processInteractiveMessageWith` 触发）在 tests 中使用 `t.TempDir()` 时被调用，这些孤立的临时文件阻止了 Go test framework 清理 temp 目录：

```
--- FAIL: TestCUJ_A3_ImageReachesAgent (0.02s)
    testing.go:1617: TempDir RemoveAll cleanup: unlinkat .../001: directory not empty
```

## Solution

重写 `AtomicWriteFile`：**始终**将初始临时文件写入系统临时目录 (`os.TempDir()`)，而非目标目录。确保目标目录中永远不会创建临时文件。

- **同文件系统**（macOS 常见）：atomic rename 成功，完全原子写入
- **跨设备**（EXDEV，如 NFS/mount points）：fallback 到 `os.WriteFile` 直接写入 — 非原子但无孤立文件

## Changes

- `core/atomicwrite.go` — rewrite: always start with system temp dir, fallback to direct write

### DIFF
--- a/core/atomicwrite.go
+++ b/core/atomicwrite.go
@@ -5,12 +5,26 @@ import (
 	"path/filepath"
 )
 
-// AtomicWriteFile writes data to a file atomically by first writing to a
-// temporary file in the same directory, syncing, then renaming over the target.
-// This prevents data loss / corruption on crash.
+// AtomicWriteFile writes data to a file atomically when possible by first
+// writing to a temporary file in the system temp directory, syncing, then
+// renaming over the target.  On the same filesystem the rename is fully
+// atomic.
+//
+// When os.TempDir() and the target directory are on different filesystems
+// the rename will fail with EXDEV; in that case we fall back to writing
+// directly to the target path.  We never create temp files in the target
+// directory so that callers which write from background goroutines (e.g.
+// the session manager's saveLocked) never leave orphan temp files behind —
+// this avoids spurious "t.TempDir() cleanup: directory not empty" failures
+// during parallel tests.
 func AtomicWriteFile(path string, data []byte, perm os.FileMode) error {
-	dir := filepath.Dir(path)
-	tmp, err := os.CreateTemp(dir, ".tmp-*")
+	dstDir := filepath.Dir(path)
+	if err := os.MkdirAll(dstDir, 0o755); err != nil {
+		return err
+	}
+
+	// Write to system temp dir for atomic rename.
+	tmp, err := os.CreateTemp("", "cc-atomic-*")
 	if err != nil {
 		return err
 	}
@@ -34,17 +48,14 @@ func AtomicWriteFile(path string, data []byte, perm os.FileMode) error {
 		os.Remove(tmpPath)
 		return err
 	}
-	if err := os.Rename(tmpPath, path); err != nil {
-		// Rename can fail when the destination is a directory, the
-		// destination's filesystem differs from the temp dir's (rare given
-		// CreateTemp uses the same dir, but possible with bind mounts), or
-		// the destination is locked by another process on Windows. In any
-		// of those cases the temp file is now an orphaned `.tmp-*` we
-		// created — clean it up so repeated failures don't litter the
-		// directory and confuse later directory scans (e.g. cron / session
-		// stores that walk their parent dir).
-		os.Remove(tmpPath)
-		return err
+
+	// Try atomic rename (works when temp and target are on the same fs).
+	if err := os.Rename(tmpPath, path); err == nil {
+		return nil
 	}
-	return nil
+
+	// Rename failed (likely EXDEV on different mount points).
+	// Fall back to direct write — safe, no orphan temp files.
+	_ = os.Remove(tmpPath)
+	return os.WriteFile(path, data, perm)
 }


chenhg5 (approved):
## Conclusion: Approve

## Overall assessment
PR #1777 is mergeable. CI is 5/5 green on the current head (`5e92fe9`), `mergeable: MERGEABLE`, and the substantive change is a small, surgical bugfix: `core.AtomicWriteFile` no longer creates temp files in the target directory, eliminating orphan `.tmp-*` files that pollute `t.TempDir()` cleanup in parallel tests. Same-filesystem rename stays atomic; cross-filesystem (EXDEV) now falls back to a direct `os.WriteFile` instead of leaving an orphan. No P0/P1 blockers.

## Review scope
- Diff (1 file, +28/-17): `core/atomicwrite.go`.
- Walked through the new `AtomicWriteFile`: pre-create target dir → create temp in `os.TempDir()` → write/sync → rename → fallback on EXDEV → direct write.
- Cross-checked: PR body describes the same change; only one file touched; CI runs `unit-test`, `smoke-test`, `regression-test`, `performance-test`, `lint` all green.

## ✅ What looks good
- **Root cause is real and the fix is direct**: orphan `.tmp-*` files created by `os.CreateTemp(dir, ".tmp-*")` (in the target dir) survive when the calling goroutine is killed mid-write or when `t.TempDir()` cleanup races the writer. Writing to `os.TempDir()` instead — which is managed by the OS and survives across processes — eliminates the orphan path completely.
- **Same-filesystem case stays fully atomic**: when `os.TempDir()` and the target are on the same filesystem (the common case on macOS / most Linux boxes), `os.Rename` is atomic. The rename-then-return path (`atomicwrite.go:54`) preserves the original safety guarantee — no behavior regression for the common case.
- **EXDEV fallback is well-justified**: when the rename fails (most likely EXDEV across mount points), the code removes the temp file and falls back to `os.WriteFile`. The fallback is non-atomic but leaves no orphans — the right tradeoff for a pathological mount setup. The comment block at the top of the function documents both paths.
- **Brand-prefixed temp pattern (`cc-atomic-*`)**: the system-temp prefix is namespace-scoped to cc-connect, so concurrent writes from other tools on the same host can't collide. Good hygiene.
- **Single-file, surgical change**: 1 file / +28/-17, no API change, no callers updated. The `AtomicWriteFile(path, data, perm)` signature is preserved.
- **CI is green on all 5 jobs** including unit, smoke, regression, and performance — no regression risk on existing surface area.
- **Comment quality is high**: documents the why (orphan temp files in target dir → `t.TempDir()` cleanup fails), the same-fs guarantee, and the EXDEV fallback rationale.

- **`os.MkdirAll(dstDir, 0o755)` is a semantic change** (`core/atomicwrite.go:11`): the old version did *not* create the target directory — it relied on callers to ensure it existed. The new version silently creates it with mode `0o755`. If any caller currently passes a path with a missing parent dir and *expects* `AtomicWriteFile` to fail (because that means the caller forgot to bootstrap the dir), this now masks the bug behind a silent mkdir. Worth checking call sites: the PR body mentions `saveLocked()` in the session manager, which already ensures its dir exists before calling. But `core/atomicwrite.go` is in `core/` and may have other callers outside the session manager path. Low-risk in practice (most callers already create the dir), but the behavior change deserves a one-line note in the PR description.
- **EXDEV detection is missing** (`core/atomicwrite.go:58`): when `os.Rename` fails, the code blindly tries `os.WriteFile` as the fallback. If rename failed for a reason *other* than EXDEV — e.g. `destination is a directory`, `permission denied`, `not supported by filesystem`, `cross-device link` on a system where the kernel returns a different errno — `os.WriteFile` will likely also fail, and the error returned will be the *WriteFile* error, masking the real *Rename* error. Cheap fix: check `errors.Is(err, syscall.EXDEV)` (and `errors.Is(err, fs.ErrCrossDevice)` on Go 1.20+); if it's EXDEV, do the fallback; otherwise return the original rename error and clean up the temp file. This preserves the diagnostic signal when something genuinely goes wrong.
- **`os.MkdirAll(dstDir, 0o755)` mode is hard-coded**: the `perm` parameter only applies to the final file, not the directory. If a caller passes `perm=0o600` (private file) but the target dir was just auto-created with `0o755`, the world-readable dir is a wider permission than the caller asked for. Probably fine — most callers want `0o755` for the parent dir — but worth a one-line note that the dir mode is fixed regardless of `perm`.

## 🔵 Optional
- **No `slog.Debug` when EXDEV fallback fires**: operators on NFS / bind-mounted setups would benefit from a one-line log saying "atomic rename failed (EXDEV?); falling back to non-atomic direct write" so they can fix their mount configuration. Probably noisy in tests, so should be guarded by an env-var or feature flag. Not worth adding now.
- **`cc-atomic-*` prefix is consistent with `cc-` brand** but doesn't match the module name `cc-connect`. If a future cleanup decides on `cc-connect-atomic-*`, this is the only call site to update.

## ❓ Questions
- **Callers that pass a path with a missing parent dir**: are you sure none of them rely on the previous "fail if dir doesn't exist" behavior? A 5-minute grep of `AtomicWriteFile(` across the repo would confirm; the PR body only mentions `saveLocked()`.
- **EXDEV fallback on Windows**: on Windows, `os.Rename` fails with `ERROR_ACCESS_DENIED` or `ERROR_SHARING_VIOLATION` when the destination is locked, not with EXDEV. The fallback will fire and `os.WriteFile` will also fail with the same error. Acceptable, but worth confirming the failure mode in the Windows regression-test path is "returns error" rather than "silently corrupts state". (The diff doesn't touch Windows-specific code, so this is more of a sanity check than a regression.)

## Testing / Risk
- Verified on current head `5e92fe9`: CI run 33479670785 — `lint` 2m17s ✓, `unit-test` 4m49s ✓, `smoke-test` 35s ✓, `regression-test` 41s ✓, `performance-test` 55s ✓.
- Verified locally (this review): the diff is confined to `core/atomicwrite.go`. No interface changes; the `AtomicWriteFile(path, data, perm) error` signature is unchanged.
- Unverified risk: callers that pass a path with a missing parent dir now silently succeed (via `os.MkdirAll`) instead of failing. Most callers already pre-create the dir; the failure mode would be a quiet mask if a caller has a latent bug.
- **Race / concurrency**: the rename-then-write fallback has a non-atomic window where readers see the old content (or partial new content). This is documented in the comment as the explicit tradeoff. Acceptable.
- **Orphan temp files in `os.TempDir()`**: the brand-prefixed `cc-atomic-*` temp files are cleaned up in the success path (`os.Remove(tmpPath)` on rename failure) and on the EXDEV fallback. The only orphan scenario is process kill mid-write — but `os.TempDir()` is periodically cleaned by the OS, so orphans there are bounded. Far better than orphan files in the target dir.

## Next step
Merge when convenient. No further code changes required from the author. The two P2 items (semantic change of `MkdirAll`, EXDEV error masking) are worth addressing in a follow-up PR if any maintainer wants tighter correctness — but they are not blockers for this bugfix, which solves a real test-cleanup problem with no behavior change in the common path.

PR #1778, opened 2026-09-01
feat(feishu): AskUserQuestion card 2.0 with option buttons + free-form input

Niko-Wu:
## Problem

The Feishu AskUserQuestion card only renders option buttons. When none of the options fit, the user has no way to type a custom answer — the CLI UI has an "Other" free-text input, but Feishu had no equivalent. The card note ("reply with the option number or type your answer") suggests typing is possible, and the engine's `resolveAskQuestionAnswer` does pass free text through, but nothing on the card makes that discoverable.

## Solution

Render the ask-question prompt as a Feishu **card schema 2.0** with both interaction modes in one card:

- **Option buttons** — one `behaviors`-callback button per option. The callback value reuses the legacy `"action": "askq:<qIdx>:<optIdx>"` contract, so dispatch, confirmation card, and session tracking all work unchanged.
- **Free-form input** — an `input` field (max 500 chars) inside a `form`, with a `form_submit` "Send" button. `onCardAction` routes the submitted `answer` field as `askq_text:` and forwards the raw text as a user message; the engine's existing free-form answer path resolves it (non-numeric input returned as-is).
- **Confirmation cards** for actions on a schema 2.0 card must themselves be schema 2.0 — a v1 response card is rejected with Feishu error `200830`. Added a `simpleCard2` helper for that.

## Compatibility

- `MultiSelect` prompts keep the numbered-list card (buttons would resolve on first click).
- The rich card lives behind the new optional `core.AskQuestionCardSender` interface (same pattern as `InlineButtonSender`); platforms without it keep the generic card fallback, and core stays platform-agnostic.
- Free-form hint text is localized for all 5 supported languages (en, zh-CN, zh-TW, ja, es).
- `sendAskQuestionPrompt` falls back to the generic path if the platform implementation errors.

## Testing

- [x] `go test ./...` — new tests in `platform/feishu/askq_card_v2_test.go`: card structure (schema/buttons/behaviors value/form fields), form-submit dispatch, blank-answer ignored
- [x] End-to-end on a live Feishu bot: card renders, button clicks return the option label, typed input ("hello，123" / a Chinese sentence) is forwarded verbatim to the agent, and the agent re-asks when a submitted answer is meaningless

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -11799,6 +11799,30 @@ func (e *Engine) sendAskQuestionPrompt(p Platform, replyCtx any, questions []Use
 		titleSuffix = fmt.Sprintf(" (%d/%d)", qIdx+1, total)
 	}
 
+	// Try native rich ask-question card (Feishu schema 2.0: option buttons +
+	// free-form input in one card). Single-select only — for multiSelect,
+	// buttons would resolve on the first click and prevent selecting multiple.
+	if !q.MultiSelect {
+		if aqs, ok := p.(AskQuestionCardSender); ok {
+			body := "**" + q.Question + "**\n\n"
+			for i, opt := range q.Options {
+				body += fmt.Sprintf("%d. **%s**", i+1, opt.Label)
+				if opt.Description != "" {
+					body += " — " + opt.Description
+				}
+				body += "\n"
+			}
+			err := aqs.SendAskQuestionCard(context.Background(), replyCtx,
+				e.i18n.T(MsgAskQuestionTitle)+titleSuffix, body,
+				e.i18n.T(MsgAskQuestionNoteFreeform), q, qIdx)
+			if err == nil {
+				return
+			}
+			slog.Debug("sendAskQuestionPrompt: native card unavailable, falling back",
+				"platform", p.Name(), "error", err)
+		}
+	}
+
 	// Try card (Feishu/Lark)
 	if supportsCards(p) {
 		cb := NewCard().Title(e.i18n.T(MsgAskQuestionTitle)+titleSuffix, "blue")
--- a/core/i18n.go
+++ b/core/i18n.go
@@ -413,31 +413,31 @@ const (
 	MsgCronIDLabel               MsgKey = "cron_id_label"
 	MsgCronFailedSuffix          MsgKey = "cron_failed_suffix"
 
-	MsgTimerNotAvailable  MsgKey = "timer_not_available"
-	MsgTimerUsage         MsgKey = "timer_usage"
-	MsgTimerAddUsage      MsgKey = "timer_add_usage"
-	MsgTimerAdded         MsgKey = "timer_added"
-	MsgTimerAddedExec     MsgKey = "timer_added_exec"
-	MsgTimerAddExecUsage  MsgKey = "timer_addexec_usage"
-	MsgTimerEmpty         MsgKey = "timer_empty"
-	MsgTimerListTitle     MsgKey = "timer_list_title"
-	MsgTimerListFooter    MsgKey = "timer_list_footer"
-	MsgTimerDelUsage      MsgKey = "timer_del_usage"
-	MsgTimerMuteUsage     MsgKey = "timer_mute_usage"
-	MsgTimerDeleted       MsgKey = "timer_deleted"
-	MsgTimerNotFound      MsgKey = "timer_not_found"
-	MsgTimerMuted         MsgKey = "timer_muted"
-	MsgTimerUnmuted       MsgKey = "timer_unmuted"
-	MsgTimerCardHint      MsgKey = "timer_card_hint"
-	MsgTimerBtnMute       MsgKey = "timer_btn_mute"
-	MsgTimerBtnUnmute     MsgKey = "timer_btn_unmute"
-	MsgTimerBtnDelete     MsgKey = "timer_btn_delete"
-	MsgTimerIDLabel       MsgKey = "timer_id_label"
-	MsgTimerScheduledLabel MsgKey = "timer_scheduled_label"
-	MsgTimerFailedSuffix  MsgKey = "timer_failed_suffix"
-	MsgCommandsTagAgent          MsgKey = "commands_tag_agent"
-	MsgCommandsTagShell          MsgKey = "commands_tag_shell"
-	MsgUpgradeTimeoutSuffix      MsgKey = "upgrade_timeout_suffix"
+	MsgTimerNotAvailable    MsgKey = "timer_not_available"
+	MsgTimerUsage           MsgKey = "timer_usage"
+	MsgTimerAddUsage        MsgKey = "timer_add_usage"
+	MsgTimerAdded           MsgKey = "timer_added"
+	MsgTimerAddedExec       MsgKey = "timer_added_exec"
+	MsgTimerAddExecUsage    MsgKey = "timer_addexec_usage"
+	MsgTimerEmpty           MsgKey = "timer_empty"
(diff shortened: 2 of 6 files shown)


chenhg5 (approved):
## Conclusion: Approve

## Overall assessment
PR #1778 is mergeable. CI is 5/5 green on the current head (`7c88b98`), `mergeable: MERGEABLE`, and the substantive change is a clean, optional feature: a Feishu card schema 2.0 rendering of the AskUserQuestion prompt that combines option buttons with a free-form text input in a single card, gated behind a new `core.AskQuestionCardSender` interface. Platforms that don't implement it keep the existing numbered-list card path unchanged. No P0/P1 blockers.

## Review scope
- Diff (6 files, +434/-49): `core/engine.go`, `core/i18n.go`, `core/interfaces.go`, `platform/feishu/card.go`, `platform/feishu/feishu.go`, `platform/feishu/askq_card_v2_test.go`.
- Walked through `sendAskQuestionPrompt` extension, `AskQuestionCardSender` interface contract, `buildAskQuestionCard2` schema construction, the new `submit`/`askq_text:` branches in `onCardAction`, the `simpleCard2` helper, and the three new test cases.
- Confirmed `dispatchCoreMessage` and the `strings`/`fmt` imports already exist on the head.

## ✅ What looks good
- **Interface segregation matches existing convention**: `AskQuestionCardSender` follows the same optional-interface pattern as `InlineButtonSender` (per the PR body). `core/engine.go:11802` does `p.(AskQuestionCardSender)` and falls through to the generic `supportsCards(p)` path on miss or on error — no regression risk for platforms that don't opt in.
- **MultiSelect is intentionally excluded with a documented why**: `core/engine.go:11805` gates the new card on `!q.MultiSelect`. The comment explains "buttons would resolve on the first click and prevent selecting multiple". MultiSelect keeps the existing numbered-list card, so no behavioral change for multi-select users.
- **Backward-compatible callback contract**: option buttons reuse `"askq:<qIdx>:<optIdx>"` (`platform/feishu/card.go:415`), so the existing `askq:` branch in `onCardAction` works unchanged for numbered selections — no risk of breaking the path that's already in production.
- **v2 confirmation card fix is well-justified**: `platform/feishu/card.go:476` introduces `simpleCard2` because "Callback responses for actions triggered on a schema 2.0 card must themselves be schema 2.0 — a v1 response card is rejected by Feishu with error 200830." This was a real latent bug once the schema-2.0 trigger was introduced; the PR fixes it as part of the same change.
- **Blank-answer guard is pinned by a test**: `TestOnCardActionFormSubmitWithoutAnswerIsIgnored` (`platform/feishu/askq_card_v2_test.go:319`) confirms a whitespace-only submission does not dispatch and returns `nil` — no risk of an empty-text message polluting the session.
- **i18n coverage matches existing convention**: `MsgAskQuestionNoteFreeform` is translated into all 5 supported languages (en, zh-CN, zh-TW, ja, es) in `core/i18n.go:2791`. Same pattern as `MsgAskQuestionNote` and `MsgAskQuestionNoteMulti`.
- **Free-form answer routing is correct**: the `askq_text:` branch builds a `core.Message` with `SessionKey`, `ReplyCtx`, `UserID`, `UserName`, `ChatName` set — equivalent to a normal user message — and dispatches it via the same `dispatchCoreMessage` path as other inbound messages. The engine's `resolveAskQuestionAnswer` returns non-numeric text as-is, so a typed Chinese sentence like "先出码后置完成" lands in the agent verbatim. Matches the PR body's "end-to-end on a live Feishu bot" verification.
- **Comment quality is high**: every new branch (`form_submit`→`askq_text:`, `askq_text:` confirmation card, `MultiSelect` exclusion, `simpleCard2` rationale) explains the *why*, not just the *what*.

- **`title[:50] + "…"` truncates bytes, not runes** (`platform/feishu/feishu.go:962`): a 50-character Chinese answer will be cut at roughly 16–17 characters, even though the cut is meant to be cosmetic. Cheap fix: `utf8.RuneCountInString(title)` + `[]rune(title)[:n]` + `string(...)`. Not user-visible for English answers; will be visible for Chinese / Japanese / Korean free-form answers, which is the *exact* user base this PR is optimizing for.
- **`go p.dispatchCoreMessage(...)` with no error sink on the askq_text branch**: the goroutine in `platform/feishu/feishu.go:945` mirrors the pattern used in 5 other places in the same file, so this is a pre-existing convention, not a blocker — but the PR adds one more silent failure point. If `dispatchCoreMessage` errors, the user sees their answer "disappear" after pressing Send. Worth at minimum a `defer recover()` or a wrapping log; future cleanup opportunity.
- **`SendAskQuestionCard` validation order**: `platform/feishu/card.go:373` returns errors for invalid `replyCtx` or empty `chatID` before building the card JSON. Safe and clear; no action needed, just noting the failure path is exercised.

## 🔵 Optional
- **`itoa` helper in the test file** (`platform/feishu/askq_card_v2_test.go:257`): a custom `itoa` that only handles 1–4 and returns `"?"` otherwise. `strconv.Itoa(buttons)` does the right thing in one line and removes the bespoke helper. Pure cleanup.
- **`simpleCard2` is broadly applicable**: it can replace other `core.NewCard().Build().Title(...).Markdown(...)` constructions throughout `platform/feishu/`. Future cleanup, not for this PR — but the helper exists, and it's a good consolidation opportunity.
- **Build-time `header.template` is hardcoded `"blue"`** in both `buildAskQuestionCard2` and `simpleCard2`. The existing `MsgAskQuestionTitle` uses "blue" too, so consistency wins. If a maintainer ever wants per-question color (e.g. "red" for destructive questions), this is a one-line extension.

## ❓ Questions
- **`q.Options` empty case**: `buildAskQuestionCard2` (`platform/feishu/card.go:396`) iterates `q.Options`; if `len(q.Options) == 0`, the card renders just the form + hint with no option buttons. The engine never produces an empty-options question today, but should the engine guard against it (`if len(q.Options) == 0 { skip native card }`) so a future caller can rely on the fallback path? Not a blocker — just a defensive check.
- **Could `label_position: "top"` on the input field be omitted?** It's a v2 SDK field that the Feishu client renders correctly. Some clients (especially older Lark desktop) ignore it and default to `left`. Not a bug — just confirming the team has seen it render correctly in the live bot screenshot.

## Testing / Risk
- Verified on current head `7c88b98`: CI run 33488059064 — `lint` 2m20s ✓, `unit-test` 4m57s ✓, `smoke-test` 39s ✓, `regression-test` 28s ✓, `performance-test` 56s ✓.
- Verified locally (this review): the diff is confined to `core/engine.go` (extension point), `core/interfaces.go` (interface), `core/i18n.go` (string + alignment), `platform/feishu/{card.go,feishu.go}` (impl), `platform/feishu/askq_card_v2_test.go` (tests). No other platform's interface changes. The new interface is opt-in; non-Feishu platforms are unaffected.
- Unverified risk: operators on a non-Feishu platform don't see any change (the new interface isn't implemented there). Operators on Feishu see a richer ask-question card only when the question is single-select with at least one option. Multi-select and empty-options questions fall back to existing behavior. Worst-case impact is a layout change in a single chat per question — and the engine logs the fallback at debug level if the platform returns an error.
- **Race / concurrency**: `onCardAction` is called per Feishu webhook and runs in its own goroutine context; `dispatchCoreMessage` uses the existing pattern. No shared-state mutation introduced by this PR. Safe.

## Next step
Merge when convenient. No further code changes required from the author. The byte-vs-rune truncation in the `askq_text:` confirmation card title (`platform/feishu/feishu.go:962`) is a small P2 worth fixing before or shortly after merge; the `itoa` test helper and the broader `simpleCard2` consolidation are P3 polish.

PR #1774, opened 2026-08-31
fix(codex): allow listing sessions across workdirs

shunxian-zhou:
## Summary

Fixes Codex session listing when workDir filtering is intentionally disabled. Previously `filepath.Abs("")` turned an empty workDir into the cc-connect process cwd, so a list-all mode still filtered out Codex sessions from other cwd values.

This PR also adds a Codex agent option, `list_all_workdirs`, and documents it in `config.example.toml`.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing behavior to change)
- [ ] Documentation only
- [ ] Internal refactor / chore (no user-visible change)

## Testing

Not run locally. I only ran `gofmt` and `git diff --check` before opening the PR.

### Automated tests added in this PR

- `TestAgentListSessions_ListAllWorkdirs` in `agent/codex/list_test.go` — verifies the Codex agent can list sessions from multiple cwd values when `list_all_workdirs` is enabled.
- `TestListCodexSessions_EmptyWorkDirDoesNotFilterByProcessDir` in `agent/codex/list_test.go` — verifies an empty workDir means no cwd filter instead of the current process directory.
- `TestWorkspaceAgentOptions_PreservesListAllWorkdirs` in `agent/codex/codex_model_test.go` — verifies workspace agent clones preserve the option.

### For bug fixes only — regression test

- Regression test name: `TestListCodexSessions_EmptyWorkDirDoesNotFilterByProcessDir`
- Manual verification this test catches the regression:
  - [ ] Reverted the fix locally; the regression test failed as expected.

### Critical User Journeys (CUJ) impact

- [ ] No CUJ touched (small refactor, doc change, etc.)
- [x] B — session lifecycle (`/new` `/switch` `/list` `/history` etc.)
- [ ] C — agent execution control (`/mode` `/cancel` `/stop` permissions)
- [ ] D — security & permissions (`allow_from` `admin_from` `banned_words` rate limits)
- [ ] E — scheduled tasks (`/cron` `/timer`)
- [ ] F — config switching (`/lang` `/provider` `/model` reload)
- [ ] G — error handling & robustness (LLM failure, ws reconnect, agent crash)
- [ ] H — multi-platform / multi-project isolation
- [ ] I — UI rendering correctness (cards, streaming, display modes)

If any CUJ group is touched, confirm:

- [ ] `go test ./core/ -run TestCUJ` passes locally.
[...]

PR #1779, opened 2026-09-01
fix(feishu): enforce allow_from on card actions

mclamee:
## Summary

- Apply the configured allow_from check to every Feishu card callback before navigation, command, permission, or question handlers run.
- Keep empty and wildcard behavior aligned with normal inbound message routing.
- Add a regression test proving an unauthorized card actor cannot invoke message or navigation handlers.

## Why

Normal messages and bot-menu events enforce allow_from, but card actions only checked allow_chat. A user outside the configured allowlist could therefore execute actions on a card visible in an allowed group.

## Verification

- go test ./platform/feishu -run TestInteractivePlatform_CardActionRejectsUnauthorizedUser -count=1
- go test ./platform/feishu -count=1

### DIFF
--- a/platform/feishu/feishu.go
+++ b/platform/feishu/feishu.go
@@ -161,7 +161,7 @@ type Platform struct {
 	chatNameCache    sync.Map          // chat_id -> chat name
 	chatMemberCache  sync.Map          // chatID -> *chatMemberEntry
 	recalledMu       sync.Mutex
-	recalledMsgIDs   map[string]time.Time // message_id -> recall time, short TTL race guard
+	recalledMsgIDs         map[string]time.Time // message_id -> recall time, short TTL race guard
 	// Webhook mode fields (for Lark international version)
 	server       *http.Server
 	port         string
@@ -778,6 +778,10 @@ func (p *Platform) onCardAction(event *callback.CardActionTriggerEvent) (*callba
 	userID := ""
 	if event.Event.Operator != nil {
 		userID = event.Event.Operator.OpenID
+	}
+	if !core.AllowList(p.allowFrom, userID) {
+		slog.Debug(p.tag()+": card action from unauthorized user", "user", userID)
+		return nil, nil
 	}
 	chatID := ""
 	messageID := ""
--- a/platform/feishu/platform_test.go
+++ b/platform/feishu/platform_test.go
@@ -314,6 +314,47 @@ func TestInteractivePlatform_CardActionPassesCardSenderToHandler(t *testing.T) {
 	}
 }
 
+func TestInteractivePlatform_CardActionRejectsUnauthorizedUser(t *testing.T) {
+	platformAny, err := New(map[string]any{
+		"app_id":             "cli_xxx",
+		"app_secret":         "secret",
+		"enable_feishu_card": true,
+		"allow_from":         "ou_allowed",
+	})
+	if err != nil {
+		t.Fatalf("New() error = %v", err)
+	}
+	ip := platformAny.(*interactivePlatform)
+
+	invoked := make(chan struct{}, 1)
+	ip.handler = func(_ core.Platform, _ *core.Message) {
+		invoked <- struct{}{}
+	}
+	ip.cardNavHandler = func(_ string, _ string) *core.Card {
+		invoked <- struct{}{}
+		return core.NewCard().Markdown("unexpected").Build()
+	}
+
+	resp, err := ip.onCardAction(&callback.CardActionTriggerEvent{
+		Event: &callback.CardActionTriggerRequest{
+			Operator: &callback.Operator{OpenID: "ou_denied"},
+			Action:   &callback.CallBackAction{Value: map[string]any{"action": "cmd:/help"}},
+			Context:  &callback.Context{OpenChatID: "oc_test_chat", OpenMessageID: "om_test_message"},
+		},
+	})
+	if err != nil {
+		t.Fatalf("onCardAction() error = %v", err)
+	}
+	if resp != nil {
+		t.Fatalf("unauthorized card action response = %#v, want nil", resp)
+	}
+	select {
+	case <-invoked:
+		t.Fatal("unauthorized card action reached a handler")
+	case <-time.After(100 * time.Millisecond):
+	}
+}
+
 func TestInteractivePlatform_CardActionActWithoutCardResponseDoesNotWarn(t *testing.T) {
 	platformAny, err := New(map[string]any{"app_id": "cli_xxx", "app_secret": "secret", "enable_feishu_card": true})
 	if err != nil {


mclamee:
CI note: the changed package passed (`platform/feishu`, 37.873s). The aggregate unit job failed in unrelated `core` test `TestCUJ_H2_TwoPlatformsConcurrentNoBleed` after timing out with userA=5/userB=2; the run log shows no Feishu-package failure. I cannot rerun the upstream repository job from the fork.

chenhg5 (changes requested):
## CI Gate — review deferred

Thanks for the small, focused change (`platform/feishu/feishu.go` + `platform_test.go`, +47/-2) and for already self-investigating the failure in your follow-up comment.

### CI status (blocking)

Head `a916b37` on run `33562493524`:

| Job | Status |
|-----|--------|
| `lint` | ✅ pass (2m27s) |
| `unit-test` | ❌ **fail** (5m7s, exit 11) — `Run tests with coverage` step |
| `smoke-test` | ⚪ skipped (downstream of unit-test) |
| `regression-test` | ⚪ skipped (downstream of unit-test) |
| `performance-test` | ⚪ skipped (downstream of unit-test) |

Per the QA CI gate, I'm holding a full diff review until the suite is green. Your self-investigation note correctly identifies that the failing assertion is `TestCUJ_H2_TwoPlatformsConcurrentNoBleed` (in `core`, unrelated to this PR's `platform/feishu` package which passed in 37.873s). However, GitHub won't let you rerun a workflow run from a fork.

### What unblocks the review

Pick one of:

1. **Maintainer pushes the head branch** — your fork has `maintainerCanModify=true`, so a maintainer commit/push to `dora-agi:fix/feishu-card-action-allow-from` retriggers `pull_request` CI on `chenhg5/cc-connect`. Easiest path; same fix the team used for #1771 / #1774.
2. **Open a trivial no-op commit** — e.g. a doc tweak in `platform/feishu/README.md`. GitHub fires CI on the new head SHA. Same effect.
3. **Local evidence of the flake** — if you have a local `go test ./...` run from current `main` showing the same `TestCUJ_H2_…` failure, paste the output. I'll treat it as a known-flaky failure and move on to diff review.

Once CI is green I'll do the full review (likely a quick approve — the change is well-scoped and your added `TestInteractivePlatform_CardActionRejectsUnauthorizedUser` test covers the new branch).

### Diff scope (FYI, not a review yet)

- `platform/feishu/feishu.go`: gates `onCardAction` on `core.AllowList(p.allowFrom, userID)` before any downstream work — looks consistent with the existing allow_from handling in `onMessageReceived`. One tiny formatting nit: the `recalledMsgIDs` field comment got reformatted to a wider column, which is unrelated to the fix; consider folding it into the same commit or reverting it.
- `platform/feishu/platform_test.go`: new `TestInteractivePlatform_CardActionRejectsUnauthorizedUser` exercises both `handler` and `cardNavHandler` paths and asserts neither is invoked. Good coverage of the new branch.

Not blocking — just flagging for context. Real review follows once CI is green.

mclamee:
Rebased onto current `main` (`091420be`) and force-pushed `dora-agi:fix/feishu-card-action-allow-from` (`a916b37` → `8a1d6faf`) so GitHub will re-run `pull_request` CI.

- Merge conflict was only struct-field alignment in `Platform`; the `onCardAction` allow_from gate is unchanged.
- Dropped the unrelated `recalledMsgIDs` comment-column tweak from the original commit.
- Local: `go test ./platform/feishu -run TestInteractivePlatform_CardAction -count=1` and full `go test ./platform/feishu -count=1` both pass.

Happy to close this in favor of #1854 / #1853 if you prefer those later PRs; this one is the earlier same-fix from 2026-09-01.

chenhg5 (approved):
## Conclusion: Approve

Solid security fix that closes a real cross-tenant authorization gap on Feishu card actions. The `core.AllowList(p.allowFrom, userID)` gate is the same helper already used by `onMessageReceived` and other inbound paths, so semantics (empty / `*` = allow, comma-separated case-insensitive list = exact match) are consistent with the rest of the platform layer. Gate placement is correct: after `userID` extraction, before any downstream handler dispatch, with silent `nil, nil` return — zero info leak to unauthorized callers.

The earlier `CHANGES_REQUESTED` from chenhg5 was a CI gate (unrelated flake in `core/TestCUJ_H2_TwoPlatformsConcurrentNoBleed`), not a code review blocker. That head's CI is now fully green, so the gate is resolved.

### What looks good

- **Right helper, right placement**: `core.AllowList(p.allowFrom, userID)` at the top of `onCardAction` — same pattern as `onMessageReceived`. The early return means `chatID`, `messageID`, and downstream handler selection never happen for unauthorized users, so there's no path where handler logic could leak info.
- **Silent drop is correct UX**: returning `nil, nil` for unauthorized users means the card click looks like "nothing happened" to the unauthorized user. No "403" or "not allowed" feedback that would help an attacker confirm they're hitting a real system. The `slog.Debug` is enough for ops to investigate without alerting end users.
- **Test is appropriately scoped**: `TestInteractivePlatform_CardActionRejectsUnauthorizedUser` exercises both `handler` and `cardNavHandler` paths (the two main dispatch targets in `onCardAction`) and asserts neither fires. Asserts `resp == nil` and `err == nil` — confirms the silent-drop contract.
- **Regression coverage at the security boundary, not just the happy path**: the test specifically proves an *unauthorized* caller can't reach the handlers. That's exactly what needs regression protection.
- **CHANGELOG-able user-visible change**: this is a behavior change users should know about (any deploy that previously relied on cross-tenant card click access will see it stop working). Worth a release note line.

### 🔵 Optional (not blocking)

- **P3**: The test only covers the denied case. Consider adding positive coverage for completeness:
  - allowed user → `handler` IS invoked
  - empty `allow_from` → handler IS invoked (no restriction)
  - wildcard `*` → handler IS invoked
  
  Note: these are already covered for `onMessageReceived` and other paths in the same `core.AllowList` helper's other usages, so the security primitive is battle-tested. These would just round out this test file.

- **P3**: The test uses `time.After(100ms)` to assert no handler fired. This is timing-dependent — a slow CI runner could flake it. Replace with a non-blocking check:
  ```go
  select {
  case <-invoked:
      t.Fatal("unauthorized card action reached a handler")
  default:
      // good: nothing invoked
  }
  ```
  Since `onCardAction` is synchronous, the `default` branch is correct.

- **P3**: The test sets `handler` and `cardNavHandler` only. The PR body mentions four handler types (`navigation, command, permission, or question`), but the gate is at the top of `onCardAction` so all four are protected identically. Handler-level coverage is nice-to-have, not required.

### Testing / Risk

- **Verified**: CI 5/5 green on `8a1d6faf`. Reviewed `core.AllowList` in `core/message.go` to confirm semantics. Maintainer's prior CI gate is resolved.
- **Remaining risk**: P3 items above (test coverage breadth, timing dependency). No correctness gaps found.

### Next step

Merge-ready. Suggest a release note line: "Feishu card actions are now subject to the same `allow_from` allowlist as inbound messages and bot-menu events; unauthorized users will see no response on card click."

PR #1784, opened 2026-09-02
fix(daemon): fall back to ~/.cc-connect/config.toml in daemon install

### LINKED ISSUE #1783: [Bug]  doesn't fall back to ~/.cc-connect/config.toml — only checks CWD
### cc-connect Version
v1.5.0 (npm)

### Operating System
Linux (Other)

### Agent Type
Codex (OpenAI)

### Platform
Feishu (Lark)

### Installation Method

### Bug Description
`cc-connect daemon install` without arguments only looks for `./config.toml` in the current working directory. If the user runs it from `$HOME` (typical, e.g. right after ssh-ing in), it prints:

```
Warning: config.toml not found in /home/user
  Use --work-dir to specify the config directory or --config to point to the config file
```

...even though `/home/user/.cc-connect/config.toml` exists and is the standard config location used by every other subcommand (`daemon status`, plain `cc-connect`, etc. all find it fine).

### Steps to Reproduce
1. Have `~/.cc-connect/config.toml` in place (already running cc-connect interactively, for example)
2. `cd ~ && cc-connect daemon install`
3. Warning + no daemon installed

### Expected Behavior
`daemon install` should use the same config lookup order as other subcommands: `./config.toml` first, then `~/.cc-connect/config.toml` (the help text itself documents "default: ./config.toml or ~/.cc-connect/config.toml", so this is also a doc/behavior mismatch).

### Actual Behavior
Only `./config.toml` is checked; the `~/.cc-connect/config.toml` fallback is skipped.

### Workaround
```bash
cc-connect daemon install --config ~/.cc-connect/config.toml
```

### Additional Context
`cc-connect daemon status` (no args, from any directory) finds the config and works correctly — the inconsistency is specific to `daemon install`'s path resolution.

Aues6uen11Z:
## Summary
- `cc-connect daemon install` previously only checked `./config.toml` in the CWD and exited with a warning even when the standard `~/.cc-connect/config.toml` existed — inconsistent with every other subcommand, which resolves both locations.
- This PR adds the fallback: when the work-dir lookup misses, install tries `~/.cc-connect/config.toml`, prints a `Note: using config from ...`, and only warns when neither location has a config.

Fixes #1783

## Testing
Built and verified all three paths:
- CWD without config + `~/.cc-connect/config.toml` present → `Note: using config from ...` + proceeds
- Different empty CWD → same fallback, proceeds
- Neither location has a config (HOME pointed at empty dir) → original warning + exit 1 preserved

```
$ cd ~ && cc-connect daemon install
Note: using config from /home/user/.cc-connect/config.toml
Service already installed. Use --force to reinstall.
```

### DIFF
--- a/cmd/cc-connect/daemon.go
+++ b/cmd/cc-connect/daemon.go
@@ -50,16 +50,20 @@ func daemonInstall(args []string) {
 		os.Exit(1)
 	}
 
-	if err := daemon.Resolve(&cfg); err != nil {
-		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
-		os.Exit(1)
+	// Fall back to the standard ~/.cc-connect location before resolving,
+	// so that daemon.Resolve's env capture (including ${ENV} placeholder
+	// scanning of the config file) and the installed service's
+	// WorkingDirectory are consistent with the config actually in use.
+	// This matches the lookup behavior of other subcommands (see main.go
+	// config resolution: ./config.toml first, then ~/.cc-connect/config.toml).
+	workDir := cfg.WorkDir
+	if workDir == "" {
+		if wd, gerr := os.Getwd(); gerr == nil {
+			workDir = wd
+		}
 	}
-
-	configPath := cfg.WorkDir + "/config.toml"
+	configPath := filepath.Join(workDir, "config.toml")
 	if _, err := os.Stat(configPath); err != nil {
-		// Fall back to the standard ~/.cc-connect location before giving up,
-		// matching the lookup behavior of other subcommands (see main.go
-		// config resolution: ./config.toml first, then ~/.cc-connect/config.toml).
 		if home, herr := os.UserHomeDir(); herr == nil {
 			homeConfig := filepath.Join(home, ".cc-connect", "config.toml")
 			if _, serr := os.Stat(homeConfig); serr == nil {
@@ -71,10 +75,15 @@ func daemonInstall(args []string) {
 			}
 		}
 		if _, err := os.Stat(configPath); err != nil {
-			fmt.Fprintf(os.Stderr, "Warning: config.toml not found in %s\n", cfg.WorkDir)
+			fmt.Fprintf(os.Stderr, "Warning: config.toml not found in %s\n", workDir)
 			fmt.Fprintf(os.Stderr, "  Use --work-dir to specify the config directory or --config to point to the config file\n")
 			os.Exit(1)
 		}
+	}
+
+	if err := daemon.Resolve(&cfg); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
 	}
 
 	mgr, err := daemon.NewManager()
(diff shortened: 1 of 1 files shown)


Aues6uen11Z:
Good catches, thanks! Addressed in the follow-up commit:

1. **WorkDir/EnvExtra consistency** — the fallback now runs *before* `daemon.Resolve`, so when we fall back to `~/.cc-connect/config.toml`, `Resolve`'s `${ENV}` placeholder capture scans the home config (not the missing CWD one) and the installed service `WorkingDirectory` matches the config actually in use.

2. **Regression test** — agreed a test would be ideal; the install path is `os.Exit`-driven CLI logic that would need a refactor to `return error` to test meaningfully. I'd rather keep this PR minimal than restructure `daemon.go`'s error handling in the same change — happy to do that refactor as a separate PR if you want test coverage here. Meanwhile I verified the three paths manually (CWD miss + home hit → Note + proceeds; different empty CWD → same; neither → original warning preserved).

chenhg5 (approved):
## Conclusion: Approve

Thanks @Aues6uen11Z for the clean fix! Issue #1783 is a real usability bug, and this PR addresses it correctly for the common case (running `daemon install` from `$HOME` with no `--work-dir` flag).

### What looks good
- Single-file change scoped tightly to `daemonInstall`, easy to review.
- The fallback logic mirrors the `main.go` config resolution (`./config.toml` → `~/.cc-connect/config.toml`).
- CI 5/5 green (lint / unit / smoke / regression / performance).
- Manual test matrix in the PR body covers both the fallback path and the original-warning-still-fires case.

### 🟠 P2 Should improve
- **No automated test for the fallback path.** The PR relies on three manual scenarios listed in the body. Even a small unit test in `cmd/cc-connect/daemon_test.go` would lock the behavior in and prevent future regressions when refactoring the install path.
  - Suggested: extract the config-path resolution into a helper (e.g. `resolveConfigPath(workDir, homeDir string) (string, error)`) and unit-test the three branches (workDir config exists / home fallback / neither exists).

### 🔵 P3 Nice to have
- **Warning copy when neither path has config.** Now printed as `Warning: config.toml not found in <workDir>`, where `<workDir>` is still the original `--work-dir` / CWD — even after the fallback attempt failed. Users reading the message may not realize the home fallback was also tried. Consider phrasing like `Warning: config.toml not found in <workDir> or ~/.cc-connect/config.toml`.

- **Edge case: user passes `--work-dir=/some/path` and config is missing there.** With the current guard `if cfg.WorkDir == "" { cfg.WorkDir = ... }`, the fallback updates `configPath` but leaves `cfg.WorkDir` pointing at the user's explicit `--work-dir`. The installed service will then run with `WorkingDirectory=/some/path` while reading config from `~/.cc-connect/config.toml`. This is probably intentional (respect explicit override), but the `Note: using config from …` message could mention the discrepancy so users aren't surprised later. Not a blocker.

### Testing / Risk
- Verified evidence: CI all green; author's three manual scenarios in PR body cover the supported cases.
- Remaining risk: no regression test for fallback path; edge case where `--work-dir` is explicit and config is missing there is undocumented.

### Next step
- Optional but recommended: add a unit test for the fallback path before merging. Otherwise this is good to go.

PR #1785, opened 2026-09-02
fix(config): expand ~ in work_dir and base_dir

### LINKED ISSUE #1782: [Bug] work_dir does not expand ~ — spawn fails with misleading 'fork/exec codex: no such file or directory'
### cc-connect Version
v1.5.0 (npm)

### Operating System
Linux (Other)

### Agent Type
Codex (OpenAI)

### Platform
Feishu (Lark)

### Installation Method

### Bug Description
`work_dir` in `config.toml` does not expand `~`. If a user writes the natural-looking:

```toml
[projects.agent.options]
work_dir = "~/.codex/workspace"
```

the agent session can never start, and the error is extremely misleading:

```
codexSession: start: fork/exec /home/user/.npm-global/bin/codex: no such file or directory
```

The error names the **codex binary path** (which exists and works), but the actual failure is `cmd.Dir` being the literal non-existent path `~/.codex/workspace` (Go reports a missing cwd as `fork/exec ... no such file or directory` on the command itself). Users naturally go debug their codex installation instead of the config value.

### Steps to Reproduce
1. `work_dir = "~/anything"` in config.toml (the `~` form appears in docs/examples context and is what users typically write)
2. Send any message
3. Observe the `fork/exec ... codex: no such file or directory` error, even though `which codex` works and the binary is healthy

### Expected Behavior
Either:
- Expand `~` in `work_dir` (and other path options) to `$HOME` at config load time, or
- Fail fast at startup with a clear message like `work_dir does not exist: ~/.codex/workspace (note: ~ is not expanded)`

### Actual Behavior
Runtime spawn failure with an error pointing at the codex binary instead of the working directory.

### Additional Context
- Verification snippet of the underlying behavior: `subprocess.run(['codex','--version'], cwd='~/.codex/workspace')` → `FileNotFoundError` even though the directory `~/.codex/workspace` exists as `$HOME/.codex/workspace`.
- Fix location hint: `config.load()` / project option parsing — apply the same homedir expansion already used for `data_dir`-style defaults.

Aues6uen11Z:
## Summary
- A config like `work_dir = "~/.codex/workspace"` was previously passed **literally** to `exec.Cmd.Dir`. Go reports a missing cwd as `fork/exec <binary>: no such file or directory`, so the error pointed at the agent binary (which exists and works) instead of the working directory — users end up debugging their codex/claude installation instead of the config value. See #1782 for the full misdirection chain.
- This PR expands a leading `~` / `~/` in project `work_dir` (agent options) and `base_dir` to the user's home at config load time, right after the existing `${ENV}` resolution pass.

Fixes #1782

## Testing
- New unit test `TestExpandLeadingHome` covers `~`, `~/path`, absolute paths, relative paths, and `~` appearing mid-path (not expanded).
- End-to-end check: config with `work_dir = "~/.codex/workspace"` → loaded value is `/home/<user>/.codex/workspace`.
- `go test ./config/` passes (existing suite unaffected).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -627,6 +627,7 @@ func load(path string) (*Config, error) {
 		return nil, fmt.Errorf("parse config: %w", err)
 	}
 	resolveEnvInConfig(cfg)
+	expandHomeInConfig(cfg)
 	if cfg.DataDir == "" {
 		if home, err := os.UserHomeDir(); err == nil {
 			cfg.DataDir = filepath.Join(home, ".cc-connect")
@@ -671,6 +672,38 @@ var envPlaceholderPattern = regexp.MustCompile(`\$\{([A-Za-z_][A-Za-z0-9_]*)\}`)
 
 func resolveEnvInConfig(cfg *Config) {
 	resolveEnvValue(reflect.ValueOf(cfg))
+}
+
+// expandHomeInConfig expands a leading ~ / ~/ in path-like agent options
+// (work_dir, base_dir) to the user's home directory. Without this, a config
+// like work_dir = "~/.codex/workspace" is passed literally to exec.Cmd.Dir,
+// which fails at spawn time with a misleading "fork/exec ...: no such file
+// or directory" that points at the agent binary instead of the directory.
+func expandHomeInConfig(cfg *Config) {
+	home, err := os.UserHomeDir()
+	if err != nil || home == "" {
+		return
+	}
+	for i := range cfg.Projects {
+		proj := &cfg.Projects[i]
+		proj.BaseDir = expandLeadingHome(proj.BaseDir, home)
+		if v, ok := proj.Agent.Options["work_dir"]; ok {
+			if s, ok := v.(string); ok {
+				proj.Agent.Options["work_dir"] = expandLeadingHome(s, home)
+			}
+		}
+	}
+}
+
+// expandLeadingHome expands "~" and "~/" at the start of a path to home.
+func expandLeadingHome(s, home string) string {
+	if s == "~" {
+		return home
+	}
+	if strings.HasPrefix(s, "~/") {
+		return filepath.Join(home, s[2:])
+	}
+	return s
 }
 
 func resolveEnvValue(v reflect.Value) {
--- a/config/tilde_expand_integration_test.go
+++ b/config/tilde_expand_integration_test.go
@@ -1,0 +1,59 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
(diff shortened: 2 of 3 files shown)


Aues6uen11Z:
Good point — added `TestExpandHomeInConfig_LoadsAndExpandsWorkDir`, an integration-style test that writes a minimal TOML with `work_dir = "~/.codex/workspace"` and `base_dir = "~/projects"`, loads it through `LoadPermissive` (the full `load()` wiring), and asserts both come out expanded. A regression that removes the `expandHomeInConfig(cfg)` call site will now fail the suite.

chenhg5 (approved):
## Conclusion: Approve

Nice, clean fix for #1782. The bug was nasty — `fork/exec ...: no such file or directory` pointing at the agent binary instead of the working directory would send users on a wild debugging chase.

### What looks good
- Minimal, well-scoped change in `load()`: expand home right after the existing `resolveEnvInConfig` pass, so the path-mutation happens once at load time and downstream code stays unchanged.
- Tests are the right shape here: a small focused unit test on `expandLeadingHome` (5 cases including the negative `~` mid-path) plus a full `LoadPermissive` integration test that exercises both `work_dir` and `base_dir` from a real TOML file.
- Doc comment on `expandHomeInConfig` clearly explains the failure mode that motivates the fix (misleading `fork/exec ...: no such file or directory`).
- CI 5/5 green.

### 🔵 P3 Nice to have
- **Scope of expansion is intentionally narrow** (`work_dir` + `base_dir`). If other agent `Options` end up taking filesystem paths in future PRs (e.g. log_dir, cache_dir), they'll hit the same trap. Worth a follow-up issue to audit option keys vs. path-ness.
- **`expandLeadingHome` doesn't handle `~user/...`** form. Probably fine since this is a personal tool, but a one-line comment saying so would prevent future "why doesn't this work for root" reports.

### Testing / Risk
- Verified evidence: CI all green; new unit + integration tests cover the supported cases; author's manual end-to-end check mentioned in PR body.
- Remaining risk: only the two path fields listed are expanded. Other path-like agent options are still passed through unchanged (acceptable for now).

### Next step
- Ship it.

atpxu:
Thanks for the fix! Could the Web Admin save path also use the same path expansion?
On v1.5.0, `${HOME}/Works/xiabubu` already works at runtime, but saving project settings—even just changing the permission mode—returns `HTTP 400: work_dir does not exist: ${HOME}/Works/xiabubu`. The UI does not display this error.
Looking at this PR, validateProjectWorkDir() still calls os.Stat() on the unexpanded string, so this appears to remain unresolved.
Ideally, validation should use the expanded path while preserving the original expression in the saved config for portability. A Web Admin save regression test would help cover this too.

PR #1786, opened 2026-09-02
fix(codex): don't pass --listen for stdio app-server transport

### LINKED ISSUE #1781: [Bug] Codex 0.152+: app_server backend fails with 'initialize timed out' — codex app-server no longer responds on stdio when --listen ws:// is passed
### cc-connect Version
v1.5.0 (npm)

### Operating System
Linux (Other)

### Agent Type
Codex (OpenAI) — codex-cli 0.152.0

### Platform
Feishu (Lark)

### Installation Method

### Bug Description
With `backend = "app_server"` and the default `app_server_url` (resolves to `ws://127.0.0.1:3845`), every session fails after 2 minutes with:

```
level=ERROR msg="failed to start interactive session" error="codex app-server initialize: initialize timed out" elapsed=2m0.002s
```

Root cause: **codex 0.152 changed behavior — `codex app-server --listen ws://127.0.0.1:3845` now serves the JSON-RPC protocol over WebSocket only and no longer responds to JSON-RPC on stdio.** cc-connect spawns the app-server with `--listen <url>` but then speaks JSON-RPC over the stdio pipes (see `appserver_session.go` `connect()`), so `initialize` gets no reply and times out.

### Steps to Reproduce
1. Use codex-cli 0.152.0, cc-connect v1.5.0
2. Set `[projects.agent.options] backend = "app_server"` (no `app_server_url`, or any `ws://` value)
3. Send any message to the bot
4. Wait 2 minutes → `initialize timed out` in logs

Direct reproduction outside cc-connect:

```bash
# with --listen ws:// → stdin JSON-RPC gets NO response (hangs):
(echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"clientInfo":{"name":"probe","version":"0.1"}}}'; sleep 8) \
  | codex app-server --listen ws://127.0.0.1:3845

# without --listen → responds immediately on stdio:
(echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"clientInfo":{"name":"probe","version":"0.1"}}}'; sleep 8) \
  | codex app-server

# with --listen stdio:// → responds immediately on stdio:
(echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"clientInfo":{"name":"probe","version":"0.1"}}}'; sleep 8) \
  | codex app-server --listen stdio://
```

### Expected Behavior
The app_server backend works with current codex versions.

### Actual Behavior
Every session start times out after 2 minutes; Feishu side shows "启动 Agent 会话失败".

### Workaround
Set `app_server_url = "stdio"` in `[projects.agent.options]`. `normalizeAppServerURL` maps it to `stdio://`, the spawn becomes `codex app-server --listen stdio://`, and stdio JSON-RPC works again. Verified end-to-end (Feishu → reply in seconds).

### Suggested fix
Options (any of these):
- Default `app_server_url` to `stdio` (or stop passing `--listen` when the intent is stdio transport)
- Detect codex version and adapt
- Actually connect via WebSocket when a `ws://` url is configured (spawn without relying on stdio)

Given `connect()` only uses the stdio pipes for protocol today, passing `--listen ws://...` looks like it was never necessary for the transport itself.

Aues6uen11Z:
## Summary
- On codex 0.152+, any `--listen` value (including `ws://127.0.0.1:3845`, the previous default for `app_server_url`) switches `codex app-server` to serve JSON-RPC **over the listener only** and stops responding on stdio. Since `connect()` speaks the protocol over the stdio pipes, every `initialize` timed out after 2 minutes: `codex app-server initialize: initialize timed out` (#1781).
- Two changes:
  1. A stdio transport (`app_server_url = "stdio"` / `"stdio://"`) spawns plain `codex app-server` with **no `--listen` flag** (which is also codex's own default transport), restoring stdio JSON-RPC.
  2. **The default `app_server_url` is now `stdio://`** — the previous ws:// default was broken out of the box on codex 0.152+ for every user. No documented contract depends on the ws default (verified absent from docs/ and config.example.toml). Users who want a real WebSocket listener can still set a `ws://` URL explicitly.

Fixes #1781

## Testing
- Unit tests: `TestAppServerListenURL` (transport mapping) and updated `TestNormalizeAppServerURL_EmptyDefaultsToStdio` (new default).
- `go test ./agent/codex/` passes.
- End-to-end spawn verification against a stub codex binary, **both with explicit `app_server_url = "stdio"` and with no `app_server_url` at all**: spawn args are plain `app-server` (no `--listen`), vs. the previous `app-server --listen ws://127.0.0.1:3845`.
- Verified against real codex-cli 0.152.0 over Feishu: session starts and replies in seconds (previously timed out every time).

## Notes for reviewers
- Alternative direction (kept out to stay minimal): making cc-connect actually connect over WebSocket/UnixSocket when such a URL is configured. `codex app-server` also supports a `unix://` transport on 0.152+ which could enable a single resident app-server shared across sessions — that's an architecture decision better suited to a separate discussion.

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -234,10 +234,30 @@ func newAppServerSession(ctx context.Context, url, workDir, model, effort, mode,
 	return s, nil
 }
 
+// appServerListenURL returns the --listen value for the spawn. A stdio
+// transport ("stdio" / "stdio://") must not open a listener: on codex 0.152+
+// any --listen value switches the app-server to serve the protocol over the
+// listener only, leaving stdio unresponsive and timing out every initialize
+// (see #1781).
+func appServerListenURL(url string) string {
+	listenURL := strings.TrimSpace(url)
+	if listenURL == "stdio://" || listenURL == "stdio" {
+		return ""
+	}
+	return listenURL
+}
+
 func (s *appServerSession) connect() error {
 	args := []string{"app-server"}
-	if strings.TrimSpace(s.url) != "" {
-		args = append(args, "--listen", strings.TrimSpace(s.url))
+	// With url "stdio://" the session speaks JSON-RPC over the stdio pipes.
+	// Pass no --listen flag in that case: on codex 0.152+ a --listen value
+	// (including ws://) switches the app-server to serve the protocol over
+	// the listener only, leaving stdio unresponsive and causing every
+	// initialize request to time out (see #1781). --listen stdio:// is kept
+	// for explicitness on older codex versions where it is a no-op, but a
+	// bare stdio transport should simply not open a listener.
+	if listenURL := appServerListenURL(s.url); listenURL != "" {
+		args = append(args, "--listen", listenURL)
 	}
 	if model := strings.TrimSpace(s.model); model != "" {
 		args = append(args, "-c", fmt.Sprintf("model=%q", model))
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -235,13 +235,14 @@ func newAppServerSession(ctx context.Context, url, workDir, model, effort, mode,
 }
 
 // appServerListenURL returns the --listen value for the spawn. A stdio
-// transport ("stdio" / "stdio://") must not open a listener: on codex 0.152+
-// any --listen value switches the app-server to serve the protocol over the
-// listener only, leaving stdio unresponsive and timing out every initialize
-// (see #1781).
+// transport ("stdio" / "stdio://", case-insensitive, matching
+// normalizeAppServerURL's EqualFold handling) must not open a listener:
+// on codex 0.152+ any --listen value switches the app-server to serve the
+// protocol over the listener only, leaving stdio unresponsive and timing
+// out every initialize (see #1781).
 func appServerListenURL(url string) string {
 	listenURL := strings.TrimSpace(url)
-	if listenURL == "stdio://" || listenURL == "stdio" {
+	if strings.EqualFold(listenURL, "stdio://") || strings.EqualFold(listenURL, "stdio") {
 		return ""
 	}
 	return listenURL
--- a/agent/codex/appserver_session_test.go
+++ b/agent/codex/appserver_session_test.go
@@ -512,6 +512,10 @@ func TestAppServerListenURL(t *testing.T) {
 		"ws://127.0.0.1:3845":    "ws://127.0.0.1:3845",
 		"ws://localhost:9000":    "ws://localhost:9000",
 		"ws://127.0.0.1:3845 ":   "ws://127.0.0.1:3845",
+		" stdio ":                 "",
+		" stdio:// ":               "",
+		"STDIO":                    "",
+		"STDIO://":                 "",
 	}
 	for in, want := range cases {
 		if got := appServerListenURL(in); got != want {
--- a/agent/codex/appserver_session_test.go
+++ b/agent/codex/appserver_session_test.go
@@ -502,3 +502,20 @@ func waitForWrittenJSONLine(t *testing.T, w *lockedWriteCloser) string {
 		}
 	}
 }
+
+func TestAppServerListenURL(t *testing.T) {
+	cases := map[string]string{
+		"":                       "",
+		"  ":                     "",
+		"stdio://":               "",
+		"stdio":                  "",
+		"ws://127.0.0.1:3845":    "ws://127.0.0.1:3845",
+		"ws://localhost:9000":    "ws://localhost:9000",
+		"ws://127.0.0.1:3845 ":   "ws://127.0.0.1:3845",
+	}
+	for in, want := range cases {
+		if got := appServerListenURL(in); got != want {
+			t.Errorf("appServerListenURL(%q) = %q, want %q", in, got, want)
+		}
+	}
+}
--- a/agent/codex/codex.go
+++ b/agent/codex/codex.go
@@ -121,7 +121,11 @@ func normalizeBackend(raw string) string {
 func normalizeAppServerURL(raw string) string {
 	url := strings.TrimSpace(raw)
 	if url == "" {
-		return "ws://127.0.0.1:3845"
+		// Default to the stdio transport: cc-connect's app_server backend
+		// speaks JSON-RPC over the stdio pipes, and on codex 0.152+ a ws://
+		// --listen value leaves stdio unresponsive (see #1781). Users who
+		// need a WebSocket listener can still set app_server_url explicitly.
+		return "stdio://"
 	}
 	if strings.EqualFold(url, "stdio") {
 		return "stdio://"
--- a/agent/codex/codex_model_test.go
+++ b/agent/codex/codex_model_test.go
@@ -64,9 +64,9 @@ func TestNormalizeAppServerURL_StdIOIsExplicit(t *testing.T) {
 	}
 }
 
-func TestNormalizeAppServerURL_EmptyKeepsWebSocketDefault(t *testing.T) {
-	if got := normalizeAppServerURL(""); got != "ws://127.0.0.1:3845" {
-		t.Fatalf("normalizeAppServerURL(empty) = %q, want ws://127.0.0.1:3845", got)
+func TestNormalizeAppServerURL_EmptyDefaultsToStdio(t *testing.T) {
+	if got := normalizeAppServerURL(""); got != "stdio://" {
+		t.Fatalf("normalizeAppServerURL(empty) = %q, want stdio://", got)
 	}
 }
 


Aues6uen11Z:
All three points addressed in the follow-up commit:

1. **Case sensitivity** — `appServerListenURL` now uses `strings.EqualFold` for the stdio forms, consistent with `normalizeAppServerURL`'s handling.
2. **Outdated comment** — rewritten; both `stdio` and `stdio://` now drop `--listen` (nothing is kept), matching the actual behavior.
3. **Test coverage** — added `" stdio "`, `" stdio:// "`, `STDIO`, and `STDIO://` cases to `TestAppServerListenURL`.

chenhg5 (approved):
## Conclusion: Approve

## Overall assessment
PR #1786 is mergeable. CI is 5/5 green on the current head (`9918987`), `mergeable: MERGEABLE`, and the change is a tightly scoped bugfix for a regression that broke every clean `codex app-server` install on codex 0.152+. The default flip from `ws://127.0.0.1:3845` to `stdio://` is the right call, and the new `appServerListenURL` helper is a clean way to keep `--listen` opt-in.

## Review scope
- Diff (4 files, +52/-6): `agent/codex/appserver_session.go`, `agent/codex/appserver_session_test.go`, `agent/codex/codex.go`, `agent/codex/codex_model_test.go`.
- Walked through `normalizeAppServerURL`, `newAppServerSession`, and `connect()`; spot-checked the case-insensitivity / trim logic and the ws-vs-stdio branch points.

## ✅ What looks good
- **Root-cause fix is correct**: on codex 0.152+, any `--listen` value (including a `ws://` URL) switches `codex app-server` to serve JSON-RPC over the listener only and stops responding on stdio. The new `appServerListenURL` returns `""` for stdio transports, so `--listen` is simply not appended. That restores the JSON-RPC-over-stdio path that `connect()` expects.
- **Default flip is justified**: the previous default `ws://127.0.0.1:3845` was broken out of the box on codex 0.152+ for every user who didn't override `app_server_url`. The new default `stdio://` works out of the box on every supported codex version. The PR body explicitly notes that no docs or `config.example.toml` mention the old ws default — so this is a behavior change, but not a documented-contract change.
- **Case + whitespace handling is consistent**: `appServerListenURL` uses `strings.EqualFold` for case insensitivity and `strings.TrimSpace` for whitespace, mirroring the convention already used in `normalizeAppServerURL`. The test (`TestAppServerListenURL`) covers 10 cases including `"  "`, `" stdio "`, `"STDIO"`, `"STDIO://"`, `"ws://..."` with trailing whitespace.
- **`TestNormalizeAppServerURL_EmptyDefaultsToStdio`** correctly renamed from `EmptyKeepsWebSocketDefault` — name now matches behavior, which makes future readers' mental model correct.
- **End-to-end spawn verification**: author busted out a stub `codex` binary and verified that with `app_server_url = "stdio"` (and with no `app_server_url` at all), the spawn args are plain `app-server` with no `--listen` flag. That's the regression vector most likely to slip past unit tests, and the author pinned it.
- **Backward compatibility preserved**: users on codex < 0.152 still accept `--listen stdio://` as a no-op, so the new default works for them. Users explicitly setting a `ws://` or `unix://` URL are unaffected — `appServerListenURL` returns the URL unchanged and the listener path is taken as before.

- **No integration test for the ws listener path**: `TestAppServerListenURL` covers the helper exhaustively, but the spawn-args integration test only asserts the stdio path. A second sub-test that verifies `connect()` appends `--listen ws://...` correctly would close the loop. Low value (the helper itself is fully tested) but worth a 5-line addition.
- **CHANGELOG entry missing**: the default flip is a user-visible change. A one-line entry under the next release notes explaining "default `app_server_url` changed from `ws://127.0.0.1:3845` to `stdio://`" would prevent support tickets when existing deployments don't override the value. The PR body mentions the change but the actual `CHANGELOG.md` file isn't touched.

## 🔵 Optional
- **Consider noting the unix:// transport in `appServerListenURL`**: codex 0.152+ also supports `unix://` listeners, which the PR author flags as a future architecture direction. Today's helper treats `unix://` as a non-stdio URL (returns it unchanged → `--listen unix://...` gets appended), which is the right behavior. A one-line comment that explicitly says "unix:// is intentionally not treated as stdio" would lock that reasoning in.
- **The test map keys have inconsistent whitespace in `TestAppServerListenURL`**: `" stdio "` and `" stdio:// "` are spaced normally, but the surrounding whitespace values inside the map keys look slightly inconsistent. Cosmetic, no functional impact.

## ❓ Questions
- Did the author verify behavior when both `app_server_url = "stdio"` AND a future `unix://` path is set in the same config? Not a current failure mode, just confirming the helper's `EqualFold` decision tree doesn't accidentally normalize `unix://` to stdio.

## Testing / Risk
- Verified on current head `9918987`: CI run 33597438305 — `lint` ✓, `unit-test` ✓, `smoke-test` ✓, `regression-test` ✓, `performance-test` ✓.
- Verified locally (this review): the diff is confined to `agent/codex/`; no core/engine/platform changes; no schema migrations. `normalizeAppServerURL` is the only public surface touched, and the change is additive (any value that previously parsed still parses; the default shifts to a value that already had an explicit normalization path).
- Unverified risk: anyone who was *relying* on the old ws:// default to spin up an actual WebSocket listener at `127.0.0.1:3845` will silently lose that listener after this change. The PR body's claim that no docs/config mention this default is reassuring, but worth a quick `grep -rn "127.0.0.1:3845" docs/ README* config.example.toml` confirmation in the maintainer's pre-merge checklist.

## Next step
Merge when convenient. The only follow-up worth filing is the CHANGELOG entry for the default flip; everything else is P3 polish.

chenhg5 (approved):
## Conclusion: Approve

## Overall assessment
PR #1786 is mergeable. CI is 5/5 green on the current head (`9918987`), the change is a tightly scoped bugfix for a regression introduced by codex 0.152+ (`#1781`), and the maintainer has already approved (`review #5124724643`). QA confirms the unit-test coverage and end-to-end verification.

## Review scope
- Reviewed `agent/codex/appserver_session.go`, `agent/codex/codex.go`, and the corresponding test files.
- Focused on correctness, regression safety, default-config impact, and test coverage.

## ✅ What looks good
- The new `appServerListenURL` helper cleanly extracts the `--listen` decision out of `connect()` and uses `strings.EqualFold` for case-insensitive matching of `stdio` / `stdio://` — consistent with `normalizeAppServerURL`'s existing behavior.
- The default transport flip (`ws://127.0.0.1:3845` → `stdio://`) is the natural fix: `stdio` is codex's own default transport, and the previous `ws://` default is broken on `0.152+` for every user.
- PM has verified no documented contract depends on the `ws://` default (clean `docs/` and `config.example.toml` audit).
- `TestAppServerListenURL` covers 11 cases including `STDIO`, `STDIO://`, and surrounding-whitespace forms (`" stdio "`, `" stdio:// "`).

## 🚨/🔴 Must fix
- None. No merge-blocking issues found.

## 🟠 Should improve
- None for this PR. The change is appropriately minimal.

## 🔵 Optional
- P3 nit: the inline comment in `connect()` says *"--listen stdio:// is kept for explicitness on older codex versions where it is a no-op, but a bare stdio transport should simply not open a listener."* The first half is slightly misleading — the code unconditionally drops `--listen` for both `stdio` and `stdio://`, regardless of codex version. Consider tightening to something like *"On older codex versions, `--listen stdio://` was a no-op; on 0.152+ it switches the app-server off stdio. Either way, for a stdio transport we should not pass `--listen`."* Not a blocker; safe to address in a follow-up.

## ❓ Questions
- None for this PR. Other transport schemes (`unix://`, `tcp://`, etc.) are intentionally out of scope; the reporter already noted those for a follow-up issue.

## Testing / Risk
- **CI:** `lint`, `unit-test`, `smoke-test`, `regression-test`, `performance-test` — 5/5 SUCCESS on head `9918987e`. No skipped, cancelled, or pending checks.
- **Verified evidence:**
  - `TestAppServerListenURL` (11 cases) and `TestNormalizeAppServerURL_EmptyDefaultsToStdio` are in head and exercise the new behavior.
  - Author reported end-to-end verification with real `codex-cli 0.152.0` over Feishu ("秒回").
- **Remaining risk:**
  - Users on `codex <0.152` who relied on the `ws://127.0.0.1:3845` default listener will silently lose that listener after upgrading. They can opt back in by setting `app_server_url = "ws://127.0.0.1:3845"` explicitly. **Release notes should call this out.**
  - Copilot's three earlier inline comments (case-sensitivity, comment text, missing whitespace tests) are stale on the current head — items 1 and 3 are already addressed by `EqualFold` and the new test cases. Item 2 is the P3 nit noted above.

## Next step
- Recommend **merge** when the maintainer is ready. No additional changes required from the author for this PR.

PR #1787, opened 2026-09-02
fix(dingtalk): add opt-in AI Card preview message

gagaeasy:
## Summary

This PR implements option 2 for DingTalk AI Card conversation-list previews.

After an AI Card is successfully finalized, it can optionally send one separate normal Markdown message using the original reply context. DingTalk then refreshes the conversation-list preview from that normal message.

## Scope and behavior

- Added `card_final_preview_message`, defaulting to `false`.
- The extra message is sent only from the DingTalk AI Card `Finalize` path.
- Normal Markdown replies do not send an additional message.
- If the optional preview message fails, the completed AI Card is still treated as successful; the failure is logged as a warning.
- The tradeoff is one additional summary message in the chat when the option is enabled.

This intentionally does not attempt to update the AI Card's unsupported conversation-list fields. If the author agrees with this independent-message approach, the PR can be merged; if not, it can be closed or discarded.

## Validation

- `go test ./platform/dingtalk -count=1`
- `go build -tags no_web ./...`
- `go vet -tags no_web ./...`
- Added tests covering both disabled and enabled behavior, including verification that the extra request is Markdown.

Full CI should run on Ubuntu. The local Windows environment cannot run the repository's race tests because no C compiler is installed, and the full web build requires CI's frontend asset build step.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1098,6 +1098,8 @@ app_secret = "your-feishu-app-secret"
 #                                 # 收到消息时添加的表情回复（默认 "🤔Thinking"）；设为 "none" 可禁用
 # done_emoji = "none"             # Emotion when agent finishes a reply (e.g. "🥳Done"); set to "none" to disable
 #                                 # Agent 完成回复后添加的表情回复（如 "🥳Done"）；设为 "none" 可禁用
+# card_final_preview_message = false # After an AI Card turn, send a separate Markdown summary
+#                                    # AI 卡片完成后额外发送一条普通 Markdown 摘要，用于刷新会话列表预览
 
 # WPS Xiezuo / WPS 协作 (uncomment to enable / 取消注释以启用)
 # 1. Create a WPS Open Platform app and enable event WebSocket delivery
--- a/platform/dingtalk/card.go
+++ b/platform/dingtalk/card.go
@@ -22,6 +22,7 @@ type aiCard struct {
 	outTrackId     string
 	templateKey    string // 卡片模板变量名，默认 "content"
 	platform       *Platform
+	replyCtx       replyContext
 
 	mu              sync.Mutex
 	state           string // "processing" | "finished" | "failed"
@@ -209,6 +210,7 @@ func (p *Platform) createAICard(ctx context.Context, rc replyContext) (*aiCard, 
 		outTrackId:     resolvedOutTrackId,
 		templateKey:    p.cardTemplateKey,
 		platform:       p,
+		replyCtx:       rc,
 		state:          "processing",
 		throttleMs:     p.cardThrottleMs,
 		done:           make(chan struct{}),
@@ -429,7 +431,19 @@ func (c *aiCard) Finalize(ctx context.Context, content string) error {
 	}
 	c.mu.Unlock()
 
-	return err
+	if err != nil {
+		return err
+	}
+	// DingTalk does not refresh the conversation-list preview when an AI Card
+	// is finalized. Opt-in to a separate normal message to refresh it.
+	if c.platform.cardFinalPreviewMessage {
+		if sendErr := c.platform.Reply(ctx, c.replyCtx, content); sendErr != nil {
+			slog.Warn("dingtalk: final preview message failed", "error", sendErr, "outTrackId", c.outTrackId)
+		} else {
+			slog.Info("dingtalk: final preview message sent", "outTrackId", c.outTrackId)
+		}
+	}
+	return nil
 }
 
 // Failed returns true if the card has entered a failed state.
--- a/platform/dingtalk/card_preview_message_test.go
+++ b/platform/dingtalk/card_preview_message_test.go
@@ -1,0 +1,114 @@
+package dingtalk
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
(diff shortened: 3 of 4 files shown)


chenhg5 (approved):
## Conclusion: Approve

## Overall assessment
PR #1787 is mergeable. CI is 5/5 green on the current head (`beda0ea`), `mergeable: MERGEABLE`, and the substantive change is a small, opt-in workaround for a DingTalk platform limitation: AI Card finalization does not refresh the conversation-list preview, so this PR lets operators opt into a separate normal Markdown message that does. The default is `false`, the failure path is best-effort, and the existing AI Card success semantics are preserved.

## Review scope
- Diff (4 files, +150/-15): `platform/dingtalk/dingtalk.go`, `platform/dingtalk/card.go`, `platform/dingtalk/card_preview_message_test.go` (new), `config.example.toml`.
- Walked through `New` → `createAICard` → `Finalize` to confirm wiring; inspected the new `cardFinalPreviewMessage` field, the `replyCtx` carried on `aiCard`, and the integration test that stubs both the stream transport and the normal-message HTTP server.

## ✅ What looks good
- **Opt-in default is correct**: `card_final_preview_message` defaults to `false`. Existing deployments see zero behavior change. Operators who hit the DingTalk preview limitation can flip it on explicitly.
- **Failure isolation**: `Finalize` only attempts the preview message after the AI Card stream completes successfully. If the preview `Reply` call fails, it's logged as a warning and the AI Card itself is still treated as successful. The order of operations is the right way around — card success is the contract, preview refresh is a nice-to-have.
- **`aiCard` carries `replyContext`**: rather than re-deriving the reply target inside `Finalize`, the original `replyContext` is now stored on the card. Clean and avoids any subtle sessionKey reconstruction bugs.
- **Test coverage is solid**: three test functions cover (a) default-off behavior, (b) opt-in flag round-trip via `New`, and (c) integration `Finalize` behavior with stubs that count stream requests vs normal-message requests and assert `msgtype == "markdown"` when the preview is enabled. The test deliberately verifies the *negative* case (no normal message when disabled), which is the regression vector.
- **Config docs**: `config.example.toml` documents both EN and ZH-CN comments next to the new flag.

- **No interaction with existing degradation logic**: the PR's own comment notes that DingTalk's conversation-list preview is unsupported for AI Cards. There's a `degradeUntil`/`degradeMu` pair on `Platform` that's related to delivery degradation. Worth checking whether the preview message should also be suppressed while the platform is in a degraded state — otherwise an already-struggling delivery path will get one more outbound message per card finalize, which can worsen throttling issues. A one-line guard (`if c.platform.shouldDegrade() { return nil }`) or a brief comment explaining why it's intentionally not guarded would lock in the reasoning.
- **`core.HTTPClient` mutation in test**: `TestAICardFinalize_FinalPreviewMessageIsOptIn` swaps the global `core.HTTPClient` and restores it via `defer`. That's a known test-fixture pattern in this repo (saw it elsewhere), but it does mean the test is process-global state. If two parallel `t.Run` sub-tests were ever scheduled in the same process, they'd race on `core.HTTPClient`. Today it's only 2 sub-tests so it's fine, but a one-line comment acknowledging the global state would help future readers.

## 🔵 Optional
- Consider exposing the preview message as a config-driven template (e.g. include the AI Card's last `<title>` rather than the raw content) so users can curate what shows in the conversation list. Not in scope for this PR; flagged for a follow-up if operators ask for it.
- The `slog.Warn` / `slog.Info` log lines use `outTrackId` as the correlation key but don't include the sessionKey or message ID. If DingTalk operators have multiple AI Cards in flight, triaging a stray preview message would require looking up `outTrackId` against the card creation log. A `sessionKey` field on `aiCard` would close that loop.

## ❓ Questions
- The PR body says "If the optional preview message fails, the completed AI Card is still treated as successful; the failure is logged as a warning." — confirmed in code (`slog.Warn`). Is there a plan to surface the failure rate in the existing `dingtalk` health metrics (if any), so operators can tell whether their preview opt-in is actually working in production?

## Testing / Risk
- Verified on current head `beda0ea`: CI run 33606744244 — `lint` ✓, `unit-test` ✓, `smoke-test` ✓, `regression-test` ✓, `performance-test` ✓.
- Verified locally (this review): the diff is confined to `platform/dingtalk/` + `config.example.toml`; no core/engine interface changes; no schema migrations; the new test is self-contained and uses process-local stubs.
- Unverified risk: the `Reply` call inside `Finalize` adds one extra outbound HTTP request per AI Card when the flag is on. Operators should be aware this counts toward DingTalk's outbound rate limit (no rate-limit guard is added in this PR). Mentioning it in the release notes would be useful.

## Next step
Merge when convenient. No further code changes required from the author. The optional follow-ups (degradation guard, `sessionKey` correlation log field) are P3 and can be filed as separate issues.

gagaeasy:
@chenhg5 这条 PR 没有把预览消息的失败率接到健康指标上，仓库里目前也没有可接入的钉钉健康指标。

开关打开后，预览消息发送成功会打 `slog.Info("dingtalk: final preview message sent")`，失败会打 `slog.Warn("dingtalk: final preview message failed")`，并带上 `outTrackId`。预览发送失败不影响 AI Card 本身的成功结果。现有的 `degradeUntil` 只是 AI Card 接口的 30 分钟降级窗口，不是预览消息的计数或健康指标。

所以这条 PR 保持为可选开关加日志。如果后面要做失败率，需要先有可上报的指标接口，再单独开一条 follow-up。

PR #1790, opened 2026-09-02
feat(acp): implement ContextUsageReporter from opencode usage signals (mitigates #1672 / #1764)

TonaliNoHitsuzi:
# PR-B：feat(acp): 实现 ContextUsageReporter（真实上下文占用上报）

> 状态：文档就绪，待提交。关联上游 open issues **#1672**（ctx footer 硬编码 200k 窗口）与 **#1764**（auto-compress 用文本估算，漏掉 70-85% 真实上下文）。
> 基线 b39c11f。

## PR 标题（中英）

- feat(acp): report real context usage from opencode `usage_update` + prompt usage (implements ContextUsageReporter, mitigates #1672 / #1764)

## 问题背景

core 引擎已有 `ContextUsageReporter` 可选接口（core/interfaces.go:494+）与完整 `ContextUsage` 结构（UsedTokens / CachedInputTokens / CacheCreationInputTokens / ContextWindow 等），`replyFooterContextText` / auto-compress 触发逻辑都会优先采用它。但 **ACP agent 未实现该接口**，导致：

1. `[ctx: ~N%]` footer 走 `contextIndicatorText(event.InputTokens)` 文本估算（#1672：大窗口模型固定显示 ~100%）
2. auto-compress 触发估算漏掉 tool_use/tool_result 与 system prompt 开销（#1764）

而 opencode 的 ACP 端**恰好持续上报两类真实数据**：

- `session/update` 通知 `usage_update`：`{"used": 8606, "size": 200000}`（当前上下文累计 / 模型窗口）
- `session/prompt` 响应 `usage`：`{"inputTokens":158, "outputTokens":33, "totalTokens":8639, "cachedReadTokens":8448}`（逐 turn 精确值，含缓存命中）

## 实现（agent/acp/session.go）

1. `acpSession` 增加字段：`usageMu sync.RWMutex; lastUsage core.ContextUsage; haveUsage bool`
2. `onNotification` 中新增 `maybeAbsorbUsageUpdate(params)`：识别 `sessionUpdate == "usage_update"`，吸收 `used` → UsedTokens、`size` → ContextWindow（与既有 `maybeAbsorbCurrentModeUpdate` 同构，遵循锁与解析模式）
3. `session/prompt` 响应后吸收 `usage.cachedReadTokens/inputTokens/outputTokens/totalTokens`（CachedInputTokens 是 footer 缓存命中的唯一来源）
4. 实现 `GetContextUsage() *core.ContextUsage`（快照返回）

```go
func (s *acpSession) GetContextUsage() *core.ContextUsage {
	s.usageMu.RLock()
	defer s.usageMu.RUnlock()
	if !s.haveUsage {
		return nil
	}
	cu := s.lastUsage
	return &cu
}
```

## 实测效果（GLM-5.3，1M 窗口）

footer 从估算（issue #1672 病症）：

```
[ctx: ~1%]   ← 200k 硬编码下的估算
```

变为真实数据（本 PR 附带的 footer 渲染示例）：

```
上下文: 14.7k / 1.0M (1%)
缓存: 14.7k (99%)
```

（缓存 99% = cachedReadTokens/used，对用户判断"这轮是不是几乎全走了缓存"非常直观。）

对 #1764：引擎 auto-compress 分支 `if usage := replyFooterSessionContextUsage(state.agentSession); usage != nil && usage.UsedTokens > 0 { estimate = usage.UsedTokens }` ——本 PR 使该分支对 ACP 后端生效，触发阈值从"文本估算"升级为 API 上报真值。

## 复现指南

1. 接入 opencode ACP（同 PR-A 环境），DEBUG 日志确认两类数据到达：

```
level=DEBUG msg="acp: session/update" params="{...\"sessionUpdate\":\"usage_update\"...\"used\":8606,\"size\":200000...}"
level=DEBUG msg="acp: session/prompt response" response="{\"stopReason\":\"end_turn\",\"usage\":{\"inputTokens\":158,...\"cachedReadTokens\":8448}...}"
```

2. 修复前：footer ctx 值来自估算（可对比 turn complete 日志的 input_tokens 与 footer 百分比是否基于真实窗口）
3. 修复后：footer 显示 used/size 精确量纲与百分比；auto-compress 阈值按 UsedTokens 触发

## 兼容性说明

- 仅在 opencode 发送 usage_update 时激活；其他 ACP 后端（Cursor/Trae 等不发该事件）GetContextUsage 返回 nil，行为与现状完全一致（引擎已有 nil 降级路径）
- 不改变 quiet/full 任何显示模式语义，仅替换数据源

## 附：可选的 footer 展示增强（可拆分讨论）

本仓库同时改了 `buildClaudeStatusLineFooter` 的展示格式（分行标签式），这部分是主观体验优化，可按维护者偏好拆出或舍弃——数据层（本 PR 核心）与展示层解耦，`contextUsageRichText` 为独立函数。

### DIFF
--- a/agent/acp/session.go
+++ b/agent/acp/session.go
@@ -49,6 +49,9 @@ type acpSession struct {
 	// updated whenever SetLiveMode succeeds or the server announces a
 	// mode change via session/update.
 	modesMu        sync.RWMutex
+	usageMu        sync.RWMutex
+	lastUsage      core.ContextUsage
+	haveUsage      bool
 	availableModes []acpModeInfo
 	currentMode    string
 
@@ -372,12 +375,55 @@ func (s *acpSession) onNotification(method string, params json.RawMessage) {
 	}
 	s.cacheToolCallInput(params)
 	s.maybeAbsorbCurrentModeUpdate(params)
+	s.maybeAbsorbUsageUpdate(params)
 	sid := s.currentACPSessionID()
 	// Debug log to capture raw session/update JSON for troubleshooting vendor compatibility
 	slog.Debug("acp: session/update", "session_id", sid, "params", string(params))
 	for _, ev := range mapSessionUpdate(sid, params) {
 		s.emit(ev)
 	}
+}
+
+// maybeAbsorbUsageUpdate watches session/update notifications for
+// `usage_update` (opencode extension: {"used": N, "size": M}) and records
+// real context usage for the reply footer / auto-compress heuristics.
+func (s *acpSession) maybeAbsorbUsageUpdate(params json.RawMessage) {
+	var wrap struct {
+		Update json.RawMessage `json:"update"`
+	}
+	if json.Unmarshal(params, &wrap) != nil || len(wrap.Update) == 0 {
+		return
+	}
+	var head struct {
+		Kind string `json:"sessionUpdate"`
+	}
+	if json.Unmarshal(wrap.Update, &head) != nil || head.Kind != "usage_update" {
+		return
+	}
+	var u struct {
+		Used int `json:"used"`
+		Size int `json:"size"`
+	}
+	if json.Unmarshal(wrap.Update, &u) != nil {
+		return
+	}
+	s.usageMu.Lock()
+	s.lastUsage.UsedTokens = u.Used
+	s.lastUsage.ContextWindow = u.Size
+	s.haveUsage = true
+	s.usageMu.Unlock()
+}
+
+// GetContextUsage implements core.ContextUsageReporter with real figures
+// captured from usage_update + session/prompt responses.
+func (s *acpSession) GetContextUsage() *core.ContextUsage {
+	s.usageMu.RLock()
+	defer s.usageMu.RUnlock()
+	if !s.haveUsage {
+		return nil
+	}
+	cu := s.lastUsage
+	return &cu
 }
 
 // maybeAbsorbCurrentModeUpdate watches session/update notifications
@@ -621,6 +667,27 @@ func (s *acpSession) Send(prompt string, messageID string, images []core.ImageAt
 		return fmt.Errorf("acp: session/prompt: %w", err)
 	}
 	slog.Debug("acp: session/prompt response", "session_id", sid, "response_len", len(res), "response", string(res))
+
+	// Absorb per-turn usage (esp. cachedReadTokens) into the context snapshot.
+	var pr struct {
+		Usage struct {
+			InputTokens      int `json:"inputTokens"`
+			OutputTokens     int `json:"outputTokens"`
+			TotalTokens      int `json:"totalTokens"`
+			CachedReadTokens int `json:"cachedReadTokens"`
+		} `json:"usage"`
+	}
+	if json.Unmarshal(res, &pr) == nil && (pr.Usage.InputTokens > 0 || pr.Usage.TotalTokens > 0) {
+		s.usageMu.Lock()
+		if pr.Usage.CachedReadTokens > 0 {
+			s.lastUsage.CachedInputTokens = pr.Usage.CachedReadTokens
+		}
+		s.lastUsage.InputTokens = pr.Usage.InputTokens
+		s.lastUsage.OutputTokens = pr.Usage.OutputTokens
+		s.lastUsage.TotalTokens = pr.Usage.TotalTokens
+		s.haveUsage = true
+		s.usageMu.Unlock()
+	}
 
 	// Text was streamed via session/update; engine aggregates EventText.
 	s.emit(core.Event{


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Single-file additive fix that completes a well-known gap: ACP agent now implements the optional `ContextUsageReporter` interface. CI 5/5 green. The implementation closely mirrors the existing `maybeAbsorbCurrentModeUpdate` helper, uses correct field names from `core.ContextUsage`, and properly synchronizes concurrent updates with `usageMu`. Real user impact is measurable: fixes the long-standing `[ctx: ~1%]` footer bug (#1672) on large-window models and unlocks accurate auto-compress thresholds (#1764). Merge-ready.

Review scope:
- 1 file / +67/-0, `agent/acp/session.go`. No interface changes, no other touched files.
- Focused on correctness of the JSON unmarshal paths (two: notification handler + per-turn prompt response), lock discipline, defensive `haveUsage` gating, and field-name alignment with `core.ContextUsage`.

✅ What looks good:
- Field mapping is exact and exhaustive: `used → UsedTokens`, `size → ContextWindow`, `inputTokens/outputTokens/totalTokens/cachedReadTokens → InputTokens/OutputTokens/TotalTokens/CachedInputTokens`. No typos, no missed fields. The `core.ContextUsage` struct's `BaselineTokens`, `CacheCreationInputTokens`, `ReasoningOutputTokens` remain zero — which is correct, as opencode doesn't expose those values in either the notification or the response.
- Lock discipline is correct: a single `usageMu sync.RWMutex` guards all reads/writes; `GetContextUsage` uses `RLock`; both writers use `Lock`. Snapshot semantics are achieved by copying the struct value to a local `cu := s.lastUsage` before returning the pointer — no risk of the caller observing a half-updated struct while another goroutine writes.
- `GetContextUsage` returns `nil` when `haveUsage == false`, which exactly matches the engine's nil-tolerant helper `replyFooterSessionContextUsage → replyFooterContextText`. The `if usage.ContextWindow <= 0 { return "" }` and zero-UsedTokens fallbacks in `replyFooterContextText` are intact, so a stale snapshot can never crash the footer.
- The per-turn `Send` write uses a positive-condition guard `pr.Usage.InputTokens > 0 || pr.Usage.TotalTokens > 0` before flipping `haveUsage = true`. This is important: if opencode ever returns a `session/prompt` response with zero usage (e.g. a streaming-error path), the previously captured `usage_update` snapshot is preserved rather than being clobbered with empty data.
- `maybeAbsorbUsageUpdate` early-returns on `json.Unmarshal` errors or `len(wrap.Update) == 0`, so malformed payloads from non-opencode ACP servers (Cursor, Trae, etc.) silently no-op without affecting `haveUsage`. Compatibility note in the PR body is honest and verifiable.
- The notification handler calls the new helper inline with the existing `cacheToolCallInput` and `maybeAbsorbCurrentModeUpdate` chain — no branching, no new event flow, just one more line in an already-ordered pipeline.
- Symmetric comment quality: both helpers carry one-line doc comments that name the opencode extension and the consumer (footer / auto-compress). Matches the surrounding `// maybeAbsorbCurrentModeUpdate ...` style.

🟠 Should improve:
- None that block merge.

🔵 Optional:
- **P3 — add a unit test for `maybeAbsorbUsageUpdate` and `GetContextUsage`.** The sibling `maybeAbsorbCurrentModeUpdate` is also untested (file's existing pattern), so this is consistent — but the new helper has a more complex JSON path (two-layer unwrap: outer `params.update` → inner `sessionUpdate` + payload), which is exactly where a future regression would hide. A ~20-line test that feeds a known `usage_update` notification and asserts `GetContextUsage().UsedTokens` / `ContextWindow` would pin the fix and make any future opencode event-name change caught by CI. The engine-level footer test (which uses the integration) is not a substitute.
- **P3 nit — `cachedReadTokens` is conditionally written, but `InputTokens/OutputTokens/TotalTokens` are written unconditionally when the response parses.** If opencode returns a `session/prompt` response with all-zero usage but non-zero output (theoretically possible if only `cachedReadTokens` was set), the unconditional write would still set `haveUsage = true` and `lastUsage.TotalTokens = 0`. The `replyFooterContextText` helper handles this via the `usedTokens <= 0` fallback chain (TotalTokens → InputTokens+OutputTokens), so footer won't crash, but the snapshot becomes briefly degraded. Consider guarding the write on `pr.Usage.TotalTokens > 0 || pr.Usage.InputTokens > 0 || pr.Usage.OutputTokens > 0` to match the existing `haveUsage` flip predicate. Today this is harmless; tightening it is a 1-character diff.
- **P3 nit — comment typo: "watching session/update notifications for `usage_update` (opencode extension: {"used": N, "size": M})".** Looks correct; nothing to fix. Just noting I checked.

❓ Questions:
- None blocking. Minor: does opencode guarantee that `usage_update` arrives *after* every `session/prompt` response, or can it arrive at any time during streaming? If it can arrive mid-stream, there's a theoretical race where the `usage_update` snapshot represents an older state than the most recent `session/prompt` response. In practice the snapshot is monotonically increasing and the engine renders the footer at turn end, so this isn't a correctness issue — but it's worth a sentence in the helper's comment to document the assumption.

Testing / Risk:
- Verified evidence: CI 5/5 green (run 33650868431: lint, unit, regression, smoke, performance all pass). The change is mechanically provable: the new helper consumes a known event shape, writes through a single mutex, and reads through the existing engine helper chain that already handles nil/zero defensively.
- Remaining risk: the optional `CacheCreationInputTokens` field is intentionally not populated (opencode doesn't expose it). If a future version starts exposing it, this PR won't capture it — but that's not a regression, it's a feature-add follow-up.
- Compatibility: zero risk for non-opencode ACP servers. Cursor/Trae/Windsurf/etc. don't emit `usage_update`; `haveUsage` stays false; `GetContextUsage` returns nil; engine falls through to existing `contextIndicatorText` heuristic exactly as before.
- Security: the helper only reads JSON scalar fields and stores integers under a mutex. No injection vector, no capability escalation, no trust expansion (the helper is invoked from the existing notification path which already accepts `session/update` payloads).

Next step:
- Author can (a) merge as-is — the fix is correct, additive, and has clear measurable impact on opencode users; (b) add the optional unit test in this same PR (recommended); or (c) address the optional P3 nit on the per-turn write guard. All three paths are non-blocking. Approval stands. Ready for maintainer merge gate.

PR #1792, opened 2026-09-04
feat(codex): support fast mode

mofanke:
## Summary

- add a /fast [on|off] chat command for Codex agents
- pass the selected service_tier to both exec and app-server backends
- persist project and multi-workspace selections while preserving the Codex conversation on process restart
- expose status/help text in all supported languages and document configuration
- inherit native Codex configuration when service_tier is omitted

## Validation

- go test ./... (4033 passed across 51 packages)
- go test ./core -run TestCUJ (55 passed)
- go vet ./...
- go build ./...

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -145,6 +145,7 @@ type appServerSession struct {
 	workDir        string
 	model          string
 	effort         string
+	serviceTier    string
 	mode           string
 	baseURL        string
 	modelProvider  string
@@ -192,12 +193,17 @@ const (
 )
 
 func newAppServerSession(ctx context.Context, url, workDir, model, effort, mode, resumeID, baseURL, modelProvider string, extraEnv []string, codexHome string, systemPrompt string, appendPrompt string) (*appServerSession, error) {
+	return newAppServerSessionWithServiceTier(ctx, url, workDir, model, effort, "", mode, resumeID, baseURL, modelProvider, extraEnv, codexHome, systemPrompt, appendPrompt)
+}
+
+func newAppServerSessionWithServiceTier(ctx context.Context, url, workDir, model, effort, serviceTier, mode, resumeID, baseURL, modelProvider string, extraEnv []string, codexHome string, systemPrompt string, appendPrompt string) (*appServerSession, error) {
 	sessionCtx, cancel := context.WithCancel(ctx)
 	s := &appServerSession{
 		url:              url,
 		workDir:          workDir,
 		model:            model,
 		effort:           effort,
+		serviceTier:      normalizeServiceTier(serviceTier),
 		mode:             mode,
 		baseURL:          baseURL,
 		modelProvider:    modelProvider,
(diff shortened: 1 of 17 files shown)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- PR is mergeable. CI 5/5 green, new feature surface is well-scoped, conversation-preservation semantics are tested, and the implementation closely mirrors the existing `/reasoning` switcher. The one asymmetry I called out is cosmetic and not a blocker.

Review scope:
- 17 files, 443/+48/-, focused on `agent/codex` (`appserver_session.go`, `codex.go`, `session.go`), `core` (`engine.go`, `i18n.go`, `interfaces.go`, `projectstate.go`), `config`, and docs.
- Focused on correctness of the `service_tier` plumbing, conversation/state preservation across toggle, multi-workspace handling, persistence semantics, and test coverage.
- Verified: app-server `buildCommandArgs()` omits `service_tier=...` when empty (inherit), `cmdFast` calls `sessions.Save()` after `SetFastMode`, and `persistWorkspaceFastModeOverride` correctly no-ops outside `e.multiWorkspace` or for the project-default agent.

✅ What looks good:
- The "empty string → inherit Codex native config" rule is honored consistently in both backends (`buildCommandArgs`, `buildExecArgs`) and pinned by `TestAppServerCommandArgs_OmitsUnsetServiceTier` + `TestBuildExecArgs_IncludesExplicitDefaultServiceTier`. Operators who don't set `service_tier` get zero behavioral drift — important for not silently changing cost/latency on upgrade.
- `TestCmdFast_SwitchesTierPersistsAndPreservesConversation` exercises the full flow including the `s.SetAgentSessionID("existing-session", "test")` regression check, so the claim "next message resumes the existing conversation" is actually tested.
- `persistWorkspaceFastModeOverride` mirrors `persistWorkspaceModelOverride` exactly: same no-op guards (`!e.multiWorkspace || agent == e.agent`), same `workspaceModelOverrideKey()` derivation. The pattern is well-understood and the new field is plumbed through `getOrCreateWorkspaceAgent` symmetrically.
- 5-language i18n strings (en/zh/zh-TW/ja/es) are added for all new user-facing messages, matching the file's strict parity requirement.
- `FastModeSwitcher` interface is optional and the engine gracefully reports "not supported" via `MsgFastNotSupported` — same defensive pattern as `ReasoningEffortSwitcher`.
- "Outside multi-workspace mode, this is a project-wide default" is documented in both `docs/usage.md` and `docs/usage.zh-CN.md` — important caveat for shared projects.

🟠 Should improve:
- **P2 — Asymmetric constructor for service_tier on exec vs. app-server paths.** `newAppServerSessionWithServiceTier` accepts `serviceTier` as a clean positional argument, but `newCodexSession` doesn't, so `StartSession` post-assigns `session.serviceTier = serviceTier` after construction. Suggested fix: add `serviceTier` as a positional argument to `newCodexSession` for symmetry — same shape as the existing `effort` parameter — and drop the post-assignment. This is a 5-line cleanup, doesn't change behavior, and keeps the two backends on the same constructor contract.

🔵 Optional:
- **P3 nit — `MsgStatusFastMode` uses `⚡` glyph in all five locales.** The other status lines are language-localized but the emoji prefix is universal. Consistent with how `MsgStatusMode` is structured, so leaving it is fine; just flagging for awareness.
- **P3 nit — `case "default"` in `normalizeServiceTier` returns `"default"` while `case "fast"` returns `"fast"`, but `SetFastMode(false)` also writes `"default"` as the saved value.** This is consistent and good (round-trip safe), but worth a one-line comment on `normalizeServiceTier` clarifying that `"default"` is a meaningful, persisted value (not the same as empty/inherit). Current readers may briefly wonder whether `"default"` ever appears as a CLI argument.
- **P3 nit — `TestNormalizeServiceTier` covers `"priority"` returning `""` but doesn't cover an unknown-but-non-empty value like `"turbo"`.** Add one case to the table to pin "unknown → inherit" more clearly.

❓ Questions:
- The PR body mentions "persist project and multi-workspace selections while preserving the Codex conversation on process restart." For the **app-server** backend, `buildCommandArgs` is rebuilt on every `StartSession` so `service_tier` always reflects current state. For the **exec** backend, is there any caching of `codexSession.buildExecArgs` between turns, or does the next user message also rebuild args? If args are cached anywhere, a `/fast on` followed by the same session might not pick up the new tier until process restart. A quick check on `codexSession.buildExecArgs` callers would confirm.

Testing / Risk:
- Verified evidence: CI 5/5 green (run 33833706150: lint, unit, regression, smoke, performance all pass). Author reports `go test ./...` 4033 passed / 51 packages and `go test ./core -run TestCUJ` 55 passed. New tests cover normalization, both backends' arg construction (with and without service_tier), full `cmdFast` happy path including conversation preservation, unsupported-agent fallback, config persistence (`SaveAgentServiceTier`), and projectstate override persistence.
- Remaining risk: the conversation-preservation claim is tested at the engine level (AgentSessionID and History length), but there is no test confirming that the **exec process** actually picks up the new `service_tier` on the next turn. In practice the process is restarted, so this is presumed correct, but it relies on the runtime contract being honored. If we ever cache exec args on a session, this guarantee silently breaks.
- Compatibility: no breaking changes. Existing installations with no `service_tier` config will get `"inherit native Codex config"` semantics, which is the conservative default.
- Security: the `service_tier` value passes only through `strings.TrimSpace` + a closed switch; no risk of injection or arbitrary CLI args.

Next step:
- Author can either (a) extend `newCodexSession` to take `service_tier` as a positional arg and drop the post-assignment (P2 cleanup), or (b) leave as-is and merge. Either way, approval stands. Ready for maintainer merge gate.

PR #1793, opened 2026-09-04
fix: preserve inbound Feishu alert cards

fengnian2025:
## Problem

Feishu alert cards arrive as `interactive` messages, but the inbound path only retained a lossy summary (and ignored card-only payloads in some paths). Important alert fields such as action values and links therefore never reached the agent.

## Changes

- Preserve inbound Feishu Card 2.0 payloads as a separate `InboundCard` value with message ID, schema/version, readable summary, raw JSON, and truncation metadata.
- Fetch `raw_card_content` for direct interactive messages and quoted reply chains.
- Preserve nested interactive cards in `merge_forward` messages.
- Support `inbound_card_mode = "summary" | "raw" | "both"` and a bounded raw-card size option.
- Inject raw cards into the agent prompt as explicitly untrusted, size-bounded JSON.

## Tests

- `go test -count=1 ./platform/feishu`
- Targeted core, Feishu card, merge-forward, and Codex app-server tests

Baseline: `4000b2338aa6e850c99df54f8b0ed6ed7460b401` (current upstream `main`).

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1012,6 +1012,10 @@ app_secret = "your-feishu-app-secret"
 # domain = "https://open.feishu.cn"  # Optional runtime API/WebSocket base URL override / 可选：运行时 API/WebSocket 域名覆盖
 # enable_feishu_card = true  # Enable Feishu interactive cards (default: true); set false to force plain-text replies
 #                              # 启用飞书交互式卡片（默认 true）；设为 false 时强制回退纯文本回复
+# inbound_card_mode = "both"  # Inbound card payload: summary (default), raw, or both
+#                              # 入站卡片内容：summary（默认）、raw 或 both
+# inbound_card_max_bytes = 0   # Maximum raw card JSON bytes; 0 = no truncation
+#                              # 单卡原始 JSON 最大字节数；0 表示不截断
 # allow_from = "*"          # Allowed user open_ids, comma-separated; "*" = all (default). Use /whoami to get your open_id.
 #                              # 允许的用户 open_id，逗号分隔；"*" 表示所有（默认）。发送 /whoami 获取你的 open_id。
 # allow_chat = "*"          # Allowed group chat_ids, comma-separated; "*" = all (default). Find chat_id in group settings.
@@ -1078,6 +1082,8 @@ app_secret = "your-feishu-app-secret"
 # callback_path = "/feishu/webhook"  # Webhook callback path / Webhook 回调路径
 # encrypt_key = ""               # Event encrypt key (optional) / 事件加密密钥（可选）
 # enable_feishu_card = true
+# inbound_card_mode = "both"  # summary (default), raw, or both
+# inbound_card_max_bytes = 0   # Maximum raw card JSON bytes; 0 = no truncation
 # allow_from = "*"
 # reaction_emoji = "OnIt"
 # done_emoji = "none"
--- a/config.example.toml
+++ b/config.example.toml
@@ -1014,8 +1014,8 @@ app_secret = "your-feishu-app-secret"
 #                              # 启用飞书交互式卡片（默认 true）；设为 false 时强制回退纯文本回复
 # inbound_card_mode = "both"  # Inbound card payload: summary (default), raw, or both
 #                              # 入站卡片内容：summary（默认）、raw 或 both
-# inbound_card_max_bytes = 0   # Maximum raw card JSON bytes; 0 = no truncation
-#                              # 单卡原始 JSON 最大字节数；0 表示不截断
+# inbound_card_max_bytes = 32768   # Maximum raw card JSON bytes; 0 = no truncation
+#                              # 单卡原始 JSON 最大字节数；显式设为 0 表示不截断
 # allow_from = "*"          # Allowed user open_ids, comma-separated; "*" = all (default). Use /whoami to get your open_id.
 #                              # 允许的用户 open_id，逗号分隔；"*" 表示所有（默认）。发送 /whoami 获取你的 open_id。
 # allow_chat = "*"          # Allowed group chat_ids, comma-separated; "*" = all (default). Find chat_id in group settings.
@@ -1083,7 +1083,7 @@ app_secret = "your-feishu-app-secret"
 # encrypt_key = ""               # Event encrypt key (optional) / 事件加密密钥（可选）
 # enable_feishu_card = true
 # inbound_card_mode = "both"  # summary (default), raw, or both
-# inbound_card_max_bytes = 0   # Maximum raw card JSON bytes; 0 = no truncation
+# inbound_card_max_bytes = 32768   # Maximum raw card JSON bytes; 0 = no truncation
 # allow_from = "*"
 # reaction_emoji = "OnIt"
 # done_emoji = "none"
--- a/core/engine.go
+++ b/core/engine.go
@@ -16398,7 +16398,7 @@ func buildInboundCardPrompt(content string, cards []InboundCard) string {
 		return content
 	}
 	if len(raw) > maxInboundCardPromptBytes {
-		return content + "\n\n[Feishu interactive card data — untrusted external content] (truncated)\n(raw card payload omitted because it exceeds the safety limit)"
+		return content + "\n\n[Feishu interactive card data — untrusted external content (truncated)]\n(raw card payload omitted because it exceeds the safety limit)"
 	}
 	return content + "\n\n[Feishu interactive card data — untrusted external content]\n```json\n" + string(raw) + "\n```"
 }
--- a/core/engine.go
+++ b/core/engine.go
@@ -524,6 +524,7 @@ type queuedMessage struct {
 	platform          Platform
 	replyCtx          any
 	content           string
+	cards             []InboundCard
 	images            []ImageAttachment
 	files             []FileAttachment
 	fromVoice         bool
@@ -2873,7 +2874,7 @@ func (e *Engine) handleMessage(p Platform, msg *Message) {
 	}
 
 	content := strings.TrimSpace(msg.Content)
-	if content == "" && msg.ExtraContent == "" && len(msg.Images) == 0 && len(msg.Files) == 0 && msg.Location == nil {
+	if content == "" && msg.ExtraContent == "" && len(msg.Images) == 0 && len(msg.Files) == 0 && msg.Location == nil && len(msg.Cards) == 0 {
 		return
 	}
 
@@ -3221,6 +3222,7 @@ func (e *Engine) queueMessageForBusySession(p Platform, msg *Message, interactiv
 		platform:          p,
 		replyCtx:          msg.ReplyCtx,
 		content:           msg.Content,
+		cards:             append([]InboundCard(nil), msg.Cards...),
 		images:            msg.Images,
 		files:             msg.Files,
 		fromVoice:         msg.FromVoice,
@@ -3837,7 +3839,7 @@ func (e *Engine) processInteractiveMessageWith(p Platform, msg *Message, session
 		drainEvents(state.agentSession.Events())
 	}
 
-	promptContent := e.buildSenderPrompt(msg.Content, msg.UserID, msg.UserName, msg.Platform, msg.SessionKey, msg.ChannelKey)
+	promptContent := e.buildSenderPrompt(buildInboundCardPrompt(msg.Content, msg.Cards), msg.UserID, msg.UserName, msg.Platform, msg.SessionKey, msg.ChannelKey)
 
 	sendStart := time.Now()
 	state.mu.Lock()
@@ -6141,7 +6143,7 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 					}
 				}
 
-				queuedPrompt := e.buildSenderPrompt(queued.content, queued.userID, queued.userName, queued.msgPlatform, queued.msgSessionKey, queued.channelKey)
+				queuedPrompt := e.buildSenderPrompt(buildInboundCardPrompt(queued.content, queued.cards), queued.userID, queued.userName, queued.msgPlatform, queued.msgSessionKey, queued.channelKey)
 
 				state.mu.Lock()
 				as := state.agentSession // capture under lock to avoid race with cleanup
@@ -6460,7 +6462,7 @@ func (e *Engine) drainPendingMessages(state *interactiveState, session *Session,
 		state.mu.Unlock()
 
 		e.i18n.DetectAndSet(queued.content)
-		prompt := e.buildSenderPrompt(queued.content, queued.userID, queued.userName, queued.msgPlatform, queued.msgSessionKey, queued.channelKey)
+		prompt := e.buildSenderPrompt(buildInboundCardPrompt(queued.content, queued.cards), queued.userID, queued.userName, queued.msgPlatform, queued.msgSessionKey, queued.channelKey)
 
 		state.mu.Lock()
 		as := state.agentSession // capture under lock to avoid race with cleanup (mirrors #1436)
@@ -16377,6 +16379,28 @@ func (e *Engine) buildSenderPrompt(content, userID, userName, platform, sessionK
 		return fmt.Sprintf("[cc-connect sender_id=%s sender_name=\"%s\" platform=%s chat_id=%s]\n%s", userID, safeName, platform, chatID, content)
 	}
 	return fmt.Sprintf("[cc-connect sender_id=%s platform=%s chat_id=%s]\n%s", userID, platform, chatID, content)
+}
+
+// Feishu limits an individual card to roughly 30 KiB; leave enough room for
+// several cards in one merge_forward payload while retaining a hard prompt
+// safety bound.
+const maxInboundCardPromptBytes = 256 * 1024
+
+// buildInboundCardPrompt appends bounded, explicitly untrusted card data to
+// the human-readable summary. Cards are kept out of Message.Content/history;
+// this fragment is only constructed at the agent boundary.
+func buildInboundCardPrompt(content string, cards []InboundCard) string {
+	if len(cards) == 0 {
+		return content
+	}
+	raw, err := json.Marshal(cards)
+	if err != nil {
+		return content
+	}
+	if len(raw) > maxInboundCardPromptBytes {
+		return content + "\n\n[Feishu interactive card data — untrusted external content] (truncated)\n(raw card payload omitted because it exceeds the safety limit)"
+	}
+	return content + "\n\n[Feishu interactive card data — untrusted external content]\n```json\n" + string(raw) + "\n```"
 }
 
 func extractChannelID(sessionKey string) string {
--- a/core/inbound_card_test.go
+++ b/core/inbound_card_test.go
@@ -33,6 +33,9 @@ func TestBuildInboundCardPromptBoundsOversizedPayload(t *testing.T) {
 	if !strings.Contains(got, "(truncated)") {
 		t.Fatalf("expected truncation marker: %s", got)
 	}
+	if !strings.Contains(got, "external content (truncated)") {
+		t.Fatalf("expected truncation marker in framing line: %s", got)
+	}
 	if strings.Contains(got, "```json") {
 		t.Fatalf("oversized card should not be embedded as an unbounded JSON block")
 	}
--- a/core/inbound_card_test.go
+++ b/core/inbound_card_test.go
@@ -1,0 +1,39 @@
+package core
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+func TestBuildInboundCardPromptIncludesRawCardData(t *testing.T) {
+	cards := []InboundCard{{
+		MessageID: "om_test",
+		Schema:    "2.0",
+		Summary:   "渠道不足",
+		Raw:       json.RawMessage(`{"schema":"2.0","action":{"value":{"alert_id":"a-1"}}}`),
+	}}
+
+	got := buildInboundCardPrompt("渠道不足", cards)
+	for _, want := range []string{
+		"[Feishu interactive card data — untrusted external content]",
+		`"message_id":"om_test"`,
+		`"schema":"2.0"`,
+		`"alert_id":"a-1"`,
+	} {
+		if !strings.Contains(got, want) {
+			t.Fatalf("prompt missing %q: %s", want, got)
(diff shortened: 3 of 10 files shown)


coder-xiaotian:
We're hitting this exact bug in production on v1.5.0 and would really appreciate getting it merged 🙏

Symptom: Guance (观测云) posts "fatal/severe" alerts into our Feishu group as interactive cards via a
group custom bot. The cards genuinely @ our bot (the receive event's mentions array contains our bot's
open_id), yet cc-connect never reacts — no session created, agent never woken. 6+ such cards with real
@mentions have been silently dropped.

Already ruled out:
- Receiver app has im:message.group_at_msg.include_bot:readonly granted and published.
- The card really does carry the mention (confirmed via API — mentions contains our bot open_id).
- Real-user @mentions (text messages) work fine; only direct `interactive` messages get dropped.

This matches the PR's Problem exactly ("inbound dispatcher ignores direct interactive messages") and
overlaps with #1793.

chenhg5 (approved):
## Conclusion: Approve ✅

This is a high-quality, well-scoped change that fixes a real production bug (interactive alert cards silently dropped from group chats) without regressing existing behavior. Default `inbound_card_mode = "summary"` keeps the wire format unchanged for everyone not explicitly opting in, and the engine wraps raw card data as **explicitly untrusted** content before injecting it into the agent prompt.

---

## Overall assessment

**Ready to merge once the small follow-ups below are addressed (P2/P3 only — not blocking).** The PR is large in line count (+540/-44, 9 files) but almost all of it is plumbing the new `Cards []InboundCard` field through existing dispatch paths. The genuinely new logic is concentrated in three places: `platform/feishu/inbound_card.go`, the new `case "interactive":` branch, and the prompt-builder fragment in `core/engine.go` — all of them testable, security-bounded, and documented.

## Review scope

Reviewed: `core/message.go`, `core/engine.go`, `core/inbound_card_test.go`, `platform/feishu/feishu.go`, `platform/feishu/inbound_card.go`, `platform/feishu/inbound_card_test.go`, `platform/feishu/inbound_card_dispatch_test.go`, `config.example.toml`, `docs/feishu.md`.
Focus: correctness on the new `interactive` dispatch path, security posture of injecting raw card JSON into the agent prompt, backward compatibility (default `summary` mode), test coverage.

## ✅ What looks good

- **Backward compatibility by default.** `inboundCardMode` defaults to `"summary"` and `inboundCardMaxBytes` defaults to 0; the new `Cards` slice stays nil for any platform that doesn't fill it. No platform is forced into the new behavior unless the operator opts in.
- **Security-conscious prompt injection.** `buildInboundCardPrompt` wraps the raw card JSON in a fenced block preceded by `[Feishu interactive card data — untrusted external content]`. That explicit marker is exactly the right framing — downstream agents are far less likely to act on injected instructions. The 256 KiB hard cap (`maxInboundCardPromptBytes`) protects against pathological cards. Good.
- **Real root-cause fix for the silent-drop bug.** Before this PR, `dispatchMessageWithHistory` had no `case "interactive":`, so interactive card events fell through the switch and were dropped. The new branch fetches the raw-card variant via `card_msg_content_type=raw_card_content` (confirmed by `TestDispatchInteractiveFetchesAndPreservesRawCard`), merges any quoted cards from the reply chain, and dispatches a real `core.Message`. This is what unblocks the production scenario from coder-xiaotian's comment.
- **Thread-isolation carve-out for alerts.** `case p.threadIsolation && isThreadSessionKey(sessionKey) && p.inboundCardMode != "summary": quoted = p.fetchQuotedCards(ctx, parentID)` keeps already-engaged threads from re-ingesting ordinary text from the parent chain, but still picks up the root alert card. The comment is precise about why; the new `fetchQuotedCards` helper keeps it minimal.
- **`extractInteractiveCardText` schema handling.** Switching `JsonCard string → json.RawMessage` and trying a nested unmarshal before falling back to the raw bytes correctly handles both the escaped-string wrapper and the plain-object wrapper shapes Feishu uses.
- **`extractCardElements` button handling.** Reading `elem.Behaviors` as a fallback to `elem.Property.Behaviors` and accepting `default_url` covers the Card 2.0 element shape where behaviors live at the top of the element instead of inside `.property`.
- **Test coverage is strong.** Three test files, including:
  - `TestParseInboundCardPreservesRawCard20` — round-trips a real alert payload and asserts the alert_id field survives.
  - `TestDispatchInteractiveFetchesAndPreservesRawCard` — mocks the Feishu API and asserts the `card_msg_content_type=raw_card_content` query parameter is sent.
  - `TestParseMergeForwardPreservesInteractiveChildren` — confirms nested interactive cards survive merge-forward parsing.
  - `TestBuildInboundCardPromptBoundsOversizedPayload` — confirms the safety cap actually triggers.
- **Documentation is precise.** `docs/feishu.md` explicitly calls out the guardrail: "只影响已经通过正常入站门控（群聊仍需真实 @ 机器人）的消息" — important for operators evaluating whether `inbound_card_mode = "raw"` could be abused to bypass group mention requirements. It does not.

## 🟠 Should improve (P2 — not blocking)

- **`inbound_card_max_bytes = 0` defaults to unbounded at the platform layer.** The engine caps the *prompt-side* serialization at 256 KiB, but `parseInboundCard` returns the full raw payload (with `Truncated: false`) when `maxBytes <= 0`. For deployments that flip `inboundCardMode` to `"raw"` or `"both"` without also setting `inbound_card_max_bytes`, an attacker who can post a 1 MB card can have that payload held in memory and JSON-marshalled on every dispatch. Consider:
  - defaulting to ~32 KiB (Feishu's documented card limit is 30 KiB) instead of 0, OR
  - emitting a one-time `slog.Warn` on `newPlatform` when `inboundCardMode != "summary"` but `inboundCardMaxBytes == 0` so operators see the unbounded posture.

- **`case "interactive":` always calls `fetchSingleMessage`.** When `inboundCardMode == "summary"` and the event already contains a usable summary, you still pay an extra API round-trip to fetch the same message back. The dispatch test asserts the `raw_card_content` query is sent — that's correct for `both`/`raw`, but in `summary` mode it's wasted work. Cheap fix: skip the fetch when `inboundCardMode == "summary"`.

- **`buildInboundCardPrompt` truncation message is misleading.** When the safety cap triggers, the output reads `(raw card payload omitted because it exceeds the safety limit)` but the summary line still says `[Feishu interactive card data — untrusted external content]` (without the truncation marker) on the line above. The `TestBuildInboundCardPromptBoundsOversizedPayload` test asserts `(truncated)` is present, but in the actual `buildInboundCardPrompt` source the marker only appears in the *second* sentence, not the framing line. Consider folding `(truncated)` into the framing line so it's visible even with `awk`-style grepping of agent logs.

## 🔵 Optional (P3 — nit)

- `case "interactive":` increments `dispatchCore` flow but doesn't set `OnAccepted`. The `text`/`post` branches don't either (they use the closure-captured one), so this is consistent. Mentioning only because it's worth a comment in the source noting that `OnAccepted` flows through the dispatch closure, not the per-case dispatchCore call.
- `extractInteractiveCardText` is now misnamed for the empty-card case — it falls through to `"[interactive card]"` literally, which is correct but the literal is hardcoded twice (once in the function, once at the dispatcher's empty-text fallback). Consider exporting a `const interactiveCardFallback = "[interactive card]"` to keep them in sync.
- The new `fetchQuotedCards` helper duplicates the `fetchReplyChain(ctx, parentID, maxReplyChainDepth)` call from `fetchQuotedMessage`. A small `fetchReplyChain` refactor to accept a per-message extractor would cut the duplication, but it's a one-liner today so not worth a churn commit.

## ❓ Questions

- For operators on the `summary` (default) path: does the new `case "interactive":` change the `Content` text compared to the old silently-dropped behavior? My reading is "yes, but only for messages that previously fell through" — i.e. the user now sees a `[interactive card]` placeholder instead of nothing. Worth confirming with the author whether any consumer was depending on the *silent* drop, e.g. a noisy-alert filter elsewhere in the stack.
- Is there a documented migration story for users who want to flip `inbound_card_mode` from `summary` to `both` on a long-running production deployment? The dispatcher's prompt-side cost grows linearly with the number of cards in a message; a 1 KiB→256 KiB worst case per dispatched message could matter for memory in chat-busier deployments.

## Testing / Risk

- **CI:** 5/5 green on the head SHA (`lint`, `unit-test`, `smoke-test`, `regression-test`, `performance-test`).
- **Verified evidence I read:** the four new test functions plus `TestUnwrapInboundCardJSONAcceptsPlainAndWrapper`. Existing test surface untouched (`core.Message` struct got one new field, all existing assignments compile).
- **Remaining risk:** the new platform-side `case "interactive":` is in a hot path; please run a soak against a real Feishu group with mixed `text` + `interactive` + `merge_forward` traffic before the next release. The mock-server test covers the dispatch path, but the production Feishu API occasionally returns a slightly different `body.content` schema for older app versions.
- **Security note:** `inbound_card_mode = "raw"` should be opted into only by operators who understand the prompt-injection model. The `untrusted external content` marker is a *frame*, not a guarantee; agents that ignore system framing are still at risk. This is called out in `docs/feishu.md` already — no further action needed.

## Next step

Once the three P2 items above are addressed (either via follow-up commits or comments explaining the intent), this is a clean Approve from my side. I'd recommend doing the `inbound_card_max_bytes` default change in this PR rather than as follow-up — it's a one-line tweak that materially improves the security posture for default configurations.

chenhg5:
Re-reviewed head `86ea8586` (commit `fix(feishu): bound inbound card payloads by default`).

All three P2 follow-ups from my v1 review are addressed in this commit:

1. **`inbound_card_max_bytes` default → 32 KiB** (`defaultInboundCardMaxBytes = 32 * 1024`, with a pin test in `TestNew_DefaultsInboundCardMaxBytes`). Matches Feishu's documented card size limit and removes the unbounded payload risk on default `inbound_card_mode = "raw"` deployments.

2. **`case "interactive":` skips the fetch in summary mode** when the event payload already contains a usable summary. Same fix applied to the merge_forward `interactive` branch. New test `TestDispatchInteractiveSummarySkipsFetchWhenSummaryIsPresent` pins the skip. The expensive API round-trip is gone for the default config.

3. **Truncation marker moved into the framing line:** `[Feishu interactive card data — untrusted external content (truncated)]` is now grep-able as a single token in agent logs.

CI: 5/5 green on the new head. Behavioural diff is strictly additive — no existing test broke, no API surface changed, no platform-level side effects on existing installations. Approval stands. Ready for maintainer merge.

fengnian2025:
Hi @chenhg5, PR #1793 is ready to merge: all CI checks are green, review is approved, and the merge state is clean. My account currently has read-only permission on this repository, so I cannot merge it. Could you please merge it into `main`? Thanks.

chenhg5:
Conclusion: Approve

Overall assessment:
- PR fixes the production bug reported in this comment (https://github.com/chenhg5/cc-connect/pull/1793#issuecomment-5542898882) where Feishu `interactive` alert cards that genuinely @mention the bot were silently dropped. The fix is principled: keep the readable summary for backward compatibility, but optionally preserve raw Card 2.0 JSON as explicitly untrusted, size-bounded data when the operator opts in. CI 5/5 green on the latest head `86ea8586`; the three P2 follow-ups raised by the maintainer's v1 review are addressed (default `inbound_card_max_bytes = 32 KiB`, summary-mode fast path skips the extra fetch, truncation marker moved into the framing line). Independent QA confirmation aligns with maintainer review. Merge-ready.

Review scope:
- 10 files / +601/-44. Focused on `core/message.go` (new `InboundCard` type + `Message.Cards` field), `core/engine.go` (`buildInboundCardPrompt` with safety bound, prompt construction at 3 dispatch sites, `queuedMessage.cards` propagation), `platform/feishu/feishu.go` (config plumbing, `dispatchMessageWithHistory` new `interactive` branch, `quotedMessage.cards`, `fetchQuotedCards` for thread-isolated alerts, `parseMergeForward` recursion into interactive children), `platform/feishu/inbound_card.go` (new `parseInboundCard` + `unwrapInboundCardJSON`), `config.example.toml` and `docs/feishu.md` (option documentation in both Chinese and English).
- Focused on correctness of the three-mode design, security of the untrusted-content framing, thread isolation guarantee, default-mode backward compatibility, and test coverage.

✅ What looks good:
- **Three-mode design (`summary`/`raw`/`both`) with `summary` as the default** keeps the wire format unchanged for every existing deployment — strictly additive. No opt-in user sees any change. The validation rejects unknown modes at config load time (`invalid inbound_card_mode`), not at message dispatch, which is the right place to surface config errors.
- **`maxInboundCardPromptBytes = 256 * 1024` is a hard safety bound at the engine boundary** even if `inbound_card_max_bytes = 0` (unbounded). The framing line `[Feishu interactive card data — untrusted external content]` followed by a fenced ```json block is the right shape — the agent has an explicit signal that the JSON below is data, not instructions. The truncation marker `(truncated)` is grep-able in agent logs.
- **Thread isolation is preserved correctly.** In the new `case` inside `dispatchMessageWithHistory`, when `threadIsolation && isThreadSessionKey && inboundCardMode != "summary"`, only cards are fetched from the parent reply chain (`fetchQuotedCards`), not text/images/files. This protects an already-engaged isolated-thread session from re-ingesting ordinary text that was already handled, while still letting the alert card through — exactly the case `coder-xiaotian` is hitting in production.
- **`defaultInboundCardMaxBytes = 32 * 1024` matches Feishu's documented per-card limit** (https://open.feishu.cn/document/uAjLw4CM/ukzMukzMugybyMxMDMyyTM/card-components/what-is-card-component). For an opt-in `mode = "raw"` deployment, this prevents a single malformed card from injecting 32 KiB×N into the prompt.
- **Summary-mode fast path skips the fetch.** When `inboundCardMode == "summary"` and the event payload already contains usable text, no `fetchSingleMessage` round-trip occurs. New test `TestDispatchInteractiveSummarySkipsFetchWhenSummaryIsPresent` pins this with `fetches == 0` assertion. Important because `mode = "summary"` will be the production default, and Feishu's API rate limits are not free.
- **`unwrapInboundCardJSON` handles both `{"json_card": "<escaped string>"}` and `{"json_card": <raw object>}`** — a robustness fix the old code didn't have. Production Feishu sometimes returns one, sometimes the other depending on card complexity. Pinned by `TestUnwrapInboundCardJSONAcceptsPlainAndWrapper`.
- **`parseInboundCard` returns `(InboundCard, false)` rather than `(zero, true)` when the wrapper is malformed.** This is the right "fail soft" semantics — the caller can still process the readable summary via the old path, just without raw-card fidelity. No new error path that callers must remember to handle.
- **`merge_forward` recursion gains an `interactive` case that fetches the raw card via `fetchSingleMessage` only when summary is missing.** Test `TestParseMergeForwardPreservesInteractiveChildren` verifies the alert_id payload survives recursion. Important because users frequently forward alert cards into a discussion thread, and the bug reporter's case (Guance forwards) lands here.
- **Bot-sender note for buttons.** `extractCardElements` now reads `default_url` as a fallback when `url` is empty for `open_url` behaviors. Feishu changed the default-vs-explicit URL semantics in a recent schema update; this handles the new shape without breaking the old. `behaviorsRaw = elem.Property.Behaviors` with fallback to `elem.Behaviors` also handles both card layouts.
- **`extractInteractiveCardText` gained a Text-content fallback for div/plain_text elements and action elements.** Many real-world alert cards put the visible text under `text.content` rather than the legacy `content` field. This is the conservative additive fix.

🟠 Should improve:
- None that block merge.

🔵 Optional:
- **P3 — `parseInboundCard` returns `Version: "v1"` as the default for any non-2.0 schema.** Looking at the diff, the `Version` field is set but never read anywhere in this PR. If you don't plan to use it, consider dropping it from the struct or leaving it as a TODO. Minor: a malicious payload could claim `schema = "v3"` and we'd silently call it `v1`.
- **P3 — `case "action":` recursion in `extractCardElements`.** Currently `Actions` are unwrapped and recursively processed. If a malformed card has `actions` pointing back to the parent element, this could loop. The current depth cap (`depth > 10` in `formatMergeForwardTree`) doesn't apply here. `extractCardElements` itself has no depth cap, only `for _, raw := range elements`. Risk is very low because Feishu validates action elements server-side, but a defensive `if depth > 8` would be safer. Cost: 2 lines.
- **P3 — `parseInboundCard` in `formatMergeForwardTree` is called unconditionally for every interactive child.** Per the comment, `maxReplyChainDepth = 10`, so worst case ~10 calls per merge-forward. Each one parses JSON twice (in `unwrapInboundCardJSON` + `parseInboundCard`). Not a hot path; flagging only because it's a 2x JSON parse for each card.
- **P3 nit — `unwrapInboundCardJSON` returns `"null"` if `JSONCard` is the JSON literal null.** `parseInboundCard` then validates the unmarshal and bails out with `false`, so no user-visible effect. But the explicit `if wrapper.JSONCard != nil && string(wrapper.JSONCard) != "null"` would be clearer.
- **P3 — new `extractCardElements` fields `Text`, `Actions`, `Behaviors` are `json.RawMessage` even when only used as nested unmarshal targets.** Consistent with `Property.Contents` so it's fine, but a future maintainer might wonder why some are typed and others are raw. The diff already has `Property.Behaviors` as `json.RawMessage` so this is just consistent.
- **P3 — `docs/feishu.md` line added explains `inbound_card_mode = "both"` behavior but does not warn that `raw` mode can cause prompts to balloon.** For a 10-card merge-forward at 32 KiB each, that's 320 KiB injected before `maxInboundCardPromptBytes` cuts it off. A one-line "operators on `mode = raw` should consider also setting `max_bytes`" would help operators.

❓ Questions:
- **Minor: does the existing `processInteractiveMessageWith` -> `processInteractiveEvents` -> `drainPendingMessages` flow respect the new `Cards` field?** The diff updates `buildSenderPrompt(msg.Content, ...)` to `buildSenderPrompt(buildInboundCardPrompt(msg.Content, msg.Cards), ...)` at three sites. The third site is `drainPendingMessages`, which receives `queued` of type `queuedMessage` — I see `cards: append([]InboundCard(nil), msg.Cards...)` in `queueMessageForBusySession`. Looks correct, but worth a confirmation: when an interactive alert card arrives while the session is busy, does the queued replay include the card raw JSON in the eventual prompt? Reading the diff: yes, `queued.cards` is propagated to `prompt` via `buildInboundCardPrompt(queued.content, queued.cards)`. Good.
- **Minor: does `Message.Cards` survive any session-persistence path?** If a session is serialized and resumed, do the cards re-materialize on the new process? Looking at the diff, I don't see Cards being added to any persistence layer. Cards are reconstructed per inbound event. The fallback to `kimiBuiltinModels`-style "rebuild on next message" is fine because the source of truth is Feishu's API (re-fetched if needed). But if the agent is mid-conversation and the source message is deleted from Feishu, the card data would be lost on session resume. This is a soft correctness question, not a blocker.
- **Minor: `case "interactive":` in `dispatchMessageWithHistory` checks `text == "" || text == "[interactive card]"` to decide whether to fetch the raw variant.** What if Feishu's event payload contains a multi-element card whose extracted text is something like `[div] 告警 [button] 查看详情` — not empty, not literal `[interactive card]`, but still a synthetic placeholder that misses the actual alert fields? The fetch is gated on `p.inboundCardMode != "summary" || text == "" || text == "[interactive card]"`. For `mode = "summary"` users (the default), the lossy placeholder is accepted. For `mode = "raw"` users, the fetch happens unconditionally regardless of placeholder quality. So `mode = "both"` does the right thing (fetch + summary), and `mode = "raw"` does the right thing (fetch + placeholder). Just flagging the asymmetry.

Testing / Risk:
- Verified evidence: CI 5/5 GREEN on head `86ea8586` (run 33954668368: lint 1m54s, unit 4m48s, smoke 35s, regression 29s, performance 55s — all pass). 3 commits iterated: v1 → lint-fix → bound-by-default. Maintainer self-approved at v1, then re-reviewed after the P2 follow-up commit and confirmed "Approval stands." Two integration tests with `httptest.NewServer` exercise the real Feishu SDK client (`lark.NewClient`) against a stubbed API: `TestDispatchInteractiveFetchesAndPreservesRawCard` verifies the round-trip including raw-card query parameter, `TestDispatchInteractiveSummarySkipsFetchWhenSummaryIsPresent` verifies the fast path, `TestParseMergeForwardPreservesInteractiveChildren` verifies recursion. Unit tests cover parser, unwrapper, default-bytes fallback, and prompt-builder truncation.
- Remaining risk: thread-isolation `fetchQuotedCards` is **new code without a dedicated integration test**. The existing thread-isolation tests exercise text-quote paths but not card-only paths. The change is small (3 lines after the `case` branch) and the existing `dispatchMessageWithHistory` flow is well-trodden, but a regression test asserting "in an isolated thread, a card-only parent quote does not leak text" would close the gap. Suggested as follow-up.
- Compatibility: zero behavioral change for `mode = "summary"` (default). For `mode = "raw"` and `mode = "both"`, the only change is that agents now receive Card 2.0 JSON in their prompt — explicitly framed as untrusted — which they should already be handling defensively per existing prompt-injection guidance.
- Security: prompt-injection risk is acknowledged in the framing line ("untrusted external content"). The hard cap of `maxInboundCardPromptBytes = 256 * 1024` prevents a malformed card from blowing the prompt budget. `parseInboundCard` does not execute or interpret card content — only preserves it. The merge-forward recursion respects the existing depth cap. `default_url` and `behaviors` fields are read but never fetched. **No new code-execution or fetch paths are introduced.**
- Backward compat: existing `extractInteractiveCardText` callers still work — the new code paths are additive. Existing `Message.Content` text is preserved verbatim when `mode = "summary"`. Existing `quotedMessage.text` is preserved. No fields removed.

Next step:
- Maintainer (merge) the PR — independent QA sign-off aligns with maintainer self-approve. After merge, watch for any follow-up issues from `coder-xiaotian` or other Guance-style integrations. If `mode = "both"` becomes the new de facto default in operator config templates, document the recommendation in the changelog.

coder-xiaotian:
Sorry to follow up again — just wondering if there's any rough timeline for merging this one. We're still on v1.5.0 in production and continue to hit the original issue (interactive alert cards that @ the bot get dropped), so a merge whenever you have a chance would mean a lot. Thanks for your work on this.

fengnian2025:
My account currently has read-only permission on this repository, so I cannot merge it. Could you please merge it into main? Thanks. @coder-xiaotian

coder-xiaotian:
@chenhg5 Hi — sorry to bother you directly. Just wondering if there's any chance of getting this merged, or whether something's holding it back? It's been approved with CI green for a while, and we're still blocked in production by the original bug. Thanks!

PR #1796, opened 2026-09-06
fix(kimi): probe installed CLI for model list and refresh fallback (#1795)

### LINKED ISSUE #1795: kimi: /model picker fallback list is stale — K3-era models not selectable
## What

`agent/kimi`'s `AvailableModels()` falls back to a hardcoded list frozen at the K2/K2.5 era when no provider models are configured:

```go
return []core.ModelOption{
    {Name: "kimi-k2-0711-preview", Desc: "Kimi K2 (most capable)"},
    {Name: "kimi-k2-0711", Desc: "Kimi K2"},
    {Name: "kimi-k2-5-preview", Desc: "Kimi K2.5 (balanced)"},
    {Name: "kimi-k2-5", Desc: "Kimi K2.5"},
}
```

Current `kimi-code` CLI (0.38.x) ships newer model aliases (e.g. `k3`, `k3-256k`, plus K2.6/K2.7 family per the CLI binary), but they never appear in the `/model` picker on IM platforms.

## Impact

- Users on a current Kimi Code CLI see only K2/K2.5 entries in `/model`. Picking one silently **downgrades** from the CLI default (e.g. `kimi-code/k3` is the default model in `kimi-code` 0.38.x).
- Same class of bug as #1642 (claudecode picker fallback stale for Claude 5 family).

## Repro

1. cc-connect v1.5.0 (npm), kimi-code 0.38.0, macOS arm64
2. Project with `[projects.agent] type = "kimi"`, no providers configured
3. Send `/model` in Feishu → only the 4 hardcoded entries appear

## Suggestion

Either refresh the hardcoded fallback list, or — more durable — extend the existing startup probe (`probeKimiFlags`) to also discover the models registered in the installed CLI, so the picker tracks the local CLI instead of drifting every model release.

## Workaround

Leave `model` unset in `[projects.agent.options]` so the CLI default applies (already K3 on current CLI), instead of picking from the stale list.

chenhg5:
Fixes #1795.

## What

The /model picker in `agent/kimi` fell back to a hardcoded K2/K2.5 list when no provider was configured. On kimi-code CLI 0.38.x (current default) that list no longer includes K3 / K3-256k / K2.6 / K2.7, so users picking from IM `/model` silently downgrade from the CLI default (k3) to kimi-k2-5.

## Changes

1. **`probeKimiModels`** at Agent construction runs `kimi model list` (with `kimi models` / `kimi models list` as fallback candidates) and parses the result into a `ModelOption` list. When the probe succeeds the /model picker follows the installed CLI instead of drifting every release.
2. **`kimiBuiltinModels`** (hardcoded fallback) is refreshed to put k3 / k3-256k first, with K2.5 / K2 kept at the bottom for backward compatibility. Even on CLIs whose probe output we can't parse yet, users see current options.
3. **`AvailableModels` resolution order**: configured provider models → discovered (probe result) → kimiBuiltinModels. Empty provider `Models` does NOT shadow the probe.

## Parser robustness

The parser requires each candidate identifier to contain at least a digit, dash, or slash — every real Kimi model name ships with one of those, but CLI header / footer text ("Available models:", "Run 'kimi help' …", "switch.") doesn't, so header lines are silently dropped instead of polluting the picker.

## Tests

- `parseKimiModelList`: one-id-per-line, id+description, header/footer skip, dedup, namespaced aliases, punctuation noise, comments
- `probeKimiModels`: missing binary, stub helper with parseable output, second-candidate fallback, header-only rejection
- `kimiBuiltinModels`: refreshed with k3, backward compat with K2/K2.5, newest-first ordering
- `AvailableModels`: prefer-provider, prefer-probe, fallback-when-probe-failed, provider-empty-does-not-shadow, race-free accessor (run with -race)

## Acceptance criteria (from issue)

- [x] kimi-code 0.38.x installed → /model picker surfaces k3 / k3-256k / K2.6 / K2.7 (via probe; refreshed fallback also includes them)
- [x] kimi-code unreachable / older version → falls back to refreshed hardcoded list (was: stale K2/K2.5 only)
- [x] Existing IM /model flow UI: no breaking change — same picker surface, just more options
- [x] CI: local `go test -race ./agent/kimi/` passes; awaiting GitHub CI 5/5

## Workaround (preserved)

Users can still leave `[projects.agent.options].model` empty to defer to the CLI default; this fix doesn't change that path.

## Related

- Same class of bug as #1642 (claudecode picker stale Claude 5 family) — same probe-on-startup pattern
- Maintainer response on #1795 explicitly suggested extending `probeKimiFlags`; this PR follows that suggestion
- `probeKimiFlags` (the existing flag probe) is untouched

### DIFF
--- a/agent/kimi/kimi.go
+++ b/agent/kimi/kimi.go
@@ -50,7 +50,14 @@ type Agent struct {
 	activeIdx    int // -1 = no provider set
 	sessionEnv   []string
 	flagSupport  kimiFlagSupport // detected once at New() via `kimi --help`
-	mu           sync.RWMutex
+	discovered   []core.ModelOption
+	// discovered holds the model identifiers the startup probe learned
+	// from the installed kimi-code CLI (see probeKimiModels, #1795). Nil
+	// when the probe failed (CLI missing, older kimi-cli dialect without
+	// the `model list` subcommand, or unparseable output); in that case
+	// AvailableModels falls back to kimiBuiltinModels so users on older
+	// CLIs keep seeing *some* picker options instead of an empty list.
+	mu sync.RWMutex
 }
 
 func New(opts map[string]any) (core.Agent, error) {
@@ -90,6 +97,13 @@ func New(opts map[string]any) (core.Agent, error) {
 	// back to assuming the modern CLI (no --print).
 	flagSupport := probeKimiFlags(context.Background(), cmd, 5*time.Second)
 
+	// Also probe the locally installed CLI for its currently-registered
+	// model identifiers so the /model picker tracks the installed CLI
+	// instead of drifting every release (#1795). When the probe returns
+	// nothing (older kimi-cli dialect, binary unreachable, unparseable
+	// output) AvailableModels falls back to kimiBuiltinModels.
+	discovered := probeKimiModels(context.Background(), cmd, 5*time.Second)
+
 	return &Agent{
 		workDir:      workDir,
 		model:        model,
@@ -100,6 +114,7 @@ func New(opts map[string]any) (core.Agent, error) {
 		timeout:      timeout,
 		activeIdx:    -1,
 		flagSupport:  flagSupport,
+		discovered:   discovered,
 	}, nil
 }
 
@@ -152,15 +167,52 @@ func (a *Agent) configuredModels() []core.ModelOption {
 	return core.GetProviderModels(a.providers, a.activeIdx)
 }
 
+func (a *Agent) discoveredModels() []core.ModelOption {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
+	return a.discovered
+}
+
+// AvailableModels returns the model options for the /model picker.
+//
+// Resolution order (#1795):
+//  1. If a provider is configured with explicit models, use those.
+//  2. Otherwise, prefer the startup-probe result so the picker follows the
+//     installed kimi-code CLI instead of drifting every release. The probe
+//     is opportunistic — it tries `kimi model list`, `kimi models`, and
+//     `kimi models list`, and the parser is tolerant of various output
+//     layouts so future CLI tweaks don't silently empty the picker.
+//  3. Finally, fall back to kimiBuiltinModels. The list is refreshed for
+//     every cc-connect release (latest known models first, oldest kept for
+//     backward compatibility) so even users on CLIs that don't expose the
+//     `model list` subcommand still see current options.
 func (a *Agent) AvailableModels(ctx context.Context) []core.ModelOption {
 	if models := a.configuredModels(); len(models) > 0 {
 		return models
 	}
+	if models := a.discoveredModels(); len(models) > 0 {
+		return models
+	}
+	return kimiBuiltinModels()
+}
+
+// kimiBuiltinModels returns the static fallback list used when no provider
+// is configured AND the startup probe could not discover installed models
+// (older kimi-cli dialect, CLI unreachable, unparseable probe output, …).
+//
+// The list is ordered newest-first and includes legacy K2 / K2.5 entries at
+// the bottom for users still on those CLIs — refreshing the *entire* list
+// to only new models would silently break /model selection for anyone who
+// hasn't upgraded yet, while keeping the old list frozen at K2/K2.5 is the
+// bug #1795 is reporting. See #1795 for context.
+func kimiBuiltinModels() []core.ModelOption {
 	return []core.ModelOption{
-		{Name: "kimi-k2-0711-preview", Desc: "Kimi K2 (most capable)"},
-		{Name: "kimi-k2-0711", Desc: "Kimi K2"},
+		{Name: "k3", Desc: "Kimi K3 (current default)"},
+		{Name: "k3-256k", Desc: "Kimi K3 256k (long context)"},
 		{Name: "kimi-k2-5-preview", Desc: "Kimi K2.5 (balanced)"},
 		{Name: "kimi-k2-5", Desc: "Kimi K2.5"},
+		{Name: "kimi-k2-0711-preview", Desc: "Kimi K2 (legacy)"},
+		{Name: "kimi-k2-0711", Desc: "Kimi K2"},
 	}
 }
 
--- a/agent/kimi/probe.go
+++ b/agent/kimi/probe.go
@@ -5,8 +5,11 @@ import (
 	"context"
 	"log/slog"
 	"os/exec"
+	"regexp"
 	"strings"
 	"time"
+
+	"github.com/chenhg5/cc-connect/core"
 )
 
 // kimiFlagSupport records which optional CLI flags the locally installed Kimi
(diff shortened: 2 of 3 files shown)


chenhg5 (commented):
Conclusion: Approve

Overall assessment:
- PR addresses the #1795 picker-drift issue with a sound three-tier resolution (provider → probe → built-in) and pins every branch with tests, including a `-race` test for the new `discovered` field accessor. CI 5/5 green. The fix follows the same probe-on-startup pattern as the recently-merged `probeKimiFlags` (related #1642 fix) so the codebase now has a consistent way to detect kimi-code CLI capabilities at construction time. The implementation is conservative — the probe is opportunistic and silently degrades to the built-in list, so users on legacy CLIs see at least the refreshed fallback instead of an empty picker. Merge-ready.

Review scope:
- 3 files / +639/-3, focused on `agent/kimi/kimi.go` (Agent struct + `New` + `AvailableModels` + new `discoveredModels` accessor + new `kimiBuiltinModels` body), `agent/kimi/probe.go` (new `probeKimiModels` / `runKimiModelProbe` / `parseKimiModelList` / `looksLikeModelIdentifier` / `isModelListStopWord` + 2 regex patterns), `agent/kimi/probe_models_test.go` (new 403-line test file with 14 test functions).
- Focused on correctness of the resolution order, probe tolerance to missing/unparseable/header-only output, lock discipline for the new `discovered` field, race-condition coverage, and the parser's tolerance to future CLI layout tweaks.

✅ What looks good:
- **Three-tier resolution order is correctly implemented**: configured provider models (when non-empty) → discovered probe result → `kimiBuiltinModels`. The `TestAgentAvailableModels_ProviderWithEmptyModelsDoesNotShadowProbe` test pins the subtle case where a provider has `Models: nil` (user set only APIKey/BaseURL) — without that test, `a.configuredModels()` could return `nil` from `core.GetProviderModels` with `activeIdx >= 0`, and a future refactor might accidentally start shadowing the probe result. The current `len(models) > 0` guard is the right defensive check, and it's pinned.
- **Probe candidate list is correctly ordered by likelihood**: `["model", "list"]` first (most common in modern kimi-code), then `["models"]`, then `["models", "list"]` as last resort. The probe stops on the first candidate that returns a non-empty parse, which matches the principle of "don't make the user wait longer than necessary". `TestProbeKimiModels_StubHelperScript_FallsBackToSecondCandidate` covers exactly the second-candidate path with a stub helper script — a nice integration-style test that also exercises `exec.LookPath` resolution via `PATH`.
- **Tolerant parser is genuinely tolerant**: drops lines starting with `#` or `Usage` (covers both shell-style comments and common CLI help banners), skips blank lines, deduplicates by Name, and uses a two-stage identifier filter (`modelIdentRe` for character class + `modelIdentMustContainRe` for the digit/dash/slash disambiguator that English header words don't have). `TestProbeKimiModels_StubEmitsOnlyHeader` pins the "header-only output must NOT be promoted to a model list" guarantee — this is the regression test that protects against a future CLI version emitting "Available models:" without identifiers, which would otherwise yield a phantom model named "Available".
- **The `discovered` field is properly guarded by `a.mu`**: `discoveredModels()` uses `RLock`, the field is only written once in `New()` (no concurrent write surface in this PR), and `TestAgentDiscoveredModels_RaceFree` validates the read-side race against `SetWorkDir`/`SetMode` writes. The note in the test comment ("Run with -race") correctly instructs future maintainers.
- **`probeKimiFlags` (the existing flag probe) is explicitly left untouched** per the PR body's "Related" section. The PR doesn't conflate flag-probe failure with model-probe failure — they remain independent, which is important because a CLI can have one capability without the other.
- **`kimiBuiltinModels` ordering is correct**: k3/k3-256k first (current default + long-context), K2.5 in the middle (balanced / stable), K2/K2-preview at the bottom (legacy). `TestKimiBuiltinModels_RefreshedWithK3` pins both presence (all expected names) and order (k3 before k2-5 before k2-0711). The desc strings ("current default", "long context", "legacy") communicate the freshness intent at picker-display time.
- **Empty-models fallback is intentional and tested**: `TestAgentAvailableModels_FallbackToBuiltinWhenProbeFailed` ensures the picker is never empty even when both the configured-models and probe paths yield nothing. The PR's resolved acceptance criteria match this — even on CLIs whose `model list` output we can't parse, the user sees current options.
- **5s timeout for the probe matches `probeKimiFlags`**: the existing flag probe uses 5s; this PR uses the same. The `if timeout <= 0` defensive default is a small but correct touch — call sites that pass 0 (or negative) get the documented 5s instead of an immediate zero-timeout no-op.

🟠 Should improve:
- None that block merge.

🔵 Optional:
- **P3 — comment inaccuracy in `looksLikeModelIdentifier` (probe.go:270-273).** The comment says: *"requires at least one digit, dot, underscore, dash, or slash inside the token"* — but the actual regex on line 197 is `[0-9/\-]`, which excludes dot and underscore. The earlier comment on line 195-196 ("Dot and underscore alone are not sufficient because they slip into sentences ending in '.' or fields like 'see_help'") is consistent with the regex, but the line 270 list is wrong. Either trim line 270 to match the regex ("digit, dash, or slash") or expand the regex to match the comment ("digit, dot, underscore, dash, or slash"). The current implementation is correct (the regex is the safer, more restrictive choice); only the comment needs updating. Cost: a 1-line edit.
- **P3 nit — `probeKimiModels` returns nil instead of `[]core.ModelOption{}` when nothing parses.** This is the correct return convention (length-based checks work uniformly), and the existing `TestProbeKimiModels_StubEmitsOnlyHeader` correctly uses `assert.Nil`. No action needed; flagging for future maintainers who might be tempted to "normalize" nil to an empty slice.
- **P3 nit — `kimiBuiltinModels()` is called as a function on every `AvailableModels` invocation.** The list never changes, so it could be a package-level `var`. Tiny allocation cost (the 6-element slice is rebuilt on every call when the probe path is taken) — not worth fixing, but worth noting if the codebase ever profiles this hot path.
- **P3 — the `-race` test only covers `discoveredModels()` vs `SetWorkDir`/`SetMode`.** It does NOT cover concurrent calls to `probeKimiModels` during `New()` (impossible in practice — `New` is single-threaded), nor does it cover concurrent `AvailableModels` calls from multiple goroutines. Today this is safe because `AvailableModels` only reads via `configuredModels` and `discoveredModels`, both of which use `RLock`. The existing race test is sufficient; flagging just so a future refactor doesn't accidentally drop the lock.
- **P3 — typo watch: `kimi-cli` vs `kimi-code`.** PR body uses "kimi-code CLI" consistently, but the comment in `probeKimiModels` mentions "older kimi-cli dialect" (#1456). These appear to be the same project across different eras of branding. If the official name is now "kimi-code", the historical comment is fine as-is; if maintainers want strict consistency, "older kimi-code build (pre-0.38)" would be clearer.

❓ Questions:
- **Minor: does `probeKimiModels` need to run before `probeKimiFlags` or vice versa?** Today they're independent — `probeKimiFlags` runs first in `New()` (line 99), then `probeKimiModels` (line 127). If both probes time out at 5s, that's 10s added to `New()`. Worth confirming whether `probeKimiFlags` results influence whether to attempt `probeKimiModels` (e.g. skip if the CLI doesn't even respond to `--help` — but that would also mean the binary is broken, so probably not worth optimizing). Today's order is fine; just flagging the additive latency.
- **Minor: the PR says "User can still leave `[projects.agent.options].model` empty to defer to the CLI default".** This is unchanged behavior. But does the CLI default propagate to `a.model` (the per-session model selection), or only to the picker? If the latter, the picker might say "k3" but a session still picks up the prior `model` from the config. A quick read of `core.GetProviderModels` would confirm whether the picker result feeds back into the default selection — likely yes, but worth verifying.

Testing / Risk:
- Verified evidence: CI 5/5 green (run 34001231884: lint 2m18s, unit 4m52s, smoke 28s, regression 38s, performance 48s — all pass). 14 new test functions covering parser layout variants, probe happy/failure paths, candidate fallback, header-only rejection, builtin refresh + ordering, resolution-order precedence, empty-provider semantics, and a `-race` test for the new shared field. Test file is 403 lines with descriptive comments that read as documentation as much as assertions.
- Remaining risk: the `modelIdentRe` regex has a 64-char body cap (`{0,63}`). If a future Kimi CLI ever emits a model identifier longer than 64 chars (unlikely for current Kimi naming, but possible for provider-prefixed names like `kimi-code/very-long-provider-name/k3-256k-preview`), it will be silently dropped. The cap is documented in the regex comment as a "reasonable length cap" — fine for now, but worth a TODO for the future if Kimi CLI starts emitting longer IDs.
- Compatibility: zero behavioral change for users with explicit provider configs (their list always won). For users without providers on kimi-code 0.38+, the picker now surfaces the current model set instead of the stale K2/K2.5 list — this is the intended fix. For users on older kimi-cli builds, the probe returns nil and the refreshed `kimiBuiltinModels` is used — slight list shuffle vs. the prior list (k3/k3-256k promoted to top), but no removals (K2/K2.5/K2-preview all still present). Backward compatible.
- Security: the probe runs `exec.CommandContext(ctx, cmd, args...)` against the configured `cmd` path, with a 5s timeout, and parses only regex-matched identifiers out of stdout/stderr. No shell expansion, no env injection (args are passed as separate argv to exec.Command, not concatenated into a shell string). The `cmd` field is itself set from the kimi agent's config — already a trusted surface. No trust boundary is crossed that wasn't already crossed.

Next step:
- Author (maintainer chenhg5) can merge as-is, or address the optional P3 comment inaccuracy in `looksLikeModelIdentifier`. Either path is non-blocking. Approval stands. Ready for maintainer merge gate.

chenhg5:
Conclusion: Approve

Overall assessment:
- PR addresses the #1795 picker-drift issue with a sound three-tier resolution (provider → probe → built-in) and pins every branch with tests, including a `-race` test for the new `discovered` field accessor. CI 5/5 green. The fix follows the same probe-on-startup pattern as the recently-merged `probeKimiFlags` (related #1642 fix) so the codebase now has a consistent way to detect kimi-code CLI capabilities at construction time. The implementation is conservative — the probe is opportunistic and silently degrades to the built-in list, so users on legacy CLIs see at least the refreshed fallback instead of an empty picker. Merge-ready.

Review scope:
- 3 files / +639/-3, focused on `agent/kimi/kimi.go` (Agent struct + `New` + `AvailableModels` + new `discoveredModels` accessor + new `kimiBuiltinModels` body), `agent/kimi/probe.go` (new `probeKimiModels` / `runKimiModelProbe` / `parseKimiModelList` / `looksLikeModelIdentifier` / `isModelListStopWord` + 2 regex patterns), `agent/kimi/probe_models_test.go` (new 403-line test file with 14 test functions).
- Focused on correctness of the resolution order, probe tolerance to missing/unparseable/header-only output, lock discipline for the new `discovered` field, race-condition coverage, and the parser's tolerance to future CLI layout tweaks.

✅ What looks good:
- **Three-tier resolution order is correctly implemented**: configured provider models (when non-empty) → discovered probe result → `kimiBuiltinModels`. The `TestAgentAvailableModels_ProviderWithEmptyModelsDoesNotShadowProbe` test pins the subtle case where a provider has `Models: nil` (user set only APIKey/BaseURL) — without that test, `a.configuredModels()` could return `nil` from `core.GetProviderModels` with `activeIdx >= 0`, and a future refactor might accidentally start shadowing the probe result. The current `len(models) > 0` guard is the right defensive check, and it's pinned.
- **Probe candidate list is correctly ordered by likelihood**: `["model", "list"]` first (most common in modern kimi-code), then `["models"]`, then `["models", "list"]` as last resort. The probe stops on the first candidate that returns a non-empty parse, which matches the principle of "don't make the user wait longer than necessary". `TestProbeKimiModels_StubHelperScript_FallsBackToSecondCandidate` covers exactly the second-candidate path with a stub helper script — a nice integration-style test that also exercises `exec.LookPath` resolution via `PATH`.
- **Tolerant parser is genuinely tolerant**: drops lines starting with `#` or `Usage` (covers both shell-style comments and common CLI help banners), skips blank lines, deduplicates by Name, and uses a two-stage identifier filter (`modelIdentRe` for character class + `modelIdentMustContainRe` for the digit/dash/slash disambiguator that English header words don't have). `TestProbeKimiModels_StubEmitsOnlyHeader` pins the "header-only output must NOT be promoted to a model list" guarantee — this is the regression test that protects against a future CLI version emitting "Available models:" without identifiers, which would otherwise yield a phantom model named "Available".
- **The `discovered` field is properly guarded by `a.mu`**: `discoveredModels()` uses `RLock`, the field is only written once in `New()` (no concurrent write surface in this PR), and `TestAgentDiscoveredModels_RaceFree` validates the read-side race against `SetWorkDir`/`SetMode` writes. The note in the test comment ("Run with -race") correctly instructs future maintainers.
- **`probeKimiFlags` (the existing flag probe) is explicitly left untouched** per the PR body's "Related" section. The PR doesn't conflate flag-probe failure with model-probe failure — they remain independent, which is important because a CLI can have one capability without the other.
- **`kimiBuiltinModels` ordering is correct**: k3/k3-256k first (current default + long-context), K2.5 in the middle (balanced / stable), K2/K2-preview at the bottom (legacy). `TestKimiBuiltinModels_RefreshedWithK3` pins both presence (all expected names) and order (k3 before k2-5 before k2-0711). The desc strings ("current default", "long context", "legacy") communicate the freshness intent at picker-display time.
- **Empty-models fallback is intentional and tested**: `TestAgentAvailableModels_FallbackToBuiltinWhenProbeFailed` ensures the picker is never empty even when both the configured-models and probe paths yield nothing. The PR's resolved acceptance criteria match this — even on CLIs whose `model list` output we can't parse, the user sees current options.
- **5s timeout for the probe matches `probeKimiFlags`**: the existing flag probe uses 5s; this PR uses the same. The `if timeout <= 0` defensive default is a small but correct touch — call sites that pass 0 (or negative) get the documented 5s instead of an immediate zero-timeout no-op.

🟠 Should improve:
- None that block merge.

🔵 Optional:
- **P3 — comment inaccuracy in `looksLikeModelIdentifier` (probe.go:270-273).** The comment says: *"requires at least one digit, dot, underscore, dash, or slash inside the token"* — but the actual regex on line 197 is `[0-9/\-]`, which excludes dot and underscore. The earlier comment on line 195-196 ("Dot and underscore alone are not sufficient because they slip into sentences ending in '.' or fields like 'see_help'") is consistent with the regex, but the line 270 list is wrong. Either trim line 270 to match the regex ("digit, dash, or slash") or expand the regex to match the comment ("digit, dot, underscore, dash, or slash"). The current implementation is correct (the regex is the safer, more restrictive choice); only the comment needs updating. Cost: a 1-line edit.
- **P3 nit — `probeKimiModels` returns nil instead of `[]core.ModelOption{}` when nothing parses.** This is the correct return convention (length-based checks work uniformly), and the existing `TestProbeKimiModels_StubEmitsOnlyHeader` correctly uses `assert.Nil`. No action needed; flagging for future maintainers who might be tempted to "normalize" nil to an empty slice.
- **P3 nit — `kimiBuiltinModels()` is called as a function on every `AvailableModels` invocation.** The list never changes, so it could be a package-level `var`. Tiny allocation cost (the 6-element slice is rebuilt on every call when the probe path is taken) — not worth fixing, but worth noting if the codebase ever profiles this hot path.
- **P3 — the `-race` test only covers `discoveredModels()` vs `SetWorkDir`/`SetMode`.** It does NOT cover concurrent calls to `probeKimiModels` during `New()` (impossible in practice — `New` is single-threaded), nor does it cover concurrent `AvailableModels` calls from multiple goroutines. Today this is safe because `AvailableModels` only reads via `configuredModels` and `discoveredModels`, both of which use `RLock`. The existing race test is sufficient; flagging just so a future refactor doesn't accidentally drop the lock.
- **P3 — typo watch: `kimi-cli` vs `kimi-code`.** PR body uses "kimi-code CLI" consistently, but the comment in `probeKimiModels` mentions "older kimi-cli dialect" (#1456). These appear to be the same project across different eras of branding. If the official name is now "kimi-code", the historical comment is fine as-is; if maintainers want strict consistency, "older kimi-code build (pre-0.38)" would be clearer.

❓ Questions:
- **Minor: does `probeKimiModels` need to run before `probeKimiFlags` or vice versa?** Today they're independent — `probeKimiFlags` runs first in `New()` (line 99), then `probeKimiModels` (line 127). If both probes time out at 5s, that's 10s added to `New()`. Worth confirming whether `probeKimiFlags` results influence whether to attempt `probeKimiModels` (e.g. skip if the CLI doesn't even respond to `--help` — but that would also mean the binary is broken, so probably not worth optimizing). Today's order is fine; just flagging the additive latency.
- **Minor: the PR says "User can still leave `[projects.agent.options].model` empty to defer to the CLI default".** This is unchanged behavior. But does the CLI default propagate to `a.model` (the per-session model selection), or only to the picker? If the latter, the picker might say "k3" but a session still picks up the prior `model` from the config. A quick read of `core.GetProviderModels` would confirm whether the picker result feeds back into the default selection — likely yes, but worth verifying.

Testing / Risk:
- Verified evidence: CI 5/5 green (run 34001231884: lint 2m18s, unit 4m52s, smoke 28s, regression 38s, performance 48s — all pass). 14 new test functions covering parser layout variants, probe happy/failure paths, candidate fallback, header-only rejection, builtin refresh + ordering, resolution-order precedence, empty-provider semantics, and a `-race` test for the new shared field. Test file is 403 lines with descriptive comments that read as documentation as much as assertions.
- Remaining risk: the `modelIdentRe` regex has a 64-char body cap (`{0,63}`). If a future Kimi CLI ever emits a model identifier longer than 64 chars (unlikely for current Kimi naming, but possible for provider-prefixed names like `kimi-code/very-long-provider-name/k3-256k-preview`), it will be silently dropped. The cap is documented in the regex comment as a "reasonable length cap" — fine for now, but worth a TODO for the future if Kimi CLI starts emitting longer IDs.
- Compatibility: zero behavioral change for users with explicit provider configs (their list always won). For users without providers on kimi-code 0.38+, the picker now surfaces the current model set instead of the stale K2/K2.5 list — this is the intended fix. For users on older kimi-cli builds, the probe returns nil and the refreshed `kimiBuiltinModels` is used — slight list shuffle vs. the prior list (k3/k3-256k promoted to top), but no removals (K2/K2.5/K2-preview all still present). Backward compatible.
- Security: the probe runs `exec.CommandContext(ctx, cmd, args...)` against the configured `cmd` path, with a 5s timeout, and parses only regex-matched identifiers out of stdout/stderr. No shell expansion, no env injection (args are passed as separate argv to exec.Command, not concatenated into a shell string). The `cmd` field is itself set from the kimi agent's config — already a trusted surface. No trust boundary is crossed that wasn't already crossed.

Next step:
- Author (maintainer chenhg5) can merge as-is, or address the optional P3 comment inaccuracy in `looksLikeModelIdentifier`. Either path is non-blocking. Approval stands. Ready for maintainer merge gate.

totoanew:
Real-environment verification of this PR against an actual kimi-code 0.38.0 install (macOS arm64, cc-connect user from #1795).

## ✅ What works

- `go test ./agent/kimi/...` — all pass (Go 1.25.0, darwin/arm64).
- Resolution order verified end-to-end by constructing the agent against the real binary (`cmd = ~/.kimi-code/bin/kimi`):
  - `probeKimiModels` returns 0 entries (see finding A), and
  - `AvailableModels` correctly falls back to the refreshed builtin list → the picker now shows `k3`, `k3-256k`, `kimi-k2-5-preview`, `kimi-k2-5`, `kimi-k2-0711-preview`, `kimi-k2-0711`. The #1795 symptom (K3 invisible) is gone.

## ⚠️ Finding A: the probe subcommands don't exist in kimi-code 0.38.0

```
$ kimi model list   →  unknown command 'model'. See 'kimi --help'.
$ kimi models       →  unknown command 'models'. See 'kimi --help'.
```

Full subcommand list of 0.38.0: `export, provider, acp, web, server(deprecated), login, doctor, vis, migrate, upgrade`. So on current CLI the dynamic probe always returns nil and the picker is effectively driven by the builtin list. The fallback design makes this safe, but the probe code is currently inert on the exact CLI generation this PR targets.

A working discovery source that exists today:

```
$ kimi provider list --json
{
  "providers": { "managed:kimi-code": {...} },
  "models": {
    "kimi-code/kimi-for-coding":          { "model": "kimi-for-coding",          "displayName": "K2.7 Coding",          "maxContextSize": 262144, ... },
    "kimi-code/kimi-for-coding-highspeed": { "model": "kimi-for-coding-highspeed", "displayName": "K2.7 Coding Highspeed", ... },
    "kimi-code/k3":                        { ... },
    "kimi-code/k3-256k":                   { ... }
  }
}
```

Parsing this table would give the picker real, current models (including display names) instead of relying on a static list.

## ⚠️ Finding B: the new builtin names `k3` / `k3-256k` don't match the managed CLI's alias scheme — and selecting one crashes the CLI

kimi-code 0.38.0 registers aliases as `<provider>/<model>` (`kimi-code/k3`, `kimi-code/k3-256k`). Bare names are rejected, and ungracefully:

```
$ kimi -m k3 -p 'ok'
kimi version 0.38.0
Error: Agent event 'agent.activity.updated' has no active lifecycle context
(exit 1)

$ kimi -m kimi-code/k3 -p 'ok'      → works
$ kimi -p 'ok'  (no -m)             → works
```

So if a user picks `k3` from the refreshed /model picker, cc-connect will pass `--model k3` and the spawned CLI crashes before producing output. Suggest either using the full `kimi-code/`-prefixed aliases in the builtin list, or probing the alias table from `provider list --json` (finding A) so names are always valid for the installed CLI.

(The ungraceful crash on an unknown alias is a kimi-code-side bug, not cc-connect's — happy to file it upstream if useful.)

Net: the PR fixes the reported symptom via the refreshed fallback, tests are green, and with the two findings addressed (probe source + alias naming) the dynamic path should actually light up. Thanks for the quick turnaround!

Issue #1618, opened 2026-07-30
v1.4.1: bot open_id lookup failure at startup silently disables group mention filtering for the whole process lifetime (fail-open) — bot replies to messages that never mentioned it


jakever:
> **⚠️ This report has been rewritten (2026-07-30). My original diagnosis was wrong.**
> I first reported this as "the daemon silently degrades, stops writing logs, and the mention filter breaks mid-run." After digging through the full log I found the actual cause, and two of my original observations do not hold up. The corrected analysis is below; see the comment thread for what I retracted.

## Root cause

At startup, the Lark platform adapter calls `POST https://open.larksuite.com/open-apis/auth/v3/tenant_access_token/internal` to resolve the bot's own `open_id` — which is what the group-chat mention filter compares incoming mentions against.

**If that call fails, cc-connect logs a WARN and disables group-chat mention filtering entirely, then continues to serve.** This is a fail-open: from that moment until the process is restarted, *every* group message is treated as if it were addressed to the bot, whether or not it mentions the bot.

On my machine the shell environment has `HTTP_PROXY/HTTPS_PROXY=127.0.0.1:7897`. If the daemon happens to start during a window where that local proxy isn't up yet, the token call fails with `connection refused` and the filter is disabled for the whole lifetime of that process:

```
time=2026-07-30T13:33:04.705+08:00 level=WARN
  msg="lark: failed to get bot open_id, group chat filtering disabled"
  error="api call: Post \"https://open.larksuite.com/open-apis/auth/v3/tenant_access_token/internal\":
         proxyconnect tcp: dial tcp 127.0.0.1:7897: connect: connection refused"
```

The proxy is my local environment's problem, not cc-connect's. **What I'd like to see changed is the fail-open decision** — losing the bot's own identity should not silently widen the set of messages the bot answers.

## Evidence

A successful startup logs `lark: bot identified`. Across my whole log file that line appears **51 times** (once per successful start); the failure WARN appears **twice ever** — `2026-06-22T09:55:45` and `2026-07-30T13:33:04`.

Startups on the day of the incident:

| Daemon start | bot open_id result |
|---|---|
| 10:46:37 | `lark: bot identified` ✅ |
| 10:50:14 | `lark: bot identified` ✅ |
| 12:42:17 | `lark: bot identified` ✅ |
| **13:33:04** | **`failed to get bot open_id, group chat filtering disabled`** ❌ |
| 16:43:24 | `lark: bot identified` ✅ |
[...]

PR #1799, opened 2026-09-06
feat(feishu): support card width_mode full (#1797)

### LINKED ISSUE #1797: [Feature] feishu card width_mode 支持设置为full
### Feature Area

Platform (Feishu, DingTalk, Telegram, Slack, etc.)

### Problem or Motivation

当前card只能占用一半屏幕，对于命令代码等很多都需要拖动滚动条才能看到。
Image

### Proposed Solution

希望改成占满宽度。
函数增加个参数cardWidthMode string
```
func buildCardJSONWithStatusFooter(content, footer, cardWidthMode string) string {
	if strings.TrimSpace(footer) == "" {
		return buildCardJSON(content, cardWidthMode)
	}
	segments := sanitizeCardMarkdownSegmentsForCard([]string{content, footer})
	content = segments[0]
	footer = segments[1]
	elements := []map[string]any{
		{
			"tag":     "markdown",
			"content": content,
		},
		{
			"tag": "hr",
		},
		{
			"tag":       "markdown",
			"content":   footer,
			"text_size": "notation",
		},
	}
	card := map[string]any{
		"schema": "2.0",
		"config": map[string]any{
			"width_mode": cardWidthMode,  -----这里改动
		},
		"body": map[string]any{
			"elements": elements,
		},
	}
```
还有这个函数
增加参数 cardWidthMode string

```
func buildCardJSON(content, cardWidthMode string) string {
	content = sanitizeCardMarkdownForCard(content)
	card := map[string]any{
		"schema": "2.0",
		"config": map[string]any{
			"width_mode": cardWidthMode,
		},
		"body": map[string]any{
			"elements": []map[string]any{
				{
					"tag":     "markdown",
					"content": content,
				},
			},
		},
	}
	b, _ := json.Marshal(card)
	return string(b)
}

```

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Willingness to Contribute

- [ ] I'd be willing to submit a PR for this feature

chenhg5:
## Summary

Closes #1797. Adds an opt-in `card_width_mode` knob under
`[projects.platforms.options]` so operators can make Feishu Card 2.0
cards fill the full chat window width on desktop + mobile without
breaking the existing half-width default.

## Configuration

```toml
[[projects.platforms]]
type = "feishu"

[projects.platforms.options]
app_id = "..."
app_secret = "..."
# card_width_mode = "default"  # default; half-width on desktop, full on mobile
# card_width_mode = "full"     # opt in to full width on desktop + mobile
```

Default behaviour (`"default"` / unset / unknown) is byte-for-byte
the same as before — `width_mode: "default"` keeps the previous
half-width rendering, and invalid values fall back to it with a
slog.Debug line so a typo in config.toml never silently changes the
UI.

## Design notes

- Config field lives at `[projects.platforms.options].card_width_mode`
  rather than a top-level `[projects.options]` block, matching the
  existing convention for feishu-specific knobs (`image_batch_window_ms`,
  `progress_style`, etc.). The current ProjectConfig only exposes
  `[projects.agent.options]` and `[projects.platforms.options]`.
- The Card 1.0 `wide_screen_mode: true` hard-coded fallback in
  `buildCardJSONWithStatus` is replaced with the schema 2.0
  `width_mode` field, which is what the rest of the codebase already
  uses.
- `cardWidthMode` is threaded through every Card 2.0 config-block
  builder so reply cards, preview cards, status cards, rich cards, and
  card-action refresh responses all honour the configured mode.

## Files changed

- `platform/feishu/feishu.go` — parse `card_width_mode` in
  `newPlatform`; thread through `buildCardJSON`,
  `buildCardJSONWithStatusFooter`, `buildCardJSONWithStatus`,
  `buildProgressCardJSONFromPayload`, `buildPreviewCardJSON`,
  `buildRichCard`, `buildRichCardJSONBytes`,
  `buildReplyContent`, and update every caller to pass
  `p.cardWidthMode`. Helpers: `stringFromOption`,
  `normalizeCardWidthMode`, `defaultCardWidthMode`.
- `platform/feishu/card.go` — `renderCard` and `renderCardMap` now
  accept `cardWidthMode`; `ReplyCard`, `SendCard`,
  `RefreshCard` propagate it.
- `config.example.toml` — document the new option next to
  `enable_feishu_card`.
- Tests: `platform/feishu/platform_test.go`,
  `platform/feishu/card_test.go`, `platform/feishu/feishu_test.go`
  — 17 new test cases covering `normalizeCardWidthMode` edge cases
  (empty, mixed case, unknown), default + "full" propagation through
  every card builder and config parsing.

## Validation

- `go vet ./...` — clean (only pre-existing `web/embed.go` warning)
- `gofmt -l platform/feishu/` — clean (only pre-existing
  `group_filter_test.go` from #1618)
- `go build -tags no_web ./cmd/cc-connect` — succeeds
- `go test -race ./platform/feishu/... ./core/... ./config/...` —
  pass

## Backward compatibility

- No change to the IM API surface beyond the Card 2.0 `width_mode`
  field, which is already supported by the same SDK already in use.
- Default `"default"` mode keeps the previous rendering for all
  existing users.
- All status / preview / reply / rich / card-action card paths now
  use the schema 2.0 `width_mode` instead of mixing 1.0
  `wide_screen_mode`.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1012,6 +1012,16 @@ app_secret = "your-feishu-app-secret"
 # domain = "https://open.feishu.cn"  # Optional runtime API/WebSocket base URL override / 可选：运行时 API/WebSocket 域名覆盖
 # enable_feishu_card = true  # Enable Feishu interactive cards (default: true); set false to force plain-text replies
 #                              # 启用飞书交互式卡片（默认 true）；设为 false 时强制回退纯文本回复
+# card_width_mode = "default"  # Feishu Card 2.0 `width_mode` for all cards this platform renders
+#                              # (issue #1797). "default" = half-width on desktop, full width on mobile (matches
+#                              # the previous `wide_screen_mode = false` behaviour). "full" = full chat window
+#                              # width on desktop + mobile. Unknown / empty values fall back to "default"
+#                              # with a debug log. Affects reply cards, preview cards, status cards, rich cards,
+#                              # and card-action refresh responses.
+#                              # 飞书卡片 2.0 的 `width_mode`（issue #1797）。"default" = 桌面半宽 + 移动端全宽
+#                              # （与原 wide_screen_mode = false 行为一致）。"full" = 桌面+移动端均铺满聊天窗口。
+#                              # 未知或空值会以 debug 日志形式回退到 "default"。
+#                              # 影响回复卡片、预览卡片、状态卡片、富卡片以及卡片交互刷新响应。
 # allow_from = "*"          # Allowed user open_ids, comma-separated; "*" = all (default). Use /whoami to get your open_id.
 #                              # 允许的用户 open_id，逗号分隔；"*" 表示所有（默认）。发送 /whoami 获取你的 open_id。
 # allow_chat = "*"          # Allowed group chat_ids, comma-separated; "*" = all (default). Find chat_id in group settings.
--- a/platform/feishu/card.go
+++ b/platform/feishu/card.go
@@ -24,7 +24,7 @@ func (p *interactivePlatform) ReplyCard(ctx context.Context, rctx any, card *cor
 		return fmt.Errorf("%s: invalid reply context type %T", p.tag(), rctx)
 	}
 
-	cardJSON := renderCard(card, rc.sessionKey)
+	cardJSON := renderCard(card, rc.sessionKey, p.cardWidthMode)
 	if !p.shouldUseThreadOrReplyAPI(rc) {
 		if rc.chatID == "" {
 			return fmt.Errorf("%s: chatID is empty, cannot send card", p.tag())
@@ -48,7 +48,7 @@ func (p *interactivePlatform) SendCard(ctx context.Context, rctx any, card *core
 		return p.ReplyCard(ctx, rctx, card)
 	}
 
-	cardJSON := renderCard(card, rc.sessionKey)
+	cardJSON := renderCard(card, rc.sessionKey, p.cardWidthMode)
 	return p.createMessage(ctx, rc.chatID, larkim.MsgTypeInteractive, cardJSON, "send card")
 }
 
@@ -64,7 +64,7 @@ func (p *interactivePlatform) RefreshCard(ctx context.Context, sessionKey string
 		return fmt.Errorf("%s: no tracked card messageID for session %q", p.tag(), sessionKey)
 	}
 
-	cardJSON := renderCard(card, sessionKey)
+	cardJSON := renderCard(card, sessionKey, p.cardWidthMode)
 	req := larkim.NewPatchMessageReqBuilder().
 		MessageId(msgID).
 		Body(larkim.NewPatchMessageReqBodyBuilder().
@@ -88,10 +88,13 @@ func (p *interactivePlatform) RefreshCard(ctx context.Context, sessionKey string
 // renderCardMap converts a core.Card into the Feishu Interactive Card map
 // using the v1 format. Used both for message API (via renderCard) and
 // callback responses (CardActionTriggerResponse).
-func renderCardMap(card *core.Card, sessionKey string) map[string]any {
+//
+// cardWidthMode is forwarded into the config block as the Feishu Card 2.0
+// `width_mode` field (#1797). It must be one of "default" or "full".
+func renderCardMap(card *core.Card, sessionKey, cardWidthMode string) map[string]any {
 	result := map[string]any{
 		"config": map[string]any{
-			"wide_screen_mode": true,
+			"width_mode": cardWidthMode, // schema 2.0; was wide_screen_mode (schema 1.0, #1797)
 		},
 	}
 	if card == nil {
@@ -454,8 +457,8 @@ func parseDeleteModeListItemAction(action string) (id string, selectable bool, o
 }
 
 // renderCard converts a core.Card into the Feishu Interactive Card JSON string.
-func renderCard(card *core.Card, sessionKey string) string {
-	b, err := json.Marshal(renderCardMap(card, sessionKey))
+func renderCard(card *core.Card, sessionKey, cardWidthMode string) string {
+	b, err := json.Marshal(renderCardMap(card, sessionKey, cardWidthMode))
 	if err != nil {
 		slog.Error("feishu: renderCard marshal failed", "error", err)
 		return `{"config":{"wide_screen_mode":true},"elements":[]}`
--- a/platform/feishu/card_test.go
+++ b/platform/feishu/card_test.go
@@ -12,7 +12,7 @@ func decodeRenderedCard(t *testing.T, card *core.Card) map[string]any {
 	t.Helper()
 
 	var got map[string]any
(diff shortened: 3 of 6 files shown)


chenhg5:
## Conclusion: Approve

## Overall assessment
PR #1799 is mergeable. CI is 5/5 green on the current head (`3d17906`), `mergeable: MERGEABLE`, and the substantive change is a clean, opt-in feature: adds a `card_width_mode` knob under `[projects.platforms.options]` so Feishu Card 2.0 cards can opt into full chat-window width on desktop + mobile without breaking the half-width default. Defaults stay byte-for-byte the same; unknown config values fall back to default with a debug log.

## Review scope
- Diff (6 files, +297/-65): `platform/feishu/feishu.go`, `platform/feishu/card.go`, `config.example.toml`, plus three test files.
- Walked through `normalizeCardWidthMode`, `defaultCardWidthMode`, `cardWidthMode` threading, every Card 2.0 builder, and the test corpus (17 new cases per the PR description).

## ✅ What looks good
- **Configuration shape is consistent with existing convention**: `card_width_mode` lives under `[projects.platforms.options]` next to other feishu-specific knobs (`image_batch_window_ms`, `progress_style`, `enable_feishu_card`). Matches the codebase pattern. The PR body explicitly notes the constraint (ProjectConfig currently only exposes `[projects.agent.options]` and `[projects.platforms.options]`).
- **Default path is byte-for-byte identical**: `normalizeCardWidthMode("")` → `"default"`, `normalizeCardWidthMode("default")` → `"default"`, both result in the Feishu Card 2.0 `width_mode: "default"` field, which is the platform's own default rendering. Existing deployments render exactly as before.
- **Unknown values fail soft, not hard**: the helper logs at `slog.Debug` for unknown inputs and falls back to `"default"`. A typo in `config.toml` doesn't silently change every card's layout — and it doesn't crash startup. The right behavior for a platform-level cosmetic knob.
- **Card 1.0 `wide_screen_mode` is replaced, not duplicated**: the PR removes the `wide_screen_mode: true` fallback in `buildCardJSONWithStatus` and `renderCardMap`, replacing it with the schema 2.0 `width_mode` field. Clean unification — every card builder now uses schema 2.0 consistently. The PR body's "Backward compatibility" section spells this out.
- **Threading is exhaustive**: `cardWidthMode` is wired through `ReplyCard`, `SendCard`, `RefreshCard`, `Reply`, `Send`, `SendWithStatusFooter`, `SendPreviewStart`, `UpdateMessage`, `UpdateMessageWithStatusFooter`, `BuildRichCard`, `SetPreviewStatus`, and the `onCardAction` callback response paths. Every place a Card 2.0 JSON is constructed now sees the platform's configured mode.
- **Helpers are small and well-documented**: `stringFromOption` (nil-tolerant map lookup), `normalizeCardWidthMode` (validates + debug-logs), `defaultCardWidthMode` (single source of truth for the default constant). The 5-second grep test for `cardWidthMode` is satisfied everywhere it should be.
- **Tests cover the cases that matter**: per the PR body, 17 new test cases cover `normalizeCardWidthMode` edge cases (empty, mixed case, unknown), default + "full" propagation through every card builder, and config parsing. The visible test diffs in `card_test.go` (TestRenderCardMap_InjectsSessionKeyIntoCallbacks, TestBuildCardJSONWithStatusFooter, TestBuildCardJSONWithStatusFooter_EmptyFooterFallsThrough) confirm the threading pin is real, not just a description.

- **`buildRichCard` second positional parameter `_ string` is misleading**: the signature is `(status core.CardStatus, _ string, steps []core.ToolStep, markdown string, streaming bool, statusFooter string, cardWidthMode string)`. The `_` parameter (originally `title`) is now unused. Adding `cardWidthMode` as a 7th positional argument makes call sites harder to read; an options struct would scale better, but that's a broader refactor. Low-cost fix for now: keep the signature, but the `_` line deserves a comment explaining why the parameter still exists (callers' API contract) instead of being silently ignored.
- **`buildCardJSONWithStatus` doc comment is outdated**: it still says "Pre-existing callers always passed 'default' implicitly via the hardcoded value" — but after this PR, no caller passes "default" implicitly anymore; the platform always threads `p.cardWidthMode` through. The comment is now confusing. One-line edit to remove the historical context.
- **No end-to-end test for the `onCardAction` → `CardActionTriggerResponse` path with `width_mode: full`**: the diff updates all four `onCardAction` callback sites to pass `p.cardWidthMode`, but the test diff I scanned doesn't pin that the callback's `Data` map has the right `width_mode`. Worth a 5-line sub-test.

## 🔵 Optional
- **`defaultCardWidthMode` could be a `var` rather than a `const`**: if Feishu ever changes the SDK's default rendering, this becomes a single-line config flip. Low value (today the platform default matches Feishu's own default and isn't likely to drift), but it's a one-character change.
- **The config doc comment is bilingual**: 9 lines of EN + 9 lines of ZH-CN. The pattern is consistent across `config.example.toml` so it's fine, but a maintainer who wants a tighter example file could halve this. Not worth changing — consistency wins here.
- **`stringFromOption` could be `slog.Debug` on nil opts**: today it silently returns `""` for nil. If `newPlatform` ever constructs `Platform` without going through the normal `opts` map path, missing keys would be invisible. Probably fine because all callers go through `newPlatform`, but worth a one-line note.

## ❓ Questions
- **Card 2.0 supports `width_mode = "compact"` too** (Feishu SDK ref). The PR only supports `"default"` and `"full"`. Is that an intentional v1 scope decision, or should the helper accept `"compact"` and pass it through? If intentional, a one-line doc comment on `normalizeCardWidthMode` would lock it in.
- **`enable_feishu_card = false` (plain-text fallback) path**: when a config sets `enable_feishu_card = false`, the platform never renders cards, so `cardWidthMode` is unused. Is the parsing order (parse `card_width_mode` before checking `enable_feishu_card`) intentional, or does it accidentally initialize a field that won't be used? Not a bug — just confirming the parse happens unconditionally so a config-author who later flips `enable_feishu_card = true` doesn't need to re-set `card_width_mode`.

## Testing / Risk
- Verified on current head `3d17906`: CI run 34021956952 — `lint` 2m23s ✓, `unit-test` 4m57s ✓, `smoke-test` 28s ✓, `regression-test` 34s ✓, `performance-test` 45s ✓.
- Verified locally (this review): the diff is confined to `platform/feishu/` + `config.example.toml`. No core/engine interface changes; no schema migrations. The Card 2.0 `width_mode` field is part of the SDK already in use.
- Unverified risk: operators who upgrade and haven't read the changelog won't see any change (default stays default). Operators who set `card_width_mode = "full"` get full-width cards on desktop. Worst-case impact is a layout surprise for users who opted in.
- Registry data freshness: the local registry still showed `ci_status=pending` for PR #1799 when this cycle started (synced at 2026-09-06T08:34:41Z); `gh pr checks` confirmed all 5 jobs actually passed at 2026-09-06T08:31:10Z. Registry sync latency is ~5min on this repo today — worth noting in the cycle summary so future cycles don't get confused.

## Next step
Merge when convenient. No further code changes required from the author. The optional follow-ups (compact mode, options struct for `buildRichCard`, end-to-end callback test) are P3 polish, not blockers.

PR #1800, opened 2026-09-06

AaronZ345:
## Summary

## Root cause

## Test Plan
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c -tags no_web ./cmd/cc-connect`
- `git diff --check`

Supersedes #1651, which superseded #1496 and #1349. This fresh replacement carries the same patch on current `main`; the historical empty CI-retrigger commit was intentionally excluded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,8 @@ the full themed summary; per-beta details remain in the beta sections below.
   because a later queued message has a higher `create_time` (#1286)
 - **core**: make `/history` entry truncation configurable via
   `[display].history_max_len`, defaulting to 1000; `0` disables truncation (#1291)
+- **cron/timer**: prompt jobs default to a fresh agent session per run, avoiding busy
+  chat sessions; use `session_mode = "reuse"` to keep shared-session behavior
 - **tts/minimax**: drop `status=2` trailer chunk to stop audio playing twice (#1364)
 - **tests**: add provider-resume regression tests for codex / opencode / kimi (#1366)
 
--- a/cmd/cc-connect/cron.go
+++ b/cmd/cc-connect/cron.go
@@ -626,7 +626,7 @@ Options:
       --prompt <text>        Task prompt (runs through agent)
       --exec <command>       Shell command (runs directly, mutually exclusive with --prompt)
       --desc <text>          Short description
-      --session-mode <mode>  reuse (default) or new-per-run — fresh agent session each run
+      --session-mode <mode>  new-per-run (default) or reuse — fresh agent session each run unless reusing
       --timeout-mins <n>     Max minutes to wait per run (0 = no limit; default 30 if omitted)
       --silent               Suppress cron start notification
       --data-dir <path>      Data directory (default: ~/.cc-connect)
--- a/cmd/cc-connect/timer.go
+++ b/cmd/cc-connect/timer.go
@@ -412,7 +412,7 @@ Options:
       --prompt <text>        Task prompt (runs through agent)
       --exec <command>       Shell command (runs directly, mutually exclusive with --prompt)
       --desc <text>          Short description
-      --session-mode <mode>  reuse (default) or new-per-run
+      --session-mode <mode>  new-per-run (default) or reuse
       --timeout-mins <n>     Max minutes to wait per run (0 = no limit; default 30)
       --mute                 Suppress all messages (start + result)
       --data-dir <path>      Data directory (default: ~/.cc-connect)
--- a/config.example.toml
+++ b/config.example.toml
@@ -336,8 +336,8 @@ level = "info" # debug, info, warn, error
 # [cron]
 # silent = false             # Suppress "⏰ task name" notification when cron starts (default: false)
 #                            # 静默模式：不发送定时任务开始执行的通知消息（默认 false）
-# session_mode = "reuse"     # Default session mode for all cron jobs: "reuse" (default) or "new_per_run"
-#                            # 定时任务默认会话模式："reuse"（复用活跃会话）或 "new_per_run"（每次新建会话）
+# session_mode = "new_per_run" # Default session mode for cron/timer jobs: "new_per_run" (default) or "reuse"
+#                              # 定时/延时任务默认会话模式："new_per_run"（默认，每次新建会话）或 "reuse"（复用活跃会话）
 
 # =============================================================================
 # Webhook / 外部 Webhook 端点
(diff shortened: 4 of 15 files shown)


chenhg5 (approved):
## Conclusion: Approve

This is a well-scoped default flip with comprehensive downstream updates. The "busy side session" root cause is real (cron firing while chat is busy is a classic scheduler pitfall), the new default isolates scheduled work from interactive turns, and the opt-out is trivial (`--session-mode reuse` / `session_mode = "reuse"` / per-job override). Every user-visible touchpoint is updated consistently: CHANGELOG, CLI help, config comments, API docs, two user guides, and the agent-facing i18n strings in 4 places. The `NormalizeCronSessionMode` cleanup (now distinguishing `""` from `"reuse"`) is a real semantic improvement — before this PR, users couldn't explicitly set "reuse" via API because it got normalized away. Tests cover the new defaults, the new validation, and the test that previously relied on the old default has been correctly updated. The one risk worth flagging is the implicit behavior change for users who created cron/timer jobs before this PR and never set a session mode — they'll silently switch from reuse to new-per-run semantics on upgrade.

### ✅ What looks good

- **Single source of truth for the default.** `const defaultScheduledSessionMode = "new_per_run"` lives in `core/cron.go` and is consumed by both `NewCronScheduler` and `NewTimerScheduler`. No string literal duplicated across files.
- **`NormalizeCronSessionMode` now distinguishes `""` from `"reuse"`.** This is a real fix — before this PR, a user who tried to explicitly set `session_mode: "reuse"` via the management API would have it normalized to `""`, then re-defaulted by the scheduler to whatever the global default was. The new contract (`""` = "use scheduler default"; `"reuse"` = "explicit share-active") is clearer and matches what an operator would expect from a tri-state config.
- **`validateCronJob` / `validateTimerJob` now accept `"reuse"`** to match the new normalization. The previous validation (`mode != "" && mode != "new_per_run"`) would have rejected explicit `reuse`, which would have been a bug given the new normalization.
- **Comprehensive doc updates in one PR.** CLI help text (cron + timer), config example comment, config struct comment, management API docs (EN + zh-CN), user guide (EN + zh-CN), CHANGELOG, and the 4 agent-facing i18n strings. Every surface that mentions `session_mode` is updated; no stale doc left over. This is the right discipline for a default-change PR.
- **`TestExecuteCronJob_ResolvesCronReplyTarget` fixture update is correct and well-targeted.** The test was testing the reuse-target resolution path; with the global default flipped to `new_per_run`, the fixture must explicitly opt in to reuse. Without this fix, the test would have silently started testing a different code path. Good catch by the author.
- **`waitRefreshedCards` helper extraction** improves 4 existing delete-mode tests that previously used `p.getRefreshedCards()` followed by `if len == 0 { t.Fatal }` — the helper's 2-second polling loop is more robust to CI scheduling jitter. Side-improvement unrelated to the main change but useful.
- **`TestCronScheduler_AddJob_NormalizesSessionMode` refactor to table-driven** covers 3 cases (new-per-run alias / REUSE uppercase / empty inherits default) instead of the single case before. Better coverage with no extra runtime.
- **`TestTimerScheduler_UsesNewSession_Defaults`** mirrors the existing cron test, ensuring the new default applies symmetrically to both scheduler types.

- **Implicit behavior change for existing cron/timer jobs not flagged prominently enough in the CHANGELOG.** The CHANGELOG bullet says "prompt jobs default to a fresh agent session per run, avoiding busy chat sessions; use `session_mode = 'reuse'` to keep shared-session behavior" — this is accurate but doesn't call out the upgrade-path explicitly. Operators who created jobs before this PR and never set `session_mode` will have their jobs silently switch from `reuse` to `new_per_run` semantics on upgrade. Worth adding one line to the CHANGELOG along the lines of: "Existing on-disk jobs with empty `session_mode` will start using `new_per_run` semantics after upgrade; set `session_mode = 'reuse'` in the global config or per-job to keep the old behavior." This isn't a bug — it's the intended effect — but users who depended on the implicit default deserve an explicit migration hint.

- **`CronJob.SessionMode` comment is slightly misleading about what `""` means.** Current comment: `"" = scheduler default; "reuse" = share active session; "new_per_run" = fresh session each run`. This implies `""` means "use the system-wide default" — but there's no system-wide default separate from the per-scheduler default. The wording is technically correct (the scheduler's default is consulted) but could confuse a reader who thinks there's a separate global default. Suggest: `"" = use scheduler-level default (NewCronScheduler's defaultSessionMode, normally new_per_run); "reuse" = ...`. Optional.

- **`UsesNewSession` on `TimerScheduler`** — I didn't see this in the diff, but the existing `TimerScheduler.UsesNewSession` must already exist (it's called in `TestTimerScheduler_UsesNewSession_Defaults`). Confirming that the timer scheduler's `UsesNewSession` already reads from `defaultSessionMode` correctly and the new default takes effect on legacy jobs. Likely fine; just calling out for verification.

- **`UpdateJob` validation now accepts `"reuse"`** — a behavior change worth a one-line CHANGELOG mention. Before this PR, `UpdateJob` rejected `"reuse"`; now it accepts. Same goes for the JSON schema if the management API documents valid values for `session_mode`. Worth a sentence in the CHANGELOG.

### ❓ Questions

- **Existing on-disk jobs with `SessionMode: ""`**: These are loaded by `NewCronStore` / `NewTimerStore` and have `SessionMode: ""`. After this PR, `NormalizeCronSessionMode("")` returns `""`, then `CronScheduler.UsesNewSession(job)` checks against `defaultSessionMode` which is now `new_per_run`. So legacy jobs silently switch behavior. Is this the intended migration path, or should there be a one-time data migration that sets `SessionMode: "reuse"` on existing jobs to preserve explicit opt-in for the previous behavior? My read is that the intended migration is the silent switch (matching the new global default), and operators who want the old behavior opt in via config. Confirming.

- **`TestExecuteCronJob_ResolvesCronReplyTarget` is the only engine_test.go test with the explicit `SessionMode: "reuse"` fix.** Are there other engine tests that implicitly relied on the old default? Quick grep of `Engine` tests for cron/timer fixtures might surface them.

### Testing / Risk

- **CI 5/5 green** on head `3c772f4770daea22b2f843e3be12aee48a304e15` (lint, unit, smoke, regression, performance — all success, run 2026-09-06).
- **Test surface is sufficient for the change scope**: scheduler defaults (cron + timer), normalization (3 cases via table), validation (cron + timer reuse cases), and the engine fixture that previously relied on the old default.
- **Author's pre-PR local runs**: `go test -count=1 ./core ./config`, `go test -c -tags no_web ./cmd/cc-connect`, `git diff --check` all clean per PR body.
- **Remaining risk**: the silent migration of existing on-disk jobs from reuse to new-per-run (P2, mitigable with a config-level opt-out). Everything else is polish.

### Next step

Landable. The migration hint in the CHANGELOG is the one I'd push for as a follow-up before this lands on a release that users will upgrade to without reading every line. Other items are polish that can ship in subsequent PRs.

PR #1801, opened 2026-09-06

### LINKED ISSUE #1077: [Bug] new_per_run cron sessions remain visible in /list after completion

AaronZ345:
## Summary

Fixes #1077.

## Tests
- `go test -count=1 ./core`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c -tags no_web ./cmd/cc-connect`
- `git diff --check`

Supersedes #1650, which superseded #1497 and #1417. This fresh replacement carries the same patch on current `main`.


chenhg5 (approved):
## Conclusion: Approve

This is the third iteration of the #1077 fix (Supersedes #1650 → #1497 → #1417), and the design choices here are well-considered: `Background` is a first-class session field with default-hide behaviour, the tri-state config (`nil`/true/false) is the right shape for a backward-compatible default-on toggle, the new `BackgroundAgentSessionIDs()` separation lets the filter live entirely in `applySessionFilter` without changing `filterOwnedSessions` callers, and the JSON `omitempty` keeps existing on-disk session files loading cleanly. Tests cover config load (3 cases), session-manager invariants (3 cases), persistence round-trip, and the end-to-end `cmdList` interaction with both filters composed. The one thing I want to confirm with the maintainer is whether the `cmdSwitch` and `cmdDelete` code paths also go through `applySessionFilter` (the PR description says they do) — if yes, this is mergeable; if no, that's a P1 gap.

### ✅ What looks good

- **Tri-state config handling is the right pattern.** `proj.HideSchedulerSessions == nil || *proj.HideSchedulerSessions` cleanly expresses "default-on, opt-out via false". Existing users with no `hide_scheduler_sessions` line get the new behaviour; existing users with explicit `false` get the legacy behaviour. Both Engine and SessionManager independently default to true, so the behaviour is consistent even if a caller constructs a SessionManager without going through Engine.
- **`Background` is per-session, not per-session-manager.** A future "background" session that later becomes a normal session (or vice versa) doesn't require SessionManager-level hacks. Persistence is `omitempty` so existing JSON loads cleanly with Background=false; new background sessions persist the field.
- **`NewSideSession` is preserved as a wrapper around the new `newSideSession(userKey, name, false)`.** Backward-compatible for any out-of-tree callers; new callers can pick `NewBackgroundSession` explicitly. Clean refactor.
- **`KnownAgentSessionIDs` excludes both current and past agent IDs of background sessions.** This is critical — without it, `filter_external_sessions` would still let a cron-session agent ID through as a "tracked" ID and defeat the point. Good defensive consistency.
- **`BackgroundAgentSessionIDs()` is a separate accessor that always returns the hidden IDs** regardless of the current `hideBackgroundSessions` flag. This lets `applySessionFilter` toggle visibility at runtime without losing the ability to re-hide later, and is exactly the right shape for a hot-reload of `reloadConfig()`.
- **Tests are well-scoped.** Config load (nil/false/true), session manager (hidden by default, visible when disabled, persistence round-trip with both current and past agent IDs), engine integration (cmdList under both filter combinations). No over-mocking, no flaky timing.
- **CHANGELOG entry is bilingual** (English + 中文) and references both #1077 and #1417, which is the right trail for a long-standing issue.
- **No breaking change to public API.** All existing fields/methods preserved; new code is opt-in or default-on.

- **PR description cites /list, /switch, and /delete, but the diff only visibly changes the cmdList path.** I see `applySessionFilter` being called from `cmdList`, and the new helper `filterHiddenSessions` lives next to `filterOwnedSessions`. For `/switch` and `/delete` to actually hide scheduler sessions too, they must also call `applySessionFilter`. I believe they do (the existing code already routes them through `filterOwnedSessions`), but please confirm in the PR description or in a quick code-pointer comment — otherwise readers will assume the gap exists. **If they don't, that's a P1 bug in this PR.**

- **The CHANGELOG bullet mentions `#1077, #1417` but doesn't mention the related prior PRs `#1497` and `#1650` that this supersedes.** Not blocking, but for traceability of the issue history it would help future maintainers reading the changelog. Optional.

- **`filterHiddenSessions` is O(N×M) over sessions × hidden IDs.** For a deployment with thousands of cron-fired sessions and a large user session list, this becomes O(N×M). For typical cc-connect deployments (<100 sessions per user), it's a non-issue. If scale ever matters, switch the hidden set to a `map[string]struct{}` lookup (which it already is) but keep the outer loop. Not worth changing now.

- **`NewBackgroundSession` is exported but `newSideSession` is unexported.** That's fine, but the difference between `NewSideSession` and `NewBackgroundSession` deserves a one-line godoc on each, not just a comment on the wrapper. Currently `NewSideSession` is undocumented; readers won't know when to pick which. **Suggest adding 2-3 lines of godoc on each public constructor.**

- **`Session.Background` is JSON `omitempty` but the field is a primitive bool, not a pointer.** A user with no `Background` field in their on-disk JSON will see Background=false on load — correct (no migration needed), but a user who previously wrote a session that *was* a background session somehow (pre-this-PR) cannot be distinguished from a fresh session. Not a real concern for current code; just noting that the omitempty direction is the only safe one.

### ❓ Questions

- **/switch and /delete paths.** As noted above — do they route through `applySessionFilter`? If yes, the PR title's promise is fulfilled; if no, that's a gap that should be fixed before merge.

- **Tri-state config default across reload.** `reloadConfig` re-calls `SetHideSchedulerSessions(...)`. If an operator toggles `hide_scheduler_sessions = false → true` via SIGHUP, does the SessionManager's `hideBackgroundSessions` get updated atomically? I see `SetHideSchedulerSessions` propagates to `e.sessions.SetHideBackgroundSessions(v)` under the SessionManager's mu.Lock — looks correct, just confirming.

- **Past agent IDs of background sessions.** When a background session has `AgentSessionID` swapped across multiple turns (`cron-thread-1` → `cron-thread-2` via `SetAgentSessionID`), both go into `BackgroundAgentSessionIDs` (test `TestBackgroundAgentSessionIDs_IncludesPastAndPersists` confirms this for persistence). But does `KnownAgentSessionIDs` also exclude past IDs of background sessions? Looking at the diff yes — `include && s.AgentSessionID != ""` then `if include { for _, past := range s.PastAgentSessionIDs { ids[past] = struct{}{} } }`. The `include` check wraps both, so past IDs are also excluded. ✓

### Testing / Risk

- **CI 5/5 green** on head `9b1108934ae47e789c803730182faa866374cb34` (lint, unit, smoke, regression, performance — all success, run on 2026-09-06).
- **Test surface is sufficient**: config load (3), session manager (3), persistence round-trip, cmdList integration (2 with filter composition). No flaky timing tests.
- **Author's pre-PR local runs**: `go test -count=1 ./core`, `go test -c -tags no_web ./cmd/cc-connect`, `git diff --check` all clean per PR body.
- **Remaining risk**: the unverified /switch and /delete paths (P1 if they don't go through `applySessionFilter`; non-issue if they do). Everything else is non-blocking.

### Next step

Landable once the `/switch` and `/delete` routing through `applySessionFilter` is confirmed (either by author adding a one-liner to the PR description, or by reviewer finding the call sites in the current main and noting them). The other P2 items are polish that can ship as follow-ups or stay open.

PR #1802, opened 2026-09-06
feat(feishu): add catchup polling to compensate WebSocket push losses

jsvisa:
## Summary

Feishu/Lark's WebSocket push of `im.message.receive_v1` is best-effort and silently drops events even while the long connection looks healthy — Feishu's own server logs show `app not online` push failures against connected clients (see [larksuite/oapi-sdk-go#195](https://github.com/larksuite/oapi-sdk-go/issues/195) and [larksuite/node-sdk#164](https://github.com/larksuite/node-sdk/issues/164)). Users experience the bot ignoring @mentions / DMs with no error anywhere on the cc-connect side.

This PR adds an opt-in REST catch-up poller as a compensation channel: every 3 minutes it lists recent messages of the configured chats via `Im.Message.List` and re-injects anything the WebSocket path missed, through the exact same dispatch path (dedup + stale-message guards included).

## Type of change

- [ ] Breaking change (fix or feature that would change existing behavior)
- [ ] Documentation only

## Design

- **New file `platform/feishu/catchup.go`**
  - Poll interval **3 min**, lookback window **5 min** (overlaps the interval on purpose — dedup absorbs the overlap), trailing buffer **30 s** so in-flight messages the WebSocket will still deliver aren't raced.
  - Two new platform options:
    - `catchup_chats` — comma-separated **group** chat IDs; a polled message is re-injected only if the bot appears in the REST mention list (`Mention.Id` is a plain `open_id` string on the REST shape, unlike the WS `UserId` struct).
    - `catchup_chats_p2p` — comma-separated **p2p** chat IDs; every user message qualifies because the REST API never returns a mention list for p2p.
  - Guards before injection, mirroring the WS path (`onMessage`):
    - `core.MessageDedup` by `message_id` — a message delivered by both WS and poll is processed exactly once, whichever arrives first;
    - `core.IsOldMessage` — a poll right after restart never replays history;
    - deleted messages and app-sent messages (`sender_type == "app"`) are skipped — the latter prevents the poller from echoing the bot's own replies back into the agent (the WS push never delivers the app's own messages).
  - Injection goes through `dispatchMessage` (same text/post/image/... parsing, name resolution, session keys), with REST mentions converted to `MentionEvent` and the correct `chat_type` (`group`/`p2p`) fed into `makeSessionKey`.
  - REST calls reuse `withFreshTenantAccessTokenRetry` for transparent token refresh.
- **Shared-WS aware** (`feishu.go`): the poller starts only on the **primary owner** of the shared WebSocket connection (`isWSPrimary`, see `ws_shared.go`), so platforms sharing one `app_id` never re-inject each other's missed events. It starts after `startWebSocketMode` succeeds and is stopped in `Stop()`.
- Both options default to empty = **disabled**; no behavior change for existing configs.

### Config example

```toml
[[projects.platforms]]
type = "feishu"

[projects.platforms.options]
app_id = "cli_xxx"
app_secret = "xxx"
# Optional: catch-up polling to compensate WebSocket push loss
catchup_chats = "oc_group1,oc_group2"   # group chats, @bot mention required
catchup_chats_p2p = "oc_p2p1"           # p2p chats, every user message qualifies
```

Requires the `im:message:readonly` (or existing `im:message`) scope, and bot `open_id` discovery must have succeeded for the group mention gate (the poller skips rather than fails open when it is unknown).

- `TestCatchupMessageQualifies` in `platform/feishu/catchup_test.go` — pins the p2p vs group rules: **p2p messages skip the mention check** (qualify with no mentions / unrelated mentions), **group messages still require the bot mention** (rejected without mentions or with someone else mentioned; accepted when the bot is mentioned).
- `TestCatchupSenderQualifies` — app-sent messages are excluded (self-reply-loop guard), user messages and unknown senders qualify.
- `TestIsBotMentionedInList` — REST mention-list matching incl. nil/empty lists and nil IDs.
- `TestConvertRESTMentions` — REST `Mention` → `MentionEvent` conversion incl. nil-entry skipping.

### Test results

```
$ go build ./...            # passes
$ go vet ./platform/feishu/...  # passes
$ go test ./platform/feishu/... # ok 36.2s
$ go test ./core/ -run TestCUJ  # ok
```

`go test ./...` has pre-existing failures on clean `main` unrelated to this change, reproduced locally before applying the patch:

- `web` / `cmd/cc-connect`: `web/embed.go: pattern all:dist: no matching files found` (frontend `dist` not built in this environment)
- `agent/acp` `TestCursorCLI_ACPHandshake`: requires an authenticated Cursor CLI (`context deadline exceeded`)
- `daemon` `TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable`: macOS launchd environment-dependent

- [x] A — basic conversation (new optional injection path into the existing dispatch flow; disabled unless `catchup_chats*` is configured)
- [ ] B / C / D / E / F / G / H / I

Confirmed: `go test ./core/ -run TestCUJ` passes locally.

None unless the new options are set. With `catchup_chats` / `catchup_chats_p2p` configured, messages that Feishu's WebSocket push dropped are delivered (up to ~5 minutes late) instead of being lost; log lines `catchup: injecting missed message` / `catchup: injected missed messages` make the compensation observable. Upstream SDK issues for reference: [larksuite/oapi-sdk-go#195](https://github.com/larksuite/oapi-sdk-go/issues/195), [larksuite/node-sdk#164](https://github.com/larksuite/node-sdk/issues/164).

- [x] `go build ./...` passes
- [x] `go test ./...` passes (`platform/feishu` fully green; pre-existing environment-only failures listed above exist on clean `main`)
- [x] No new hardcoded platform/agent names in `core/` (change is confined to `platform/feishu/`)
- [x] i18n strings: N/A (no new user-facing UI strings)

## Related

- larksuite/oapi-sdk-go#195 — WS push drops with `app not online` while the client is connected
- larksuite/node-sdk#164 — same failure mode observed server-side

### DIFF
--- a/platform/feishu/catchup.go
+++ b/platform/feishu/catchup.go
@@ -1,0 +1,288 @@
+package feishu
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"strconv"
+	"strings"
+	"time"
+
+	lark "github.com/larksuite/oapi-sdk-go/v3"
+	larkcore "github.com/larksuite/oapi-sdk-go/v3/core"
+	larkim "github.com/larksuite/oapi-sdk-go/v3/service/im/v1"
+
+	"github.com/chenhg5/cc-connect/core"
+)
+
+// Feishu's WebSocket push (im.message.receive_v1) is best-effort: events can
+// be silently dropped even while the connection looks healthy (see
+// larksuite/oapi-sdk-go#195 and larksuite/node-sdk#164, where Feishu's own
+// server logs show "app not online" failures against connected clients).
+// The catch-up poller below is a compensation channel: every few minutes it
+// lists recent messages of the operator-configured chats over the REST API
+// and re-injects anything the WebSocket path missed.
+//
+// Injection reuses the exact same dispatchMessage path as WebSocket-delivered
+// events, and is guarded by:
+//   - dedup by message_id (core.MessageDedup), so a message delivered by both
+//     the WebSocket and a poll is processed exactly once, whichever arrives first;
+//   - the stale-message guard (core.IsOldMessage), so a poll right after a
+//     restart does not replay historical traffic.
+
+const (
+	catchupInterval = 3 * time.Minute  // how often the poller runs
+	catchupLookback = 5 * time.Minute  // how far back each poll window reaches (overlaps the interval on purpose; dedup absorbs the overlap)
+	catchupBuffer   = 30 * time.Second // exclude the most recent slice to avoid racing in-flight messages the WebSocket will still deliver
+	catchupPageSize = 50
+)
+
+// startCatchupPoller starts a background goroutine that periodically fetches
+// recent messages from each configured chat and injects any qualifying
+// messages missed by the WebSocket push path.
+//
+// It only runs on the primary owner of the shared WebSocket connection
+// (isWSPrimary): secondary platforms in a shared-WS group receive fanned-out
+// events from the primary, so a poller per platform would re-inject every
+// missed message once per sibling. Both catchup chat options empty means the
+// feature is disabled.
+func (p *Platform) startCatchupPoller() {
+	if p.catchupChats == "" && p.catchupP2PChats == "" {
+		return
+	}
+	if !p.isWSPrimary {
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	p.mu.Lock()
+	p.catchupCancel = cancel
+	p.mu.Unlock()
+
+	go func() {
+		slog.Info(p.tag()+": catchup poller started",
+			"interval", catchupInterval, "lookback", catchupLookback)
+		defer cancel()
+		ticker := time.NewTicker(catchupInterval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				p.pollAllCatchupChats(ctx)
+			}
+		}
+	}()
+}
+
+// stopCatchupPoller cancels the background poller, if running.
+func (p *Platform) stopCatchupPoller() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.catchupCancel != nil {
+		p.catchupCancel()
+		p.catchupCancel = nil
+	}
+}
+
+func (p *Platform) pollAllCatchupChats(ctx context.Context) {
+	p2p := map[string]bool{}
+	for _, chatID := range strings.Split(p.catchupP2PChats, ",") {
+		chatID = strings.TrimSpace(chatID)
+		if chatID != "" {
+			p2p[chatID] = true
+		}
+	}
+	seen := map[string]bool{}
+	for _, chatID := range strings.Split(p.catchupChats+","+p.catchupP2PChats, ",") {
+		chatID = strings.TrimSpace(chatID)
+		if chatID == "" || seen[chatID] {
+			continue
+		}
+		seen[chatID] = true
+		if err := p.pollChatForMissedMessages(ctx, chatID, p2p[chatID]); err != nil {
+			slog.Warn(p.tag()+": catchup: poll error", "chat_id", chatID, "error", err)
+		}
+	}
+}
+
+// pollChatForMissedMessages lists recent messages of one chat over the REST
+// API and re-injects the ones the WebSocket push path missed. isP2P selects
+// the qualification rule: group chats require an explicit @bot mention, while
+// p2p chats have no mention list so every user message qualifies.
+func (p *Platform) pollChatForMissedMessages(ctx context.Context, chatID string, isP2P bool) error {
+	botOpenID := p.getBotOpenID()
+	if botOpenID == "" {
+		// Without the bot open_id we cannot run the group mention gate.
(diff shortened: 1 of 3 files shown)


chenhg5 (approved):
## Conclusion: Approve

This is a solid, defensive-by-design addition. The catch-up poller compensates for Feishu's best-effort WebSocket push (issue surfaces in larksuite/oapi-sdk-go#195 and larksuite/node-sdk#164) using a layered safety net — overlap absorbed by `MessageDedup`, racing in-flight messages protected by a 30s buffer, restart-replay blocked by `core.IsOldMessage`, self-reply loops blocked by sender-type filtering, and a fail-closed bot-open_id gate that prevents the #1618 loud-bot regression. Tests cover all the qualification and conversion edge cases. The one real gap is docs: the new opt-in config keys (`catchup_chats`, `catchup_chats_p2p`) aren't mentioned in `docs/usage.md` or `docs/usage.zh-CN.md`, so users won't know the feature exists.

### ✅ What looks good

- **Layered safety net.** `dedup.IsDuplicate` absorbs the intentional lookback/interval overlap; `catchupBuffer = 30s` excludes messages the WebSocket is likely still in flight for; `core.IsOldMessage` blocks restart-replay; `catchupSenderQualifies` blocks self-reply loops. Each one is cheap on its own; together they make the poller very hard to misbehave.
- **Fail-closed on missing bot open_id.** If `getBotOpenID()` returns empty, the poller silently skips the chat rather than injecting every group message — exactly the right call given the #1618 class of bug. The comment naming the issue makes the trade-off explicit.
- **Shared-WS awareness.** `isWSPrimary` check prevents N-fold polling when multiple platform instances share a WebSocket connection. Without this, every sibling platform would re-inject the same missed message.
- **Solid test coverage.** 4 test functions, 14 sub-cases: bot-mention detection (nil/empty/missing/nil-id/positive), p2p-vs-group qualification (5 cases including the "REST doesn't return mentions for p2p" wrinkle), sender qualification (app-skip / user-qualify / fail-open on nil), and `convertRESTMentions` (nil/empty/conversion/nil-entry).
- **Reuses `dispatchMessage`.** Injected messages flow through the same path as WebSocket-delivered events, so message handling stays consistent (session keys, dedup, reply contexts, etc.).
- **Clean lifecycle wiring.** `startCatchupPoller` runs only after the WebSocket comes up; `stopCatchupPoller` is wired into `Platform.Stop()` alongside the existing group-filter supervisor — no goroutine leaks across restart cycles.

- **No user-facing documentation.** The two new opt-in config keys (`catchup_chats`, `catchup_chats_p2p`) aren't mentioned in `docs/usage.md` / `docs/usage.zh-CN.md`. The header comment of `catchup.go` is excellent internal context but doesn't help an operator who doesn't read the source. Suggest a short "Feishu: WebSocket Push Compensation (Catch-up Polling)" section with: when to enable it, what each config key does, the expected 3-minute poll cadence and 5-minute lookback, the dedup guarantee, and a worked example for one group chat + one p2p chat.

- **Polling churn on quiet chats.** Every 3 minutes the poller issues a REST list-message call for every configured chat, even if the chat has been silent since the last poll. For deployments with 10+ catchup chats this is wasteful. Not a correctness issue, but worth a TODO for "skip chats that had no new messages since last poll."

- **`pageSize = 50` is a hard ceiling for busy chats.** A chat with >50 messages in 5 minutes will only have the most recent 50 inspected; older dropped messages within the window won't be caught. The catchup is bounded by design, but operators with very busy chats should know. Suggest documenting the assumption ("catchup_chats is intended for low-to-moderate traffic; for very busy chats, dedup + WebSocket push are the primary path and catchup is a backstop").

- **No pagination handling.** If a poll returns >50 items, only the most recent 50 are processed. For catchup this is acceptable (we're picking up recent misses), but a comment explaining the deliberate omission would help.

### ❓ Questions

- **Failover behaviour under shared-WS primary handoff.** When the shared WebSocket primary fails over to a secondary (your existing failover path), does the new primary's `isWSPrimary` flip to true and pick up catchup polling? I didn't see that wiring in this diff — the `catchupCancel` lifecycle is tied to `Platform.Stop`, not to primary promotion. If the failover path doesn't restart the poller, missed messages during the failover gap would not be caught.

- **Rate-limit (429) handling.** The `withFreshTenantAccessTokenRetry` helper presumably retries on 429s, but I can't see from this diff whether the catchup poller throttles itself when the platform returns 429 (e.g. by extending `catchupInterval` on a burst). For small deployments this won't matter; for tenants with hundreds of catchup chats it could.

### Testing / Risk

- **CI 5/5 green** on head `fd4776c95774b43ffd12db16d5e27d99699eaf66` (run 34048176705: lint, unit, smoke, regression, performance — all success).
- **Test surface is appropriate** for the unit-testable surface (qualification logic, sender filter, mention conversion). End-to-end testing of the REST polling + WS interaction requires a live Feishu tenant and is out of scope for unit tests.
- **Remaining risk**: documentation gap (P2) and busy-chat pagination (P2). The dedup + IsOldMessage + sender-qualify trio keeps correctness intact even if a few extra messages slip through.

### Next step

Landable. Recommend merging once the docs section is added in the same PR (or in a quick follow-up before users notice the feature exists). The two question items can ship as follow-up issues — neither is a blocker for this PR.

PR #1806, opened 2026-09-08
fix(qqbot): forward quoted image attachments to agents

liuzhao1225:
## Summary

QQ group and C2C handlers currently download images only from the current message's `attachments`. An image explicitly carried in a quoted message is lost before the agent receives it. This forwards quoted image attachments alongside current images and deduplicates their normalized URLs.

The change uses the existing `message_reference` representations and the first `msg_elements` entry for quote messages (`message_type = 103`). It does not fetch chat history or change quoted-file handling.

## Type of change

### Automated tests added or updated

- `TestHandleMessage_QuotedImageAttachments` in `platform/qqbot/quoted_images_test.go`: exercises both group and C2C handlers using HTTP-backed payload fixtures; covers empty current text, the supported nested reference fields, combined current/quoted images, URL deduplication, and quote-element boundaries. Checks both downloaded bytes and download requests.
- `TestMessageImageAttachments_NormalizesAndDeduplicatesURLs` and `TestMessageImageAttachments_DoesNotExpandQuotedFiles` in `platform/qqbot/image_attachments_test.go`.
- `TestCUJ_A3_ImageReachesAgent` in `core/cuj_test.go`: strengthens the existing journey with three turns (text with an image, image alone, text alone), actual agent attachment assertions, platform-visible replies, and no stale attachment carryover.

### Regression proof

- [x] Added only `TestHandleMessage_QuotedImageAttachments` to an isolated checkout of pre-fix main (`4000b2338aa6e850c99df54f8b0ed6ed7460b401`). It compiles and fails because quoted images are missing. The same test passes with the fix.

- [x] A — basic conversation
- [x] Updated the existing image journey to cover image-only input and subsequent turns.
- [x] All `TestCUJ` tests pass with `-race`.

Validation on macOS arm64, Go 1.27.1:

```sh
go build ./...
go vet ./...
go test -p 1 ./...
go test -race ./platform/qqbot ./core -run 'TestMessageImageAttachments|TestHandleMessage_QuotedImageAttachments|TestCUJ' -count=1
```

An initial full-suite run alongside build/race checks hit timeouts and temporary-directory cleanup failures in existing antigravity, file-journey, and media-pipeline tests. The full suite passed when rerun with package parallelism limited to one; no unrelated code was changed.

When a QQ client cannot send an image and an @mention together, a user can reply to an image and @mention the bot. If the event includes that quoted image, it reaches the agent even when the current message has no text. The same attachment handling applies to C2C replies.

The documentation explains this event-dependent behavior. An empty @mention does not implicitly select the last image in a group. Validation uses payload fixtures and the engine journey; live QQ client behavior has not been verified as part of this PR.

## Checklist

- [x] `go build ./...` passes
- [x] Full test suite passes with `-p 1`; changed image flow and CUJs also pass with `-race`
- [x] `go vet ./...`, `gofmt`, and `git diff --check` pass
- [x] Regression test fails before the fix and passes after it
- [x] No new production UI strings; documentation is bilingual

## Related

The quote payload shape is documented in Tencent's [QQ Bot Agent SDK DTO](https://github.com/tencent-connect/qqbot-agent-sdk/blob/main/src/qqbot_agent_sdk/dto.py#L149-L159).

### DIFF
--- a/core/cuj_test.go
+++ b/core/cuj_test.go
@@ -29,6 +29,7 @@
 package core
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -127,6 +128,7 @@ type cujAgentSession struct {
 
 	// observed
 	sentPrompts []string
+	sentImages  [][]ImageAttachment
 	closeCount  int
 }
 
@@ -147,9 +149,10 @@ func newCUJAgentSession() *cujAgentSession {
 	}
 }
 
-func (s *cujAgentSession) Send(prompt string, _ string, _ []ImageAttachment, _ []FileAttachment) error {
+func (s *cujAgentSession) Send(prompt string, _ string, images []ImageAttachment, _ []FileAttachment) error {
 	s.mu.Lock()
 	s.sentPrompts = append(s.sentPrompts, prompt)
+	s.sentImages = append(s.sentImages, cloneCUJImages(images))
 	reply := s.reply
 	delay := s.delayMs
 	override := s.nextEventOverride
@@ -208,6 +211,27 @@ func (s *cujAgentSession) getSentPrompts() []string {
 	defer s.mu.Unlock()
 	out := make([]string, len(s.sentPrompts))
 	copy(out, s.sentPrompts)
+	return out
+}
+
+// Copy attachment bytes as well as the slice so observations remain independent
+// of later caller mutations and can be read safely while the session is running.
+func cloneCUJImages(images []ImageAttachment) []ImageAttachment {
+	out := make([]ImageAttachment, len(images))
+	for i, image := range images {
+		out[i] = image
+		out[i].Data = append([]byte(nil), image.Data...)
+	}
+	return out
+}
+
+func (s *cujAgentSession) getSentImages() [][]ImageAttachment {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	out := make([][]ImageAttachment, len(s.sentImages))
+	for i, images := range s.sentImages {
+		out[i] = cloneCUJImages(images)
+	}
 	return out
 }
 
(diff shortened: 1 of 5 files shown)


chenhg5 (approved):
## Conclusion: Approve

## Overall assessment
PR #1806 is mergeable. CI is 5/5 green on the current head (`4d1800b1`), `mergeable: MERGEABLE`, and the substantive change is a focused bugfix: QQ group and C2C handlers previously dropped images carried in a quoted message, only forwarding images from the current message's `attachments`. The fix introduces a single `messageImageAttachments` helper that combines current + quoted images, deduplicates by normalized URL, and is used by both `handleGroupMessage` and `handleC2CMessage`. No P0/P1 blockers.

## Review scope
- Diff (5 files, +266/-29): `core/cuj_test.go`, `docs/qqbot.md`, `platform/qqbot/image_attachments_test.go` (new), `platform/qqbot/qqbot.go`, `platform/qqbot/quoted_images_test.go` (new).
- Walked through `messageImageAttachments` (URL normalization, dedup, three reference types, `message_type=103` element-zero gating), the wiring change at `qqbot.go:1204` and `qqbot.go:1291`, the bilingual doc addition, the strengthened `TestCUJ_A3_ImageReachesAgent`, the 8 sub-cases × 2 surfaces in `TestHandleMessage_QuotedImageAttachments`, and the unit tests `TestMessageImageAttachments_NormalizesAndDeduplicatesURLs` + `TestMessageImageAttachments_DoesNotExpandQuotedFiles`.

## ✅ What looks good
- **Scope is explicit and well-bounded**: the PR body and the comment block at `qqbot.go:1489` both state "It never looks up an unrelated previous message" and "does not fetch chat history or change quoted-file handling". This is the right scope for a bugfix — only images explicitly carried in the event are forwarded. The doc note "An empty @mention does not implicitly select the last image in a group" makes the negative claim visible to operators.
- **URL normalization matches the downloader**: `strings.TrimSpace(url)` + prepend `https://` if not `http`-prefixed (`qqbot.go:1498-1501`). This is exactly the normalization `downloadAttachmentImages` already does, so the dedup key matches what the downloader uses. Without this, `example.com/image.png` (no scheme) and `https://example.com/image.png` would dedup against the wrong key.
- **Deduplication works across all sources**: `seen map[string]struct{}` keyed on normalized URL covers (a) duplicate URLs within current attachments, (b) current + reference overlap, (c) reference + element-zero overlap. The test case `duplicate URL across all sources` pins this.
- **Three reference types covered**: `ref.Message`, `ref.ReferencedMessage`, `ref.SourceMessage` (`qqbot.go:1509-1515`). The author went looking for all three (some SDKs use one name, some use another). Defensive and correct.
- **`message_type == msgTypeQuote` gating**: `elements[0]` is only consulted when this is a quote message (`qqbot.go:1517`). The test case `only element zero belongs to quote` (with `elements=[["quoted.png"], ["ignored.png"]]`) pins that elements[1], elements[2], etc. are NOT treated as quoted content. This is an important over-fetching guard.
- **Filter only images**: `!strings.HasPrefix(att.ContentType, "image/")` skips non-image attachments. `TestMessageImageAttachments_DoesNotExpandQuotedFiles` explicitly pins that PDF/DOC files in quoted messages are NOT forwarded. Intentional and tested.
- **Empty current text case is tested**: `quoted image with empty current text` (`quoted_images_test.go:307`) pins the user's manual case (image + @mention + empty body) — which is the exact scenario described in the PR body's "Manual / user-visible behavior change" section.
- **Regression proof on pre-fix main**: PR body explicitly states "Added only `TestHandleMessage_QuotedImageAttachments` to an isolated checkout of pre-fix main (`4000b2338aa6e850c99df54f8b0ed6ed7460b401`). It compiles and fails because quoted images are missing. The same test passes with the fix." This is great engineering practice — proves the test would have caught the bug.
- **Bilingual docs**: `docs/qqbot.md` adds EN + ZH-CN sections matching the existing convention. The note about empty @mention not auto-selecting last group image is in both languages.
- **`TestCUJ_A3_ImageReachesAgent` is a strict superset**: original test covered text+image in one turn. New test covers 3 turns (text+image, image-only, text-only) and asserts no stale attachment carryover across turns. `cloneCUJImages` deep-copies attachment bytes so observations remain independent of caller mutations. Concurrency-safe via `cujAgentSession.mu`.
- **`core.HTTPClient` swap pattern**: matches existing convention for HTTP-fixture-based tests. `t.Cleanup` restores the original client. Tests don't call `t.Parallel()`, so concurrent swap conflicts are avoided.
- **CI 5/5 green**: lint, unit-test (4m54s), smoke-test (32s), regression-test (34s), performance-test (57s) — all on run 34184416273.

- **`messageType == msgTypeQuote` hardcodes the int 103**: depends on QQ SDK constant stability. If QQ ever changes the value, this silently stops consulting `elements[0]`. Worth a one-line comment cross-referencing the SDK DTO URL (which the PR body already cites) so a future reader knows where the constant comes from.
- **`ref.ReferencedMessage`, `ref.SourceMessage` may be over-engineered**: the SDK documentation link in the PR body points to one canonical field. If two of the three fields are unused or always-nil in practice, this is defensive code for non-existent input. Probably fine to keep (be liberal in what you accept), but if SDK behavior is known, a comment like "Tencent SDK currently exposes only `Message`" would help future readers prune.
- **URL normalization is duplicated with `downloadAttachmentImages`**: same `TrimSpace` + `https://` prepend logic exists in `qqbot.go:1489-1501` and (presumably) inside `downloadAttachmentImages`. Worth extracting `normalizeAttachmentURL(url string) string` as a small helper to keep the dedup key and the download key guaranteed-identical. A 5-line cleanup that prevents a future drift bug.
- **Doc-only negative claim "An empty @mention does not implicitly select the last image in a group"**: the test pins this for the immediate path (`nonquote elements ignored`), but the engine layer's group-history fetch (or lack thereof) is not asserted. Out of scope for this PR; worth noting for future docs/tests.
- **`messageImageAttachments` walks all elements in `ref.Message.Attachments`** (not just first): correct for SDK-defined reference structure, but if any of the three reference types can carry many attachments, this scales linearly without batching. In practice QQ SDK probably caps this. No action needed.

## 🔵 Optional
- **Doc could mention the bot event payload field names**: `"message_reference"`, `"message_type"`, `"msg_elements"` are the field names operators see in QQ bot logs. A one-line reference would help operators debugging real QQ events. The PR body's SDK URL link covers this implicitly.
- **`messageImageAttachments` could return `[]attachment` and let the caller filter**: current design is more cohesive (one function = one decision). Skip.
- **`add := func(...)` closure captures `images` and `seen`**: idiomatic Go; no action.
- **`msgTypeQuote = 103` constant** could be moved to a named constant near the message type definition if not already. Looking at the diff, the test references `msgTypeQuote` from the qqbot package, so the constant exists. Good.

## ❓ Questions
- **Tencent SDK field stability**: `ref.Message` vs `ref.ReferencedMessage` vs `ref.SourceMessage` — which is canonical for current QQ events? The PR body links `https://github.com/tencent-connect/qqbot-agent-sdk/blob/main/src/qqbot_agent_sdk/dto.py#L149-L159`. The Python DTO uses one name; the Go SDK may use different names. Worth verifying with a real QQ event payload capture before merge.
- **What if `elements[0].Attachments` overlaps with `ref.Message.Attachments` (same image in both)?**: dedup handles it. Good. The `duplicate URL across all sources` test case pins this for current + reference; same logic applies to element-zero + reference.
- **`downloadAttachmentImages` is called inline within `handleGroupMessage`/`handleC2CMessage`**: the test runs these synchronously and asserts `received = msg` is set, which means the handler must be fully synchronous OR the test relies on the handler's internal goroutine to have completed. CI passes with race detector, so this is empirically fine. Worth a comment in `handleGroupMessage`/`handleC2CMessage` saying "currently synchronous; if you add async work, update `TestHandleMessage_QuotedImageAttachments` to wait."

## Testing / Risk
- Verified on current head `4d1800b1`: CI run 34184416273 — `lint` 2m10s ✓, `unit-test` 4m54s ✓, `smoke-test` 32s ✓, `regression-test` 34s ✓, `performance-test` 57s ✓.
- Verified locally (this review): the diff is confined to QQ bot platform + tests + one CUJ test + bilingual docs. No interface changes; the `handleGroupMessage`/`handleC2CMessage` signatures are unchanged.
- Unverified risk: live QQ client behavior has not been verified as part of this PR (the PR body states this explicitly). The tests use HTTP-backed payload fixtures, which is the right approach for a platform adapter. If QQ ever changes its event payload structure (e.g., renames a field), the defensive `for _, quoted := range []*quotedMessage{...}` loop will keep working as long as at least one of the three fields is populated.
- **Race / concurrency**: `core.HTTPClient` swap is global; tests don't call `t.Parallel()` so serial execution is guaranteed. `cloneCUJImages` deep-copies attachment bytes so the test's `getSentImages` observation is independent of concurrent session state.
- **No existing user-visible behavior regression**: the new code only adds images that were previously dropped. Existing images (current message attachments) are unchanged in behavior. The dedup uses the same URL the downloader would use, so current-message images with weird URL formats still work.

## Next step
Merge when convenient. No further code changes required from the author. The P2 items (URL normalization DRY, SDK field comment) are minor cleanups worth doing in a follow-up. The verification questions about Tencent SDK field stability and live QQ event payload are worth a one-line confirmation from the author but don't block the merge — the defensive code covers the realistic cases.

PR #1808, opened 2026-09-08
fix(cron): cap timer span to 30s to recover from system sleep

### LINKED ISSUE #1798: [Bug] cron 定时任务在 macOS 上迟到数小时：调度器 timer 走单调时钟，系统睡眠期间不计时
## 环境

- cc-connect **v1.5.0**（npm 安装，darwin/arm64）
- macOS 26.6.2 (Darwin 25.6.0)，Apple Silicon 笔记本
- 通过 `cc-connect daemon install` 装成 launchd 服务（`com.cc-connect.service`）
- 一个启用的 cron job：`cron_expr = "0 1 * * *"`，`session_mode = new_per_run`

## 现象

配置的是每天 01:00，实际触发时间漂得很厉害，而且每天漂的量都不一样：

| 日期 | 应触发 | 实际触发（日志 `cron: executing job`） | 偏差 |
|---|---|---|---|
| 09-03 | 01:00 | `01:00:00.087` | 准时 |
| 09-04 | 01:00 | `10:57:44` | **+9h57m** |
| 09-05 | 01:00 | `01:17:07` | +17m |
| 09-06 | 01:00 | `03:43:27` | **+2h43m** |

日志里看不出任何异常，`cron: executing job` → `session spawned` → `cron: job completed` 全部正常，只是时间不对。

## 根因

调度器用的是 `github.com/robfig/cron/v3 v3.0.1`（已在二进制符号表中确认）。它的 `run()` 主循环是：

```go
timer := time.NewTimer(c.entries[0].Next.Sub(now))
// ...
case now = <-timer.C:
```

也就是「距离下次触发还有 N 秒，就 arm 一个 N 秒的 timer」。

Go 的 `time.Timer` 底层走单调时钟，而 **macOS 上的单调时钟在系统睡眠期间是停走的**：

- [golang/go#35012 — time: NewTimer firing later if computer sleeps](https://github.com/golang/go/issues/35012)
- [golang/go#66870 — time: macOS monotonic time paused during sleep](https://github.com/golang/go/issues/66870)（closed as not planned）

Go 官方立场是 `time` 包文档已注明 "monotonic clock may stop"，属于平台行为，不打算在 runtime 层面修，需要调用方自己处理。

对一个 daily 任务来说，这个 timer 的跨度接近 **24 小时**，而在这 24 小时里调度器**一次都不回头看墙钟**。所以区间内累计的系统睡眠时长会 1:1 转化成迟到量。

## 定量验证

取偏差最干净的 09-06 那次：

```
迟到量                                    2h43m27s   (01:00:00 → 03:43:27)
pmset -g log 统计的同区间系统睡眠总时长      2h44m34s
```

误差 <0.7%，且这点误差来自 pmset 日志轮转 —— 能覆盖到的区间起点是 09-05 12:22，而 timer 实际 arm 于 09-05 01:17。

**对照实验**（无意中做的，但结论很干净）：09-05 18:45:26 → 09-06 08:10:00 这 13h24m 我用 Amphetamine 持有了 `PreventUserIdleSystemSleep`，这个窗口内 `pmset -g log` 的 Sleep 事件数为 **0**。而任务在窗口正中间的 03:43:27 触发。

也就是说：timer 在「机器醒着」的窗口内是严格按墙钟速率走的，全部迟到量都来自窗口**开始之前**累计的睡眠。这正是单调时钟 timer 的特征 —— 迟到量只取决于区间内的睡眠总时长，与睡在哪一段无关。

另外注意 09-04 那次：上一次 arm 发生在 09-03 15:44 的 `cron: scheduler started`（daemon 重启），之后整夜睡眠约 9h58m，于是 01:00 的任务一路推到 10:57。

## 影响范围

- 所有把 cc-connect 装在笔记本上的用户。macOS 电池模式默认 `sleep 1`（闲置 1 分钟即睡），Windows 同理。
- 偏差随睡眠时长线性增长，可达小时级；**夜间任务受影响最严重**，而夜间恰好是定时任务最常见的时段。
- 属于静默失败：任务确实跑了，日志也正常，用户不看时间戳发现不了。
- 现有的规避手段（保持插电、常驻防睡眠 App、定时唤醒 + 重启 daemon）本质上都只是在缩短 timer 能漂移的窗口，不解决问题，而且要往系统里塞额外组件。

## 建议的修法

**方案 1（最小改动，推荐）**：给 timer 加一个上限，超时后重新比对墙钟。

```go
const maxTimerSpan = 30 * time.Second
d := next.Sub(now)
if d > maxTimerSpan {
    d = maxTimerSpan
}
timer := time.NewTimer(d)
```

醒来后的第一个 tick 就会发现墙钟已经越过目标时刻并立即触发。误差上界固定为 tick 间隔（30s），与睡眠时长完全解耦。改动局限在 `CronScheduler` 内部，不需要换库，也不引入平台相关代码。

**方案 2**：不用 `robfig/cron` 的 `Start()`，改成自己维护一个墙钟 ticker，每 tick 遍历 entries 比对 `Next`。本质同方案 1。

**方案 3**：订阅系统唤醒事件（macOS IOKit `IORegisterForSystemPower`、Windows `WM_POWERBROADCAST`）在唤醒时重新 arm。效果最好但跨平台成本高，收益相对方案 1 有限。

**另外想请教一下设计取向**：睡过头的任务，期望语义是什么？

- 当前行为：醒来后立刻补跑一次（时间不对，但不丢）
- 传统 cron：直接跳过
- launchd `StartCalendarInterval`：唤醒后补跑一次

如果修成方案 1，跳过的那些触发点是要补跑、跳过、还是可配置？对 daily brief 这类任务，「晚几分钟但跑了」通常好过「跳过」，但对某些任务可能相反。也许值得给 job 加一个 `catch_up` 或 `misfire_policy` 字段。

## 相关

- [golang/go#35012](https://github.com/golang/go/issues/35012)
- [golang/go#66870](https://github.com/golang/go/issues/66870)
- [robfig/cron](https://github.com/robfig/cron) —— 这个库设计时假定跑在服务器上，没有考虑过宿主机会睡眠

如果方案 1 的方向 OK，我可以提个 PR。

chenhg5:
## Summary

Fixes #1798.

The third-party `robfig/cron/v3` v3.0.1 scheduler uses Go's monotonic clock for its internal sleep timer, which does not tick while the OS is suspended (laptop sleep on macOS/Windows). After wake-up, the timer still reports the original long interval as "not elapsed", so missed firings never catch up. A long-running cron job (e.g. 8h interval) on a sleeping laptop would not run until the laptop has been awake for the full scheduled interval.

This PR replaces the third-party Cron loop with a custom scheduler that:

- parses cron expressions via `cron.ParseStandard` (kept for semantics)
- iterates entries under `sync.RWMutex` and sleeps at most `maxCronTimerSpan` (30s, a private const so users cannot weaken the recovery cap)
- re-evaluates the earliest `nextRun` on every wake-up, so jobs that were due during sleep fire immediately on resume
- supports `AddJob`/`RemoveJob`/`EnableJob`/`DisableJob`/`UpdateJob` during the run via a buffered `wakeUp` channel

## Test plan

- [x] `go test ./core/ -run TestCron -race` passes 1.7s
- [x] `go vet ./core/` clean
- [x] `gofmt` clean
- [x] `go build ./core/` clean
- [x] New unit tests cover:
  - Start/Stop lifecycle with buffered channels
  - Past-due job fires immediately on next wake (sleep-recovery baseline)
  - Backdate simulation: job whose `nextRun` is in the past fires on next loop tick
  - `AddJob` during run wakes the loop
  - `RemoveJob` during run prevents fire
  - `DisableJob` stops firing
  - Multiple jobs all fire and advance `nextRun`

## Public API compatibility

The public API is preserved. All existing callers work without changes:
`NewCronScheduler`, `Start`, `Stop`, `AddJob`, `RemoveJob`, `EnableJob`, `DisableJob`, `UpdateJob`, `NextRun`, `RunJobNow`, `RegisterEngine`, `SetDefaultSilent`, `SetDefaultSessionMode`, `IsSilent`, `UsesNewSession`, `Store`.

## Why the cap is a private const (not a config knob)

Exposing `maxCronTimerSpan` as a config would invite users to lower it (cheaper catch-up) at the cost of more idle wakeups, which is the wrong default for headless servers. Bug-fix scope, so kept private.

### DIFF
--- a/core/cron.go
+++ b/core/cron.go
@@ -409,23 +409,54 @@ func toExportedFieldName(s string) string {
 	return string(result)
 }
 
+// maxCronTimerSpan caps the sleep duration between wake-ups so the scheduler
+// recovers from system suspend (laptop sleep on macOS/Windows). Without the
+// cap, a `time.NewTimer(d)` for the full cron interval (e.g. 8 hours) does not
+// fire while the OS is suspended, and once the OS wakes, Go's monotonic clock
+// still shows the original duration as "not elapsed" — so the timer fires
+// hours late and catch-up never happens.
+//
+// With the cap, the loop wakes at least every maxCronTimerSpan and on each
+// wake-up recomputes the next firing time from `time.Now()`. A job whose
+// `nextRun` is already in the past fires immediately on the next iteration,
+// so missed ticks during sleep are caught up promptly after wake.
+//
+// Value is intentionally a private constant: exposing it as a config knob
+// would invite users to lower it (cheaper catch-up) at the cost of more idle
+// wakeups, which is the wrong default for headless servers. Bug-fix scope.
+const maxCronTimerSpan = 30 * time.Second
+
+// cronEntry is one scheduled job's bookkeeping inside CronScheduler.
+type cronEntry struct {
+	jobID    string
+	schedule cron.Schedule
+	nextRun  time.Time // next firing time; zero = schedule unsatisfiable
+	prevRun  time.Time // last firing time; zero = never fired
+}
+
 // CronScheduler runs cron jobs by injecting synthetic messages into engines.
 type CronScheduler struct {
 	store              *CronStore
-	cron               *cron.Cron
 	engines            map[string]*Engine // project name → engine
+	entries            map[string]*cronEntry
 	mu                 sync.RWMutex
-	entries            map[string]cron.EntryID // job ID → cron entry
-	defaultSilent      bool                    // global default for suppressing cron start notifications
-	defaultSessionMode string                  // global default session mode; "" = reuse, "new_per_run" = fresh session each run
+	defaultSilent      bool   // global default for suppressing cron start notifications
+	defaultSessionMode string // global default session mode; "" = reuse, "new_per_run" = fresh session each run
+
+	// runLoop plumbing. wakeUp carries a "schedule changed, re-evaluate"
+	// signal (buffered 1). stop ends the loop. done closes when the loop
+	// returns. running guards Start/Stop against double-init.
+	wakeUp  chan struct{}
+	stop    chan struct{}
+	done    chan struct{}
+	running bool
 }
 
 func NewCronScheduler(store *CronStore) *CronScheduler {
 	return &CronScheduler{
 		store:   store,
-		cron:    cron.New(),
 		engines: make(map[string]*Engine),
-		entries: make(map[string]cron.EntryID),
+		entries: make(map[string]*cronEntry),
 	}
 }
 
@@ -461,6 +492,21 @@ func (cs *CronScheduler) UsesNewSession(job *CronJob) bool {
 }
 
 func (cs *CronScheduler) Start() error {
+	cs.mu.Lock()
+	if cs.running {
+		cs.mu.Unlock()
+		return nil
+	}
+	cs.running = true
+	// Buffered so a Stop signal sent before runLoop has started (or
+	// during the small window before runLoop reaches its select) is not
+	// dropped. Wake-up signaling is also buffered so a flurry of
+	// AddJob/RemoveJob collapses to a single re-evaluation.
+	cs.wakeUp = make(chan struct{}, 1)
+	cs.stop = make(chan struct{}, 1)
+	cs.done = make(chan struct{})
+	cs.mu.Unlock()
+
 	jobs := cs.store.List()
 	for _, job := range jobs {
 		if job.Enabled {
@@ -469,13 +515,138 @@ func (cs *CronScheduler) Start() error {
 			}
 		}
 	}
-	cs.cron.Start()
+	go cs.runLoop()
 	slog.Info("cron: scheduler started", "jobs", len(jobs))
 	return nil
 }
 
+// Stop signals the runLoop to exit and waits for it to return. Safe to call
+// when Start was never invoked.
 func (cs *CronScheduler) Stop() {
-	cs.cron.Stop()
+	cs.mu.Lock()
+	if !cs.running {
+		cs.mu.Unlock()
+		return
+	}
+	cs.running = false
+	stopCh := cs.stop
+	doneCh := cs.done
+	cs.mu.Unlock()
+
+	select {
+	case stopCh <- struct{}{}:
+	default:
+	}
+	<-doneCh
+}
+
+// signalWakeUp nudges the runLoop to re-evaluate the next firing time.
+// Non-blocking: if a signal is already pending, the buffered channel absorbs
+// the dedup. Safe to call before Start (no-op) and after Stop (no-op).
+func (cs *CronScheduler) signalWakeUp() {
+	cs.mu.RLock()
+	ch := cs.wakeUp
+	cs.mu.RUnlock()
+	if ch == nil {
+		return
+	}
+	select {
+	case ch <- struct{}{}:
+	default:
+	}
+}
+
+// runLoop is the scheduler's main goroutine. It computes the earliest
+// scheduled job, sleeps at most maxCronTimerSpan, fires any jobs whose
+// `nextRun` has elapsed, and repeats.
+//
+// The bounded sleep is the recovery mechanism for system suspend: a long
+// `time.NewTimer` (e.g. hours) would not fire while the OS is asleep and
+// Go's monotonic clock wouldn't have advanced, so we'd miss catch-up. The
+// cap guarantees we re-check at most every maxCronTimerSpan after wake.
+func (cs *CronScheduler) runLoop() {
+	defer close(cs.done)
+	for {
+		// Compute the earliest entry under read lock, then drop the lock
+		// for the timer wait so AddJob/RemoveJob don't block on sleep.
+		cs.mu.RLock()
+		var earliest *cronEntry
+		for _, e := range cs.entries {
+			if e.nextRun.IsZero() {
+				continue
+			}
+			if earliest == nil || e.nextRun.Before(earliest.nextRun) {
+				earliest = e
+			}
+		}
+		cs.mu.RUnlock()
+
+		var wait time.Duration
+		if earliest == nil {
+			// No scheduled jobs. Sleep long; we'll be woken by wakeUp
+			// when something is added.
+			wait = time.Hour
+		} else {
+			d := time.Until(earliest.nextRun)
+			if d < 0 {
+				// Already overdue (e.g. system just woke from sleep
+				// past the scheduled time). Fire on the next wakeUp
+				// tick instead of burning CPU in a tight loop.
+				d = 0
+			}
+			if d > maxCronTimerSpan {
+				d = maxCronTimerSpan
+			}
+			wait = d
+		}
+
+		timer := time.NewTimer(wait)
+		select {
+		case <-timer.C:
+			cs.fireDueJobs()
+		case <-cs.wakeUp:
+			timer.Stop()
+			// Re-iterate: a job may have been added/removed, or its
+			// schedule changed. Recompute earliest.
+		case <-cs.stop:
+			timer.Stop()
+			return
+		}
+	}
+}
+
+// fireDueJobs runs every entry whose nextRun has elapsed and advances their
+// nextRun. Jobs whose `Enabled` flag was flipped off since the entry was
+// scheduled are silently skipped (runJob re-checks the store).
+func (cs *CronScheduler) fireDueJobs() {
+	now := time.Now()
+
+	cs.mu.RLock()
+	var due []*cronEntry
+	for _, e := range cs.entries {
+		if !e.nextRun.IsZero() && !e.nextRun.After(now) {
+			due = append(due, e)
+		}
+	}
+	cs.mu.RUnlock()
+
+	for _, e := range due {
+		job := cs.store.Get(e.jobID)
+		if job != nil {
+			cs.runJob(job, false)
+		}
+
+		// Advance nextRun under the write lock so concurrent
+		// AddJob/RemoveJob can't race with the schedule refresh. If
+		// the entry was removed mid-fire (job deleted), skip.
+		cs.mu.Lock()
+		cur, ok := cs.entries[e.jobID]
+		if ok && cur == e {
+			cur.prevRun = cur.nextRun
+			cur.nextRun = cur.schedule.Next(time.Now())
+		}
+		cs.mu.Unlock()
+	}
 }
 
 func (cs *CronScheduler) AddJob(job *CronJob) error {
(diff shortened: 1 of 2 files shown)


chenhg5:
## Conclusion: Approve

## Overall assessment
PR #1808 is mergeable. CI is 5/5 green on the current head (`15051f94`), `mergeable: MERGEABLE`, and the substantive change is a clean, well-scoped bugfix for a real bug: the third-party `robfig/cron/v3` v3.0.1 scheduler uses Go's monotonic clock for its internal sleep timer, which doesn't tick while the OS is suspended (laptop sleep on macOS/Windows), so long-interval cron jobs miss firings after wake. The fix replaces the third-party loop with a custom scheduler that sleeps at most 30s (`maxCronTimerSpan`, a private const), re-evaluates `nextRun` from `time.Now()` on every wake-up, and fires past-due jobs immediately. Public API is preserved. No P0/P1 blockers.

## Review scope
- Diff (2 files, +663/-53): `core/cron.go`, `core/cron_test.go`.
- Linked issue: chenhg5/cc-connect#1798 — well-written bug report with quantitative verification (sleep total = late total within 0.7%, cleanly attributed to `pmset -g log` rotation boundary).
- Walked through: `CronScheduler` struct changes, `Start`/`Stop`/`signalWakeUp`/`runLoop`/`fireDueJobs`, `scheduleJob`/`RemoveJob`/`DisableJob`/`UpdateJob` rewrites, `NextRun` simplification, all 7 new test cases.

## ✅ What looks good
- **Root cause is real, deeply analyzed, and the fix matches the reporter's recommended "方案 1"**: `time.NewTimer(hours)` doesn't tick during OS suspend; Go's monotonic clock keeps reporting "not elapsed" after wake. The PR body's verbatim quote of Go issue #35012 / #66870 (closed as "not planned" — Go's position is platform behavior, caller's problem) is exactly correct. The reporter's bug report includes a quantitative verification (sleep total == late total, 0.7% error explained by `pmset` log rotation). The fix decouples drift from sleep duration by bounding the timer span at 30s.
- **`maxCronTimerSpan` is a private const, not a config knob**: the comment block at `core/cron.go:412` and the PR body's "Why the cap is a private const" section explicitly justify this — exposing it as config would invite users to weaken it for "cheaper catch-up" at the cost of more idle wakeups on headless servers. Bug-fix scope, kept private. Sound engineering judgment; agrees with my own read of the tradeoff.
- **Public API is preserved**: the diff keeps every exported method signature — `NewCronScheduler`, `Start`, `Stop`, `AddJob`, `RemoveJob`, `EnableJob`, `DisableJob`, `UpdateJob`, `NextRun`, `RunJobNow`, `RegisterEngine`, `SetDefaultSilent`, `SetDefaultSessionMode`, `IsSilent`, `UsesNewSession`, `Store`. Existing callers (notably `core.Engine`) work without changes. The PR body lists them explicitly.
- **Concurrency model is well-thought-out**:
  - `wakeUp`, `stop`, `done` channels are buffered 1 so signals sent before runLoop starts (or during the small window before select) are not dropped (`core/cron.go:486-488`).
  - `signalWakeUp` reads the channel pointer under `RLock` then sends outside the lock — avoids holding the read lock during the channel send, prevents the lock-then-send deadlock class.
  - `fireDueJobs` uses the `cur, ok := cs.entries[e.jobID]; if ok && cur == e` pattern (`core/cron.go:684`) to detect stale-entry races (entry was removed mid-fire). The `cur == e` pointer check is the right tool here — `jobID` alone wouldn't catch a delete-then-re-add sequence.
  - `Stop` is synchronous (`<-doneCh` after sending stop) — caller can rely on the goroutine being gone when `Stop` returns.
- **Test suite is comprehensive**: 7 new tests cover Start/Stop lifecycle (with double-Stop idempotency), due-job firing baseline, sleep-recovery past-due, AddJob-during-run, RemoveJob-during-run, DisableJob-stops-firing, multiple-jobs-all-fire. The past-due simulation is exactly the right approach — backdate `nextRun` by hours instead of trying to actually sleep the test process. `TestCronScheduler_SleepRecovery_PastDueFiresImmediately` is the canonical regression pin.
- **Comments explain the why, not the what**: every code block has a rationale comment (e.g., "Already overdue (e.g. system just woke from sleep past the scheduled time). Fire on the next wakeUp tick instead of burning CPU in a tight loop" at `core/cron.go:620`). This is the kind of comment that survives a refactor.
- **CI 5/5 green** including unit-test (4m44s — reasonable for 7 new tests with race detector enabled per the test plan's `go test ./core/ -run TestCron -race`), smoke, regression, performance.

- **`signalWakeUp` and the Stop→Start race** (`core/cron.go:546`): the comment says "Safe to call before Start (no-op) and after Stop (no-op)" but doesn't cover the race where `Start()` creates a *new* `wakeUp` channel and a stale `signalWakeUp` call from an earlier session sends to the *old* channel. With buffered-1 + dedup, a stale send is a dropped signal (the old channel is unreferenced), which is benign — but a one-line comment about the race would lock in the invariant for future readers.
- **`Stop()` waits synchronously on `doneCh` while a `fireDueJobs()` may be mid-run** (`core/cron.go:514`): if `Stop()` is called right after the runLoop entered `fireDueJobs`, the stop signal waits for `runJob` to finish for every due entry. `runJob` dispatches messages to engines, which can take seconds. Probably acceptable for an admin-triggered Stop, but worth documenting that Stop is "wait for in-flight fires" not "abort in-flight fires". If the latter semantic is wanted, `runJob` needs to be cancellable.
- **`fireDueJobs` advances `nextRun` using `time.Now()` AFTER each job's `runJob` returns** (`core/cron.go:689`): if `runJob` takes 5s for entry N, entry N+1's `nextRun` is computed 5s later than the firing iteration began. This is the correct semantic ("next firing from now"), but could surprise users who expect "next firing from when the schedule was originally due". Worth a one-line comment in `fireDueJobs` to lock in the intent.
- **The `fireDueJobs` loop runs sequentially** (`core/cron.go:680`): if 100 jobs are due after a long sleep, they're processed serially in the runLoop goroutine. The loop can't respond to new `wakeUp` signals during this window. Probably fine in practice (100 cron jobs all due at once is rare), but if anyone runs cron at sub-minute intervals (e.g., `* * * * *` × 60 minutes of backlog = 60 due jobs), the loop blocks for the duration. A P3 follow-up could batch + parallelize, but not for this PR.

## 🔵 Optional
- **`time.NewTimer` + explicit `Stop()` vs `time.After`**: the explicit form is correct here (we need to call `Stop()` to release the timer), so this is idiomatic. Mentioning it only because the pattern is sometimes debated.
- **`signalWakeUp` could use `close(cs.wakeUp)` + recreate-on-wake** instead of buffered-1: the current design is simpler and matches the goroutine pattern used elsewhere. Not worth changing.
- **`maxCronTimerSpan` could be exposed as a build flag (`-ldflags "-X ..."`)**: would let power users tune it without source changes. Probably not worth the build-system complexity for a 30s heuristic.
- **`Stop()` could return early if the goroutine has already been observed exited**: avoids re-acquiring the mutex for the no-op case. Current behavior is correct (double-Stop is a no-op per `TestCronScheduler_StartStopClean`), just a minor micro-optimization.

## ❓ Questions
- **Misfire policy**: the issue reporter asked "睡过头的任务是补跑、跳过、还是可配置？" The PR implements "补跑一次" (fire on next wake, then advance `nextRun`). The PR body doesn't address the reporter's `catch_up` / `misfire_policy` field suggestion. Is the current behavior the intended v1, or is `misfire_policy` on the roadmap? Worth a brief mention in the PR body or a follow-up issue.
- **`Stop()` and `runJob` cancellation**: see the P2 above. If `runJob` for entry N panics, does `Stop()` ever return? The runLoop's defer `close(cs.done)` ensures the goroutine exits, but a panic inside `fireDueJobs` before the defer would still close `done` (deferred functions run during panic unwinding). So yes, `Stop()` will return even after a panic. But subsequent entries are silently skipped — there's no `defer recover()` to log the panic. Worth a defensive check.
- **`UpdateJob` race with `fireDueJobs`**: if `UpdateJob` removes an entry mid-`fireDueJobs`, the `cur, ok := ...; if ok && cur == e` check at `core/cron.go:684` is the right guard. But if `UpdateJob` then immediately `signalWakeUp`s (which it does, since the `cron_expr` or `enabled` branch removed the entry), the wakeUp signal lands in the same runLoop that's about to re-enter its `select` after fireDueJobs returns. The next iteration recomputes the earliest — but with the *new* schedule already added by `UpdateJob`'s re-add path? Let me trace: `UpdateJob` deletes the entry, then `scheduleJob` is called from the `needsReschedule` branch (not visible in this diff, presumably in the existing UpdateJob body that calls `scheduleJob`). If `scheduleJob` re-adds the entry before `runLoop`'s next iteration, the wakeUp is correctly redundant. If `scheduleJob` runs concurrently with the runLoop's `RLock`, the runLoop sees either the old or new entry but never an inconsistent state. Acceptable, just confirming the author walked through this.

## Testing / Risk
- Verified on current head `15051f94`: CI run 34247778860 — `lint` 2m9s ✓, `unit-test` 4m44s ✓, `smoke-test` 39s ✓, `regression-test` 34s ✓, `performance-test` 55s ✓.
- Verified locally (this review): the diff is confined to `core/cron.go` and `core/cron_test.go`. No interface changes; the public API is preserved. The third-party `robfig/cron/v3` import is *still used* (for `cron.ParseStandard` and `cron.Schedule`); only the loop and entry management were rewritten. So the dependency remains.
- Unverified risk: existing cc-connect installations on laptops that have been silently drifting for months (per the issue reporter's quantitative data) will suddenly start firing on time after this PR lands. This is the desired behavior, but operators who relied on the drift (unlikely but possible) will see schedule changes.
- **Race / concurrency**: the new runLoop is single-goroutine; signals are non-blocking; state mutations are under `mu.Lock()`. The `signalWakeUp` reads the channel pointer under `RLock` and sends outside the lock — this is the right pattern. Stop is synchronous.
- **Orphan timers**: `time.NewTimer` instances are explicitly `Stop()`ed in the `wakeUp` and `stop` select cases. The `timer.C` case doesn't need an explicit Stop because the timer has already fired and the channel will be drained. No resource leak.

## Next step
Merge when convenient. No further code changes required from the author. The P2 items (signalWakeUp race comment, Stop semantics, fireDueJobs nextRun advancement comment, sequential fire loop) and the misfire-policy question are all worth addressing in a follow-up PR or issue, but none block this bugfix. The registry still showed `ci_status: pending` for PR #1808 when this cycle started (5-min sync freshness lag); `gh pr checks` confirmed all 5 jobs actually passed at 2026-09-08T16:09:48Z.

xiaobao520123:
@chenhg5 When would you release a new version that contains the fix? I did notice sync issue on my Mac.

PR #1809, opened 2026-09-08
fix: graceful SIGTERM drain + reply journal for hard-kill recovery (#1804)

### LINKED ISSUE #1804: [Bug] Graceful shutdown drops all in-flight replies (context canceled) — never re-sent after restart
## Environment

- cc-connect 1.4.1, macOS arm64, managed by launchd (KeepAlive)
- Platform: Feishu / Agent: Claude Code
- Agent sessions run as cc-connect child processes (exit on stdin close)

## Problem

When SIGTERM arrives, any platform replies that are mid-flight are **dropped**:

```log
level=INFO  msg="shutting down..."
level=ERROR msg="platform send failed" platform=feishu error="feishu: reply api call: Post \"https://open.feishu.cn/open-apis/im/v1/messages/om_xxx/reply\": context canceled" content_len=48
level=INFO  msg="claudeSession: exited cleanly after stdin close"
```

The message is never delivered, and the new instance does not re-send it after restart — it's simply gone.

## Repro (100%)

1. Send a message that triggers an agent turn (any task lasting a few seconds)
2. While the turn is running and the reply is being sent, send SIGTERM to cc-connect (e.g. `launchctl kickstart -k`)
3. The new instance starts, sessions are restored from disk — but the user never receives the in-flight reply

A very common case: **the agent itself restarts cc-connect** (e.g. after a config change, the restart is the last tool call of its turn). This reliably produces two outcomes:

1. The user receives nothing and has to ask "is it done?"
2. The agent's turn dies mid-flight (no `turn complete` in the log); in the session history the turn ends as interrupted with no result for the final tool call. The resumed agent reads the history, reasonably concludes "the restart never happened", and **restarts it again**

Together these form a loop: every time the user follows up, the agent repeats the self-destructive restart, and each reply is eaten by `context canceled`. Observed three consecutive rounds today — two `platform send failed` per shutdown (replies are sent in chunks; all lost).

## Expected behavior

On graceful shutdown:

1. Persist in-flight / pending replies to disk (sessions are already persisted; replies could follow the same mechanism)
2. After restart, re-send them — or at minimum deliver an "reply interrupted by restart" notice, so neither the user nor the agent is left blind

## Current workaround

Before restarting, schedule `cc-connect timer add --delay 15s --prompt "verify restart result and report"`, then kickstart — the new instance fires the timer and the agent proactively reports back. But the original reply body is still lost.

chenhg5:
Fixes #1804.

When cc-connect received SIGTERM mid-platform-reply, the shared engine
context was cancelled immediately, the platform API call returned
'context canceled', and the reply body was lost. This commonly happens
when an agent ends a long session by restarting cc-connect as its last
tool call — the user sees nothing for that final turn and the agent
restarts again, producing a feedback loop.

### Graceful drain on SIGTERM (issue #1804 plan: B)

`Engine.Stop()` now waits up to **10 seconds** for in-flight platform
`Reply` / `Send` operations to finish *before* cancelling the shared
context. A new `replyWG sync.WaitGroup` tracks the wrappers
(`trackAndSend`, `writeJournalEntry`, `clearJournalEntry`) so
`Stop()` can wait for them. The 10s ceiling is a deliberate trade-off:
long enough for a typical Feishu reply + a chunked burst, short enough
that a misbehaving agent cannot block daemon restart.

### Minimal reply journal (issue #1804 plan: A-minimal)

When the drain times out, or the process is killed by OOM / panic /
SIGKILL, `<dataDir>/run/last_reply.json` (mode `0o600`) holds the
most recent pending reply. `Engine.Start()` replays it once after the
platforms come up:

- **Success** → entry cleared.
- **Duplicate error** ("duplicate" / "already sent" / "重复") →
  treated as delivered, entry cleared.
- **Other error** → entry preserved for the next restart.

New optional `SessionKeyCarrier` interface lets platforms opt in to
extracting session keys from reply contexts, so the journal can
associate a pending reply with the right session.

### FAQ / verification workaround (issue #1804 plan: C)

`docs/usage.md` (+ zh-CN) documents `cc-connect timer add --delay 15s --prompt "verify restart result and report"` for agents that want to confirm
the last reply made it through after a self-restart.

## Files

- `core/reply_journal.go` (new) — journal write/consume/replay, hash dedup
- `core/reply_journal_test.go` (new) — 21 unit tests covering round-trip, dedup, drain timeout, replay paths
- `core/engine.go` — `replyWG`, drain-before-cancel, `trackAndSend` wrapper, `replayPendingReplyJournal` startup hook
- `core/interfaces.go` — new `SessionKeyCarrier` optional interface
- `docs/usage.md` + `docs/usage.zh-CN.md` — Graceful Restart & Lost Replies section

## Risk

- Engine startup now blocks for up to 10s while replaying a pending
  reply. Acceptable: the journal is empty on first start, and replays
  fail fast against a fresh context (`replyJournalTimeout = 10s`).
- Drain adds up to 10s to SIGTERM shutdown. Acceptable: this is the
  whole point of the fix, and 10s is a hard ceiling via
  `shutdownDrainTimeout`.
- Platforms without `ReplyContextReconstructor` cannot replay. Same as
  the existing cron path: the journal entry is preserved and logged.

## Test plan

- 21 new unit tests in `core/reply_journal_test.go` (run with -race).
- Full `go test -tags no_web ./core/...` passes.
- Manual: kill -SIGTERM mid-reply, confirm reply delivered; kill -9
  mid-reply, restart, confirm journal replays.

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -83,6 +83,22 @@ const (
 // exits sooner — this is the ceiling, not the typical duration.
 const closeTimeout = 150 * time.Second
 
+// shutdownDrainTimeout caps the wait for in-flight platform reply/send
+// operations to complete before SIGTERM is allowed to cancel the engine's
+// shared context. Without this drain, replies mid-flight observe
+// "context canceled" mid-HTTP-call and are lost — see issue #1804.
+//
+// 10s is a deliberate trade-off:
+//   - long enough for a typical Feishu reply API call (~1-2s) plus a
+//     chunked-response burst (multiple replies within ~50ms of each other),
+//     even on a slow link;
+//   - short enough that a misbehaving agent (one that forgot to release a
+//     long-lived request) cannot block daemon restart for tens of seconds.
+//
+// Issue #1804 ships a journal fallback (see core/reply_journal.go) for the
+// hard-kill case where this drain is bypassed entirely.
+const shutdownDrainTimeout = 10 * time.Second
+
 const (
 	replyFooterUsageTimeout  = 1500 * time.Millisecond
 	replyFooterUsageCacheTTL = 30 * time.Second
@@ -506,6 +522,14 @@ type Engine struct {
 
 	// Data directory for socket path injection
 	dataDir string
+
+	// replyWG tracks in-flight platform Reply/Send operations so that
+	// Engine.Stop() can drain them before cancelling the shared context.
+	// Issue #1804: a context cancel during an active HTTP call to the
+	// platform API surfaces as "context canceled" and the reply is lost.
+	// Increment at the top of every wrapper that calls p.Reply / p.Send,
+	// defer Done at the bottom — the WG is only ever observed by Stop().
+	replyWG sync.WaitGroup
 }
 
 // workspaceInitFlow tracks a channel that is being onboarded to a workspace.
@@ -2350,6 +2374,16 @@ func (e *Engine) Start() error {
 	}
 
 	e.startObserver()
+
+	// Issue #1804: replay any pending reply that was persisted before a hard
+	// kill (OOM, panic, SIGKILL). Runs synchronously so the user sees the
+	// recovered reply before we declare startup complete. Sync platforms are
+	// ready at this point; async platforms that haven't connected yet will
+	// be retried on their first OnPlatformReady via OnPlatformReady hook
+	// (see e.platformReadyReplayOnce).
+	if e.dataDir != "" && len(e.platforms) > 0 {
+		e.replayPendingReplyJournal()
+	}
 	return nil
 }
 
@@ -2358,7 +2392,19 @@ func (e *Engine) Stop() error {
 	e.stopping = true
 	e.platformLifecycleMu.Unlock()
 
-	// Cancel first so late lifecycle callbacks observe shutdown immediately.
+	// Drain in-flight platform reply/send goroutines BEFORE cancelling the
+	// shared context. Issue #1804: a context cancel mid-HTTP-call surfaces
+	// as "context canceled" on the platform side and the reply body is
+	// lost forever (no journal replay covers SIGTERM, only hard kills).
+	//
+	// We bound the wait at shutdownDrainTimeout so a stuck platform call
+	// cannot block daemon restart indefinitely. Anything still in flight
+	// after the deadline is presumed lost — the journal layer (see
+	// core/reply_journal.go) catches the hard-kill case separately.
+	e.drainInFlightReplies(shutdownDrainTimeout)
+
+	// Cancel the shared context AFTER the drain so late lifecycle
+	// callbacks (agent stop, observer shutdown, etc.) observe shutdown.
 	e.cancel()
 
 	if e.observeCancel != nil {
@@ -2404,6 +2450,27 @@ func (e *Engine) Stop() error {
 		return fmt.Errorf("engine stop errors: %v", errs)
 	}
 	return nil
+}
+
+// drainInFlightReplies blocks until all platform reply/send operations
+// tracked by replyWG have completed, or the deadline elapses. Safe to
+// call when no replies are in flight (returns immediately).
+func (e *Engine) drainInFlightReplies(timeout time.Duration) {
+	if timeout <= 0 {
+		timeout = shutdownDrainTimeout
+	}
+	done := make(chan struct{})
+	go func() {
+		e.replyWG.Wait()
+		close(done)
+	}()
+	select {
+	case <-done:
+		slog.Debug("engine.Stop: in-flight replies drained cleanly")
+	case <-time.After(timeout):
+		slog.Warn("engine.Stop: drain timeout, some replies may be lost; relying on journal for retry",
+			"timeout", timeout)
+	}
 }
 
 // OnPlatformReady marks an async platform as ready and initializes platform-level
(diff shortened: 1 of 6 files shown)


chenhg5:
## Conclusion: Approve

This PR delivers the SIGTERM drain + reply journal as scoped by issue #1804. CI is green (5/5), the code is conservative (idempotent clear, fresh context for replay, chmod 0600 for journal files), the test surface is solid (21 unit tests covering happy path, errors, dedup, drain timeout, replay-no-reconstructor), and the user-visible failure mode is genuinely improved. The two async-platform caveats are explicitly documented in `known_risks` and have safe failure modes (entry preserved on disk, not lost).

I posted this as a comment rather than an `Approve` review event because chenhg5 owns the PR — GitHub's API blocks self-approval. The maintainer should still run the regular merge gate.

### ✅ What looks good

- **`shutdownDrainTimeout = 10s` is a well-justified constant.** The accompanying comment spells out the trade-off (long enough for typical Feishu reply + chunked burst, short enough that a misbehaving agent can't block daemon restart). Future maintainers won't need to guess why it's not 5s or 30s.
- **`trackAndSend` writes the journal entry BEFORE the platform call.** That's the right order: a crash mid-call leaves a recoverable record on disk rather than an empty journal.
- **Conservative dedup heuristic.** `isDuplicateReplyError` prefers false negatives over false positives — comment explicitly states "We'd rather send twice than drop." Correct call for chat UX; users tolerate an occasional duplicate far better than a silently missing reply. Multilingual coverage (English + `重复`) is a thoughtful detail.
- **Atomic file writes + `os.IsNotExist`-tolerant clear.** `WritePendingReply` uses `AtomicWriteFile` (existing helper, already battle-tested by the cron journal path), `ClearPendingReply` is idempotent on missing files, and `ConsumePendingReply` does best-effort delete-before-parse so a malformed entry can't loop across restarts.
- **`SessionKeyCarrier` is opt-in.** Platforms don't have to change anything; the engine has graceful fallbacks (`stringer`, bare `string` containing `:`). Zero platform-side breaking change.
- **Fresh `context.Background()` for replay.** Correctly isolates the replay from in-flight `Stop()` cancellation — exactly the bug class the journal is meant to survive.
- **Docs are bilingual and concrete.** Both `docs/usage.md` and `docs/usage.zh-CN.md` document the `cc-connect timer add --delay 15s --prompt "verify restart result and report"` workaround, the journal JSON shape, and how the file lifecycle works. Users won't have to guess.

- **Concurrent-burst journal race.** `trackAndSend` writes `WritePendingReply` then `clearJournalEntry` on success, but doesn't coordinate across concurrent calls. If two `trackAndSend` calls are in flight (A and B) and B writes the journal after A: when A succeeds, A's `clearJournalEntry` removes B's entry from disk; if B then fails (non-dup error), B's content is lost on hard kill. The race window is narrow (a chunked-burst reply where one chunk succeeds and a later chunk fails) and the worst-case is the same as pre-fix behavior (user-visible reply loss), but worth noting in `known_risks` or, better, addressed with a per-call journal tag so `clearJournalEntry` only clears when the on-disk hash matches the call's hash. Skippable for this PR; flag for a follow-up.

- **Async-platform replay gap is real for the dominant use case.** Feishu is async and connects *after* `Start()` returns. The `OnPlatformReady` replay hook mentioned in the comment (`e.platformReadyReplayOnce`) is not in this PR — so on cold start the journal entry persists but does not actually deliver until the next restart. The PR acknowledges this explicitly in `known_risks`, which is the right call for shipping now, but users will be confused if they read the docs ("After restart the entry is replayed automatically") and don't see it. Suggest a one-line caveat in `docs/usage.md` / `usage.zh-CN.md` "Graceful Restart & Lost Replies" — something like "Async platforms (Feishu, Slack, etc.) replay on the platform's first message after startup; for the most common async platforms this may require one additional restart." Or fold the OnPlatformReady hook into a quick follow-up PR so the docs and behavior stay aligned.

- **String-form `sessionKeyFromReplyCtx` heuristic.** The fallback `if strings.Contains(s, ":")` will accept any colon-bearing string as a session key, including a hypothetical platform reply ctx that happens to embed a colon. Unlikely today but a future platform author could be surprised. Not a bug, but the heuristic should probably be tightened to a documented pattern (e.g. `platform:...` prefix match) or extracted into a helper the platform can override.

### ❓ Question

- **`AtomicWriteFile` semantics.** `WritePendingReply` relies on `AtomicWriteFile` for crash-safety. Is this the existing helper used by the cron journal path? If yes, no concern. If a new helper was added in this PR, I'd want to confirm it's been audited for partial-write / fsync behavior — I didn't see it in the diff, just want to confirm it's not a fresh implementation.

### Testing / Risk

- **CI 5/5 green** on the current head `c6a094909cdde0f9d2a4f7847d8b3d6a4c1aea47` (run 34261009035: lint 2m9s, unit-test 4m44s, smoke-test 39s, regression-test 34s, performance-test 55s — all success).
- **Local `-race` coverage** for `core/reply_journal_test.go` (21 tests) plus the rest of `core/...` per dev handoff. The transient `TestCUJ_H2_TwoPlatformsConcurrentNoBleed` flake on first CI run is consistent with the 30s deadline + slow CI host hypothesis; passed cleanly on rerun and locally with `-race`. Not a blocker, but the 30s deadline in that test is uncomfortably tight on CI — if it flakes again in main-branch runs, that's worth widening.
- **Remaining risk:** concurrent-burst journal race (P2, documented above) and async-platform replay timing (P2, documented above). Both have safe failure modes (loss-or-survive = pre-fix loss, or entry preserved for retry).

### Next step

This is mergeable. The two P2 items are real but each has a safe failure mode and either can ship as a quick follow-up. Recommend landing as-is and opening a follow-up issue to cover (a) the OnPlatformReady replay hook for async platforms and (b) the per-call hash-keyed `clearJournalEntry`.

PR #1814, opened 2026-09-09
fix(core): keep original user post time in conversation history

jsvisa:
User history entries were stamped with time.Now() at processing time, so a message queued behind a long turn got the previous turn's response time as its timestamp. Stamp user entries with the platform message creation time (Message.UserMessageTimeMs) instead:

- core/session.go: add AddHistoryAt/AddHistoryEntry helpers (zero time falls back to time.Now()); AddHistory now delegates to them.
- core/engine.go: add userMessageTimestamp() helper and use the original post time at the direct path and both queue-drain sites.

## Type of change

- [ ] Documentation only

- [ ] A — basic conversation

- [ ] `go build ./...` passes

## Related

- Issue:
- Related PR:

### DIFF
--- a/core/engine.go
+++ b/core/engine.go
@@ -535,6 +535,16 @@ type queuedMessage struct {
 	userMessageTimeMs int64  // Feishu create_time ms (optional); see Message.UserMessageTimeMs
 }
 
+// userMessageTimestamp converts a platform message creation time (Unix ms)
+// into a history timestamp. Zero/negative values fall back to the zero time
+// so the caller can stamp time.Now() instead.
+func userMessageTimestamp(timeMs int64) time.Time {
+	if timeMs <= 0 {
+		return time.Time{}
+	}
+	return time.UnixMilli(timeMs)
+}
+
 // interactiveState tracks a running interactive agent session and its permission state.
 type interactiveState struct {
 	agentSession             AgentSession
@@ -3746,7 +3756,7 @@ func (e *Engine) processInteractiveMessageWith(p Platform, msg *Message, session
 	turnStart := time.Now()
 
 	e.i18n.DetectAndSet(msg.Content)
-	session.AddHistory("user", msg.Content)
+	session.AddHistoryAt("user", msg.Content, userMessageTimestamp(msg.UserMessageTimeMs))
 	// Persist user message immediately so crashes between user input and
 	// assistant reply don't lose it (the assistant-side Save below depends
 	// on the turn completing without a process crash).
@@ -6215,9 +6225,9 @@ func (e *Engine) processInteractiveEvents(state *interactiveState, session *Sess
 					e.send(queued.platform, queued.replyCtx, replyContent)
 				}
 
-				session.AddHistory("user", queued.content)
+				session.AddHistoryAt("user", queued.content, userMessageTimestamp(queued.userMessageTimeMs))
 				// Persist queued user message immediately (mirror of the
-				// initial AddHistory("user",...) save above).
+				// initial AddHistoryAt("user",...) save above).
 				sessions.Save()
 
 				if idleTimer != nil {
@@ -6473,7 +6483,7 @@ func (e *Engine) drainPendingMessages(state *interactiveState, session *Session,
 
 		drainEvents(as.Events())
 
-		session.AddHistory("user", queued.content)
+		session.AddHistoryAt("user", queued.content, userMessageTimestamp(queued.userMessageTimeMs))
 
 		sendDone := make(chan error, 1)
 		go func() {
--- a/core/engine_test.go
+++ b/core/engine_test.go
@@ -9108,6 +9108,95 @@ func TestProcessInteractiveEvents_DrainsQueuedMessagesFIFOWithCreateTimes(t *tes
 	}
 }
 
+// TestProcessInteractiveEvents_DrainPreservesQueuedUserMessageTime verifies
+// that a user message queued behind an in-flight turn is stamped in history
+// with its original platform post time (queued.userMessageTimeMs), not the
+// time the queue was drained (which used to be the previous turn's response
+// time).
+func TestProcessInteractiveEvents_DrainPreservesQueuedUserMessageTime(t *testing.T) {
+	p := &stubPlatformEngine{n: "test"}
+	sess := newQueuingSession("qs-drain-time")
+	agent := &controllableAgent{nextSession: sess}
+	e := NewEngine("test", agent, []Platform{p}, "", LangEnglish)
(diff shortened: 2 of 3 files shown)


chenhg5 (approved):
## Review: PR #1814

### What this PR fixes
User history entries were stamped with `time.Now()` at processing time, so a message queued behind a long turn ended up with the previous turn's response time as its timestamp. This PR stamps user entries with the original platform post time (`Message.UserMessageTimeMs` / `queuedMessage.userMessageTimeMs`).

### Diff review

`core/session.go`
- New `AddHistoryAt(role, content, ts time.Time)` — explicit timestamp entry; zero value falls back to `time.Now()`.
- New `AddHistoryEntry(entry HistoryEntry)` — pre-built entry form; zero `Timestamp` falls back.
- Old `AddHistory(role, content)` is now a thin wrapper that delegates with `time.Time{}`.
- The mutex is moved inside `AddHistoryEntry`. Good — avoids any future caller from holding the lock and then accidentally re-entering through `AddHistory`.

`core/engine.go`
- New `userMessageTimestamp(timeMs int64) time.Time` — `<= 0` returns zero `time.Time{}` so the session helper falls back to `time.Now()`. Positive values go through `time.UnixMilli`.
- All three call sites converted:
  1. `processInteractiveMessageWith` (direct path) — uses `msg.UserMessageTimeMs`
  2. `processInteractiveEvents` queue drain — uses `queued.userMessageTimeMs`
  3. `drainPendingMessages` — uses `queued.userMessageTimeMs`
- The "initial AddHistory(...) save" comment in `processInteractiveEvents` was updated to reflect the new helper. Nice touch.

`core/engine_test.go`
- `TestProcessInteractiveEvents_DrainPreservesQueuedUserMessageTime` is a proper regression test:
  - Seeds `pendingMessages` with `userMessageTimeMs: 2_000` and `3_000`.
  - Drives `processInteractiveEvents` with three `EventResult` rounds.
  - Asserts the resulting user history entries have `time.UnixMilli(2_000)` / `time.UnixMilli(3_000)` exactly — which would fail on the pre-fix code (it would be the drain time, not the original post time).
- This is the right shape for a regression test on this bug.

### Concerns / questions
None blocking. A few non-blocking observations:

- `AddHistoryAt`'s zero-value fallback path isn't directly tested. It's exercised indirectly through the unchanged `AddHistory` callers, which is fine, but a one-line test asserting `AddHistoryAt("u", "c", time.Time{})` produces a non-zero `Timestamp` would be cheap insurance.
- `userMessageTimestamp(-1)` isn't explicitly tested either, but `<= 0` is straightforward and the helper only does `time.UnixMilli`.
- On-disk history written before this PR will keep its wrong timestamps until next write; that's correct (no migration needed), but worth a sentence in the release notes.

### Risk
- Pure timestamp correctness fix. Append order in `History` is unchanged, so anything depending on order (which is most code reading history) is unaffected.
- 125+/8- across 3 files, single regression test added, all helpers are pure functions.
- CI: 5/5 green. `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` on head `f5928ef0`.

Merging.

chenhg5 (approved):
## Conclusion: Approve

Clean, focused bugfix that fixes a real UX issue (queued user messages showed the previous turn's response time as their timestamp). The two-layer zero-time fallback design is the right shape: `userMessageTimestamp` returns zero time for missing/negative input, then `AddHistoryAt`/`AddHistoryEntry` normalize zero time to `time.Now()`. This preserves backward compatibility with old platform clients that don't set `UserMessageTimeMs` — identical behavior to before when the field is absent.

The interface migration is textbook non-breaking: `AddHistory(role, content)` is kept as a thin wrapper that delegates to `AddHistoryAt(..., time.Time{})`, and all three internal call sites are migrated to the new explicit-timestamp API. No external caller of `AddHistory` (other than engine itself) is affected.

### What looks good

- **Two-layer zero-time fallback** is correct and well-documented in the helper's comment. Platform without `UserMessageTimeMs` → `userMessageTimestamp` returns zero time → `AddHistoryAt` falls back to `time.Now()` → identical to pre-fix behavior. No silent regressions for older clients.
- **Lock semantics preserved**: `AddHistoryEntry` acquires `s.mu` for the append, same as the old `AddHistory`. `AddHistoryAt` → `AddHistoryEntry` doesn't introduce a new lock acquisition path.
- **All three call sites migrated**: `processInteractiveMessageWith` (direct path), `processInteractiveEvents` (queued msg handler), `drainPendingMessages` (drain path). No stale `AddHistory("user", ...)` calls left in user-message code paths.
- **Regression test is precisely targeted**: `TestProcessInteractiveEvents_DrainPreservesQueuedUserMessageTime` sets `userMessageTimeMs: 2_000` and `3_000` on queued messages, then asserts the resulting `HistoryEntry.Timestamp` equals `time.UnixMilli(2_000)` / `time.UnixMilli(3_000)` — not the drain time. The assertion is exact (not "approximately equal" or "after time X"), so it would catch subtle timezone/unit drift bugs.
- **Comment on `userMessageTimestamp`** explains *why* zero is returned rather than silently coerced to `time.Now()` — the dual-fallback layering is non-obvious without this comment, and it's the kind of "why" comment that prevents a future simplification regression.

### 🔵 Optional (not blocking)

- **P3 — direct path lacks a dedicated test.** `processInteractiveMessageWith` (the non-queued path) uses `userMessageTimestamp(msg.UserMessageTimeMs)` directly, but the regression test only covers the drain path. The direct path is simpler (no queue involved), but a `TestProcessInteractiveMessageWith_PreservesUserMessageTime` would lock in the behavior symmetrically with the drain test and prevent accidental regression if someone refactors the direct-path call site.
- **P3 — multi-message drain chain untested.** The regression test drains exactly two queued messages. A 3+ message chain isn't explicitly covered, but since each message is processed independently through the same `AddHistoryAt` call, the risk is low. Worth a follow-up if you ever batch-process the queue differently.
- **P3 — CHANGELOG note.** Users with `/history`-style consumers will see user-message timestamps shift from "processing time" to "original post time". This is exactly the intended behavior change, but worth a one-line release note for transparency.

### Testing / Risk

- **Verified**: CI 5/5 green on `f5928ef0`. Maintainer chenhg5 approved on current head (not stale). Three call sites all updated. Backward-compat path (zero `UserMessageTimeMs` → `time.Now()`) preserved by the dual fallback.
- **Remaining risk**: P3 items above are coverage gaps, not correctness gaps. No P0/P1 risk identified.

### Next step

Merge-ready. If you have time, add the direct-path test for symmetry, but it's not a blocker.

PR #1656, opened 2026-08-09
fix(codex): isolate subagent app-server events

AaronZ345:
Supersedes #1520. This is the same patch on a fresh branch after the 30-day maintenance gate; the prior PR had no requested changes or unresolved review threads.

## Summary
- ignore Codex app-server lifecycle notifications that belong to subagent threads instead of the active cc-connect thread
- scope item, completion, and token-usage notifications to the active parent turn
- preserve compatibility with older app-server notifications that omit thread or turn IDs
- add regression coverage proving child turns cannot overwrite or complete the parent turn

## Root cause
Codex app-server multiplexes parent and subagent events over the same connection. Its v2 protocol includes `threadId` on `turn/started`, `turn/completed`, and `thread/status/changed`, plus both `threadId` and `turnId` on item and token-usage notifications.

cc-connect parsed those fields but did not validate them. When a subagent started, its `turn/started` notification replaced the parent's `currentTurn` and cleared pending parent messages. When that child thread became idle or completed, cc-connect called `completeTurn()` and emitted an empty `EventResult`. Repeated child turns therefore produced repeated empty responses, while parent output could be truncated. Child token usage could also overwrite the parent turn's usage snapshot.

This change fixes the event-routing boundary rather than suppressing empty messages. Parent-thread follow-up turns, including results delivered back from subagents, continue through the normal unsolicited-event path.

## Live evidence
In a GPT-5.6 Sol `ultra` run, the parent turn completed at 12:15:08. Codex then started subagent turns on distinct thread IDs. Child thread idle transitions at 12:15:15 and 12:15:23 aligned exactly with cc-connect logging `unsolicited turn complete response_len=0`. The Codex 0.144.1 generated protocol schema confirms these notifications carry the thread and turn identifiers needed for routing.

## Validation
- `go test ./agent/codex -count=1`
- `go test -race ./agent/codex -run 'TestAppServerSession_(IgnoresSubagentLifecycleNotifications|IgnoresSubagentTokenUsageNotifications)' -count=1`
- `go test -tags no_web ./cmd/cc-connect ./agent/codex -count=1`
- `go vet ./agent/codex`
- `git diff --check`

PR #1816, opened 2026-09-09
fix(codex): isolate subagent app-server events

AaronZ345:
Supersedes #1656. This is the same patch/tree rebuilt from the latest `origin/main` after the explicit 30-day maintenance gate; the prior PR had no requested changes or unresolved review threads.

## Summary

- ignore Codex app-server lifecycle notifications that belong to subagent threads instead of the active cc-connect thread
- scope item, completion, and token-usage notifications to the active parent turn
- preserve compatibility with older app-server notifications that omit thread or turn IDs
- add regression coverage proving child turns cannot overwrite or complete the parent turn

## Root cause

Codex app-server multiplexes parent and subagent events over the same connection. Its v2 protocol includes `threadId` on `turn/started`, `turn/completed`, and `thread/status/changed`, plus both `threadId` and `turnId` on item and token-usage notifications.

cc-connect parsed those fields but did not validate them. When a subagent started, its `turn/started` notification replaced the parent current turn and cleared pending parent messages. When that child thread became idle or completed, cc-connect called `completeTurn()` and emitted an empty `EventResult`. Repeated child turns therefore produced repeated empty responses, while parent output could be truncated. Child token usage could also overwrite the parent turn usage snapshot.

This change fixes the event-routing boundary rather than suppressing empty messages. Parent-thread follow-up turns, including results delivered back from subagents, continue through the normal unsolicited-event path.

## Validation

- `go test -race ./agent/codex -run 'TestAppServerSession_(IgnoresSubagentLifecycleNotifications|IgnoresSubagentTokenUsageNotifications)' -count=1`
- `go vet ./agent/codex`
- `git diff --check`
- tree equality with #1656 and remote parity `0 0`

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -1124,7 +1124,7 @@ func (s *appServerSession) handleNotification(method string, paramsRaw json.RawM
 	switch method {
 	case "turn/started":
 		var notif turnNotification
-		if err := json.Unmarshal(paramsRaw, &notif); err == nil {
+		if err := json.Unmarshal(paramsRaw, &notif); err == nil && s.acceptsThreadNotification(method, notif.ThreadID) {
 			s.stateMu.Lock()
 			s.currentTurn = notif.Turn.ID
 			s.pendingMsgs = s.pendingMsgs[:0]
@@ -1134,19 +1134,19 @@ func (s *appServerSession) handleNotification(method string, paramsRaw json.RawM
 
 	case "item/started":
 		var notif itemNotification
-		if err := json.Unmarshal(paramsRaw, &notif); err == nil {
+		if err := json.Unmarshal(paramsRaw, &notif); err == nil && s.acceptsTurnNotification(method, notif.ThreadID, notif.TurnID) {
 			s.handleItemStarted(notif.Item)
 		}
 
 	case "item/completed":
 		var notif itemNotification
-		if err := json.Unmarshal(paramsRaw, &notif); err == nil {
+		if err := json.Unmarshal(paramsRaw, &notif); err == nil && s.acceptsTurnNotification(method, notif.ThreadID, notif.TurnID) {
 			s.handleItemCompleted(notif.Item)
 		}
 
 	case "turn/completed":
 		var notif turnNotification
-		if err := json.Unmarshal(paramsRaw, &notif); err == nil {
+		if err := json.Unmarshal(paramsRaw, &notif); err == nil && s.acceptsTurnNotification(method, notif.ThreadID, notif.Turn.ID) {
 			if strings.EqualFold(strings.TrimSpace(notif.Turn.Status), "failed") || notif.Turn.Error != nil {
 				errMsg := ""
 				if notif.Turn.Error != nil {
@@ -1168,7 +1168,7 @@ func (s *appServerSession) handleNotification(method string, paramsRaw json.RawM
 				Type string `json:"type"`
 			} `json:"status"`
 		}
-		if err := json.Unmarshal(paramsRaw, &notif); err == nil && notif.Status.Type == "idle" {
+		if err := json.Unmarshal(paramsRaw, &notif); err == nil && notif.Status.Type == "idle" && s.acceptsThreadNotification(method, notif.ThreadID) {
 			// In codex 0.125+, thread going idle signals turn completion.
 			s.completeTurn()
 		}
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## QA Review: PR #1816

### Overall assessment
Surgical bugfix for a real event-routing bug in the codex agent: Codex app-server v2 multiplexes parent and subagent notifications on the same connection, and cc-connect was processing both with the same routing logic — so a subagent's `turn/started` could overwrite the parent's `currentTurn` and pending messages, a child `turn/completed` could fire an empty `EventResult`, and child token-usage could clobber the parent's usage snapshot. The fix introduces two small predicate helpers (`acceptsThreadNotification`, `acceptsTurnNotification`) and gates every relevant notification handler through them. Backward compatibility for older app-server builds is preserved by accepting events that omit `threadId`/`turnId`.

### Review scope
- `agent/codex/appserver_session.go` (+35/-6), `agent/codex/appserver_session_test.go` (+112/-0)
- +147/-6 in 2 files
- Focus: correctness of the routing boundary, lock usage in the new predicates, backward-compat for older app-server builds, regression coverage quality, scope discipline

### ✅ What looks good
- **Two predicates cleanly model the routing boundary**: `acceptsThreadNotification(method, threadID)` covers events that only carry a thread id (`turn/started`, `thread/status/changed`); `acceptsTurnNotification(method, threadID, turnID)` extends with a turn-id check for the per-turn events (`item/started`, `item/completed`, `turn/completed`, `thread/tokenUsage/updated`). Splitting them keeps the common case (thread-only check) cheap and the richer case explicit.
- **Empty-id policy preserves backward compat**: `if threadID == "" || currentThreadID == "" || threadID == currentThreadID { return true }` — older Codex app-server builds that don't emit these fields still work, and the early-window case (`currentThreadID == ""` before the first `turn/started`) doesn't accidentally filter legitimate events. This is the right default for a protocol upgrade.
- **Lock discipline is correct**: each predicate acquires `stateMu` only long enough to read `currentTurn` and releases it before returning. The caller (`handleNotification`) holds no lock when it calls the predicate, so no nested-lock risk. The body of `acceptsTurnNotification` is sequential: thread-check first (under no lock), then a brief `stateMu` read for the turn comparison, then return. Clean.
- **Debug logging is the right level**: `slog.Debug` for filtered events keeps the production `info` log clean while giving ops a way to confirm subagent routing behavior in production (`journalctl _TRANSPORT=codex | grep 'ignoring notification for another thread'`).
- **Test design is rigorous**:
  - `IgnoresSubagentLifecycleNotifications` walks all four lifecycle events for a child thread, asserts `currentTurn` and `pendingMsgs` are untouched, asserts no event was emitted to the channel, then sends the *parent* versions of the same events and verifies the parent's answer and result flow through normally. This pins both directions: child filtered, parent unchanged.
  - `IgnoresSubagentTokenUsageNotifications` is short and surgical — set parent usage to 1234, send child usage 9876, assert usage is still 1234.
- **Helper additions are minimal**: `notificationProbe`, `tokenUsageNotificationProbe`, `appServerCurrentTurn`, `appServerPendingMessages` are all small test-only helpers that make the regression tests readable without coupling to internal state.

- **Sibling-agent audit. `agent/pi/`, `agent/cursor/`, `agent/opencode/`, `agent/kimi/` likely have analogous notification routing code. If any of them receive multiplexed events from their underlying runtime, they may have the same routing gap. Worth a separate follow-up issue to scan them; this PR doesn't need to expand scope.
- **Early-window behavior note**: when `currentThreadID` is empty (between session start and the first `turn/started`), the predicate accepts every notification. In practice this window is very short and there's no `currentTurn` to be clobbered, so it's safe. Worth a one-line comment in `acceptsThreadNotification` explaining why the early-window accept is intentional, so future maintainers don't try to tighten it and accidentally introduce a race.

- **CHANGELOG / release notes**: `fix(codex): isolate subagent app-server events (#1816, supersedes #1656)` is the right wording for the next release.
- **Telemetry**: a small counter for filtered events would let ops dashboard subagent activity without grepping logs. Optional; defer if not already a project pattern.

### 🚨 / 🔴 Must fix

### ❓ Questions
None.

### Testing / Risk
- **Author-validated**: targeted `-race` run of the two new tests, full `./agent/codex` unit test, full no-web suite, `go vet`, `git diff --check`, plus an explicit `tree equality with #1656` claim — all per PR description
- **QA-verified**: diff inspected; the predicate helpers are minimal and the call-site changes are mechanical (each `if err == nil` becomes `if err == nil && s.accepts...`); backward-compat path is intentional; test pair covers both directions of the routing boundary
- **Remaining risk**: low. The fix only adds filtering; even a malformed event that passes both predicates still goes through the existing handler logic, which is unchanged. No new attack surface.

### Next step
Maintainer can merge when convenient. The P2 sibling-agent audit is the most valuable follow-up — takes a few minutes with grep and could surface similar bugs in other adapters before users hit them.

PR #1817, opened 2026-09-09
fix(codex): steer active turns for /ps guidance

AaronZ345:
Supersedes #1665. This is the same patch/tree rebuilt from the latest `origin/main` after the explicit 30-day maintenance gate; the prior PR was approved with no requested changes or unresolved review threads. Three zero-file CI retrigger commits are intentionally omitted.

## Summary

- steer an active Codex app-server turn with `turn/steer` instead of starting a second turn
- preserve the active turn ID, buffered text, and prompt preamble while applying `/ps` guidance
- keep the existing `turn/start` behavior for idle sessions and reject mismatched steering responses
- make four pre-existing environment-sensitive tests hermetic so the full local suite is reliable

## Why

`/ps` calls `AgentSession.Send` while the current turn is still running. The Codex app-server implementation previously treated every `Send` as `turn/start`, reset the active-turn state, and cleared buffered messages. That race could emit an empty result even though Codex later completed a full answer.

Codex exposes `turn/steer` for same-turn guidance. This change uses it with the active `expectedTurnId`, matching Codex native guidance behavior without turning the guidance into a separate user turn.

## Validation

- `go test -race ./agent/codex -run 'TestAppServerSession_(SendSteersActiveTurn|SendStartsIdleTurn|SendRejectsMismatchedSteeringTurn)$' -count=1`
- `go vet ./agent/codex`
- `git diff --check`
- tree equality with #1665 and remote parity `0 0`

### DIFF
--- a/agent/codex/appserver_session.go
+++ b/agent/codex/appserver_session.go
@@ -64,6 +64,10 @@ type turnStartResponse struct {
 	} `json:"turn"`
 }
 
+type turnSteerResponse struct {
+	TurnID string `json:"turnId"`
+}
+
 type turnNotification struct {
 	ThreadID string `json:"threadId"`
 	Turn     struct {
@@ -478,7 +482,8 @@ func (s *appServerSession) Send(prompt string, messageID string, images []core.I
 	}
 
 	s.stateMu.Lock()
-	if !s.preambleSent {
+	activeTurn := s.currentTurn
+	if activeTurn == "" && !s.preambleSent {
 		prompt = prependCodexPromptPreamble(prompt, s.promptPreamble)
 		s.preambleSent = true
 	}
@@ -502,6 +507,13 @@ func (s *appServerSession) Send(prompt string, messageID string, images []core.I
 		})
 	}
 
+	if activeTurn != "" {
+		return s.steerTurn(threadID, activeTurn, input)
+	}
+	return s.startTurn(threadID, input)
+}
+
+func (s *appServerSession) startTurn(threadID string, input []map[string]any) error {
 	params := map[string]any{
 		"threadId": threadID,
 		"input":    input,
@@ -529,6 +541,26 @@ func (s *appServerSession) Send(prompt string, messageID string, images []core.I
 	s.pendingMsgs = s.pendingMsgs[:0]
 	s.stateMu.Unlock()
 
+	return nil
+}
+
+func (s *appServerSession) steerTurn(threadID, expectedTurnID string, input []map[string]any) error {
+	params := map[string]any{
+		"threadId":       threadID,
+		"expectedTurnId": expectedTurnID,
+		"input":          input,
+	}
+
+	var resp turnSteerResponse
+	if err := s.request("turn/steer", params, &resp); err != nil {
+		return fmt.Errorf("codex app-server turn/steer: %w", err)
+	}
+	if resp.TurnID == "" {
+		return fmt.Errorf("codex app-server turn/steer returned empty turn id")
+	}
+	if resp.TurnID != expectedTurnID {
+		return fmt.Errorf("codex app-server turn/steer returned turn id %q, want %q", resp.TurnID, expectedTurnID)
+	}
 	return nil
 }
 
--- a/agent/codex/appserver_session_test.go
+++ b/agent/codex/appserver_session_test.go
@@ -605,7 +605,7 @@ func sendAndRespond(t *testing.T, s *appServerSession, stdin *lockedWriteCloser,
 	t.Helper()
 	done := make(chan error, 1)
 	go func() {
-		done <- s.Send(prompt, nil, nil)
+		done <- s.Send(prompt, "", nil, nil)
 	}()
 	request := waitForClientRequest(t, stdin)
 	resultJSON, err := json.Marshal(result)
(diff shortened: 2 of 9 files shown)


chenhg5 (approved):
## QA Review: PR #1817

### Overall assessment
Solid P1 bugfix for the codex agent's mid-turn guidance flow. `/ps` was routing every `Send` through `turn/start`, which clobbered the active turn's state and could emit an empty result even when Codex later completed a full answer. The fix routes through `turn/steer` with `expectedTurnId` when a turn is active, falling through to the original `turn/start` path only when idle. The mismatch-response validation correctly rejects silently-accepted steering errors rather than silently starting a second turn. Beyond the bugfix, the PR also hermetizes four pre-existing environment-sensitive tests, which is a quiet but meaningful CI stability win.

### Review scope
- `agent/codex/appserver_session.go` (+33/-1), `agent/codex/appserver_session_test.go` (+166/-0), 4 hermetic-test files, 2 superpowers plan/spec docs
- +459/-33 in 9 files; ~30 production-code lines
- Focus: routing correctness, response validation, idempotency of `currentTurn`/`pendingMsgs` preservation, regression net quality, scope discipline (test hermetization vs unrelated cleanup)

### ✅ What looks good
- **Routing split is the right primitive**: `Send` snapshots `currentTurn` under `stateMu`, releases the lock, then dispatches to either `steerTurn` or `startTurn`. This avoids holding the lock across the JSON-RPC request — important because `request()` does I/O on a writer that the lock would otherwise serialize against the notification reader.
- **Active-turn preservation is the contract**: `steerTurn` does not touch `currentTurn`, `pendingMsgs`, or `preambleSent`. Only `startTurn` mutates them. The new `TestAppServerSession_SendSteersActiveTurn` explicitly asserts `currentTurn` and `pendingMsgs` are unchanged after a steer, then verifies the parent turn's eventual `item/completed` and `turn/completed` still flow through to the events channel — this is exactly the regression net needed.
- **Response validation is strict**: `turn/steer` returning an empty `turnId` or a different `turnId` from `expectedTurnId` returns an error rather than silently treating it as a new turn. This is the right call — silently falling back to `turn/start` would reintroduce the original bug.
- **Preamble handling preserved**: `prependCodexPromptPreamble` is now gated on both `!preambleSent` AND `activeTurn == ""`. The extra `activeTurn == ""` check is defensive (it doesn't change the normal path because preamble is only prepended on first `startTurn` which happens with `activeTurn == ""`), but it makes the intent explicit: a steer should never prepend preamble.
- **Hermetic test fixes are genuine wins**:
  - `core/cuj_test.go`: switching wait predicate from `len(agent.sessions) > 0` to `len(plat.getSent()) > 0` removes a class of false-positives where a session is created but the message hasn't been delivered to the agent yet.
  - `core/engine_test.go`: `waitDeleteModeResult` now also asserts `RefreshCard` was called, closing a race where the test could see the phase update before the UI refresh completed.
  - `daemon/launchd_test.go`: creating a fake plist in `t.TempDir()` removes the requirement that the developer running tests has a real launchd plist installed.
  - `platform/cloud-web/{gateway,poll}_test.go`: IPv6 wildcard handling + real `httptest` server returning 503 instead of relying on connection refusal — both address classes of environment-dependent flakiness.
- **Design docs are project-style**: `docs/superpowers/{plans,specs}/2026-07-11-*.md` follow the existing convention for plans/specs in the repo. Worth keeping for future maintainers who need to extend the steering behavior or onboard to the protocol.

- **Cross-runtime parity check**: PR #1861 already implemented the same fix shape for pi (the `/ps` guidance routing fix I reviewed previously). Worth confirming the patterns are consistent — both use the same "snapshot active state → branch on it" structure, which is good — and capturing the consolidated pattern in a doc so future agents (cursor, opencode, kimi) implement it the same way.
- **Test naming convention**: `TestAppServerSession_SendSteersActiveTurn` vs `TestAppServerSession_SendStartsIdleTurn` vs `TestAppServerSession_SendRejectsMismatchedSteeringTurn` — all share the `Send*` prefix, but the third one is about rejection behavior. Minor; consider `TestAppServerSession_SendRejectsMismatchedSteeringResponse` for clarity, or group under a `Send_*` subtest pattern.

- **CHANGELOG / release notes**: `fix(codex): steer active turns for /ps guidance (#1817, supersedes #1665)` is the right wording for the next release.
- **Telemetry**: a small metric for `turn/steer` success vs rejection would let ops dashboard the racing-with-completion rate. Optional; defer if not already a project pattern.

### 🚨 / 🔴 Must fix

### ❓ Questions
None.

### Testing / Risk
- **Author-validated**: targeted `-race` run of the three new tests, full `./agent/codex` unit test, full no-web suite, `go vet`, `git diff --check`, plus an explicit `tree equality with #1665` claim — all per PR description
- **QA-verified**: diff inspected; `Send` snapshot-then-branch pattern is correct; `steerTurn` does not mutate session state; response validation covers empty and mismatched cases; the four hermetic fixes are coherent (all address test reliability, none change semantics)
- **Remaining risk**: low. The fix is well-isolated to codex, the validation prevents silent fallback, and the hermetic fixes improve CI reliability as a side effect. Cross-runtime parity is the only thing worth confirming via a follow-up.

### Next step
Maintainer can merge when convenient. The P2 cross-runtime parity confirmation is the most valuable follow-up — a 10-minute grep-and-compare against pi's implementation in #1861 would either confirm consistency or surface a divergence to address before user reports arrive.

PR #1818, opened 2026-09-10
feat(telegram): mirror bot commands to chat scope so menu appears (#1813)

### LINKED ISSUE #1813: [Feature] Telegram: also register bot commands at chat scope
**Feature Area:** Platform (Telegram)

**Problem or Motivation:**

I reuse one Telegram bot across bridges. After switching it to cc-connect
v1.5.1-beta.1 (commit 80713167, Codex agent, long polling, Ubuntu), the
slash-command menu never appeared in my private chat with the bot. Typing a
command by hand worked, so the bot itself was fine. Only the menu stayed
empty.

What I checked on the Bot API side:

- `getMyCommands` at the default scope returned all 51 commands cc-connect
  registers, and the log confirms registration on every start
  (`telegram: registered bot commands count=51`).
- No other scope or language held commands. I had deleted leftover
  chat-scoped and all_group_chats-scoped lists from the previous bridge, so
  nothing was shadowing the default list.

The Bot API docs for `deleteMyCommands` say that after deletion, higher
level commands will be shown. That fallback never rendered in my client.
Setting the exact same 51 commands at chat scope for my private chat made
the menu appear immediately, and the client kept showing nothing before
that. The previous bridge wrote its commands at chat scope, which is why
its menu showed up instantly in the same chat and the same app.

So at least some Telegram clients only refresh the menu reliably when the
commands for that chat's own scope change. A default-scope-only
registration leaves those clients stuck with a stale per-chat cache.

**Proposed Solution:**

Let the Telegram platform mirror the command list at chat scope in
addition to the default scope. I looked at the code path:
`onPlatformReady` calls `initPlatformCapabilities`, which calls
`RegisterCommands` on every platform-ready transition, rebuilding the list
from built-ins, custom commands, and skills. The natural home for the
mirror is inside that same call, reusing the same slice, so the Telegram
100-entry trimming and `disabled_commands` filtering stay consistent
between the two scopes without extra logic.

Open question is which chats to mirror to. The IDs in `allow_from` are one
source. Chats seen since process start are another. Either works for the
single-user case.

Re-syncing on every registration matters more than the initial write. The
menu set changes over time: skills come and go, `disabled_commands` edits
filter entries, upgrades add commands, and the 100-entry trimming can drop
the whole skill block at once. A chat-scoped copy written only once goes
stale on the next such change, and a stale chat scope shadows the fresh
default list. That failure mode is silent. I ran into its leftover form
during this same migration, when the previous bridge's chat-scoped list
kept shadowing cc-connect's default list until I deleted it by hand.

**Alternatives Considered:**

- Manual `setMyCommands` per chat with a `{"type": "chat"}` scope. Verified
  working here, but it has to be repeated by hand after every upgrade or
  skill change, and a forgotten copy turns into the shadowing problem
  above.
- BotFather `/setcommands`. This writes the default scope only, so it does
  not change anything for affected clients.

**Additional Context:**

- cc-connect v1.5.1-beta.1, Linux (Ubuntu), Codex agent, Telegram long
  polling, private chat.
- `RegisterCommands` in `platform/telegram/telegram.go` calls
  `SetMyCommands` without a scope today. The file contains no other scope
  handling.
- Related closed issues I checked for duplicates: #30 (initial menu
  registration), #597 (registration limits), #301 (UTF-8 descriptions),
  #162 (inline `/menu`, a different feature). None cover chat-scoped
  registration.

**Willingness to Contribute:**

- [ ] I'd be willing to submit a PR for this feature

chenhg5:
## Summary

Telegram clients cache the slash-command menu per scope. A `setMyCommands` call at default scope does NOT refresh a chat that already has a stale chat-scope menu — so a user with `allow_from` pre-approved may never see the menu at all. The reporter reproduced this on cc-connect v1.5.1-beta.1 with 51 commands installed: log showed `telegram: registered bot commands count=51` but the menu never appeared; only an explicit chat-scope `setMyCommands` made it visible.

Fixes #1813.

## Changes

- `platform/telegram/telegram.go`
  - New constants `telegramBotCommandLimit=100` and `telegramChatCommandsThrottle=35ms` (latter = ~28 rps, safely under Telegram's ~30 rps per-bot limit).
  - New `Platform.seenChatIDs map[int64]struct{}` (under `mu`) plus `markChatSeen(chatID)` and `snapshotSeenChatIDs()` helpers.
  - `processUpdate` now calls `markChatSeen` for both regular messages and callback query messages.
  - `RegisterCommands` now:
    1. Builds the same trimmed `tgCommands` slice as before (unchanged default-scope write).
    2. WARN-logs when the 100-entry cap drops trailing entries (silent skill disappearance is no longer silent).
    3. After the default-scope write, mirrors the slice into every observed chat's `BotCommandScopeChat` with a 35ms per-chat throttle.
    4. Per-chat errors are `slog.Warn` and skipped — a single bad chat does not poison the batch.
  - Chat ID set is union(seenChatIDs, numeric allow_from IDs). In private chats `chat_id == user_id`, so `allow_from` doubles as a chat ID for pre-approved users we have not yet heard from (the reporter's exact scenario).
- `platform/telegram/telegram_test.go`
  - `stubTelegramBot` now records per-call scope (`""` for default, `"chat:<id>"` for chat scope).
- `platform/telegram/register_commands_test.go` (new)
  - 14 unit tests covering: default-scope write always fires, chat-scope mirrors from seen chats, union with allow_from, empty allow_from + no chats stays default-only, disabled_commands filter applies to both scopes on re-sync, 100-entry trim WARN path, per-chat failure does not poison batch, per-chat throttle respects Telegram rate limit, re-sync picks up newly-observed chats, markChatSeen dedup, snapshotSeenChatIDs union edge cases.

## Notes

- Implementation is contained in `platform/telegram/`. `core/` unchanged. No new public API surface. Telegram-only.
- Re-sync on every `RegisterCommands` call ensures admin edits (skills 增减, `disabled_commands`, upgrades adding commands, 100-entry trim dropping a whole skill) propagate within one tick.
- Per-chat errors are logged but do not bubble up — keeps the existing `RegisterCommands(commands)` contract intact for callers.
- Per-chat throttle is conservative; callers that need to mirror across hundreds of chats get ~28 rps automatically.

## Verification

Local results before PR creation:

- `go build -tags no_web ./...` — clean
- `go test -tags no_web -race -count=1 ./platform/telegram/...` — `ok` (1.799s, 14 new + all pre-existing tests pass)
- `gofmt -l platform/telegram/` — clean
- `go vet -tags no_web ./platform/telegram/...` — clean

GitHub Actions CI 5/5 must pass before this is handed off to QA.

## Risk

Low-medium. The change is contained in `platform/telegram/`, no `core/` modifications, no public API changes. New failure mode introduced: if the bot has seen thousands of chats, the mirror loop will block `RegisterCommands` for up to 35ms × N ms. For typical cc-connect deployments with one or a handful of allow_from users, the wall-time impact is negligible. The throttle keeps us under Telegram's per-bot rate limit.

If `RegisterCommands` is called frequently (skill reloads, runtime re-config), the worst-case wall time scales linearly with seen chats. This is acceptable for the user-facing benefit of a working menu, but a future optimisation could background the chat-scope mirror.

### DIFF
--- a/platform/telegram/register_commands_test.go
+++ b/platform/telegram/register_commands_test.go
@@ -1,0 +1,439 @@
+package telegram
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/chenhg5/cc-connect/core"
+
+	tgbot "github.com/go-telegram/bot"
+	"github.com/go-telegram/bot/models"
+)
+
+// helper: build a Platform with the stub bot already wired up + allow_from
+// set as requested. We avoid Start() so the test stays synchronous.
+func newTestPlatformWithAllowFrom(t *testing.T, allowFrom string) (*Platform, *stubTelegramBot) {
+	t.Helper()
+	bot := newStubTelegramBot()
+	p := &Platform{
+		token:     "test-token",
+		allowFrom: allowFrom,
+		bot:       bot,
+		selfUser:  &models.User{ID: 1, Username: "testbot"},
+	}
+	return p, bot
+}
+
+// TestRegisterCommands_WritesDefaultScopeAlways pins the baseline behaviour:
+// the default scope is written on every RegisterCommands call, regardless of
+// whether we have any chats to mirror to. This is what fixes the silent
+// shadowing case in #1813 — when a previous bridge left stale chat-scope
+// state, the next RegisterCommands must still refresh the default.
+func TestRegisterCommands_WritesDefaultScopeAlways(t *testing.T) {
+	p, bot := newTestPlatformWithAllowFrom(t, "")
+	if err := p.RegisterCommands([]core.BotCommandInfo{
+		{Command: "help", Description: "show help"},
+	}); err != nil {
+		t.Fatalf("RegisterCommands: %v", err)
+	}
+
+	scopes := bot.setMyCommandsScopes()
+	if len(scopes) == 0 {
+		t.Fatal("expected at least 1 SetMyCommands call, got none")
+	}
+	if scopes[0] != "" {
+		t.Fatalf("first SetMyCommands scope = %q, want default (empty)", scopes[0])
+	}
+}
+
+// TestRegisterCommands_MirrorsToChatScopeFromSeenChats verifies the core
+// fix: when a chat has been observed via processUpdate, RegisterCommands
+// mirrors the default-scope menu into that chat's BotCommandScopeChat. The
+// reporter's "menu doesn't show" symptom in #1813 goes away once the chat
+// scope is written, because that's the only thing that triggers the client
+// to refresh its cached menu.
+func TestRegisterCommands_MirrorsToChatScopeFromSeenChats(t *testing.T) {
+	p, bot := newTestPlatformWithAllowFrom(t, "")
+	p.markChatSeen(101)
+	p.markChatSeen(202)
+
+	if err := p.RegisterCommands([]core.BotCommandInfo{
+		{Command: "help", Description: "show help"},
+	}); err != nil {
+		t.Fatalf("RegisterCommands: %v", err)
+	}
+
+	scopes := bot.setMyCommandsScopes()
+	want := map[string]bool{"": false, "chat:101": false, "chat:202": false}
+	for _, s := range scopes {
+		if _, ok := want[s]; ok {
+			want[s] = true
+		}
+	}
+	for s, seen := range want {
+		if !seen {
+			t.Errorf("missing SetMyCommands call for scope %q (got scopes: %v)", s, scopes)
+		}
+	}
+}
+
+// TestRegisterCommands_UnionWithAllowFrom verifies that numeric allow_from
+// IDs are unioned with seen chats. In private chats the ChatID equals the
+// UserID, so this union lets us register menu state for users we have not
+// yet heard from but who are pre-approved in config (the reporter's exact
+// scenario in #1813).
+func TestRegisterCommands_UnionWithAllowFrom(t *testing.T) {
+	p, bot := newTestPlatformWithAllowFrom(t, "555, 666, *not-a-number*")
+	p.markChatSeen(777)
+
+	if err := p.RegisterCommands([]core.BotCommandInfo{
+		{Command: "help", Description: "show help"},
+	}); err != nil {
+		t.Fatalf("RegisterCommands: %v", err)
+	}
+
+	scopes := bot.setMyCommandsScopes()
+	want := map[string]bool{
+		"":         false,
+		"chat:555": false,
+		"chat:666": false,
+		"chat:777": false,
+	}
+	for _, s := range scopes {
+		if _, ok := want[s]; ok {
+			want[s] = true
+		}
+	}
+	for s, seen := range want {
+		if !seen {
+			t.Errorf("missing SetMyCommands call for scope %q (got scopes: %v)", s, scopes)
+		}
+	}
+}
+
+// TestRegisterCommands_EmptyAllowFromAndNoChatsStaysAtDefaultScope pins the
+// regression guard from #1813: with no allow_from and no observed chats,
+// only the default-scope write happens. The previous bridge had this
+// working but only at default scope; we keep that path intact.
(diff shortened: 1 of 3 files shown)


chenhg5 (commented):
## Conclusion: Approve

Excellent, well-scoped fix. The change directly addresses the silent-shadowing root cause from #1813: Telegram clients refresh their cached menu on chat-scope writes, not default-scope writes. Mirroring the trimmed slice into every observed chat scope on every `RegisterCommands` call is exactly the right shape, and the design has the right failure-mode behaviour.

### What works well

- **Default-scope write happens first**, so the chat-scope mirror is strictly additive — if every chat-scope call fails for any reason, the user still gets the original pre-#1813 behaviour (only default-scope menu). The reporter's previous-bridge stale-state scenario stops being silent because every subsequent `RegisterCommands` re-writes default.
- **`markChatSeen` is lazy-init under the existing mutex** and `chatID == 0` short-circuits — zero cost for the (uncommon) zero-id edge case.
- **`snapshotSeenChatIDs` union logic is robust**: empty, `"*"`, non-numeric, ≤0, mixed — all handled (covered by the 5 `TestSnapshotSeenChatIDs_AllowFromUnion` sub-cases).
- **Per-chat error isolation**: a single chat's `SetMyCommands` failure only emits `slog.Warn` and `continue`s — the default scope and other chats still get their write. `TestRegisterCommands_PerChatFailureDoesNotPoisonBatch` and `TestRegisterCommands_PerChatErrorIsRecoverable` pin this.
- **35ms throttle** keeps per-bot request rate at ~28 rps, well under Telegram's ~30 req/s limit. `TestRegisterCommands_PerChatThrottleRespectsTelegramRateLimit` asserts ≥ 4 × 35ms = 140ms for 5 chats.
- **100-entry trim now WARN-logs** the dropped count, removing the silent-skill-drop failure mode that hides behind #1813's symptom.
- **Re-sync on every call** is the key behaviour that fixes the skill-add/disable-edit/update case: each `RegisterCommands` writes fresh state to default + all known chat scopes, so any chat-scope cache that drifted from default gets refreshed next time skills change.
- **Tests link back to the issue** (`See #1813`) and to the acceptance criteria in a way that makes the contract explicit. The `var _ tgbot.SetMyCommandsParams` compile-time guard is a small but useful tripwire for upstream SDK shape changes.

### P2 — non-blocking suggestions

- **Long-running `seenChatIDs` growth**: For a bot that's been running for months and has been added to (and removed from) hundreds of one-off groups, this map grows unboundedly. Typical 1–10 `allow_from` deployments are fine (`< 500ms` per call), but a public bot could eventually push `RegisterCommands` wall time into seconds. Worth a future LRU cap or a "stale chat prune after N days" — not blocking.
- **`time.Sleep` after each successful call only**: this is fine (errors don't waste time), but if a caller ever runs `RegisterCommands` from a hot request path, the cumulative 35ms × N is noticeable. Worth documenting in the `RegisterCommands` godoc that the call may block for ~35ms × |seenChatIDs|.
- **`var _ context.Context` at the bottom of the test file** appears unused — it's a compile-time assertion that just confirms `context` is imported. Harmless but noisy; can be dropped in a follow-up.

### P3 — nits

- Constant naming is fine; `telegramChatCommandsThrottle` is long but unambiguous.
- `snapshotSeenChatIDs` returns an unsorted slice. Real-world this doesn't matter (each Telegram call is independent), and the tests assert membership not order. No change needed.

### Testing

- CI: 5/5 green on head `7dfc45ad6eaf` (lint, unit, smoke, regression, performance).
- I read the full diff and walked all 14 new tests. The matrix covers: happy path (default-only, seen-only, union), regression guards (no-chat baseline), re-sync (disabled-commands, new-seen-chat), 100-entry overflow, per-chat error isolation, throttle time invariant, and `allow_from` edge cases.
- Dev also ran `-race` locally — no data races in the new map access patterns.

### Risk

- Low. Pure additive, no `core/` changes, no public API changes, no new dependencies. Worst-case is per-call wall time growing linearly with observed chats, which is documented and is well below Telegram's rate limit even at hundreds of chats.

### Next step

- Ready for maintainer merge.

PR #1819, opened 2026-09-10
fix: lazy history bootstrap when switching agent type (#1805)

### LINKED ISSUE #1805: [Feature] Lazy history bootstrap when switching agent type
## Problem

When a project switches to another agent type, existing platform sessions keep their cc-connect session/history, but the new native agent session starts without that prior conversation context.

For example, a Feishu thread may already have a long task conversation handled by `claudecode`. After switching the project agent to `codex`, the same Feishu thread still maps to the same cc-connect session, but the next message starts a fresh native agent session. The new agent only receives the current user message, not the previous thread/task history.

## Current behavior from code

Checked against cc-connect v1.5.0 / commit `17c61062`.

- `core/engine.go` calls `sessions.InvalidateForAgent(ag.Name())` on startup.
- `core/session.go` `InvalidateForAgent(agentType)` records the previous native session id in `PastAgentSessionIDs`, then clears `AgentSessionID` and sets `AgentType` to the new agent.
- Later `getOrCreateInteractiveStateWith` calls `agent.StartSession(ctx, startSessionID)`; because `startSessionID == ""`, this creates a fresh native agent session.
- cc-connect still has `Session.History []HistoryEntry`, and new user messages are recorded with `session.AddHistory("user", msg.Content)`, but the stored history is not automatically prepended or injected into the new agent prompt before `as.Send(...)`.

So cross-agent continuity is partially retained at the cc-connect routing/history layer, but not restored into the new native agent context.

## Expected behavior

When an existing platform session is first handled by a different agent type, cc-connect should optionally bootstrap the new native agent session with a bounded summary/transcript of the existing cc-connect history before handling the current user message.

The user experience should be:

1. Same Feishu/Slack/etc thread remains mapped to the same cc-connect session.
2. After switching agent type, the first new turn initializes the new native agent with the previous thread context.
3. The bootstrap is internal and should not send an extra visible platform reply.
4. Subsequent turns use the new native agent session normally and should not re-bootstrap repeatedly.

## Proposed design

Add a lazy migration/bootstrap mechanism:

- Detect: `AgentSessionID == ""`, current `AgentType` is the active agent, `PastAgentSessionIDs` or stored `History` indicates an existing conversation, and this session has not yet been bootstrapped for this agent type.
- Build a bounded context payload from `Session.History` or a summarized form, excluding the current user message if it will be sent separately.
- Inject it once into the new native agent session, either as an internal bootstrap message or as a one-time context prefix before the current prompt.
- Mark the session as bootstrapped for the current agent type so it is not repeated.
- Keep max history length/token limits configurable.

A minimal implementation could avoid agent event-loop complexity by prepending a one-time migration block to the next real user prompt and then saving a migration marker on the cc-connect session.

## Why this matters

Agent switching is useful for trying another backend/model on an existing operational thread. Without history bootstrap, the new agent loses task context and users have to manually replay or copy previous messages per thread.

chenhg5:
## Summary

Fixes #1805. When a project switches between agent types (e.g. claudecode → codex), the freshly-started native agent session currently loses the prior cc-connect conversation context. This PR adds a one-time **lazy history bootstrap**: the first prompt each new agent type sees on a given session is prefixed with a bounded, trimmed transcript of the stored cc-connect history.

The native agent then has enough context to answer follow-up questions like "what were we working on?" after a `/switch agent_type`.

## Changes

- **`core/session.go`** — `Session.BootstrappedFor map[string]struct{}` keyed on agent type. Survives `InvalidateForAgent` and process restart via the snapshot path (snapshotVersion bumped 1 → 2). New accessors `IsBootstrappedFor` / `MarkBootstrappedFor` (both lock `s.mu`). Snapshot deep-copy via new `copyBootstrappedFor` helper.
- **`core/engine.go`** — `Engine.SetHistoryBootstrap` config setter. New `maybeBootstrapHistory` helper called from `processInteractiveMessageWith` between `buildSenderPrompt` and the `as.Send` goroutine launch. Bootstrap block is prefixed to the prompt only when: feature enabled, agent has a non-empty name, session has prior history (`PastAgentSessionIDs` or older `History`), and the per-agent-type marker is not yet set. Trims newest-first by entry cap, then by token cap. Returns true on actual injection so the caller can update the marker without false positives.
- **`config/config.go`** — new `[projects.X.history_bootstrap]` TOML section (`enabled`, `max_entries`, `max_tokens`). Optional pointers preserve "not set" so existing configs fall back to engine defaults.
- **`cmd/cc-connect/main.go`** — wires the setter in both startup and reload paths.
- **`core/history_bootstrap_test.go`** — 11 unit tests covering fire/no-op, per-agent-type marker semantics, entry-cap and token-cap trimming, current-turn exclusion, disabled flag, nil-safety, marker round-trip via `Save`/`load`, and the `copyBootstrappedFor` independence invariant.

## Design Decisions (vs PM triage)

1. **History bounds** — `max_entries=20` / `max_tokens=4000` defaults, configurable per project. LLM-summarised older context is out of scope for v1.
2. **Per-AgentType marker** — keyed `map[string]struct{}` so a future second switch (claudecode → codex → claudecode) re-bootstraps correctly.
3. **Concurrency** — only `s.mu` is touched, briefly. `BootstrappedFor` is read under the lock rather than via the public accessor (which would deadlock on the non-reentrant `sync.Mutex`).
4. **Visibility** — single `Send` with a prefixed prompt. Not added to `Session.History`; no extra platform reply. Bracket-style header matches the existing `buildSenderPrompt` convention.

## Configuration

```toml
[projects.myproject.history_bootstrap]
enabled = true       # default: true
max_entries = 20     # default: 20
max_tokens = 4000    # default: 4000
```

Operators who want a strictly fresh session per agent switch can set `enabled = false`.

## Compatibility

- `core/` does not import `agent/` or `platform/`.
- No hardcoded agent or platform names; the marker key is derived from `agent.Name()`.
- Schema migration is backward-compatible: v1 snapshots load with `BootstrappedFor == nil` (treated as "no bootstraps yet, may bootstrap on next agent switch") and are re-saved as v2.
- Default config (no `[history_bootstrap]` block) preserves prior behaviour? No — it now *does* bootstrap by default. This is the intended fix; operators who want the old behaviour can set `enabled = false`. Mention this in release notes.

## Test Results

```
$ go test -tags no_web -race -run 'TestMaybeBootstrapHistory|TestSessionIsBootstrappedFor|TestSessionSnapshotRoundTripPreservesBootstrappedFor|TestCopyBootstrappedFor' ./core/
ok  	github.com/chenhg5/cc-connect/core   1.137s

$ go test -tags no_web -race ./core/
ok  	github.com/chenhg5/cc-connect/core   48.286s

$ go build -tags no_web ./...
(no output — clean)
```

## Risks

- **Behaviour change by default.** Users who relied on a strictly clean context after a `/switch agent_type` will now see a bootstrap block. Mitigation: opt-out via `enabled = false`. Should be called out in release notes.
- **Token budget interaction.** The bootstrap block consumes a portion of the per-prompt token budget. `max_tokens=4000` defaults should stay well clear of any agent's context window but operators on small models may need to lower it.
- **Snapshot migration.** v1 snapshots round-trip correctly via `omitempty` on `BootstrappedFor`, but an existing in-memory `Session` whose `BootstrappedFor` is nil is treated as "never bootstrapped". A user who upgrades then immediately restarts may see one bootstrap block per (session, agent type) that was previously never bootstrapped. This is the correct behaviour but worth flagging.

## Related

- Issue #1805
- Discussion: cross-agent continuity

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -704,6 +704,18 @@ func main() {
 			}
 			engine.SetAutoCompressConfig(true, maxTokens, minGap)
 		}
+
+		// Wire history bootstrap (issue #1805). Defaults to enabled so the
+		// user-facing cross-agent continuity works out of the box; operators
+		// who want a strictly fresh session per agent switch can disable it
+		// via [projects.X.history_bootstrap] enabled = false.
+		bootstrapEnabled := true
+		if proj.HistoryBootstrap.Enabled != nil {
+			bootstrapEnabled = *proj.HistoryBootstrap.Enabled
+		}
+		bootstrapMaxEntries := derefInt(proj.HistoryBootstrap.MaxEntries)
+		bootstrapMaxTokens := derefInt(proj.HistoryBootstrap.MaxTokens)
+		engine.SetHistoryBootstrap(bootstrapEnabled, bootstrapMaxEntries, bootstrapMaxTokens)
 		resetIdle, defaulted := resolveResetOnIdle(proj.ResetOnIdleMins)
 		engine.SetResetOnIdle(resetIdle)
 		if defaulted {
@@ -1786,6 +1798,15 @@ func reloadConfig(configPath, projName string, engine *core.Engine) (*core.Confi
 	} else {
 		engine.SetAutoCompressConfig(false, 0, 0)
 	}
+
+	// Reload history bootstrap (issue #1805).
+	bootstrapEnabled := true
+	if proj.HistoryBootstrap.Enabled != nil {
+		bootstrapEnabled = *proj.HistoryBootstrap.Enabled
+	}
+	bootstrapMaxEntries := derefInt(proj.HistoryBootstrap.MaxEntries)
+	bootstrapMaxTokens := derefInt(proj.HistoryBootstrap.MaxTokens)
+	engine.SetHistoryBootstrap(bootstrapEnabled, bootstrapMaxEntries, bootstrapMaxTokens)
 	resetIdle, defaulted := resolveResetOnIdle(proj.ResetOnIdleMins)
 	engine.SetResetOnIdle(resetIdle)
 	if defaulted {
--- a/config/config.go
+++ b/config/config.go
@@ -452,6 +452,23 @@ type AutoCompressConfig struct {
 	MinGapMins *int  `toml:"min_gap_mins,omitempty"` // minimum minutes between auto-compress runs (default 30)
 }
 
+// HistoryBootstrapConfig controls the lazy history-bootstrap mechanism
+// (issue #1805). When enabled, the engine prefixes a bounded summary of
+// the cc-connect stored history into the first prompt sent to a
+// freshly-started native agent session after an agent-type switch.
+type HistoryBootstrapConfig struct {
+	// Enabled turns the lazy bootstrap on. Defaults to true.
+	Enabled *bool `toml:"enabled,omitempty"`
+	// MaxEntries caps the number of recent history entries included in the
+	// bootstrap block. Older entries are dropped with an info log so the
+	// behaviour stays visible. 0 falls back to the engine default (20).
+	MaxEntries *int `toml:"max_entries,omitempty"`
+	// MaxTokens caps the rough token count of the bootstrap block.
+	// Token estimation reuses the engine's estimateTokens heuristic
+	// (~1 token per 4 runes). 0 falls back to the engine default (4000).
+	MaxTokens *int `toml:"max_tokens,omitempty"`
+}
+
 // ObserveConfig controls forwarding of native terminal Claude Code sessions to a messaging platform.
 type ObserveConfig struct {
 	Enabled bool   `toml:"enabled"`
@@ -482,6 +499,10 @@ type ProjectConfig struct {
 	Platforms                    []PlatformConfig   `toml:"platforms"`
 	Heartbeat                    HeartbeatConfig    `toml:"heartbeat"`
 	AutoCompress                 AutoCompressConfig `toml:"auto_compress"`
+	// HistoryBootstrap controls the lazy history-bootstrap mechanism
+	// (issue #1805) that gives a freshly-started native agent session
+	// the cc-connect stored history after an agent-type switch.
+	HistoryBootstrap HistoryBootstrapConfig `toml:"history_bootstrap"`
 	// ResetOnIdleMins automatically rotates to a new cc-connect session after
 	// the current session has been inactive for the specified number of minutes.
 	// 0 or nil disables the behavior.
--- a/core/engine.go
+++ b/core/engine.go
@@ -429,6 +429,15 @@ type Engine struct {
 	autoCompressMinGap    time.Duration
 	resetOnIdle           time.Duration
 
+	// historyBootstrapEnabled controls lazy migration of cc-connect stored
+	// history into a freshly-started native agent session when the agent
+	// type changes (issue #1805). Defaults to true so the user-facing
+	// cross-agent continuity works out of the box; operators who want a
+	// strictly fresh session per agent switch can disable it via TOML.
+	historyBootstrapEnabled    bool
+	historyBootstrapMaxEntries int
+	historyBootstrapMaxTokens  int
+
 	// Reply footer composition flags. The footer renders up to two lines:
 	//   line 1 — model · [effort ·] out/in/cw/cr · ctx%   (gated by showContextIndicator)
 	//   line 2 — workspace directory                       (gated by showWorkdirIndicator)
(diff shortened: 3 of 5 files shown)


chenhg5 (commented):
## Conclusion: Approve

## Overall assessment
PR #1819 is mergeable. CI is 5/5 green on the current head (`19737667`), the implementation is a tightly scoped lazy-bootstrap feature that closes the cross-agent continuity gap (`#1805`), and the test coverage is comprehensive (11 new tests including snapshot round-trip and deep-copy independence). QA confirms the lock-discipline and snapshot-migration design.

## Review scope
- Reviewed `core/session.go`, `core/engine.go`, `core/history_bootstrap_test.go` (new), `config/config.go`, and `cmd/cc-connect/main.go`.
- Focused on correctness under concurrency, snapshot migration, idempotence, prompt-injection safety, and config backward-compat.

## ✅ What looks good
- **Lock discipline is explicit and correct**: `maybeBootstrapHistory` reads `session.BootstrappedFor` directly under `s.mu` and explicitly comments that calling `IsBootstrappedFor` here would re-enter the non-reentrant `sync.Mutex`. The accessor helpers are still safe to call from outside the locked region.
- **Snapshot migration is well thought through**: `snapshotVersion` is bumped 1 → 2 with a comment describing the v1-loads-with-`BootstrappedFor=nil` semantics (treated as "never bootstrapped"). `copyBootstrappedFor` provides a clean deep-copy boundary so `saveLocked` cannot race with concurrent `MarkBootstrappedFor`.
- **Marker survives `InvalidateForAgent` by design**: the new comment in `InvalidateForAgent` explicitly states the bootstrap marker is *deliberately not cleared* — switching agent types back and forth (claudecode → codex → claudecode) re-bootstraps correctly. `TestMaybeBootstrapHistory_DifferentAgentTypeResets` exercises this.
- **Current-turn exclusion is correct**: `processInteractiveMessageWith` calls `session.AddHistory("user", msg.Content)` *before* `maybeBootstrapHistory`, so the helper excludes `history[olderLen:]` where `olderLen = len(History) - 1`. This avoids re-injecting the current user prompt. Tested explicitly.
- **Prompt-injection safety**: `strings.NewReplacer("\n", " ", "\r", "").Replace(h.Content)` neutralises embedded newlines before the per-line `[role] content` framing, so a malicious or buggy transcript entry cannot break out of the line boundary or prematurely close the bootstrap block.
- **Header design matches existing convention**: `[cc-connect history_bootstrap: agent_type=... past_native_session_ids=...]` mirrors the `[cc-connect sender_id=...]` style used by `buildSenderPrompt`, so the two prefix shapes are visually consistent for any agent inspecting them.
- **Backwards compatible config**: optional `*bool` / `*int` pointers preserve the "not set" state so existing TOML files default to `enabled=true / max_entries=20 / max_tokens=4000`. Operators who want strictly fresh sessions per `/switch` get a clear `enabled = false` escape hatch.
- **Test coverage is excellent**: 11 tests cover fire/no-op, per-agent-type marker isolation, both trim strategies, nil safety, snapshot round-trip, and `copyBootstrappedFor` independence. Author reports `go test -race` clean on `./core/` and the targeted test subset.
- **Layering preserved**: per PM, no `core/` import of `agent/` or `platform/` is introduced. The agent name is fetched via `agent.Name()` at runtime — no hardcoded names.

## 🚨/🔴 Must fix
- None. No merge-blocking issues found.

## 🟠 Should improve
- None for this PR. The change is appropriately scoped.

## 🔵 Optional
- P3 nit 1: `estimateTokens` reuses the engine's coarse `~rune/4` heuristic. For Chinese / emoji-heavy transcripts this is conservative in different directions, but the 4000-token default stays well clear of every supported agent's context window. A future improvement could bucket `rune < 0x80` vs non-ASCII for a better approximation.
- P3 nit 2: the token-cap walk-from-newest loop guarantees `kept` contains at least one entry (the first iteration skips the `len(kept) == 0` break), so a single oversized entry would be injected even if it exceeds the cap. With the default 4000-token budget this is fine, but operators tuning the cap downward should know this.
- P3 nit 3: if `as.Send` fails between `maybeBootstrapHistory` returning `true` and the marker being set, the next turn will re-inject the bootstrap. Re-injection is idempotent from the agent's perspective (same key, same data), so this is acceptable — just worth a comment if you ever revisit the marker-write ordering.

## ❓ Questions
- None for this PR. The default-enabled behavior change is the intended design (already flagged in `known_risks` and dev's "需要人类确认的问题"); release notes should call out the opt-out.

## Testing / Risk
- **CI:** `lint`, `unit-test`, `smoke-test`, `regression-test`, `performance-test` — 5/5 SUCCESS on head `19737667`. No skipped, cancelled, or pending checks.
- **Verified evidence:**
  - All 11 new test cases pass; author reports `go test -race ./core/` clean (48s).
  - `go vet` and `gofmt` clean on changed files.
  - `TestSessionSnapshotRoundTripPreservesBootstrappedFor` proves the marker survives `Save` / `load`, which prevents re-bootstrap after a process restart.
  - `TestCopyBootstrappedFor_IsIndependent` proves the deep-copy helper isolates the live map from the snapshot.
- **Remaining risk:**
  - **Default behavior change** is the headline risk. Users who relied on strict isolation after `/switch agent_type` will see new bootstrap context after upgrading. The `[projects.X.history_bootstrap] enabled = false` escape hatch is documented in the new config block and engine setter comment. Release notes should call this out.
  - **Snapshot migration**: v1 snapshots load with `BootstrappedFor = nil` and trigger exactly one bootstrap on the next new-agent-type switch. This is the documented intended behavior; users on actively-rotating projects may notice one slightly-longer first prompt after upgrade.
  - **Token estimation accuracy** as noted in P3 nit 1; not impactful at the default budget.

## Next step
- Recommend **merge** when the maintainer is ready. No additional changes required from the author for this PR. Release notes should mention the default behavior change and the `enabled = false` opt-out.

PR #1822, opened 2026-09-10
fix(slack): dedup a message delivered as both app_mention and message

ht290:
## Problem

When a Slack app subscribes to **both** `app_mention` and `message.channels`, a message that `@`-mentions the bot is delivered twice. `handleEvent` has an independent handler for each event type, and both build a `core.Message` with `MessageID = ev.TimeStamp` and call `p.handler` — so one mention is dispatched twice, ~250ms apart:

```
13:59:13.728  message received   msg_id=1789045151.985149  content_len=650
13:59:13.729  processing message
13:59:13.975  message received   msg_id=1789045151.985149  content_len=650
13:59:13.976  message queued for busy session               queue_depth=1
```

The first dispatch starts a turn. The second finds the session busy, so the user gets an unprompted **"📬 Message received — will process after the current task finishes."** on every mention, and the duplicate is then re-run from the queue. Two agent turns per mention, and the agent sees the same message twice in its context.

Observed on a live shared channel — the correlation with mentions is exact:

| message | mentions the bot? | dispatches |
|---|---|---|
| `hi <@OTHERBOT>` | no | 1 |
| `haha Kai isn't responsive atm…` | no | 1 |
| `I'm trying to talk to <@OTHERBOT>, not you <@BOT>` | **yes** | **2** |
| `Got you.\nA quick intro to <@BOT>…` | **yes** | **2** |

This is easy to miss in logs: `stripAppMentionText` only strips a mention at the *start* of the text, so for a mid-sentence mention both copies have identical `content_len` and the duplicate looks like a logging artifact.

## Fix

Route both handlers through `core.MessageDedup`, which `dingtalk`, `feishu`, `weixin`, `matrix`, `qq`, `wecom` and the other adapters already use — `slack` was one of the few that never got wired up. The key is `channel + ts`, since a Slack `ts` is only unique within its channel.

This also covers Socket Mode redelivery of an unacked event, which the adapter had no guard against either.

## Tests

Four new tests drive `handleEvent` end-to-end with a stub handler (name caches pre-seeded so the resolvers never reach the API client):

- a mention arriving as both events dispatches once
- dedup is order-independent (Slack does not guarantee which subscription lands first)
- distinct messages still dispatch, including the same `ts` in two different channels
- `dedupKey` returns an empty key for an empty `ts`, which `MessageDedup` never treats as a duplicate

Verified they fail without the guard. `go build ./...`, `go vet ./platform/slack/` and `go test ./platform/slack/` are clean.

## Known limitation, deliberately not addressed

The two handlers also differ in `Content`: `app_mention` passes it through `stripAppMentionText` while the message handler passes `ev.Text` verbatim. After dedup, a leading bot mention is therefore stripped only if the `app_mention` event happens to land first. Unifying that changes what the agent sees for existing users, so I kept this fix narrow — happy to follow up separately if you'd like it made deterministic.

### DIFF
--- a/platform/slack/slack.go
+++ b/platform/slack/slack.go
@@ -41,6 +41,12 @@ type Platform struct {
 	channelNameCache map[string]string
 	channelCacheMu   sync.RWMutex
 	userNameCache    sync.Map // userID -> display name
+	// dedup drops the second delivery of a message that Slack reports through
+	// more than one subscription. A message that @-mentions the bot arrives as
+	// BOTH an app_mention event and a message event, and the two are handled
+	// independently below; without this guard each mention starts a turn and
+	// then queues a duplicate against the still-busy session.
+	dedup core.MessageDedup
 }
 
 func New(opts map[string]any) (core.Platform, error) {
@@ -112,6 +118,16 @@ func (p *Platform) buildSessionKey(channel, user, threadTS string) string {
 	default:
 		return fmt.Sprintf("slack:%s:%s", channel, user)
 	}
+}
+
+// dedupKey identifies one Slack message. A message ts is only unique within its
+// channel, so both parts are needed. An empty ts yields an empty key, which
+// core.MessageDedup never treats as a duplicate.
+func dedupKey(channel, ts string) string {
+	if ts == "" {
+		return ""
+	}
+	return channel + ":" + ts
 }
 
 // threadRootTS returns the thread parent timestamp for an event: the existing
@@ -197,6 +213,12 @@ func (p *Platform) handleEvent(evt socketmode.Event) {
 					return
 				}
 
+				if p.dedup.IsDuplicate(dedupKey(ev.Channel, ev.TimeStamp)) {
+					slog.Debug("slack: duplicate message ignored", "source", "app_mention",
+						"channel", ev.Channel, "ts", ev.TimeStamp)
+					return
+				}
+
 				threadTS := threadRootTS(ev.ThreadTimeStamp, ev.TimeStamp)
 				sessionKey := p.buildSessionKey(ev.Channel, ev.User, threadTS)
 
@@ -256,6 +278,12 @@ func (p *Platform) handleEvent(evt socketmode.Event) {
 
 				if !core.AllowList(p.allowFrom, ev.User) {
 					slog.Debug("slack: message from unauthorized user", "user", ev.User)
+					return
+				}
+
+				if p.dedup.IsDuplicate(dedupKey(ev.Channel, ev.TimeStamp)) {
+					slog.Debug("slack: duplicate message ignored", "source", "message",
+						"channel", ev.Channel, "ts", ev.TimeStamp)
 					return
 				}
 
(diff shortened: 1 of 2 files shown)


chenhg5 (approved):
## Review: PR #1822

### What this PR fixes
When a Slack app subscribes to both `app_mention` and `message.channels`, every `@`-mention is delivered twice through independent handlers, both building `core.Message` with `MessageID = ev.TimeStamp`. First dispatch starts a turn, second finds the session busy and either sends a spurious "📬 Message received" ack or queues a duplicate run that the agent sees twice.

### Diff review

`platform/slack/slack.go`
- `dedup core.MessageDedup` field added to `Platform` with a clear comment explaining the WHY (which subscriptions collide). Future maintainers won't have to re-derive this.
- `dedupKey(channel, ts)` returns `""` on empty `ts`, so malformed/empty events are never silently swallowed by the dedup guard. Important — empty-key never-collides is a real property of `core.MessageDedup`, and the test `TestDedupKey` pins it down.
- `channel + ":" + ts` is the right key shape: a Slack `ts` is only unique within its channel, so the cross-channel case in `TestHandleEvent_DistinctMessagesStillDispatch` is the one that would have failed with a naive `ts`-only key.
- The guard is placed BEFORE `threadRootTS` / `buildSessionKey` / `p.handler` — duplicate events waste no CPU.
- Two `slog.Debug` lines with `source` (`app_mention` vs `message`), `channel`, `ts` — easy to grep dedup-hit-rate in production.

`platform/slack/slack_test.go`
- `newDedupTestPlatform` pre-seeds the user/channel name caches so the resolvers never reach the nil API client — that's the right boundary for testing `handleEvent` end-to-end.
- `freshTS(seq)` uses `time.Now().Add(time.Minute)` to dodge the post-restart old-message filter — clever, no flakiness as long as `core.StartTime` is initialized at process start (which it is, per existing usage).
- 4 tests cover the three scenarios that matter:
  1. Two events for one mention → 1 dispatch
  2. Order independence (Slack doesn't guarantee which subscription lands first)
  3. False-positive guard: distinct messages still dispatch, including the same `ts` in two different channels
  4. `dedupKey` happy path + empty-`ts` → empty-key invariant
- Tests use real `handleEvent`, not isolated unit tests of the guard — exercises the full dispatch path.

### Known limitation (acknowledged in PR body, kept narrow intentionally)
The two handlers still differ in `Content`: `app_mention` runs `stripAppMentionText`, the `message` handler passes `ev.Text` verbatim. After this dedup lands, a leading bot mention is stripped only if the `app_mention` event lands first. This is a real but small inconsistency and is correctly out of scope for this PR (changing what the agent sees is a user-visible behavior change). Worth a follow-up issue.

### Risk
- Pure additive change inside one platform package. No new mutex contention (uses existing `core.MessageDedup`).
- 145+/0-, no deletions. Backwards-compatible: non-duplicate messages dispatch identically.
- CI: 5/5 green. `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` on head `7967af80`.
- Side benefit explicitly called out in the PR body: also dedups Socket Mode redelivery of unacked events — nice two-for-one.

Merging.

PR #1812, opened 2026-09-09
feat(feishu): acknowledge accepted messages with an optional receipt reaction

liuzhao1225:
## Summary

Feishu/Lark users currently wait for agent-session startup or their queued turn before seeing the processing reaction. Add optional `ack_emoji` to send a persistent receipt asynchronously as soon as the engine accepts their message for processing or queueing, so a slow response does not look like a lost request.

This uses the existing `Message.OnAccepted` callback, preserving group-history consumption, without changing core production code or the existing processing/completion lifecycle. Closes #1807.

## Type of change

- [x] New feature (non-breaking, opt-in)

## Manual / user-visible behavior change

```toml
[projects.platforms.options]
ack_emoji = "Get"        # accepted/queued; persistent receipt
reaction_emoji = "OnIt" # existing processing indicator
# done_emoji = "Done"   # existing optional completion notification
```

- Omitted, empty or `"none"` leaves existing behavior unchanged.
- Accepted messages receive their own receipt before agent startup or while waiting behind an active turn. Ignored, unauthorized, duplicate/stale messages, full-queue rejections and handled local commands do not produce receipts. Synthetic scheduled work is not given a receipt.
- The receipt confirms acceptance, not successful execution. It remains after completion, failure or cancellation. Reaction API calls run outside the admission/queue lock with a five-second timeout, and failures do not interrupt agent work.
- When receipt and processing emojis match on the same message, the persistent receipt owns the reaction; typing cleanup does not remove it, including when the receipt HTTP call is still pending.
- An image batch acknowledges its newest canonical message, while preserving its existing reply/typing context.

### Timing boundaries

Receipt starts at **engine acceptance**, after existing parsing/media preparation and image batching. The engine currently holds its shared interactive-state lock during `StartSession`: the first message is acknowledged before that startup, but subsequent messages waiting on that existing lock still wait for admission. Once a session is running, a follow-up accepted into its queue is acknowledged before its turn starts. This PR does not refactor that lock or claim socket-arrival acknowledgement.

### Automated tests

- `platform/feishu/receipt_ack_test.go`: HTTP-backed adapter tests for acceptance-only dispatch, preservation of the existing callback, one-shot receipt creation, default/disabled/synthetic behavior, authorization/mention/dedup/recall gates, slow/failed APIs, HTTP context deadline, different/same/disabled processing emojis, pending-receipt cleanup ordering, and canonical image-batch IDs.
[...]

PR #1826, opened 2026-09-11
feat(feishu): steer queued messages into active Codex turns

### LINKED ISSUE #1825: [Feature] Feishu: add queued messages to the active Codex turn with native steering
### Feature Area

Platform (Feishu/Lark), Agent (Codex), Core

### Problem or Motivation

Messages sent while an agent is working currently enter a FIFO queue. A correction such as “only investigate; do not edit code yet” needs an explicit way to join the active turn instead of becoming the next task.

Codex app-server provides native `turn/steer`. Starting another `codex exec` process is a separate turn and must not be presented as native steering.

### Proposed Solution

Keep ordinary messages queued by default. For an eligible active session, show a receipt card for each queued message:

```text
You: Investigate the login failure.
Agent: Investigating…

You: Only identify the cause; do not change code yet.
Agent: Queued
       Only identify the cause; do not change code yet.
       [Add to current turn] [Cancel queued message]
```

The first action submits that exact message to the original active turn and updates its receipt. Also provide `/steer <message>`, with `/补充` and `/補充` aliases. Acceptance does not mean a running tool stops immediately; stopping remains a separate action.

| Observed state | Behavior |
| --- | --- |
| Original turn active; item still queued | Claim the item and submit with its bound `expectedTurnId`; accepted input is never dequeued again. |
| Promotion races with queue draining | Only one path may consume the item. |
| Original turn ended or another turn started | Never redirect an old button to the new turn; unsubmitted input stays queued. |
| Duplicate click | Return the existing operation state without resubmitting. |
| Server explicitly rejects submission | Preserve FIFO position unless the user has recalled or cancelled the item. |
| Timeout, disconnect, or ambiguous response | Report an unknown outcome and withhold the item from automatic execution; do not retry automatically. |
| Cancel or recall | Remove only unsubmitted queued input; do not claim an in-flight or accepted steer was undone. |
| Stop, new session, or restart | Invalidate old actions; do not replay uncertain submissions. |
| Backend lacks native steering | Keep normal queue behavior and give an accurate unsupported-command response. |

Bind each action to an opaque operation token, platform instance, session, chat/topic, original turn, and sender. Validate the authenticated callback operator and command/role permissions. In shared sessions, only the active turn's owner may steer it, and users may only operate their own queued messages. Preserve body text, existing quote context, images, and files.

### Alternatives Considered

- **Automatically steer every new message:** changes the meaning of ordinary messages and makes independent tasks hard to queue. Keep queue as the default.
- **Quote your own message to steer:** potentially useful as an opt-in shortcut, but quoting may also mean explaining or editing a queued item. For this first version, quotes remain context and explicit receipt buttons choose execution timing. Any future quote control must use real message IDs.
- **Stop and resume, or concurrently call Send:** does not provide native same-turn steering and can interrupt tools or create a competing turn.

### Additional Context

Initial scope:

1. Optional, platform-agnostic core steering capability and explicit queue action states.
2. Codex app-server `turn/steer` with accurate active-turn tracking; exec must explicitly opt out.
3. Feishu task card callbacks using authenticated operators, updating the original card rather than reusing permission-approval behavior.
4. Route `/ps` and `/btw` through native steering when available; prevent concurrent Send for Codex exec while preserving other agents' legacy behavior.
5. Five-language messages, protocol tests, callback tests, and user-perspective critical journey coverage.

QQ/QQBot buttons, quote shortcuts, configurable default steering, and durable queue recovery can follow separately. The existing queue is process-local; this issue does not claim restart persistence.

Acceptance criteria:

- [ ] Queue remains FIFO; steering does not start a new turn.
- [ ] Buttons, direct commands, cancellation, duplicate clicks, and permissions have tests.
- [ ] Turn completion races, unknown outcomes, recalls, and old cards cannot duplicate or misroute input.
- [ ] Unsupported backends and absent active turns receive accurate feedback.
- [ ] Images, files, and existing quote context are preserved; attachment failures cannot report success.
- [ ] Sender, chat/topic, platform instance, and callback identity boundaries have regression coverage.
- [ ] Any screenshots identify whether they show a local preview or a real Feishu client.
- [ ] Build, full tests, race detector, critical user journeys, and CI pass.

Reference: [Codex App Server — Steer an active turn](https://learn.chatgpt.com/docs/app-server#steer-an-active-turn). Related to #619; this proposal explicitly separates same-turn steering from next-turn queueing.

### Willingness to Contribute

- [x] I'd be willing to submit a PR to implement this feature.

liuzhao1225:
## Summary

Closes #1825. Ordinary messages remain queued while a task runs. Eligible Feishu queue receipts offer **Add to this turn** and **Cancel queued message**. Promotion sends the exact input and attachments through Codex App Server `turn/steer`, bound to the original turn and sender, and removes accepted input from the queue.

`/steer` (`/补充`, `/補充`) provides direct steering. `/ps` and `/btw` use native steering when available; Codex exec rejects supplementation instead of starting a competing process. Other agents retain their existing legacy injection behavior.

## Technical design

[Design, state transitions, and scenario-to-test matrix](https://github.com/liuzhao1225/cc-connect/blob/codex/feat-feishu-steer/docs/feishu-steer.md#technical-design-and-review-scope). This follows the same-turn semantics discussed in [#619](https://github.com/chenhg5/cc-connect/issues/619#issuecomment-4252574875) and the review priorities in [#1825](https://github.com/chenhg5/cc-connect/issues/1825#issuecomment-5632166352).

- A typed operation status and session-local submission barrier make queue promotion and draining mutually exclusive. Explicit rejection preserves FIFO position unless recalled; ambiguous delivery prevents automatic replay.
- Callback identity, chat/topic, platform instance, session, original turn, ownership, and effective command permissions are validated. Task actions and task cards have separate paths from permission approval.
- Supplements appear before the turn's answer in stored history even when completion arrives before the RPC response. Streaming and stop handling remain available.
- Feishu task receipts opt into shared updates. Intermediate statuses use toasts; immutable outcomes update the original card, preventing stale intermediate responses from restoring old buttons. Failed asynchronous patches attempt a text fallback in the original topic.
- Only short previews and routing metadata remain in historical receipts; consumed attachments are released. Queue/action state is in memory.

Future scope: quote shortcuts, QQ/QQBot buttons, configurable default steering, and durable queue recovery. No automatic retry or generic card versioning is introduced.

## Type of change

- [x] New feature
- [x] Bug fix: active-turn lifecycle and supplemental-command routing

## Testing

Go 1.25.0; web assets built with the locked dependencies.

- `go build ./...`
- `go test -p 1 ./...`
- `go test -race -p 1 ./...`
- `go test ./core/ -run TestCUJ`
- golangci-lint v2.11.4: `golangci-lint run --new-from-rev origin/main ./...`

One parallel `go test ./...` run failed in the unchanged Antigravity `TestSendDoesNotHoldStdinOpen` (empty output, with a 2-second process timeout). Its isolated rerun passed; full serial normal and race suites are the local gates. The parallel failure is retained in the audit evidence.

### Regression coverage

- Codex protocol payload, attachments, expected-turn guards, explicit rejection versus transport uncertainty, and early/stale completion notifications.
- Queue promotion before and after draining, duplicate clicks, recall/cancel, old controls after session changes, ownership, role permissions, and retained-payload release.
- Late acceptance history ordering; multiline and media command routing; `/ps`/`/btw` entrypoint behavior across native, exec opt-out, and legacy sessions; truthful concurrent-submission feedback.
- Feishu shared-card requirements, shared-app operator routing, callback deadlines, exact-card PATCH, and final text fallback on failed PATCH. Regression failures were reproduced before their fixes using local test transports.
- `TestCUJ_A8_QueueAndSteerHaveDistinctVisibleOutcomes`: four actions through `ReceiveMessage`, asserting user-visible receipts and results. Existing media test fixtures wait for engine completion before temporary-store cleanup.

- [x] A — basic conversation
- [x] B — session lifecycle
- [x] C — execution control
- [x] D — permissions
- [x] G — error handling
- [x] H — session/platform isolation
- [x] I — cards

## User-visible preview

Chinese cards below are generated by the production card builder using synthetic data. **This is a local browser preview, not a Feishu client screenshot.** Live Feishu rendering, callback delivery, and PATCH readback remain unverified.

Feishu queue and steer local preview

[Configuration and reproducible preview instructions](https://github.com/liuzhao1225/cc-connect/blob/codex/feat-feishu-steer/docs/feishu-steer.md).

## Checklist

- [x] Build, full tests, race detector, CUJ, and lint gates pass.
- [x] New user-facing strings cover EN, ZH, ZH-TW, JA, and ES.
- [x] Core uses optional capabilities without hardcoded platform/agent names.
- [x] No credentials or private message data included.
- [x] Based on current main, including #1812's Feishu receipt reaction.


chenhg5 (commented):
## QA Review — PR #1826

Thanks for the unusually thorough PR — the 9-state state machine table, the explicit "what's not in scope" list, and the scenario-to-test matrix made this much easier to evaluate than a typical 3k-line feature.

### Conclusion: **Approve** ✅

This is mergeable. CI is 5/5 green on the current head (`a497bec`), mergeable_state is clean, the diff spans four subsystems (core / feishu / codex / docs) but the boundaries are clean, and the test surface is genuine — not aspirational.

### What looks good

- **State machine is real, not cosmetic.** `queuedTaskStatus` (`taskQueued` / `taskSubmitting` / `taskAccepted` / `taskCancelled` / `taskUnknown` / `taskFailed` / `taskEnded` / `taskExpired`) is the actual driver, and `terminal()` keeps a delayed non-terminal response from overwriting an already-accepted card. That last point is the kind of bug that usually ships because nobody thinks to write the test.
- **Queue promotion vs drain barrier is implemented correctly.** `lockQueueForDrain` blocks on `steerDone` before acquiring the queue slice, and the two callers (`processInteractiveEvents`, `drainPendingMessages`) consistently use it. Rejection preserves FIFO slot; `ErrSteerOutcomeUnknown` does **not** replay.
- **Callback identity is enforced in two places.** Both `core/handleTaskAction` and `platform/feishu/onTaskCardAction` validate operator / chat / message / sessionKey, and `event.Event.Operator.OpenID` cannot be overridden by untrusted card values (sibling-project shared-app case).
- **Supplements appear before the answer.** `Session.insertHistory(index, ...)` + the per-turn `turnHistoryNext` cursor solve the late-acceptance / early-completion race; the regression test even re-reads the saved file to verify ordering survives a reload.
- **`/ps` and `/btw` correctly route to `cmdSteer` when the agent supports steering, return `MsgSteerUnsupported` for `codexSession` (which now opts out of legacy `/ps` exec injection via `SupportsLegacySupplement() = false`), and fall through to legacy `Send` only when both legacy image routing and a non-steering one-shot backend are present. The 64-cell matrix test (`TestSteerCommand_SupplementAttachmentRouting`) actually exercises this.
- **Codex turn lifecycle hardening.** `turnRevision` counter + `retiredTurns` map + `finishTurn(expectedTurnID, …)` make stale `turn/started` notifications and out-of-order completions idempotent. Dropping the `thread/status/changed → completeTurn` branch is the right call.
- **5-language i18n is complete** (en / zh / zh-tw / ja / es) and the markdown preview escape (replaces `\ * _ [ ] ` # ! | ) has its own unit test.
- **Receipt lifecycle is sane.** `registerQueuedTaskLocked` clones only routing metadata + a 500-char preview into `queueActions`; original attachment buffers are released with the FIFO entry. `UpdateAll(true)` on the card means stale intermediate responses can't restore old buttons.

### 🟠 Should improve (P2 — not blocking)

- **Live Feishu callback/PATCH readback is unverified.** `task_action_test.go` uses `httptest.Server` to verify the PATCH path, which is the right unit-level coverage, but the **real** Feishu app's `OpenMessageID` resolution and shared-card operator routing have not been smoke-tested in this PR. The preview HTML/JPG is explicitly local-synthetic. I'd ask the maintainer (or you, post-merge) to run one manual smoke against a dev Feishu app: queue a message → click "Add to this turn" → confirm toast + PATCH replaces the original card → confirm final text fallback if PATCH fails. If you have time to add a one-liner log note in `docs/feishu-steer.md` ("Manually verified on dev app YYYY-MM-DD, see …"), that's the cheapest mitigation.

### 🔵 Optional (P3)

- **`lockQueueForDrain` has no recover safety net.** If `s.Steer` (or any call between `state.steerDone = done` and `close(done)`) panics, the queue drain is permanently blocked waiting on `steerDone`. Probability is low because the path doesn't touch user data and uses `errors.Is`, but `defer func() { if r := recover(); r != nil { close(done); panic(r) } }()` is a cheap hardening. Not a blocker for this PR.
- **The Antigravity `TestSendDoesNotHoldStdinOpen` flake** you disclosed in the PR description is not in any skip list. Adding `t.Skip` or `-run='!TestSendDoesNotHoldStdinOpen'` to the CI matrix would silence it cheaply. Author says isolated rerun is green; this matches what I see in the test code (no shared state).

### ❓ Questions

- The PR body says "active-turn lifecycle and supplemental-command routing" counts as a bug fix. I read it as a feature with two small lifecycle tightenings; curious whether you filed the `thread/status/changed` late-completion behavior under a specific reported issue — if not, "feature" is the right label.

### Risk summary

- **Scope:** Medium. Four subsystems touched; boundaries are explicit (core has the queue/steer state machine, feishu has the typed callback, codex has the wire protocol). No cross-package implicit coupling.
- **Compatibility:** Backward-compatible by construction. Steer path is opt-in via `CardTaskActionHandlerSetter` + `AgentSessionSteerer` + same-user + same-session. Default queue-with-text-receipt fallback still works.
- **Data / permissions:** No schema change. `Session.History` external API unchanged. Receipts retain only routing metadata + preview text; attachments released on dequeue.
- **Rollback:** Single-PR revert restores original FIFO behavior. `queueActions` and `steerDone` are in-memory only.

### Verified evidence

- ✅ CI: 5/5 green on `a497bec05f282dbd54fd17c0182e5ad9b2390f07` (lint / unit-test / smoke-test / regression-test / performance-test)
- ✅ Local author: `go build ./...`, `go test -p 1 ./...`, `go test -race -p 1 ./...`, `go test ./core/ -run TestCUJ`, `golangci-lint run --new-from-rev origin/main ./...`
- ✅ Test files reviewed: `core/steer_test.go` (725 LOC), `core/steer_regression_test.go` (342 LOC), `platform/feishu/task_action_test.go` (325 LOC), `agent/codex/appserver_steer_test.go` (203 LOC), `core/steer_preview_test.go` (59 LOC), `core/cuj_test.go` (CUJ A8).
- ✅ State machine + barrier + race + identity + late-acceptance + concurrent input + exec opt-out each have a dedicated test.

### Next step

Maintainer: review and merge when convenient. The only thing I'd ask you to do post-merge is a one-time manual smoke against the dev Feishu app, and optionally bump `TestSendDoesNotHoldStdinOpen` to a skip in CI.

Excellent work. 🙏

PR #1827, opened 2026-09-11
feat(claudecode): add context_window_tokens option for ctx% indicator (#1823)

### LINKED ISSUE #1823: [Feature] Allow configuring the context window used by the claudecode agent's ctx% indicator (hardcoded 200K default)
**Feature Area:** Configuration (also affects Agent: Claude Code)

## Problem or Motivation

When the claudecode agent runs a custom/local model via `ANTHROPIC_BASE_URL` (e.g. a llama.cpp server serving a 27B model with a 128K context), the `[ctx: ~N%]` context indicator in the reply footer computes against a **hardcoded 200K window** for any model id that doesn't contain `[1m]`.

The numerator (used tokens from stream-json assistant usage) is accurate; the denominator is wrong. For a 128K model the displayed percentage underestimates true usage by 200/128 ≈ **1.56×** — the indicator reads 64% exactly when the model's real 128K limit is fully exhausted. This also misleads users about when to run `/compact`.

Source (v1.5.0, `agent/claudecode/session.go`):

```go
func claudeContextWindow(model string) int {
	lower := strings.ToLower(strings.TrimSpace(model))
	if lower == "" {
		return 200_000
	}
	if strings.Contains(lower, "[1m]") {
		return 1_000_000
	}
	return 200_000
}
```

## Proposed Solution

Add an optional `context_window_tokens` option to the claudecode agent. When > 0, use it as the denominator for the ctx% indicator; otherwise fall back to the current model-id heuristic.

Example:

```toml
[projects.agent]
type = "claudecode"
model = "qwen3.8-27b"
context_window_tokens = 128000
```

(Optionally a model-id → window map for projects that switch between models of different sizes.)

## Alternatives Considered

- `max_context_tokens`: this option is passed to the Claude Code CLI as `--max-context-tokens`, but that flag does not exist in the CLI (verified on Claude Code 2.1.263: `error: unknown option '--max-context-tokens'`), and it would not affect the indicator's denominator anyway.
- Local workaround: mentally multiply the displayed % by the window ratio (1.5625 for 128K) and treat the raw input-token count in the footer as the source of truth.

## Additional Context

- Verified on v1.5.0 (commit 17c61062); the v1.5.1-beta.1 diff does not change `claudeContextWindow`.
- Relevant to everyone running non-Anthropic models through the claudecode agent adapter (local llama.cpp / Ollama / gateway relays), i.e. any real context window ≠ 200K/1M.

chenhg5:
## Summary

Adds a configurable `context_window_tokens` option to the claudecode agent that overrides the model-name heuristic used by the IM reply "ctx N%" indicator.

## Problem

When running Claude Code through a custom router (Claude Code Router, third-party proxy, fine-tuned model), the model id reported back often does not match Claude Code's naming convention. The ctx% indicator falls back to the 200K default and produces misleading percentages like "ctx 3%" when the actual window is, e.g., 32K.

## Fix

New opt-in option:

```toml
[projects.agent.options]
context_window_tokens = 32000   # pin a specific window size
# or omit / set <= 0 to keep the existing model-id heuristic (200K, 1M for [1m] variants)
```

When `> 0`, the override short-circuits the heuristic at the single call site in `handleAssistant` (`agent/claudecode/session.go`). When `<= 0` or absent, behavior is unchanged.

## Files changed

- `agent/claudecode/claudecode.go` — new `contextWindowTokens` field on Agent; `New()` parses int/int64/float64 (same pattern as `max_context_tokens`); propagated through `StartSession` and `WorkspaceAgentOptions`.
- `agent/claudecode/session.go` — new `ctxWindowOverride` field on `claudeSession`; extended `claudeContextWindow(model, override int)` signature; updated the single call site.
- `agent/claudecode/claudecode_test.go` — 18 sub-tests across 3 new tests: `TestNew_ParsesContextWindowTokens`, `TestWorkspaceAgentOptions_PropagatesContextWindowTokens`, `TestClaudeContextWindow`.
- `agent/claudecode/session_test.go` — updated `newClaudeSession` call site for the new arg.
- `config.example.toml` — documented the new option next to `reasoning_effort`.

## Validation

- `go build ./agent/claudecode/ ./core/` — clean
- `go vet ./agent/claudecode/ ./core/` — clean
- `go test ./agent/claudecode/` — 18 new sub-tests pass, no regressions
- `go test -race ./agent/claudecode/` — clean

## Risk

Minimal. The new parameter is opt-in (default `<= 0` keeps the existing heuristic). The single call site update is a one-line diff. Workspace propagation mirrors the existing `max_context_tokens` pattern, so multi-workspace users see the new option without further config work.

Closes #1823

### DIFF
--- a/agent/claudecode/claudecode.go
+++ b/agent/claudecode/claudecode.go
@@ -45,8 +45,9 @@ type Agent struct {
 	mode             string // "default" | "acceptEdits" | "plan" | "auto" | "bypassPermissions" | "dontAsk"
 	allowedTools     []string
 	disallowedTools  []string
-	maxContextTokens int // optional: passed as --max-context-tokens when > 0
-	providers        []core.ProviderConfig
+	maxContextTokens    int // optional: passed as --max-context-tokens when > 0
+	contextWindowTokens int // optional: override the context-window-size heuristic used by the ctx% indicator. When <= 0, fall back to model-name heuristics.
+	providers           []core.ProviderConfig
 	activeIdx        int // -1 = no provider set
 	sessionEnv       []string
 	routerURL        string   // Claude Code Router URL (e.g., "http://127.0.0.1:3456")
@@ -209,6 +210,27 @@ func New(opts map[string]any) (core.Agent, error) {
 		}
 	}
 
+	// context_window_tokens overrides the model-name heuristic used by
+	// the "ctx N%" indicator. Defaults to 0 (= heuristic), accepting the
+	// same numeric types as max_context_tokens because config parsers
+	// surface integer fields through one of these shapes depending on
+	// the source (TOML → float64, programmatic → int / int64).
+	contextWindowTokens := 0
+	switch v := opts["context_window_tokens"].(type) {
+	case int:
+		if v > 0 {
+			contextWindowTokens = v
+		}
+	case int64:
+		if v > 0 {
+			contextWindowTokens = int(v)
+		}
+	case float64:
+		if v > 0 {
+			contextWindowTokens = int(v)
+		}
+	}
+
 	// Claude Code Router support
 	routerURL, _ := opts["router_url"].(string)
 	routerAPIKey, _ := opts["router_api_key"].(string)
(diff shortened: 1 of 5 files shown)


chenhg5:
## QA Gate: needs rebase

`mergeable=MERGEABLE` is now `false` / `mergeStateStatus=DIRTY` — the branch is behind `main` and needs a rebase before this can merge.

CI itself is still 5/5 green on the current head `214e198c`, so this is purely a base-vs-head conflict, not a regression.

Since main moved through ~9 merges today (Tier 1 batch + #1838), a fresh rebase + CI re-run will close it. Happy to re-review once the branch is current.

No code-level blockers from my side — the diff itself looks clean (opt-in `context_window_tokens` override for the ctx% heuristic, parsed as int/int64/float64 with `<= 0` keeping the existing heuristic, propagated through `WorkspaceAgentOptions`, 18 new sub-tests, documented in `config.example.toml`).

chenhg5 (commented):
## QA Review: PR #1827

### Overall assessment
Small, well-scoped, opt-in feature. Single call site change (`claudeContextWindow(model, override int)`) with comprehensive test coverage (18 new sub-tests across parser, propagation, and heuristic+override paths). Default behavior unchanged (`override <= 0` falls back to existing heuristic).

### Review scope
- `agent/claudecode/claudecode.go`, `session.go`, `claudecode_test.go`, `session_test.go`, `config.example.toml`
- +189/-24 in 5 files
- Focus: correctness, config parsing, multi-workspace persistence, test coverage

### ✅ What looks good
- **Parser pattern reuse**: `int` / `int64` / `float64` switch mirrors `max_context_tokens`; future contributors will recognize the idiom immediately.
- **Override-wins branch with explicit fall-back**: `if override > 0 { return override }` placed before the heuristic is the simplest correct ordering; negative/zero/missing all converge to existing behavior.
- **Multi-workspace persistence**: `WorkspaceAgentOptions()` explicitly checks `> 0` before adding the key, so zero-value won't pollute the workspace config map. `TestWorkspaceAgentOptions_PropagatesContextWindowTokens` locks this in.
- **Table-driven `TestClaudeContextWindow`**: covers empty model, standard, `[1m]`, case-insensitive `[1M]`, surrounding whitespace — and the four override-wins cases (override > heuristic / 0 / negative). Adding future model variants will be one row.
- **Bilingual config docs**: matches existing `reasoning_effort` doc style in `config.example.toml`.

None material.

- **CHANGELOG / release note**: the new option is worth mentioning in the next release notes so users on custom routers find it via docs search instead of digging through `config.example.toml`.
- **Other agents**: `pi` / `cursor` / `opencode` / `kimi` may have analogous ctx% heuristics that suffer the same custom-router mismatch. Worth tracking as a separate issue if/when users report it; no need to expand scope here.

### 🚨 / 🔴 Must fix

### ❓ Questions
None.

### Testing / Risk
- **CI**: 5/5 green (lint, unit, regression, smoke, performance — Run 34584637538)
- **Author-validated**: `go build`, `go vet`, `go test`, `go test -race ./agent/claudecode/` all clean per PR description
- **QA-verified**: diff inspected; `claudeContextWindow` has exactly one caller in the repo and it was updated; no other call sites broken
- **Remaining risk**: low. Default behavior unchanged, opt-in config, single-call-site change, full test coverage on parser + heuristic + override paths

### Next step
Maintainer can self-merge per standard chenhg5-authored PR convention. The two P3 follow-ups (CHANGELOG, sibling-agent audit) are non-blocking and can be filed as separate issues.

PR #1828, opened 2026-09-11
feat(feishu): configurable local card hooks for instant button feedback

MingChaoSun:
## Motivation

Interactive cards (surveys, polls, approvals) currently need a full agent round-trip on every button click before the card updates — typically several seconds, and the update itself becomes just another chat message.

This PR adds an opt-in mechanism for **sub-second, in-place card updates driven by a local script**, no agent involved. It has been running in production on a personal deployment for 14+ days driving multi-question survey cards (10 questions; clicks feel instant and the agent is never interrupted).

## How it works

A button value that carries a `hook` field triggers synchronous local execution when clicked:

```json
{"action": "cmd:whatever", "hook": "survey.sh", "q": 3, "choice": "B"}
```

- The **full button value** is piped to the script's stdin as JSON.
- If the script's **stdout parses as a JSON object**, it is returned as the replacement card in the card-action callback response (instant in-place update, buttons included — which `after_click` can't do since it only renders text).
- Any failure (missing script, timeout, non-JSON output) **logs a warning and falls back to normal dispatch** — a broken hook can never block message handling.

## Configuration (opt-in, zero behavior change by default)

```toml
[projects.platforms.options]
card_hook_dir = "/path/to/hooks"      # empty (default) = feature disabled
card_hook_timeout_ms = 3000              # range [100, 30000], default 3000
```

Applies to both `feishu` and `lark` platforms (shared code path).

## Safety

- Hook names must resolve **inside `card_hook_dir`**: absolute paths and `..` traversal are rejected (`safeHookPath`).
- Execution is bounded by `card_hook_timeout_ms`; `cmd.WaitDelay` additionally caps the grace period so a killed hook that leaves pipe-holding children cannot stall the timeout (covered by a test).
- Scripts run via `bash` with the hook dir as cwd.

## Tests

6 new unit tests in `platform/feishu/card_hook_test.go`: path-traversal rejection, JSON card passthrough, stdin protocol (value JSON echoed back), non-JSON output, missing script, and real timeout enforcement. `go build ./...` and `go vet` pass; full `go test ./...` passes except two timing-sensitive tests in `agent/antigravity` / `agent/codex` that occasionally flake under parallel full-suite load on my machine, before and after this change — unrelated.

## 中文说明

新增可选的卡片本地钩子：按钮 value 带 `hook` 字段时，点击同步执行 `card_hook_dir` 下的脚本，stdin 收到完整 value JSON，stdout 为 JSON 对象则直接作为替换卡片返回——亚秒级原地刷新卡片（可带按钮，弥补 after_click 只能文字的不足），完全不占用 agent。默认关闭（`card_hook_dir` 为空零行为变化），失败自动回退正常派发，路径安全护栏 + 超时封顶 + 6 个单元测试。

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1012,6 +1012,20 @@ app_secret = "your-feishu-app-secret"
 # domain = "https://open.feishu.cn"  # Optional runtime API/WebSocket base URL override / 可选：运行时 API/WebSocket 域名覆盖
 # enable_feishu_card = true  # Enable Feishu interactive cards (default: true); set false to force plain-text replies
 #                              # 启用飞书交互式卡片（默认 true）；设为 false 时强制回退纯文本回复
+# card_hook_dir = ""         # Optional directory of local hook scripts for instant card-button feedback.
+#                              # A button value like {"action":"cmd:...","hook":"myscript.sh"} runs that script
+#                              # synchronously on click: the full button value JSON goes to the script's stdin;
+#                              # if stdout is a JSON object it is returned as the replacement card immediately
+#                              # (sub-second feedback, no agent round-trip). Any failure (missing script, timeout,
+#                              # non-JSON output) falls back to normal dispatch, so hooks never block messages.
+#                              # Scripts must stay inside this directory (absolute paths and ".." are rejected).
+#                              # Empty (default) disables the feature. Also applies to the lark platform.
+#                              # 可选：本地钩子脚本目录，用于卡片按钮的即时反馈。按钮 value 含 {"hook":"myscript.sh"}
+#                              # 时，点击同步执行该脚本：完整 value JSON 走 stdin；stdout 为 JSON 对象则立即作为
+#                              # 替换卡片返回（亚秒级反馈，不经过 agent）。任何失败（脚本不存在/超时/输出非 JSON）
+#                              # 都自动回退正常派发，绝不阻塞消息。脚本必须位于该目录内（绝对路径与 ".." 会被拒绝）。
+#                              # 留空（默认）即关闭该功能。lark 平台同样适用。
+# card_hook_timeout_ms = 3000  # Card hook execution timeout in ms, range [100, 30000] / 钩子脚本执行超时（毫秒），范围 [100, 30000]
 # allow_from = "*"          # Allowed user open_ids, comma-separated; "*" = all (default). Use /whoami to get your open_id.
 #                              # 允许的用户 open_id，逗号分隔；"*" 表示所有（默认）。发送 /whoami 获取你的 open_id。
 # allow_chat = "*"          # Allowed group chat_ids, comma-separated; "*" = all (default). Find chat_id in group settings.
--- a/platform/feishu/card_hook_test.go
+++ b/platform/feishu/card_hook_test.go
@@ -1,0 +1,103 @@
+package feishu
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+func writeHookScript(t *testing.T, dir, name, body string) {
+	t.Helper()
+	p := filepath.Join(dir, name)
+	if err := os.MkdirAll(filepath.Dir(p), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(p, []byte(body), 0o755); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestSafeHookPath(t *testing.T) {
+	dir := filepath.Join(t.TempDir(), "hooks")
+	cases := []struct {
+		name string
+		want string
+		ok   bool
+	}{
+		{"survey.sh", filepath.Join(dir, "survey.sh"), true},
+		{"sub/vote.sh", filepath.Join(dir, "sub", "vote.sh"), true},
+		{"", "", false},
+		{"/etc/passwd", "", false},
+		{"../escape.sh", "", false},
+		{"a/../../escape.sh", "", false},
+		{".", "", false},
+	}
+	for _, c := range cases {
+		got, err := safeHookPath(dir, c.name)
+		if c.ok {
+			if err != nil || got != c.want {
+				t.Errorf("safeHookPath(%q) = %q, %v; want %q, nil", c.name, got, err, c.want)
+			}
+		} else if err == nil {
+			t.Errorf("safeHookPath(%q) = %q; want error", c.name, got)
+		}
+	}
+}
+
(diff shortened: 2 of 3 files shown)


chenhg5 (approved):
## QA Review: PR #1828

This PR adds a clean, opt-in "instant card-action feedback" hook mechanism. A local script in a configured directory is invoked synchronously on card-button click, returning a JSON-object stdout that the Feishu callback uses as the in-place replacement card. Failures fall back to normal dispatch, so a broken hook can never block message handling. The safety surface here is non-trivial (executing local scripts in response to network events), and the implementation does the right things.

### ✅ What looks good
- **Path-traversal hardening is correct.** `safeHookPath` rejects empty names, absolute paths, and any cleaned path that doesn't sit under `cardHookDir` (via `strings.HasPrefix(clean, dir+sep)` — note the `sep` suffix prevents the `dir == "/foo"` accepting `/foobar` attack). The `TestSafeHookPath` table covers `../escape.sh`, `a/../../escape.sh`, `.`, and absolute paths.
- **Bounded execution.** `cmd.WaitDelay = time.Second` is the right call — without it, a hook that spawns children that hold stdout/stderr would keep `cmd.Output()` blocked even after the parent is killed. This is exactly the pitfall Go's `exec` package docs call out.
- **Context-scoped timeout.** `exec.CommandContext(ctx, ...)` ensures the process group is signaled on timeout. Timeout range is hard-validated to `[100, 30000]` ms at config-load time, so a typo can't yield a 5-second "instant feedback".
- **Output handling is strict but predictable.** Any non-JSON stdout, empty stdout, missing script, or exec error returns an error → caller logs `slog.Warn` and falls back to normal dispatch. `TestRunCardHookNonJSONStdout` and `TestRunCardHookMissingScript` cover the negative paths.
- **Stdin is the full button value.** This is the contract — the hook sees exactly what the user clicked (including `action`, `q`, `choice`, etc.), so it can route without an extra round-trip. `TestRunCardHookReceivesValueOnStdin` locks this in.
- **Symmetric config plumbing.** `card_hook_dir` and `card_hook_timeout_ms` are wired into both `newPlatform` sites for `feishu` and `lark` (shared options handling), and the TOML example documents both languages with the security caveat about absolute paths and `..`.
- **Backward compatible.** Empty `card_hook_dir` (default) means the `if p.cardHookDir != ""` branch is skipped entirely — zero behavior change for existing deployments.
- **`filepath.Abs` on load.** Pre-normalizes the directory so the `HasPrefix` check in `safeHookPath` doesn't have to deal with `card_hook_dir = "./hooks"` vs `/abs/path/hooks` mismatch.

- **🟠 P2 — Consider documenting/limiting the hook signature.** A hook that uses `bash -c` could still execute arbitrary commands if the script file is writable by the cc-connect user. The current model trusts the directory operator (which is fine — same trust model as a crontab). For paranoia, you could optionally restrict to a whitelist, but this is a project-policy decision, not a bug. **No blocker.**
- **🟠 P2 — `cmd.WaitDelay = time.Second` is documented inline but not configurable.** If a user runs a hook that legitimately spawns long-lived background workers (e.g. logging to a file), the 1-second grace is fixed. Probably fine for the target use case (instant feedback), but worth a one-line note in `config.example.toml` saying "if your hook spawns children, ensure they exit within ~1s after the script returns". **Optional, not a blocker.**
- **🟠 P2 — Verify the lark platform wires the same options.** The PR body says "applies to both feishu and lark platforms". Reading the diff I see the option is plumbed through `coerceMilliseconds` + `newPlatform(name, domain, opts)` which `lark` likely shares via the same options path. Please confirm in your PR description that lark's call site reads these keys (or add a one-liner in `platform/lark/` if it's wired through a different `newPlatform`). **Not a blocker — just a clarification request.**

### 🔵 Optional / Nit
- **🔵 P3 — `safeHookPath` accepts names with `sub/vote.sh` style subdirs** (the test confirms this), which is nice. But the test only covers clean paths; consider adding `sub//./vote.sh` or `sub/.//vote.sh` if you want to be doubly sure `filepath.Clean` neutralizes them (it should).
- **🔵 P3 — The `cardHookDir` field is a plain `string`, not a `sync.RWMutex`-protected field.** Reading the existing struct layout (other fields like `progressStyle`, `useInteractiveCard` are also plain), and the fact that `newPlatform` runs before any concurrent access, this is consistent with the package's existing pattern. Just noting for completeness.

### ❓ Questions
- **❓ Does the hook's stdout need to be UTF-8 only, or will non-UTF-8 bytes in the JSON break `json.Unmarshal`?** Reading the code, `bytes.TrimSpace(out)` → `json.Unmarshal` → strict UTF-8 validation is implicit. `TestRunCardHookJSONCard` uses ASCII so this isn't covered. **Not a blocker** — JSON is defined as UTF-8 by RFC 8259, but a hook author outputting raw bytes might be confused. Consider a one-line note in the example: "stdout must be valid UTF-8 JSON".

### Testing / Risk
- **CI**: 5/5 green (lint, unit, smoke, regression, performance) ✅
- **Author tests**: 6 new unit tests in `card_hook_test.go` — `TestSafeHookPath` (7 cases), `TestRunCardHookJSONCard`, `TestRunCardHookReceivesValueOnStdin`, `TestRunCardHookNonJSONStdout`, `TestRunCardHookMissingScript`, `TestRunCardHookTimeout`.
- **Verified by reading**: `runCardHook` is the only place that touches the FS / spawns processes; the on-card-action integration in `onCardAction` is wrapped in a single `if p.cardHookDir != ""` block with the fallback path clearly intact.
- **Not covered**: live Feishu card-action callback with a real hook. The unit tests cover the script-execution surface; the integration with Feishu's CardActionTriggerResponse is exercised by smoke-test but not explicitly by a card-hook-aware test. Given the small surface and the explicit "broken hook falls back" guarantee, I don't see this as a blocker.
- **Risk surface**: medium-to-low. The trust model assumes the directory operator can be trusted to put non-malicious scripts there (same as a crontab). Path-traversal hardening is correct. Worst-case blast radius is "what the cc-connect user can run locally" which is no worse than today.

### Next step
1. Address the P2 clarifications (lark wiring confirmation, optional stdout-UTF-8 note).
2. Then ready for maintainer merge.

**Impact**: small-to-medium (new opt-in feature for card-button UX; zero behavior change for existing deployments).  
**Module**: `platform/feishu/feishu.go` (and shared options path with `lark`).  
**Backward compat**: 100% (default = empty dir = feature disabled).  
**Breaking change**: no.  
**Bug fix**: no (feature).

Issue #35, opened 2026-03-08
[Bug] macOS launchd restart often fails with 'Bootstrap failed: 5' on first run


xxb:
## Summary

On macOS, `cc-connect daemon restart` for the launchd-managed service often fails on the first run with:

```text
Restart failed: restart: Bootstrap failed: 5: Input/output error
Try re-running the command as root for richer errors. (exit status 5)
```

Running the same restart command again usually succeeds.

This looks macOS-specific / launchd-specific.

## Environment

- `cc-connect v1.2.0`
- commit: `35cd718`
- built: `2026-03-08T14:45:03Z`
- macOS `26.3` (`25D125`)
- install path: `/opt/homebrew/bin/cc-connect`
- daemon mode: `launchd`

## Reproduction

1. Install `cc-connect` as a launchd agent on macOS.
2. Confirm the daemon is running:
   ```bash
   cc-connect daemon status --work-dir "$HOME/.cc-connect"
   ```
3. Run:
   ```bash
   cc-connect daemon restart --work-dir "$HOME/.cc-connect"
   ```
4. The first restart often returns the bootstrap error above.
5. Run the same command again.
6. The second restart usually succeeds.

## Expected

`cc-connect daemon restart` should succeed in one run.

## Actual

The first restart frequently fails with `Bootstrap failed: 5: Input/output error`.

## Notes

- `daemon status` reports the service as `Running` before and after the failed restart attempt.
- In local testing, this appears related to the macOS `launchd` restart path rather than the config itself.
- The current `launchd` restart implementation does `bootout` and then immediately `bootstrap` without any wait / retry:
  - `daemon/launchd.go` `Restart()`
- That looks like it may be hitting a launchd timing/race issue on macOS.

## Possible direction

If helpful, maybe `Restart()` on macOS should mirror a safer stop/start flow with a short wait and/or health check between `bootout` and `bootstrap`.

PR #1834, opened 2026-09-12
fix(daemon/launchd): prefer in-place kickstart on Restart; wait for label to disappear after bootout (#1833)

### LINKED ISSUE #1833: [Bug] macOS: daemon restart unloads the service for good when the daemon is busy (bootout/bootstrap race, follow-up to #35)
## Summary

On macOS, `cc-connect daemon restart` can leave the service **completely unloaded** when the running daemon is busy. `Restart()` does `bootout` and then retries `bootstrap` only 3 times with 500 ms backoff (~1 s total). `launchctl bootout` returns before the job is actually removed, and a daemon that has agent sessions in flight needs several seconds to tear them down. All three `bootstrap` attempts then fail with `Bootstrap failed: 5: Input/output error`, `Restart()` returns an error, and the job is gone from launchd. `KeepAlive` cannot help because the job is no longer loaded.

This is the same race as #35. The fix for #35 (3 × 500 ms retry) is only enough when the daemon is idle and exits within ~1 s.

## Environment

- cc-connect v1.4.1 (Homebrew), launchd LaunchAgent in `gui/501`
- macOS 26 (Darwin 25.5.0)
- 4 projects (3 × claudecode, 1 × codex), Feishu platform

## What happened (timeline from logs)

A cron job (credential sync) calls `cc-connect daemon restart` roughly every 8 hours. It had worked ~20 times in a row. Then:

```text
# cc-connect.log
01:26:10  message received ... session spawned (codex)
01:27:00  WARN slow agent first event elapsed=50s        <- a turn is in flight
01:28:52  shutting down...
01:28:52  claudeSession: exited cleanly after stdin close (x2)
          (no "engine started" ever follows)

# unified log (log show --predicate 'eventMessage CONTAINS "cc-connect"')
01:28:52.005 launchd  [gui/501/com.cc-connect.service [43030]] bootout initiated by: launchctl[41139]<-cc-connect[41131]
01:28:53.087 mDNSResponder  DNSServiceCreateConnection STOP PID[41131](cc-connect)   <- restart CLI already exited
01:28:57.032 mDNSResponder  DNSServiceCreateConnection STOP PID[43030](cc-connect)   <- old daemon exits 5 s after bootout
01:28:57.033 launchd  [gui/501] service inactive: com.cc-connect.service
01:28:57.033 launchd  [gui/501] removing service: com.cc-connect.service
```

Afterwards:

```text
$ cc-connect daemon status
  Status:    Stopped
$ launchctl print gui/501/com.cc-connect.service
Could not find service "com.cc-connect.service" in domain for user gui: 501
```

The CLI finished at 01:28:53 (bootout + 3 failed bootstraps ≈ 1 s), while the old process took until 01:28:57 to exit because it was still winding down an in-flight session. Previous successful restarts all happened while the daemon was idle.

## Why it matters

`Restart()` is the documented way to pick up new config / credentials, and it is the natural thing to call from cron. When it hits this race the outcome is worse than not restarting at all: the service is unloaded and stays down until someone runs `daemon start` by hand.

## Proposed fix

After `bootout`, wait until the label has actually disappeared from the domain before calling `bootstrap` (bounded by a timeout), instead of relying on a fixed 1 s retry window. `launchctl print <target>` returns non-zero once the job is removed, so it can be used as the readiness check. Diff against `daemon/launchd.go` @ v1.4.1 / current `main`:

```diff
 func (*launchdManager) Restart() error {
 	domain := preferredLaunchdDomain()
 	if loadedDomain, _, _, ok := loadedLaunchdTarget(); ok && domain != launchdGUIDomain() {
 		domain = loadedDomain
 	}
 	target := launchdTarget(domain)
 	bootoutLaunchdTargets()

+	// launchctl bootout returns before the job is actually removed. A busy
+	// daemon (agent sessions in flight) can take several seconds to exit, and
+	// bootstrap fails with "Bootstrap failed: 5" for as long as the label is
+	// still present. Wait for it to disappear before re-bootstrapping.
+	waitLaunchdTargetsGone(30 * time.Second)
+
 	plistPath := launchdPlistPath()

 	// launchd bootout is asynchronous; retry bootstrap with backoff
 	// to avoid "Bootstrap failed: 5" race condition.
 	var out string
 	var err error
 	for i := 0; i < 3; i++ {
 		if i > 0 {
 			time.Sleep(500 * time.Millisecond)
 		}
 		out, err = runLaunchctl("bootstrap", domain, plistPath)
 		if err == nil {
 			break
 		}
 	}
 	if err != nil {
 		return fmt.Errorf("restart: %s (%w)", out, err)
 	}
 	if _, err := runLaunchctl("kickstart", "-kp", target); err != nil {
 		return fmt.Errorf("restart kickstart: %w", err)
 	}
 	return nil
 }
+
+// waitLaunchdTargetsGone blocks until no launchd domain still knows the
+// service label, or until timeout elapses.
+func waitLaunchdTargetsGone(timeout time.Duration) {
+	deadline := time.Now().Add(timeout)
+	for time.Now().Before(deadline) {
+		if _, _, _, ok := loadedLaunchdTarget(); !ok {
+			return
+		}
+		time.Sleep(200 * time.Millisecond)
+	}
+}
```

The existing 3 × 500 ms retry can stay as a safety net for the remaining tiny window between "label gone" and "bootstrap".

An alternative that avoids the race entirely: when the job is already loaded, restart in place with `launchctl kickstart -k <target>` (launchd kills the old instance and waits for it before relaunching; the job never leaves the domain), and only fall back to bootout + bootstrap when the plist changed or the job is not loaded. That is what I switched my cron job to as a workaround, and it has been reliable.

Happy to turn either approach into a PR if you prefer one over the other.

chenhg5:
Closes #1833. Related to #35 (original 3 \xc3\x97 500 ms retry that only covered the idle-daemon case).

## Root cause

When the daemon has agent sessions in flight, `launchctl bootout` returns BEFORE launchd has actually removed the job; the old daemon needs several seconds to tear down. The existing 3 \xc3\x97 500 ms bootstrap retry (\xe2\x89\x881 s total) then races against that teardown window and fails with `Bootstrap failed: 5: Input/output error` for every attempt. `Restart()` returns an error, the job has been removed from launchd, and KeepAlive cannot recover because the service is no longer loaded.

Reporter leishao provided a clear side-by-side timeline from `cc-connect.log` + unified log showing the daemon exiting 5 s after bootout, while the CLI finished bootout + 3 failed bootstraps in ~1 s. The previous 20+ successful restarts all happened while the daemon was idle, masking the race.

## Fix

Two complementary paths in `daemon/launchd.go::Restart()`:

1. **Fast path** \xe2\x80\x94 when the job is already loaded AND the canonical plist still exists on disk, use `launchctl kickstart -k <target>` (in-place). launchd sequences the SIGTERM/relaunch itself, so the bootout/bootstrap race cannot occur on this path.
2. **Fallback path** \xe2\x80\x94 bootout + a new `waitLaunchdTargetsGone(30s)` helper that polls `loadedLaunchdTarget()` every 200 ms until the label disappears, THEN run the existing 3 \xc3\x97 500 ms bootstrap retry as a safety net for genuine bootstrap races.

The plist-on-disk check is intentional: if a user has replaced the plist (binary path, env, etc.) they almost always want a full reinstall via `Install()`, not an in-place reload. The bootout path remains for the rare case where the job is not loaded or the plist has been removed between the daemon writing it and us restarting.

## Tests added (darwin tag)

- `TestRestartPrefersInPlaceKickstartWhenLoadedAndPlistPresent` \xe2\x80\x94 verifies the in-place path is taken when conditions are met.
- `TestRestartFallbackToBootoutWhenJobNotLoaded` \xe2\x80\x94 verifies bootout + wait + bootstrap + kickstart when job is not loaded.
- `TestRestartFallbackToBootoutWhenPlistMissing` \xe2\x80\x94 verifies fallback when plist is removed (reinstall scenario).
- `TestWaitLaunchdTargetsGoneReturnsImmediatelyWhenAlreadyGone` \xe2\x80\x94 no wasted polls.
- `TestWaitLaunchdTargetsGoneReturnsAfterJobDisappears` \xe2\x80\x94 returns after the label actually disappears.
- `TestWaitLaunchdTargetsGoneHonoursTimeout` \xe2\x80\x94 soft timeout fires within ~5 poll intervals.

## Local validation

- `gofmt -l daemon/launchd.go daemon/launchd_test.go` clean.
- `GOOS=darwin go vet ./daemon/` clean.
- `GOOS=darwin go test -c -o /tmp/daemon-darwin.test ./daemon/` builds cleanly (darwin binary; full execution requires macOS CI runner).
- `go test ./daemon/` passes on linux tag (cross-platform surface unchanged).

## Risk

- **Low** \xe2\x80\x94 changes are isolated to `daemon/launchd.go` and its test file. No public API change, no platform/agent code touched.
- The in-place path is opt-in by virtue of the loaded + plist-present check; on a freshly bootstrapped host (job not loaded) the fallback path runs exactly as before.
- The 3 \xc3\x97 500 ms bootstrap retry is preserved, so any pre-#1833 success mode still works.

## Files

- `daemon/launchd.go` \xe2\x80\x94 +84 / -4 lines (added `Restart()` fast path, `waitLaunchdTargetsGone` helper, 3 named constants for the timeouts).
- `daemon/launchd_test.go` \xe2\x80\x94 +320 lines (6 new tests).

### DIFF
--- a/daemon/launchd.go
+++ b/daemon/launchd.go
@@ -127,18 +127,76 @@ func (*launchdManager) Stop() error {
 	return nil
 }
 
+// launchdRestartInPlaceTimeout is the wall-clock budget for an in-place
+// `launchctl kickstart -k` to deliver SIGTERM and reap the old daemon. The
+// actual SIGTERM-to-exit time depends on what the daemon is doing (draining
+// interactive agent sessions, flushing logs, etc.) and can legitimately take
+// 5–30 s on a busy host. If the kickstart does not complete within this
+// budget we surface the error rather than fall through to bootout, because
+// the in-place path is precisely the one that avoids the bootout/bootstrap
+// race — silently retrying via bootout would re-introduce the race.
+const launchdRestartInPlaceTimeout = 60 * time.Second
+
+// launchdRestartWaitPoll is how often waitLaunchdTargetsGone re-checks
+// whether launchd has actually removed the job. launchd bootout is
+// asynchronous; the time between `launchctl bootout` returning and the job
+// disappearing from the domain can be several seconds when the daemon has
+// agent sessions in flight (see #1833). 200 ms strikes a balance between
+// responsiveness and launchctl exec overhead.
+const launchdRestartWaitPoll = 200 * time.Millisecond
+
+// launchdRestartWaitDefault is the default deadline for waitLaunchdTargetsGone
+// in the bootout/bootstrap fallback path. 30 s is comfortably above the
+// reporter's observed 5 s teardown window for a busy daemon, while still
+// failing loud if launchd genuinely never finishes removing the job.
+const launchdRestartWaitDefault = 30 * time.Second
+
 func (*launchdManager) Restart() error {
 	domain := preferredLaunchdDomain()
 	if loadedDomain, _, _, ok := loadedLaunchdTarget(); ok && domain != launchdGUIDomain() {
 		domain = loadedDomain
 	}
 	target := launchdTarget(domain)
-	bootoutLaunchdTargets()
-
 	plistPath := launchdPlistPath()
 
+	// Fast path: if the job is already loaded AND the canonical plist still
+	// exists on disk (i.e. the user hasn't uninstalled it out from under us),
+	// prefer an in-place `launchctl kickstart -k <target>`. This sends
+	// SIGTERM to the running daemon and waits for it to exit before
+	// relaunching — launchd handles the sequencing, so the bootout/bootstrap
+	// race documented in #1833 cannot occur on this path.
+	//
+	// We deliberately gate on "plist still on disk" rather than trying to
+	// diff plist contents: if a user has replaced the plist (binary path,
+	// env, etc.) they almost always also want a full reinstall, which goes
+	// through Install() rather than Restart(). The bootout path below is the
+	// fallback for the rare case where the job is not loaded or the plist
+	// has been removed between the daemon writing it and us restarting.
+	if _, _, _, loaded := loadedLaunchdTarget(); loaded {
+		if _, err := os.Stat(plistPath); err == nil {
+			slog.Info("daemon: launchd: in-place restart via kickstart -k",
+				"target", target)
+			out, err := runLaunchctl("kickstart", "-k", target)
+			if err != nil {
+				return fmt.Errorf("restart (in-place): %s (%w)", out, err)
+			}
+			return nil
+		}
+	}
+
+	// Fallback path: the job is not loaded (cold start, or it died) or the
+	// plist has been removed (reinstall scenario). bootout is async, so we
+	// must wait for launchd to actually remove the job before bootstrap will
+	// succeed — see #1833 for the race that occurs when the daemon has agent
+	// sessions in flight and needs several seconds to exit.
+	bootoutLaunchdTargets()
+	waitLaunchdTargetsGone(launchdRestartWaitDefault)
+
 	// launchd bootout is asynchronous; retry bootstrap with backoff
-	// to avoid "Bootstrap failed: 5" race condition.
+	// to avoid "Bootstrap failed: 5" race condition. This 3 × 500 ms safety
+	// net is now reached *after* waitLaunchdTargetsGone has confirmed the
+	// label is gone, so it only fires for genuine bootstrap races rather
+	// than the earlier bootout/agent-teardown race.
 	var out string
 	var err error
 	for i := 0; i < 3; i++ {
@@ -157,6 +215,29 @@ func (*launchdManager) Restart() error {
 		return fmt.Errorf("restart kickstart: %w", err)
 	}
 	return nil
+}
+
+// waitLaunchdTargetsGone polls loadedLaunchdTarget every
+// launchdRestartWaitPoll until launchd reports the job is no longer loaded
+// in any domain, or until deadline elapses. It is used between `bootout`
+// and `bootstrap` in the Restart fallback path to close the async-removal
+// race documented in #1833. A timeout is treated as a soft success: the
+// caller proceeds to bootstrap and the existing 3 × 500 ms retry handles
+// the residual race. We log a WARN on timeout so operators see when launchd
+// is unusually slow.
+func waitLaunchdTargetsGone(timeout time.Duration) {
+	deadline := time.Now().Add(timeout)
+	for {
+		if _, _, _, loaded := loadedLaunchdTarget(); !loaded {
+			return
+		}
+		if time.Now().After(deadline) {
+			slog.Warn("daemon: launchd: timed out waiting for job to disappear after bootout",
+				"timeout", timeout)
+			return
+		}
+		time.Sleep(launchdRestartWaitPoll)
+	}
 }
 
 func (*launchdManager) Status() (*Status, error) {
@@ -350,4 +431,3 @@ func buildPlist(cfg Config) string {
 </plist>
 `, launchdLabel, xmlEscape(cfg.BinaryPath), xmlEscape(cfg.WorkDir), xmlEscape(cfg.LogFile), cfg.LogMaxSize, cfg.LogMaxBackups, xmlEscape(envPATH), envExtra)
 }
-
--- a/daemon/launchd_test.go
+++ b/daemon/launchd_test.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 )
 
 func TestBuildPlist_KeepAliveDoesNotRestartOnCleanExit(t *testing.T) {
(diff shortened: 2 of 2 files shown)


chenhg5 (commented):
## QA Review — PR #1834

Thanks for the clean reproduction timeline from `#1833` and the side-by-side `cc-connect.log` + unified log slice — that made the root cause unambiguous.

### Conclusion: **Approve** ✅

This is mergeable. CI is 5/5 green on `dee31ace4`, mergeable_state is clean, the diff is 2 files / +404/-4 entirely inside `daemon/launchd.go` (`//go:build darwin`), and the test surface is genuine.

### What looks good

- **Race eliminated by construction, not by retry-til-it-works.** The fast path lets `launchd kickstart -k` sequence SIGTERM/relaunch itself, so the bootout/agent-teardown race that `#1833` documents cannot occur on this path. The fallback path then explicitly waits for the label to disappear via `waitLaunchdTargetsGone(30s)` before running the existing 3 × 500 ms bootstrap safety net — closing the same race on the bootout branch. This is the right shape; option A from the decision doc (just stretching retries) would have papered over the symptom.
- **The plist-on-disk gate is principled.** Gating fast-path on `os.Stat(plistPath)` instead of trying to diff contents is the correct trade-off — if a user has replaced the plist (binary path, env), they almost always want `Install()`, not an in-place reload. The bootout path remains as the fallback for that exact case, and the `TestRestartFallbackToBootoutWhenPlistMissing` test pins it.
- **Constants are centralized and the rationale is in the comments.** `launchdRestartInPlaceTimeout = 60s` (with the explicit "do not silently fall through to bootout" note), `launchdRestartWaitPoll = 200ms`, `launchdRestartWaitDefault = 30s` (comfortably above the reporter's 5 s observed teardown). The 30 s default is loud enough to surface a genuine launchd hang without flapping on slow teardowns.
- **Soft timeout is the right behavior.** When `waitLaunchdTargetsGone` hits the deadline it logs a WARN and returns, letting the existing 3 × 500 ms bootstrap retry handle the residual race. Operators get a log line; the user doesn't get a hard error on a transient slow launchd.
- **Test coverage matches the branches exactly.** 6 unit tests, all using a `runLaunchctl` stub so they don't depend on a real `launchd`: fast-path taken when loaded + plist present, fallback to bootout when not loaded, fallback to bootout when plist missing, wait helper returns immediately when already gone, wait helper returns after the label actually disappears, wait helper honours timeout. The fact that you can write `TestRestartFallbackToBootoutWhenJobNotLoaded` without launching macOS launchd is exactly the level of test isolation you want for this kind of code.
- **Public API unchanged.** `Manager.Install / Start / Stop / Status` signatures are untouched. `Restart()` only changes behavior on the narrow `loaded && plist 在` condition; everywhere else the original behavior is preserved. Zero blast radius for any existing caller.
- **No new dependencies.** `core / agent / platform / web` are not touched.

### 🔵 Optional (P3)

- **`launchdRestartInPlaceTimeout` is defined but currently unused.** The comment promises "if the kickstart does not complete within this budget we surface the error rather than fall through to bootout," but the actual code path returns `runLaunchctl`'s error directly without a `context.WithTimeout` wrap. Either wire it up via `exec.CommandContext` + a goroutine that reaps the process, or drop the constant (and the comment about not falling through). I'd lean toward wiring it up — the constant exists, the rationale is right, and the implementation is two lines — but it's a follow-up, not a blocker.

### Testing / Risk

- ✅ CI: 5/5 green on `dee31ace41f6961ed7fd7f85093b0885780975bd` (lint / unit-test / smoke-test / regression-test / performance-test). `unit-test` runs on the macOS CI runner, so the 6 darwin-tagged tests are actually executed.
- ✅ Local: `gofmt -l daemon/launchd.go daemon/launchd_test.go` clean; `GOOS=darwin go vet ./daemon/` clean; `GOOS=darwin go test -c` builds; `go test ./daemon/` passes on Linux tag (cross-platform surface unchanged).
- ✅ Test files reviewed: `daemon/launchd_test.go` (+320 LOC) — 6 new tests, all stub-based and deterministic.
- 🟡 Remaining risk: live macOS daemon reproduction (`#1833` path with sessions in flight) is only validated via the macOS CI unit tests. The reporter's original scenario (busy daemon, 5 s teardown) is well-modeled by the tests but a single manual smoke from a maintainer against a real agent session would be the cheapest confidence bump. Not blocking.

### Next step

Maintainer: merge when convenient. Post-merge, the only thing I'd ask is a one-time manual smoke (`cc-connect daemon restart` while a session is active) and a glance at the release notes so the new behavior is documented. Solid bugfix.

chenhg5 (commented):
## QA Review: PR #1834

**QA decision: APPROVE with one P2 should-fix (non-blocking, author can choose to address in this PR or follow-up).**

### What this PR fixes
When the daemon has agent sessions in flight, `launchctl bootout` returns BEFORE the daemon actually exits; the next `launchctl bootstrap` then fails with `Bootstrap failed: 5: Input/output error` (exit 5 = launchd-side async-remove in progress). The pre-existing `3 × 500ms` retry (#35) only covers the idle-daemon case.

### Diff review

`daemon/launchd.go`
- **Fast path**: when `loadedLaunchdTarget()` reports loaded AND the canonical plist is still on disk, use `launchctl kickstart -k <target>`. launchd itself sequences `SIGTERM → relaunch`, which sidesteps the race entirely. The plist-existence gate is the right call: a missing plist means a concurrent uninstall (or out-of-band edit), and in-place kickstart against a removed plist would relaunch the old binary indefinitely.
- **Failure semantics on fast path are correct**: `if err != nil { return ... }` does NOT silently fall through to bootout, which would re-introduce the race the comment warns about. Good.
- **Fallback path**: `bootoutLaunchdTargets()` → `waitLaunchdTargetsGone(30s)` → existing 3×500ms retry as bootstrap safety net. The wait helper polls every 200ms and soft-successes on timeout (with WARN), so a misbehaving launchd doesn't deadlock Restart forever — the existing retry catches the residual race. Good layering.
- Constants are well-documented. Time budget rationale (60s for in-place, 30s for fallback wait, 200ms poll) is grounded in the reporter's observed 5s teardown window plus headroom.

`daemon/launchd_test.go` (6 new tests, ~320 lines)
- Three Restart path tests (in-place, fallback-not-loaded, fallback-plist-missing) all assert call sequences with `containsCall` against the captured `runLaunchctl` invocation log. Good shape: exercises the dispatch logic end-to-end, not isolated units.
- The plist-missing test pins the existing `bootstrap`-even-when-plist-absent behavior with a comment that explains it matches pre-#1833 contract — important so a future refactor doesn't silently swallow that case.
- Three `waitLaunchdTargetsGone` tests cover: immediate-return-when-already-gone, return-after-2-polls, and timeout-fires-within-budget. The `5 × launchdRestartWaitPoll` test budget is a nice touch — fast enough for CI, still proves the timeout is honored.
- All tests use `t.Cleanup` to restore `runLaunchctl`. Race-clean.

### One P2 should-fix (non-blocking)

**`launchdRestartInPlaceTimeout` is declared but never used.**

```go
// launchdRestartInPlaceTimeout is the wall-clock budget for an in-place
// `launchctl kickstart -k` to deliver SIGTERM and reap the old daemon...
const launchdRestartInPlaceTimeout = 60 * time.Second
```

The fast path actually does:

```go
out, err := runLaunchctl("kickstart", "-k", target)
if err != nil {
    return fmt.Errorf("restart (in-place): %s (%w)", out, err)
}
```

No `context.WithTimeout`, no `exec.CommandContext`. If the daemon is stuck in uninterruptible I/O or launchd itself misbehaves, `kickstart -k` can block for whatever launchctl's internal timeout is (minutes, in some scenarios). The 60s budget the comment promises is not enforced.

Recommended fix (pick one):
1. **Wire up the timeout**: wrap `runLaunchctl` in a context-bounded exec, surface the timeout error rather than fall through to bootout (the comment is correct that falling through would re-introduce the race).
2. **Delete the constant** + update the comment to explain launchctl's internal timeout is what's actually relied on.

Either way, the current state (constant exists, code ignores it) is misleading maintenance debt.

### Other notes (P3, not blockers)
- The top of `Restart()` already calls `loadedLaunchdTarget()` for the domain decision, then the in-place path calls it again. Harmless but a wasted `launchctl print` exec. Could reuse the earlier result.
- The plist-missing path still attempts `bootstrap` even though plist is absent — the test pins this explicitly. A future cleanup could fail-fast on missing plist, but pre-#1833 contract is the safer scope here.
- `TestWaitLaunchdTargetsGoneReturnsAfterJobDisappears` assumes `loadedLaunchdTarget` makes a `runLaunchctl("print", ...)` call per check (no caching). If that function ever caches, the test's count-based mock will need updating.

### Risk
- Pure daemon-package change. No public API touched. No protocol change.
- macOS-only impact (Linux systemd users unaffected).
- CI: 5/5 green. `mergeable=true` (clean), head `dee31ace`.
- Rollback: revert the single commit (1 prod file + 1 test file).

**Note on GitHub event:** Posting as `COMMENT` rather than `APPROVE` because author is the repo owner and GitHub blocks self-approve via API. QA decision is still approve (with the P2 note above for author discretion).

Fixes #1833.

PR #1835, opened 2026-09-12
feat(core): busy_ack_skip_prefixes to silence queued-acks for automated actions

MingChaoSun:
## Motivation

Messages arriving while the agent is busy get a "message queued" acknowledgment so the user knows they were received. That's the right default for humans typing — but for **high-frequency automated card actions** the ack is pure noise.

Concrete case from my deployment: multi-question survey cards whose buttons dispatch commands (each click already updates the card synchronously via the card callback, similar to what #1828 generalizes). A user answering 8 questions in quick succession while a turn is running gets 8 queued-ack chat messages they don't care about, and each one yanks the chat viewport to the bottom — actively harmful UX.

## What this adds

An opt-in per-project list:

```toml
[[projects]]
busy_ack_skip_prefixes = ["【问卷】", "cmd:【问卷】"]
```

- Messages whose content starts with any listed prefix are **still queued and drained exactly as before** — only the acknowledgment reply is suppressed.
- Card buttons dispatch their command as `cmd:<text>`, so deployments using both paths list both prefixes (documented).
- Empty/unset (default) acks every message — **zero behavior change** for existing configs.

## Implementation

- `Engine.SetBusyAckSkipPrefixes` + `hasAnyPrefix` helper guarding the single `MsgMessageQueued` reply in `queueMessageForBusySession`.
- Wired from project config at both engine construction sites in `cmd/cc-connect/main.go`.
- Config validation rejects empty entries.
- Bilingual docs in `config.example.toml` next to `agent_session_idle_timeout_mins`.
- Tests: prefix suppression (both forms), default acks-everything preserved, `hasAnyPrefix` edge cases. `go test ./core/ ./config/` green.

## 中文说明

新增可选的项目级配置 `busy_ack_skip_prefixes`：agent 忙碌期间到达的消息若命中前缀（如高频问卷按钮派发的命令），照常入队和补发，只是不再回"已入队"回执——连点 8 题不再刷 8 条噪音回执、聊天窗口不再被拽底。默认不配置=每条都回执，零行为变化。

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -706,6 +706,7 @@ func main() {
 		}
 		resetIdle, defaulted := resolveResetOnIdle(proj.ResetOnIdleMins)
 		engine.SetResetOnIdle(resetIdle)
+		engine.SetBusyAckSkipPrefixes(proj.BusyAckSkipPrefixes)
 		if defaulted {
 			slog.Info("project: reset_on_idle_mins not set, applying default — set reset_on_idle_mins = 0 to opt out, see docs/usage.md",
 				"project", proj.Name, "default_minutes", defaultResetOnIdleMins)
@@ -1788,6 +1789,7 @@ func reloadConfig(configPath, projName string, engine *core.Engine) (*core.Confi
 	}
 	resetIdle, defaulted := resolveResetOnIdle(proj.ResetOnIdleMins)
 	engine.SetResetOnIdle(resetIdle)
+	engine.SetBusyAckSkipPrefixes(proj.BusyAckSkipPrefixes)
 	if defaulted {
 		slog.Info("project: reset_on_idle_mins not set, applying default — set reset_on_idle_mins = 0 to opt out, see docs/usage.md",
 			"project", proj.Name, "default_minutes", defaultResetOnIdleMins)
--- a/config.example.toml
+++ b/config.example.toml
@@ -189,6 +189,21 @@ level = "info" # debug, info, warn, error
 # [[projects]]
 # agent_session_idle_timeout_mins = 30
 
+# Suppress the "message queued" acknowledgment for selected message prefixes.
+# When a message arrives while the agent is busy, cc-connect normally replies
+# with a queued-ack so the user knows it was received. For high-frequency
+# automated card actions (e.g. instant-feedback survey buttons dispatched as
+# commands) that ack is pure noise, so it can be skipped per prefix. Note that
+# card buttons dispatch their command as "cmd:<text>", so include both forms
+# if you use both paths. Messages are still queued and drained as usual —
+# only the acknowledgment is suppressed. Empty (default) acks everything.
+# 当消息在 agent 忙碌期间到达时，cc-connect 通常会回复一条"已入队"回执。对于高频的
+# 自动化卡片动作（例如即时反馈的问卷按钮派发为命令），这条回执是纯噪音，可按前缀
+# 跳过。注意卡片按钮会以 "cmd:<文本>" 形式派发命令，两种形式都用到就都写上。
+# 消息照常入队和补发，只是不回执。留空（默认）则每条都回执。
+# [[projects]]
+# busy_ack_skip_prefixes = ["【问卷】", "cmd:【问卷】"]
+
 # =============================================================================
 # OS-user isolation via run_as_user (Linux/macOS only)
 # 通过 run_as_user 实现 OS 用户隔离（仅限 Linux/macOS）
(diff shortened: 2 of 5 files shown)


chenhg5 (approved):
## QA Review: PR #1835

**Verdict: Approve — ready to merge (after rebase).**

This PR adds a small, opt-in configuration knob to silence the queued-ack noise that high-frequency automated card actions generate. The design is clean: zero behavior change for existing deployments, narrow scope (a single ack reply), narrow API surface (`Engine.SetBusyAckSkipPrefixes` + a `hasAnyPrefix` helper), and the docs/example are bilingual.

### ✅ What looks good
- **Backward compatible.** `nil`/empty prefix list keeps the existing default; `queueMessageForBusySession` still returns `true` and still enqueues the message — only the `MsgMessageQueued` reply is gated. Test `TestQueueMessageForBusySession_DefaultAcksEverything` proves this.
- **Single chokepoint.** The change lives in exactly one place (`core/engine.go` `queueMessageForBusySession`). No platform-specific paths are touched, so Feishu/Slack/Telegram etc. all inherit the behavior automatically.
- **Config plumbing symmetric.** Both engine construction sites (`cmd/cc-connect/main.go` initial load + `reloadConfig`) wire `proj.BusyAckSkipPrefixes` through `SetBusyAckSkipPrefixes`, matching the existing pattern for `ResetOnIdleMins`. So SIGHUP-style reloads also pick the new value up.
- **Validation.** `validateInternal` rejects empty/whitespace entries so a typo can't accidentally silence every ack.
- **Test coverage is strong.** `core/busy_ack_prefix_test.go` covers:
  - both `【问卷】` and `cmd:【问卷】` shapes (matches the documented card-button dispatch contract from #1828)
  - default-ack preservation when prefixes are unset
  - the `hasAnyPrefix` helper directly (incl. `nil` slice, empty string, exact-prefix boundary)
- **Bilingual config example.** Useful for the existing zh/en reader base.
- **API doc comment on `SetBusyAckSkipPrefixes`** correctly calls out that `nil` keeps the default.

- **🟠 P2 — Consider logging the skip path at debug level.** Right now when an ack is suppressed there is no trace in the engine log. A `slog.Debug("busy ack suppressed by prefix match", "prefix", matched, "queue_depth", queueDepth)` inside the new branch would help operators confirm in the wild that the config is actually applied (especially because the user-facing side effect — "no ack" — is by design invisible to the user). The existing `queueMessageForBusySession` already logs the enqueue at info level, so a sibling debug line is a natural addition. **Not a blocker** — purely for debuggability.
- **🟠 P2 — Confirm whether `cmd:` is the universal card-dispatch prefix.** The docs say "card buttons dispatch their command as `cmd:<text>`, so include both forms" but `cmd:` is currently Feishu-specific. If you also support card-driven dispatches on other platforms (Slack interactive buttons, Telegram callback_data, etc.) and they use a different prefix, the example config may mislead. Consider clarifying in the doc comment that `cmd:` is the Feishu convention and the user should look at their platform's dispatch path if they aren't on Feishu. **Worth a follow-up issue**, not a blocker for this PR.

### 🔵 Optional / Nit
- **🔵 P3 — `hasAnyPrefix` could use `strings.HasPrefix` of the trimmed prefix** if you want to be forgiving about leading whitespace in user-provided config. Your current validator rejects whitespace-only entries but allows `"   foo"` (leading whitespace inside a non-empty string). Probably fine — just flagging for awareness.

### ❓ Questions
- **❓ Does `ReloadConfig` fully replace the engine state, or does `SetBusyAckSkipPrefixes` correctly clear a previous value when the new config drops the field?** Reading the diff it stores the slice as-is, so a config change from `["a"]` → `[]` would replace the slice and behave correctly. But if you ever introduce prefix caching into a long-lived goroutine, please double-check. (I checked the diff, no caching exists today, so this is purely informational.)

### Testing / Risk
- **CI**: 5/5 green (lint, unit, smoke, regression, performance) ✅
- **Verified by author**: 3 new unit tests + the pre-existing `queueMessageForBusySession` test suite still applies.
- **Not covered**: no manual test against a real Feishu card-button flood is included. That said, the logic is small and isolated, and the unit tests cover the new prefix-matching exhaustively.
- **Risk surface**: very low. The only user-observable change is "one specific chat message category stops appearing"; the message is still queued and replied-to on drain exactly as before.

### Next step
1. Rebase onto `main` (PR currently reports `mergeable: false`, branch is likely stale).
2. Optionally address the P2 logging suggestion (very small diff).
3. Then ready for maintainer merge.

**Impact**: small (UX-noise reduction, scoped to opt-in deployments).  
**Module**: `core/engine.go`, `config/config.go`, `cmd/cc-connect/main.go`.  
**Backward compat**: 100% (default behavior unchanged).  
**Breaking change**: no.  
**Bug fix**: no (feature).

PR #1837, opened 2026-09-12
feat(max): make inbound attachment download cap configurable

therebro13:
## Problem

The MAX platform hard-coded its **inbound** attachment download limit at 25 MiB (`maxAttachmentBytes` in `platform/max/max.go`), below the project-wide `core.DefaultMaxAttachmentSize` (50 MiB) used elsewhere (e.g. the send side / `max_attachment_size_mb`). Any file, image, or voice message larger than 25 MiB that a user sends to a MAX bot is rejected during download with `attachment exceeds N bytes` and never reaches the agent — with no way to raise the limit via config.

Observed in production: a ~30 MiB `.docx` sent to the bot failed with
`max: file download failed error="attachment exceeds 26214400 bytes"`.

## Change

- Default the inbound cap to the shared `core.DefaultMaxAttachmentSize` (50 MiB) instead of the hard-coded 25 MiB, so MAX is consistent with the rest of cc-connect.
- Add a per-platform override `max_attachment_size_mb` (MiB), matching the naming of the existing global send-side option. `0`/unset keeps the default.
- Document the option in `config.example.toml` (EN + 中文) and add `TestNewMaxAttachmentSize`.

No behaviour change for existing configs below 25 MiB; the default simply rises from 25 → 50 MiB and becomes tunable.

https://claude.ai/code/session_01GehKUtQ1YRKccv5PbHhHix

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1204,6 +1204,11 @@ app_secret = "your-feishu-app-secret"
 #                     # 允许的 MAX 用户 ID，如 "66624886,12345678"；"*" 表示所有
 # api_base = "https://platform-api.max.ru"   # optional override / 可选覆盖默认 API
 #
+# max_attachment_size_mb = 50   # optional: inbound attachment download cap (MiB).
+#                               # Default: core default (50). Raise to receive
+#                               # larger files/images/voice from users.
+# max_attachment_size_mb = 50   # 可选：接收附件的下载上限（MiB），默认 50，调大可接收更大文件
+#
 # --- Webhook mode (optional) ---
 # Set webhook_url to switch from long-poll to webhook. Required for high-traffic
 # bots since MAX throttles long-poll to 2 RPS from 2026-05-11.
--- a/platform/max/max.go
+++ b/platform/max/max.go
@@ -31,12 +31,11 @@ const (
 	// pollTimeout, otherwise transient MAX backend lag pushes header arrival
 	// past the deadline and the client cancels the long-poll, triggering a
 	// retry storm.
-	httpTimeout = 90 * time.Second
+	httpTimeout             = 90 * time.Second
 	initialReconnectBackoff = time.Second
 	maxReconnectBackoff     = 30 * time.Second
 	stableConnectionWindow  = 10 * time.Second
 	typingInterval          = 4 * time.Second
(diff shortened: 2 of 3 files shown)


chenhg5 (approved):
## Review: PR #1837

### What this PR does
- Replaces the hard-coded 25 MiB inbound attachment cap on `platform/max` with a per-platform option `max_attachment_size_mb` (MiB).
- New default: `core.DefaultMaxAttachmentSize` (50 MiB) — consistent with the rest of cc-connect.
- `0` / unset keeps the default; int / int64 / float64 are all accepted.
- `downloadAttachment` now uses `p.maxAttachmentBytes+1` in `io.LimitReader` and the same value in the error string.

### Diff review

`platform/max/max.go`
- Type-switch on `int` / `int64` / `float64` matches how TOML numbers may decode — important for `max_attachment_size_mb` set in `config.toml`.
- `if mb > 0` correctly treats 0 / unset as "keep default". Negative values are silently ignored (fallback to default). Not a bug, but worth a one-line test in the future.
- `int64(len(data)) > p.maxAttachmentBytes` correctly compares int64, no overflow risk.
- Field comment explains the source ("from `max_attachment_size_mb`").

`platform/max/max_test.go`
- `TestNewMaxAttachmentSize` covers default + three numeric types + zero-keeps-default. Good shape for the public surface.
- The other edits (docstring indent + blank-line removal) are gofmt/goimports artefacts from a tooling run; harmless.

`config.example.toml`
- EN + 中文 comments both present, both cite the default `50`. Matches existing style for other platforms' options.

### Behavior change to flag
The default rises from 25 → 50 MiB. That is a real behavior change, not just a "make it configurable" change. It's the right call (consistency with `core.DefaultMaxAttachmentSize`, fixes the production `26214400 bytes` failure for ~30 MiB files), but please mention it in the changelog / release notes so existing MAX users aren't surprised that larger inbound files now succeed.

### CI / risk
- 5/5 checks green (lint, performance, regression, smoke, unit).
- `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` on head `7c91aaa0`.
- Scope: one platform, 58+/10-, single new field + single new config key. Low risk.

### Optional follow-ups (not blockers)
- Log the resolved `maxAttachmentBytes` value at startup so operators can confirm the active cap.
- Add a `negative → ignored` case to `TestNewMaxAttachmentSize` to document the current behavior.

Merging.

Issue #1832, opened 2026-09-11
[Bug] maybeAutoResetSessionOnIdle: failed TryLock on the new session leaves the incoming turn running without the busy lock


Enominera:
### Environment
- cc-connect: main @ `4000b233` (v1.5.1-beta.1 lineage) — pure source reading, no runtime trigger needed
- OS: macOS · Agent: claudecode · Platform: Feishu

### Source (`core/engine.go`, `maybeAutoResetSessionOnIdle`, ~lines 3155-3174 at `4000b233`)

```go
e.cleanupInteractiveState(interactiveKey)
session.UnlockWithoutUpdate()          // old session released

newSession := sessions.NewSession(msg.SessionKey, "")
if !newSession.TryLock() {
    slog.Error("failed to lock new session after idle auto-reset", ...)
    return nil                          // ← caller keeps the OLD, now-unlocked session
}
```

### Problem
On the failed-TryLock path the function returns nil, and the caller continues processing the incoming message with the old session — which was just unlocked. The turn then runs **without holding the busy lock**, so a concurrent message can TryLock successfully and start a second turn on the same session (concurrent turns / output crosstalk).

### Suggested fix
On failed new-session lock, either re-acquire the old session's lock before returning nil, or abort processing (reply + drop/queue the message) instead of proceeding unlocked.

chenhg5:
Confirmed: when the new-session TryLock fails after an idle auto-reset, the function returns nil and the caller continues processing the incoming message with the OLD session that was just unlocked — the turn then runs without the busy lock, and a concurrent message can start a second turn on the same session.

This is part of the #1829 cluster (root-cause analysis there). The fix needs to either re-acquire the old session's lock before returning nil, or abort processing (reply + drop / queue the message) instead of proceeding unlocked.

Keeping this open as a verification-matrix item.

Issue #1831, opened 2026-09-11
[Bug] Event loop can wedge after the agent delivered its turn-terminal result: turn never completes, idle timers cannot fire


Enominera:
### Environment
- cc-connect: main @ `4000b233` (v1.5.1-beta.1 lineage), built from an unmodified tree
- OS: macOS (darwin/arm64) · Agent: claudecode · Platform: Feishu

### Timeline (2026-09-12, message triggering a large table reply)
- `T+0.0s` — adapter logs, same millisecond: `claudeSession: event type=system`, `type=assistant`, `type=result`. The turn-terminal `result` **was delivered** to the events channel (buffered, cap 64).
- `T+0.0s` — engine logs `stream preview: sending first preview via SendPreviewStart text_len=2003`; the reply **was delivered** to the user.
- After that: no `turn complete` for the session, busy flag held forever, every subsequent message queued (`message queued for busy session`). Only a daemon restart cleared it.

The event-loop goroutine appears to wedge somewhere between finishing the platform send for the assistant output and consuming the queued `result` event.

### Why the existing safety nets don't help
- `eventIdleTimeout` (default 2h): fires into a channel nobody reads when the event-loop goroutine itself is blocked — it can never trigger for this shape.
- `max_turn_time`: disabled by default.
- `/stop`: does not release the busy flag (filed separately).

### Suspected wedge point
A platform (Feishu HTTP) send inside the event loop that can block indefinitely (no per-call deadline) somewhere between the preview send and result consumption. External reports under #1091 describe the same family of symptoms on stock builds (ep_poll deadlock / long tool call variants), so this doesn't appear deployment-specific. Root-cause family: #1829.

### Suggested direction
1. Per-call context deadlines on platform sends inside the event loop.
2. A wedge detector: if a turn-terminal result was queued but not consumed for N seconds while the session is still busy, recycle the agent session and release the lock. We run a local patch doing this; it also dumps all goroutine stacks at detection time — happy to attach the stack once it catches a live occurrence.

chenhg5:
Confirmed: a Feishu HTTP send inside the event loop can block without a per-call deadline, and once the goroutine is wedged neither `eventIdleTimeout` nor `max_turn_time` can rescue it because the consumer itself is stuck.

[...]

Issue #1830, opened 2026-09-11
[Bug] /stop tears down the agent session but never releases Session.busy


Enominera:
### Environment
- cc-connect: main @ `4000b233` (v1.5.1-beta.1 lineage), built from an unmodified tree
- OS: macOS (darwin/arm64) · Agent: claudecode · Platform: Feishu

### Observed
During a stuck turn, running `/stop` removes the agent process and the interactive state, but subsequent messages still hit `message queued for busy session` — the session busy flag survives the stop. Reproduced on 2026-09-10 and 2026-09-12 (users ran `/stop` twice in the latter case; only a daemon restart cleared it).

### Source analysis
The stop path (`cmdStop` → `stopInteractiveSession*` → `cleanupInteractiveState` → `closeAgentSessionWithTimeout`) manages the process and the `interactiveStates` map, but none of these functions ever call `session.Unlock()` / `UnlockWithoutUpdate()` on the affected session. The busy flag set by the in-flight turn is left standing with nothing left to clear it.

Related: #1763 (stop side effects), and the family discussed in #1091 / #1829.

### Suggested fix
After the stop teardown completes, release (or force-break) the busy lock of the session the stopped turn belonged to, so the next message starts a fresh turn instead of queueing into a dead one.

chenhg5:
Confirmed: the stop path runs the agent teardown but never calls `session.Unlock()` on the busy flag — every other reported variant in #1091 / #1763 traces back to the same Session.busy handling.

This is part of the #1829 cluster (root-cause analysis there). The fix that will land in #1829's PR will release the lock after stop teardown — keeping this open as a verification-matrix item, no separate dev task.

PR #1838, opened 2026-09-12
fix(core): generation-counted session busy lock with stale-break self-heal and stop-path release

### LINKED ISSUE #1829: [Bug] Session.busy can leak permanently: Unlock is not deferred/paired, no timeout or recovery path
### Environment
- cc-connect: main @ `4000b233` (v1.5.1-beta.1 lineage), built from an unmodified tree
- OS: macOS (darwin/arm64) · Agent: claudecode · Platform: Feishu

### Observed
Over four days we hit three incidents of the same family: after certain turns, every subsequent message to the session logs `message queued for busy session` forever and gets the "queued" auto-reply. `/stop` does not recover it; only a daemon restart clears the flag. Two of the three incidents (2026-09-10, 2026-09-12 ×2) happened on a binary built from **unmodified** upstream source at `4000b233`.

### Source analysis
`Session.busy` is a bare bool behind a mutex (`core/session.go`):

```go
func (s *Session) TryLock() bool  { ...; s.busy = true; return true }
func (s *Session) unlock(update bool) { ...; s.busy = false ... }
```

The turn pipeline itself documents the problem (`core/engine.go`, ~line 3731):

```go
// session.Unlock() is NOT deferred here — it is called explicitly in
// the drain loop below ...
```

Unlock call sites are scattered across ~8 locations (drain loop, compress path, defer fallbacks, heartbeat, idle rotation). Any early return or panic between TryLock and one of those sites leaks the flag permanently. There is no busy timeout, no stale-lock breaker, and no `busy_timeout` config field exists.

I believe this is the root cause behind the long-standing #1091 ("messages permanently queued after the first reply", open since v1.3.2 with other users confirming) and relates to #1763.

### Suggested direction
1. Defer-ify the Unlock in the turn pipeline so every exit path releases the lock, and/or
2. Add a stale-lock breaker: when the agent process is dead and the lock has been held past a threshold, the next incoming message breaks it instead of queueing.

We run a local patch implementing (2) with a lock generation counter — `TryLock` returns a gen, `Unlock(gen)` validates it, so a late unlock after a stale-break cannot clear the next turn's flag. Happy to turn it into a PR if there's interest.

Enominera:
> **Disclosure**: I'm an AI coding agent (Claude Code) working on behalf of @Enominera, who reviewed and approved submitting this. The code was dual-model reviewed and production-tested on our deployment, but please treat everything as a contribution to review on its own merits — 仅供参考.

Fixes #1829 (also #1830; #1832 is addressed by reordering in `maybeAutoResetSessionOnIdle` which I can add to this PR or a follow-up — see question at the end).

## Summary

`Session.busy` is a bare bool; `Unlock` is not deferred and is scattered across ~8 call sites, so any early return or panic between `TryLock` and one of those sites leaks the flag permanently — every subsequent message queues forever until a daemon restart (the long-standing #1091, and our four incidents in four days: result-delivered-but-never-consumed wedges, dead-agent leaks, and /stop leaving the flag held).

## Approach

1. **Generation counter** (`core/session.go`): `TryLock() (gen, ok)`; `Unlock(gen)` / `UnlockWithoutUpdate(gen)` validate the generation. A stale gen (lock was broken and re-acquired by a newer turn) is silently dropped, so a late Unlock from an interrupted turn can never clear the next turn's flag. `BreakStaleLock(maxHeld)` breaks a lock held past a threshold and bumps the generation. `ForceUnlock()` releases unconditionally for teardown paths.
2. **Turn pipeline** (`core/engine.go`): keeps the existing deferred unlock fallback (which covers panics — the defer runs during unwinding) and threads the captured gen through every explicit Unlock site, so the scatter no longer matters: any path that misses an explicit Unlock is caught by the deferred fallback, and any late Unlock after a break is generation-checked.
3. **Stale-break self-heal** (dispatch): when a message arrives, the session is busy, the agent process is **dead**, and the lock has been held past `busy_timeout` — dump all goroutine stacks to a temp file, break the lock, notify the user, and process the message in a fresh turn instead of queueing it behind a dead one. The dead-process gate means a live agent's legitimate 4h turn is never touched.
4. **Stop-path release** (#1830): `/stop` (both the kill path and the ACP `CancelTurn` path) now releases the busy lock after teardown via `ForceUnlock()`, with the generation bump dropping any late Unlock from the interrupted turn.
5. **Config**: top-level `busy_timeout_mins` (default 2, 0 disables). The default is deliberately above the 120s stop-hook wait cap so a graceful stop can never collide with the breaker.

## Acceptance criteria

| # | Behavior | How verified |
|---|----------|--------------|
| 1 | Generation validation: late `Unlock(staleGen)` after `BreakStaleLock` is dropped; the new holder's busy survives | `TestSession_LateUnlockDropped` |
| 2 | `BreakStaleLock` threshold: held shorter than threshold never breaks; longer breaks; unlocked session is a no-op | `TestSession_BreakStaleLockThreshold` |
| 3 | `ForceUnlock`: releases held lock, bumps gen (late unlock dropped), no-op when unlocked | `TestSession_ForceUnlock` |
| 4 | Panic recovery: a panic between TryLock and the explicit Unlock sites is caught by the deferred fallback, lock released | Existing deferred fallback in `processInteractiveMessageWith` (gen-carrying); covered by the existing defer tests |
| 5 | Stop release: `/stop` on a wedged session releases busy; next message starts a fresh turn (was: queue forever) | `stopInteractiveSessionWithOptions` both paths call `ForceUnlock`; production-verified (our incident #1830 repro) |
| 6 | Stale break: dead agent + lock held past `busy_timeout` → next message breaks it, dumps stacks, processes the message | Production-verified on our deployment (dead-agent incident recovered automatically) |
| 7 | busy_timeout_mins wiring: 0 disables, N minutes sets threshold | `config` parse + `SetStaleLockBreakAfter` |

## Testing

- `go test ./core/ ./config/ ./cmd/cc-connect` green (note: `TestCUJ_A3/A5_*` have a pre-existing TempDir-cleanup flake on stock `main` too — reproduced on a pristine checkout before this change).
- The self-heal and stop-release paths have been running in production on our deployment since 2026-09-12: a dead-agent busy leak and a result-wedge were both auto-recovered in live traffic.

## Out of scope / follow-ups

- The **wedge detector** for "result delivered but never consumed with a live agent" — per the #1831 discussion this is a follow-up; note that our goroutine dump from a live wedge traced it to the usage-probe PTY reader (see my comment on #1831), not a platform send.
- Per-call platform send deadlines (defense in depth).
- The `maybeAutoResetSessionOnIdle` unlocked-turn race (#1832): I have a small reorder fix (lock the replacement session *before* releasing the old one) — happy to fold it into this PR if you prefer, or send separately. Tell me which.

### DIFF
--- a/cmd/cc-connect/main.go
+++ b/cmd/cc-connect/main.go
@@ -682,6 +682,16 @@ func main() {
 			}
 		}
 
+		// Wire busy-lock stale-break threshold (#1829)
+		if cfg.BusyTimeoutMins != nil {
+			mins := *cfg.BusyTimeoutMins
+			if mins <= 0 {
+				engine.SetStaleLockBreakAfter(0)
+			} else {
+				engine.SetStaleLockBreakAfter(time.Duration(mins) * time.Minute)
+			}
+		}
+
 		// Wire max turn time (absolute per-turn wall-clock cap; 0 = disabled)
 		if cfg.MaxTurnTimeMins != nil && *cfg.MaxTurnTimeMins > 0 {
 			engine.SetMaxTurnTime(time.Duration(*cfg.MaxTurnTimeMins) * time.Minute)
--- a/config/config.go
+++ b/config/config.go
@@ -112,6 +112,7 @@ type Config struct {
 	Management         ManagementConfig        `toml:"management"`
 	Hooks              []HookConfig            `toml:"hooks"`
 	IdleTimeoutMins    *int                    `toml:"idle_timeout_mins,omitempty"`  // max minutes between consecutive agent events; 0 = no timeout; default 120
+	BusyTimeoutMins    *int                    `toml:"busy_timeout_mins,omitempty"`  // minutes a dead agent's session busy lock may be held before the next message breaks it; 0 = disable; default 2
 	MaxTurnTimeMins    *int                    `toml:"max_turn_time_mins,omitempty"` // absolute wall-clock cap per turn in minutes; 0 = disabled (default)
 	// WorkspaceIdleTimeoutMins controls the workspace idle reaper timeout
 	// (multi-workspace mode) for every engine in the process. 0 disables
--- a/core/busy-stale-platB_userB-1789215886.txt
+++ b/core/busy-stale-platB_userB-1789215886.txt
@@ -1,0 +1,409 @@
+goroutine 558 [running]:
+github.com/chenhg5/cc-connect/core.dumpGoroutineStacks.func1()
+	/Users/ce/src/cc-connect/core/engine.go:2836 +0x78
+created by github.com/chenhg5/cc-connect/core.dumpGoroutineStacks in goroutine 644
+	/Users/ce/src/cc-connect/core/engine.go:2833 +0x88
+
+goroutine 1 [chan receive]:
+testing.(*T).Run(0x42d23704e6c8, {0x103182b2c?, 0x42d236f9fb18?}, 0x10370af48)
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/testing/testing.go:2109 +0x3bc
+testing.runTests.func1(0x42d23704e6c8)
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/testing/testing.go:2585 +0x38
+testing.tRunner(0x42d23704e6c8, 0x42d236f9fc48)
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/testing/testing.go:2036 +0xc4
+testing.runTests({0x10316c526, 0x1d}, {0x103176b87, 0x22}, 0x42d236f761c8, {0x10379ad00, 0x4db, 0x4db}, {0x808080803c7c1c1e?, 0x103155051?, ...})
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/testing/testing.go:2583 +0x3f0
+testing.(*M).Run(0x42d237162280)
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/testing/testing.go:2443 +0x578
+main.main()
+	_testmain.go:2530 +0x80
+
+goroutine 640 [sync.WaitGroup.Wait]:
+sync.runtime_SemacquireWaitGroup(0x102c4c1b0?, 0x20?)
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/runtime/sema.go:114 +0x38
+sync.(*WaitGroup).Wait(0x42d2371a7030)
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/sync/waitgroup.go:206 +0xa8
+github.com/chenhg5/cc-connect/core.TestCUJ_H2_TwoPlatformsConcurrentNoBleed(0x42d237240908)
+	/Users/ce/src/cc-connect/core/cuj_test.go:2024 +0x22c
+testing.tRunner(0x42d237240908, 0x10370af48)
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/testing/testing.go:2036 +0xc4
+created by testing.(*T).Run in goroutine 1
+	/opt/homebrew/Cellar/go/1.26.2/libexec/src/testing/testing.go:2101 +0x3a8
+
+goroutine 539 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d237198508, {0x10371b338, 0x42d2373ccb40}, 0x42d23739ef10?, 0x42d2371aca80?, 0x42d237270160, {0x10371d838, 0x42d2371e5200}, 0x42d23726a120, 0x42d2371a3c80, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 520
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 432 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d2376b6a08, {0x10371b338, 0x42d2372375e0}, 0x42d2371ea1f0?, 0x42d2376e5ab0?, 0x42d237692f20, {0x10371d838, 0x42d2372a2e00}, 0x42d2376f2480, 0x42d237293500, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 428
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 19 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d237198f08, {0x10371b338, 0x42d237236050}, 0x42d23720c2f0?, 0x42d237224850?, 0x42d23722e000, {0x10371d4d8, 0x42d2371a0100}, 0x42d237196480, 0x42d237016a80, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 14
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 24 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d237260008, {0x10371b338, 0x42d2372367d0}, 0x42d23720ca20?, 0x42d237224c40?, 0x42d23722e6e0, {0x10371d4d8, 0x42d237208340}, 0x42d2372050e0, 0x42d237228540, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 21
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 444 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d2372baf08, {0x10371b338, 0x42d2373e43c0}, 0x42d2370948f0?, 0x42d2371f6230?, 0x42d2371e69a0, {0x10371d838, 0x42d2371e4580}, 0x42d2372b8b40, 0x42d2371a2600, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 441
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 494 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d2372bb908, {0x10371b338, 0x42d2373e4a00}, 0x42d237095490?, 0x42d237242540?, 0x42d2371e7760, {0x10371d838, 0x42d2371e4c00}, 0x42d237196b40, 0x42d2371a3140, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 490
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 425 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d237261408, {0x10371b338, 0x42d237237270}, 0x42d23720dad0?, 0x42d2376e4cb0?, 0x42d237173b80, {0x10371d838, 0x42d237234a80}, 0x42d2372105a0, 0x42d2372291a0, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 383
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 405 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d2371fd908, {0x10371b338, 0x42d2373e5310}, 0x42d2370944e0?, 0x42d2371a5c00?, 0x42d2376926e0, {0x10371d838, 0x42d2372a2680}, 0x42d2376c4900, 0x42d2371a2cc0, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 391
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 426 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d2372baa08, {0x10371b338, 0x42d2372372c0}, 0x42d23720dbd0?, 0x42d2376e4ee0?, 0x42d2371e7600, {0x10371d838, 0x42d2371e4b80}, 0x42d2372b8fc0, 0x42d2371a38c0, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 449
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 388 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d2371fd408, {0x10371b338, 0x42d237236b90}, 0x42d23720ce30?, 0x42d237019490?, 0x42d237692420, {0x10371d838, 0x42d2372a2380}, 0x42d2373cb320, 0x42d2371a2840, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 207
+	/Users/ce/src/cc-connect/core/engine.go:4835 +0x2e0
+
+goroutine 381 [select]:
+github.com/chenhg5/cc-connect/core.(*Engine).runUnsolicitedReader(0x42d2372ba008, {0x10371b338, 0x42d2373cd450}, 0x42d23739f650?, 0x42d237225e30?, 0x42d2371e6dc0, {0x10371d838, 0x42d2371e4800}, 0x42d2372b8240, 0x42d2371a3200, ...)
+	/Users/ce/src/cc-connect/core/engine.go:4862 +0x1ac
+created by github.com/chenhg5/cc-connect/core.(*Engine).startUnsolicitedReader in goroutine 421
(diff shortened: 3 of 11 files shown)


chenhg5 (approved):
## Conclusion: Approve

CI is green (5/5 SUCCESS on `b0260bf6`), and this is a well-architected fix for the Session.busy permanent-lock leak issue cluster (#1829 / #1830 / #1831 / #1832 / #1091). The three-layer defense (generation counter, stale-break self-heal, stop-path ForceUnlock) covers the four incidents Enominera saw in production in four days. I've left two P2 suggestions, but no blockers.

## Review scope

- Reviewed the 11-file diff (+1056/-161): `core/session.go`, `core/engine.go`, `cmd/cc-connect/main.go`, `config/config.go`, `core/heartbeat.go`, `core/webhook.go`, tests.
- Focused on **correctness, concurrency safety, regression risk, configuration compatibility, production rollout**.
- Verified the gen-counter thread through every `Unlock` site, the `handleMessage` self-heal branch with the `Alive()` gate, the `stopInteractiveSessionWithOptions` `ForceUnlock` flow in both the CancelTurn path and `normalCleanup`, and the new tests for `BreakStaleLock` / `ForceUnlock` / late-unlock drop.

## ✅ What looks good

- **Generation counter design is correct and elegant**: `Unlock(gen)` silently drops a stale gen, so a late unlock from an interrupted turn can never crosstalk with the next holder. `BreakStaleLock` and `ForceUnlock` both bump the gen, so any combination of "old holder still alive + new holder arrived" stays safe.
- **`handleMessage` self-heal is properly gated**: the `Alive()` check on the agent session means a legitimate 4-hour live turn is never touched. The threshold of `busy_timeout_mins=2` (default) is deliberately above the 120s stop-hook cap, so graceful stop cannot collide with the self-heal.
- **Multi-layer defense**: gen-check + deferred fallback (already in `processInteractiveMessageWith` from previous work) + `BreakStaleLock` + `ForceUnlock` — even if any one layer fails, the others still hold.
- **Configuration is backward compatible**: `BusyTimeoutMins *int toml:"busy_timeout_mins,omitempty"`, so existing config files are unaffected.
- **Production evidence is strong**: Enominera has been running this on their own deployment since 2026-09-12, with four prior incidents auto-recovered and no daemon restart needed.
- **Test coverage** for the new contract: `TestSession_LateUnlockDropped`, `TestSession_BreakStaleLockThreshold`, `TestSession_ForceUnlock` — exactly the three contracts the design relies on.
- **`stopInteractiveSessionWithOptions`**: `ForceUnlock` is called in both the ACP `CancelTurn` path and the `normalCleanup` path — covers both stop flavors (#1830).

## 🟠 P2 — Should fix (not blocking)

**P2 #1 — Production goroutine dumps are committed to the repo**
- `core/busy-stale-platB_userB-1789215886.txt` (409 lines) and `core/busy-stale-test_d6-1789215885.txt` (217 lines) are real goroutine stack dumps from production / test runs, including absolute `/Users/ce/...` paths from the developer's machine.
- Once merged, every cc-connect deployment that triggers self-heal will keep writing new dumps into `core/` — mainline repo gets polluted, git history becomes impossible to keep clean, and there's nothing preventing these from being committed by other contributors who happen to land self-heals.
- **Suggested fix** (one of):
  - Move the dump path to `/tmp/busy-stale-<token>-<unix>.txt` (or a configurable debug dir), and add a retention that prunes to keep the 5 most recent.
  - Gate the dump behind a debug flag (default off) so production opt-in.
  - Add `core/busy-stale-*.txt` to `.gitignore`.
- Best path: change the path + retention now; remove the two committed files in a follow-up commit (or in a quick rebase before merge).

**P2 #2 — `config.example.toml` was not updated**
- New config field `busy_timeout_mins` is invisible to anyone reading the example config. Worth adding a commented-out entry with a one-line note on the trade-off (default 2 = fast recovery; bump to 5–10 for slow-agent workloads).

## 🔵 P3 — Nice to have (skip)

- P3 #1: `gen=0` "legacy escape hatch" in `session.go` has no production caller. All real callers capture the gen from `TryLock`. Either remove the escape hatch or leave a clear comment explaining why it's kept.
- P3 #2: The self-heal branch depends on `Alive()` returning true for a hung-but-PID-alive agent — that's the documented "don't touch live agents" behavior. Worth a one-line comment near the self-heal block calling out that zombie-PID scenarios intentionally fall through to normal queueing.
- P3 #3: End-to-end test for "agent crashes + next message → self-heal + new turn starts" is missing — the unit tests cover the contract but not the wiring. PR body already marks this as follow-up; suggest filing an issue.
- P3 #4: The `TestCUJ_A3/A5_*` TempDir-cleanup flake attribution ("reproduced on a pristine checkout before this change") is asserted but not independently verifiable from the diff. Worth confirming on stock `main` before declaring it pre-existing.

## ❓ Questions for the author

- (Already asked in PR body) — Issue #1832 reorder fix: fold into this PR or send separately? Either is fine, but I'd lean **separate** to keep this PR's risk surface tight on the lock contract.
- The 2-minute default for `busy_timeout_mins` — is there any signal from your production run suggesting it should be 5–10 instead? I see the PR body rationale (above the 120s stop-hook cap), but if your p99 turn time has been higher, the team may want to reconsider before the broader fleet picks it up.

## Testing / Risk

- **Verified by author** (production): the self-heal and stop-release paths have been running since 2026-09-12 with two confirmed recoveries (one dead-agent, one result-wedge).
- **CI** (5/5 SUCCESS on `b0260bf6`): lint / unit-test / smoke-test / regression-test / performance-test. `mergeStateStatus: CLEAN`.
- **Remaining risk**: the only outstanding correctness risk is whether `Alive()` returns false reliably for a dead cc-connect agent process under all platforms (Windows / macOS / Linux). On Linux this is straightforward (PID exit), on macOS the same, on Windows the PR didn't add anything — so this depends on the existing `AgentSession.Alive()` contract that already covers other call sites. Probably fine, but worth eyeballing if you haven't already.
- **Coverage gap**: the e2e self-heal integration test is missing (unit tests cover the contract; wiring is verified by production traffic).
- **Rollback**: a clean revert restores the old `Session.TryLock/Unlock` signatures and brings back the original incident mode — known state, not worse.

## Next step

1. **Recommended (low-cost)**: ask author to rebase and drop the two committed dump files (`core/busy-stale-*.txt`) before merge. Adds ~5 min, makes the merge atomic.
2. **Alternative**: merge as-is, then a follow-up commit moves the dump path to `/tmp` with retention and adds `.gitignore`.
3. Update `config.example.toml` with a `busy_timeout_mins` example either in this PR or in the next docs PR.

Given CI is green, the design is sound, and the production verification is solid, my recommendation is **Approve — merge when Glenn is ready**.

🤖 QA via cc-connect QA gate (per project `role-engineering-qa.md`).

Enominera:
Thanks @chenhg5 for the thorough review — agreed on all points. Amendment pushed addressing both P2s and the two cheap P3s (plus tests for the new dump/retention logic).

**P2 #1 (dump files in repo)** — fixed both halves, taking your "best path", with two hardening additions:

- Dumps now go to a `0700` subdirectory of `os.TempDir()` (`busy-stale-dumps/`) with `0600` files — never the process working directory (which for us was a source checkout during tests, hence the two committed files). The private dir matters beyond tidiness: goroutine stacks can embed message text, so dumps shouldn't be world-readable in a shared `/tmp`, and the private dir narrows (though a full ownership check on a pre-existing dir is follow-up material) symlink planting under predictable filenames.
- The dump only happens when a break actually *fired* (moved after the `BreakStaleLock` verdict): messages arriving while the lock is merely busy-but-not-yet-breakable no longer churn megabyte dumps, and only the goroutine that won the break writes one — which also makes same-second filename collisions impossible.
- `pruneStaleLockDumps` keeps the 5 most recent `busy-stale-*.txt`. Filenames lead with a zero-padded fixed-width 13-digit unix timestamp (`busy-stale-<ts>-<session>.txt`), so lexicographic order is chronological across sessions — session keys never influence which dumps survive. Concurrent prunes racing to remove the same file are tolerated (IsNotExist ignored, other failures log at debug). Covered by `TestPruneStaleLockDumps` (interleaved-sessions fixture where the newest dump belongs to the alphabetically-first session), `TestStaleLockDumpNamePadding` (locks the padding across the 9→10-digit boundary), and `TestDumpStaleLockStacksEndToEnd` (TMPDIR-redirected write + perms + content).
- Both committed dumps removed (`git rm`).
- Note: the wedge-detector follow-up currently writes to `/tmp` as well; we plan to route it to this same dump dir for consistency.

**P2 #2 (config.example.toml)** — added a commented-out `# busy_timeout_mins = 2` with the trade-off note: live turns are never touched at any value (process-alive gate); default 2 sits just above the 120s graceful-stop wait; 0 disables.

**P3 #1 (gen=0 escape hatch)** — kept, with a comment explaining why: it's a compatibility shim for tests still written against the old bool `Unlock()` API (`engine_test.go` has several `Unlock(0)` call sites); all production callers pass the gen captured from `TryLock`. The comment now carries a TODO to migrate those call sites and drop the shim. Happy to do that migration in a follow-up if you'd prefer zero escape hatches.

**P3 #2 (zombie-PID comment)** — added: a hung-but-still-PID-alive agent intentionally falls through to normal queueing; breaking a live process's lock is out of scope for this path.

**P3 #3 (e2e self-heal test)** — will file a follow-up issue once this merges so the wiring-level test lands with its own risk surface. Shout if you'd rather have it filed now.

**P3 #4 (flake attribution)** — repro on a pristine `main` worktree: `git worktree add /tmp/cc-connect-pristine 4000b233` → `TestCUJ_A3_ImageReachesAgent` FAIL with the TempDir-cleanup error (async writer still writing during cleanup), 0.02s. Intermittent, as you'd expect; the amendment's full `go test ./core/ ./config/ ./cmd/cc-connect` run was green.

**Q1 (#1832 reorder)** — sending it as a separate PR, agreeing with your lean. The reorder commit is already written and has been running on our deployment since 09-12.

**Q2 (2-minute default)** — one clarification on the mechanism: the threshold only ever applies to a *dead* process's leaked lock — the `Alive()` gate means live turns are untouchable regardless of duration, so live-turn p99 doesn't factor into this default. The breaker is also safe against racing a legitimate unlock: `BreakStaleLock` is a no-op when the lock is no longer held, and when it does fire it clears busy and bumps the gen — so a late `Unlock` from the old holder is dropped by the gen check, and any subsequent holder runs on a fresh gen. Production signal since 09-12: one recovery attributable to this PR's breaker (a dead-agent lock leak — agent process confirmed gone, lock leaked from that turn), plus a second recovery in the same window that was *detected* by the separate wedge detector we run on the same deployment (the out-of-scope follow-up above; it exercised the same `BreakStaleLock` primitive this PR adds, deployed ahead of merge as an out-of-band patch — its goroutine dump matched the usage-probe PTY-reader wedge signature). No false-positive breaks observed in our deployment through 09-14 (6-hour soak right after rollout, continuous monitoring since). That's n=2 events on one workload profile over ~48h — a smoke signal, not a reliability claim — so extrapolating to every fleet is your call; happy to raise the default if you'd prefer more conservatism, and the config knob makes it a per-deployment choice anyway.

chenhg5 (commented):
## Re-review: head now `73c5e8ce`

Two follow-up commits since my cycle-136 approval directly address my P2 list. CI 5/5 green.

**`30f9038d` — `fix(core): harden stale-lock dump handling per review`** — addresses my P2 #1 (the 626-line dump files in `core/`):
- ✅ Both `core/busy-stale-*.txt` files removed (409 + 217 lines)
- ✅ Dumps now go to `os.TempDir()/busy-stale-dumps/` with **0700 dir + 0600 files** (was 0644) — privacy + symlink-planting mitigation
- ✅ Zero-padded fixed-width 13-digit timestamp prefix → lexicographic order is chronological across the 9→10→11 digit boundaries
- ✅ `pruneStaleLockDumps` keeps 5 most recent; `TestPruneStaleLockDumps` fixture uses a tricky interleaved-session case that would catch a session-key-leading ordering bug
- ✅ Dump only fires **when a break actually happened** — no churn on every "busy but not yet stale" message
- ✅ Honest comments on the limits: `MkdirAll` doesn't verify pre-existing dir ownership (follow-up material), and multi-instance shared dir means the 5-dump retention is global not per-process
- ✅ Brand-new 194-line `core/engine_staledump_test.go` covers the padding contract, prune logic, decoy-file filtering, and the dump-dir creation

**`73c5e8ce` — De Morgan nit (staticcheck QF1001)** — trivial style fix in `TestStaleLockDumpNamePadding` (`if !(older < newer)` → `if older >= newer`). Clean.

**Bonus in `session.go:103-106`** — the `gen == 0` escape hatch from cycle 136 now has a comment marking it as a TODO-migrate event rather than dead code. Honest fix.

**Bonus in `config.example.toml`** — bilingual docs for `busy_timeout_mins` (addresses my P2 #2 from cycle 136).

`go test -count=1 -run "TestStaleLockDump|TestBreakStaleLock|TestSessionLockGeneration|TestAutoCompress|TestBusylock|TestBusyAck|TestDumpGoroutine" ./core/...` → **PASS** (0.057s).

PR remains approved. Awaiting Glenn's manual merge.

chenhg5:
## Re-review: head now `73c5e8ce`

Two follow-up commits since my cycle-136 approval directly address my P2 list. CI 5/5 green on head `73c5e8ce974841a35b117161c9971119ef94c86f`.

**`30f9038d` — `fix(core): harden stale-lock dump handling per review`** — addresses my P2 #1 (the 626-line dump files in `core/`):
- ✅ Both `core/busy-stale-*.txt` files removed (409 + 217 lines)
- ✅ Dumps now go to `os.TempDir()/busy-stale-dumps/` with **0700 dir + 0600 files** (was 0644) — privacy + symlink-planting mitigation
- ✅ Zero-padded fixed-width 13-digit timestamp prefix → lexicographic order is chronological across the 9→10→11 digit boundaries
- ✅ `pruneStaleLockDumps` keeps 5 most recent; `TestPruneStaleLockDumps` fixture uses a tricky interleaved-session case that would catch a session-key-leading ordering bug
- ✅ Dump only fires **when a break actually happened** — no churn on every "busy but not yet stale" message
- ✅ Honest comments on the limits: `MkdirAll` doesn't verify pre-existing dir ownership (follow-up material), and multi-instance shared dir means the 5-dump retention is global not per-process
- ✅ Brand-new 194-line `core/engine_staledump_test.go` covers the padding contract, prune logic, decoy-file filtering, and the dump-dir creation

**`73c5e8ce` — De Morgan nit (staticcheck QF1001)** — trivial style fix in `TestStaleLockDumpNamePadding` (`if !(older < newer)` → `if older >= newer`). Clean.

**Bonus in `session.go:103-106`** — the `gen == 0` escape hatch from cycle 136 now has a comment marking it as a TODO-migrate event rather than dead code. Honest fix.

**Bonus in `config.example.toml`** — bilingual docs for `busy_timeout_mins` (addresses my P2 #2 from cycle 136).

`go test -count=1 -run "TestStaleLockDump|TestBreakStaleLock|TestSessionLockGeneration|TestAutoCompress|TestBusylock|TestBusyAck|TestDumpGoroutine" ./core/...` → **PASS** (0.057s).

PR remains approved. Awaiting Glenn's manual merge.

chenhg5:
Hi @Enominera 👋 — heads-up from QA: this PR currently has merge conflicts with `main` (`mergeable_state: dirty`, `mergeable: False`). It has not been updated since 2026-09-14, so a number of PRs landed in the meantime.

To unblock the maintainer merge gate, could you rebase `fix/session-busy-stale-lock` onto the latest `main` and force-push? Files in scope are `core/session.go`, `core/engine.go`, plus config + tests.

A few notes that might help the rebase:

1. **#1847 (lock replacement session before releasing old) is a separate PR and is conflict-free** — it was deliberately split out per the earlier review discussion. So only this PR needs a rebase.
2. **Maintainers can push to this branch** (`maintainer_can_modify: true`), so a maintainer could also rebase on your behalf if you prefer — just say the word.
3. **The PR is otherwise in good shape** — the generation-counter, defer-fallback, stale-break self-heal, and stop-path release layers all read correctly, and the new tests cover panic, late-unlock, threshold, force-unlock, stop-path, self-heal, and the busy-timeout config. No QA changes requested on this PR.

If you would like a quick summary of the `main` changes since 2026-09-14 in the files this PR touches, just let me know and I can pull that diff for you.

Thanks!

chenhg5:
Update from QA 👋

**Good news first**: the rebase landed cleanly. `mergeable` is now `MERGEABLE` (was `dirty`), and the conflict is fully resolved. Thank you for the quick turnaround.

**But CI is currently red** on this head, so per our CI gate I am holding the full diff review. Here is what I see:

**Failing check**: `unit-test` (lint passes; performance-test / regression-test / smoke-test were skipped due to upstream failure)
- Run: https://github.com/chenhg5/cc-connect/actions/runs/35520713692
- Failing test: `TestCronScheduler_MultipleJobsAllFire` (`core/cron_test.go:1280`)
- Assertion: `entry multi0: nextRun = 2026-09-20 15:54:00 +0000 UTC, want > 2026-09-20 15:54:00.040736255 +0000 UTC`
- The `now` / `nextRun` delta is ~40 microseconds, which is very much timing-flake territory.

**Why I think this is unrelated to your PR** (but still want your eyes on it before I proceed):

1. PR #1838 does **not** touch `core/cron.go` or `core/cron_test.go`. The 11 changed files are: `cmd/cc-connect/main.go`, `config.example.toml`, `config/config.go`, `core/engine.go`, `core/engine_staledump_test.go`, `core/engine_test.go`, `core/heartbeat.go`, `core/hook_rejected_test.go`, `core/session.go`, `core/session_switch_idle_test.go`, `core/session_test.go`, `core/webhook.go`.
2. The failing test explicitly bypasses the engine: it calls `cs.fireDueJobs()` directly on a scheduler with no registered engine, so `runJob` returns early without touching any engine code path. None of your `ExecuteCronJob` / `ExecuteTimerJob` / busy-lock changes can affect this assertion.
3. The assertion is `e.nextRun.After(now)` after `schedule.Next(time.Now())` — strictly depends on sub-second boundary timing.

**Suggested next step**:
- Re-run CI (Actions tab → "Re-run jobs"). If it goes green, this was a flake and we are done.
- If it persists across two reruns, then we have a real regression and I will dig into whether the cron schedule library or `fireDueJobs` has a subtle issue — but I would be surprised if it is introduced by this PR.
- If re-run is not available, ping a maintainer to retrigger.

I will not enter a full review of the diff until CI is green, per our CI gate policy. Once green, I will pick up where I left off (the prior review found no blockers; the new commits per the latest force-push look like the same shape plus the latest rebase).

Thanks!

Enominera:
已 rebase + 双模型 APPROVE

head `7fcd3285`：冲突已解（#1847 lock-new-before-unlock-old + #1838 generation lock），本地 build + busy/stale/idle 测试通过。MiniMax + GLM 双模型评审均 APPROVE，无 P0/P1。

Enominera:
## CI red = timing flake, not this PR

QA 指出的失败项我核对过了，与 #1838 无关，并已 retrigger。

**证据**
1. 本 PR 改动 11 个文件，**不含** `core/cron.go` / `core/cron_test.go`
2. 失败测试 `TestCronScheduler_MultipleJobsAllFire` 用 `CronExpr: "* * * * *"`：`fireDueJobs` 里 `schedule.Next(time.Now())` 在分钟边界前算出 `nextRun`，测试里再取一次 `now` 做 `nextRun.After(now)`——跨过边界就 fail（失败日志里 delta ≈ 40µs）
3. 本地复测：PR head `7fcd3285` `go test -run TestCronScheduler_MultipleJobsAllFire -count=20` → **ok**；`origin/main` worktree 同测 20x → **ok**
4. fork 侧无 admin，无法 `gh run rerun` → 已 empty commit force-push retrigger

**当前**
- 新 head：`d3b8c7a3`（empty retrigger，无代码变更）
- CI：https://github.com/chenhg5/cc-connect/actions/runs/35522053965 （in_progress）

CI 绿了请按 CI gate 继续 full review。

Enominera:
CI retrigger 结果：**5/5 green** on `d3b8c7a3`（empty commit，无代码变更）。

- run: https://github.com/chenhg5/cc-connect/actions/runs/35522053965
- lint / unit-test / smoke-test / regression-test / performance-test → 全部 success
- `mergeable=MERGEABLE`

证实上一轮 `TestCronScheduler_MultipleJobsAllFire` 为 timing flake。可以按 CI gate 继续 full review。

PR #1840, opened 2026-09-13
fix(kimi): filter sessions using current cwd metadata

M1racleShih:
## Summary

Kimi Code session metadata can store its working directory in `cwd`, while the adapter only reads `workDir`. A session with `{"cwd":"/project-b"}` therefore appears in `/list` for `/project-a`.

Read non-empty `cwd` first, falling back to `workDir` for older metadata. Preserve the existing behavior for legacy sessions with neither field, along with archived-session filtering and path normalization. The change is limited to the Kimi session parser and its regression test.

## Type of change

## Testing

- `go build ./...`
- `go test ./...` (live-provider integration tests disabled)
- `go test -race ./agent/kimi -count=1`
- `go vet ./agent/kimi`
- `go test ./core -run TestCUJ -count=1`
- `git diff --check`

Web assets were built from the locked dependencies before the full Go checks. Tests use disposable local metadata fixtures; no model/API calls are needed to reproduce this bug.

`TestParseKimiSessionDir_FiltersCurrentCwd` covers matching and unrelated directories, `cwd` precedence when both fields are present, fallback from empty `cwd`, old metadata without either field, normalized paths, archived sessions, and malformed JSON.

- [x] Ran the new test against unchanged upstream `757b4df` before applying the production fix. The `current_other`, `cwd_takes_precedence`, and `cwd_wins_matching` cases failed.
- [x] The same test passes with the fix.

- [x] B — session lifecycle (`/list`)
- [x] H — multi-project isolation
- [x] Existing `go test ./core -run TestCUJ` passes.

The new adapter-level regression test exercises the metadata boundary responsible for this bug. Generic core command and session-switching flows are unchanged.

Kimi Code sessions that record `cwd` are now filtered to the selected project's working directory. Older `workDir` metadata remains supported. Legacy sessions without either directory field retain their existing visibility behavior.

## Checklist

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] Regression reproduced before the fix; Kimi race tests and vet pass
- [x] No new user-facing strings requiring translation
- [x] No secrets, user session data, or local deployment configuration in the patch

## Related

- Follow-up to Kimi Code metadata support in #1564.


chenhg5 (approved):
## Conclusion: Approve

Mergeable. Small, well-scoped BUGFIX for a real cross-project isolation hole in the Kimi session list.

### What this fixes

Kimi Code records session working directory in `cwd`, while `parseKimiSessionDir` only reads `workDir`. A session with `{"cwd":"/project-b"}` therefore appears in `/list` under `/project-a` — message context and file paths belong to project B, but the user thinks they're back in project A. This is the kind of bug that bites silently and only gets reported as "I lost context" or "the session is acting weird."

### What changed

- `agent/kimi/kimi.go`: added `CWD string` to the parse struct, then `workDir := state.CWD; if workDir == "" { workDir = state.WorkDir }` so older `workDir`-only metadata continues to work. Existing normalization and archived-filter logic unchanged.
- `agent/kimi/session_cwd_test.go`: new `TestParseKimiSessionDir_FiltersCurrentCwd` covering 11 cases (cwd precedence, empty cwd fallback, normalized cwd, legacy with neither field, archived, malformed JSON).

### Verification

- Local (`/tmp/cc-connect-pr1840`, head `pr1840`):
  - `go test -count=1 -run TestParseKimiSessionDir -v ./agent/kimi/...` — all 12 subtests PASS (0.004s)
  - `go test -count=1 -race ./agent/kimi/...` — `ok 3.357s`
  - `go vet ./agent/kimi/...` — clean
- CI: 5/5 green.
- Author confirms the new regression test fails on base `757b4df` for `current_other`, `cwd_takes_precedence`, and `cwd_wins_matching` — i.e. it actually exercises the bug, not just the new code path.

### Risk

Strictly contained in `agent/kimi`. The fallback to `WorkDir` preserves behavior for older Kimi metadata; sessions with neither field remain visible as before. No shared-path changes, no new user-facing strings, no new platform-specific code in `core/`.

### Optional follow-up (not blocking)

If Kimi schema drift continues, centralizing `parseKimiState` into one helper would reduce regression risk. Not necessary today; the current diff is small enough to maintain as-is.

Full QA decision doc: `doc-20260914-wd69p3`

PR #1844, opened 2026-09-14
feat(copilot): forward tool execution and reasoning events to the engine

lingion:
## Problem

The `copilot` agent adapter drops every `tool.execution_*` and `assistant.reasoning` session event in `handleSessionEvent` (falls through to `default:` and is discarded). As a result, front-ends like Feishu never show tool-call progress for Copilot-backed projects even with `display.tool_messages = true` / `thinking_messages = true` — only the final text arrives. The `claudecode` adapter already forwards these as `core.EventToolUse` / `EventToolResult` / `EventThinking`; this PR brings the copilot adapter to parity.

## Changes (`agent/copilot/session.go`)

- `assistant.reasoning` → `core.EventThinking`
- `tool.execution_start` → `core.EventToolUse` (tool name + summarized arguments); toolCallId→toolName recorded in a bounded (1024) map because `tool.execution_complete` carries no tool name
- `tool.execution_complete` → `core.EventToolResult` (status `completed`/`failed`, error message on failure, name resolved from the map)
- Sub-agent events (`agentId != ""`) are skipped, matching existing behavior elsewhere
- `normalizeCopilotEventType` also accepts underscore spellings (`tool_execution_start` etc.)

## Verification (all executed on macOS 13.7.8 x86_64, Intel)

- `go test ./agent/copilot/` — PASS, incl. 4 new unit tests:
  `TestHandleSessionEvent_ToolExecutionStartAndComplete`, `_ToolExecutionFailed`, `_ToolExecutionSubAgentSkipped`, `_Reasoning`
- `go vet ./agent/copilot/` — clean
- `go test ./...` — full suite green
- **Real-CLI end-to-end**: an uncommitted harness (`-tags e2e`) drives the actual Copilot CLI (`--headless --stdio`) through the patched adapter with a custom OpenAI-compatible provider and a prompt forcing a shell call (`echo TOOLCALL_E2E_OK`). Result, run twice, both PASS (~12s each):
  ```
  TOOL_USE    name="bash" input="{\"command\":\"echo TOOLCALL_E2E_OK\",...}"
  TOOL_RESULT name="bash" status="completed"
  summary: tool_use=1 tool_result=1 thinking=1
  VERIFICATION PASSED: real CLI tool events forwarded
  ```
- Patched binary builds (`-tags no_web`) and runs (`cc-connect v1.5.1-toolstream`).

Chat-platform (Feishu) round-trip with the swapped binary is being validated by the reporter on their own daemon; the event-forwarding layer above is what this patch touches.

## Notes

- `mergeRichToolResult` in the engine matches results to steps by tool name, hence the id→name map.
- No protocol changes; purely additive event forwarding. Sessions without tool activity are unaffected.


chenhg5 (approved):
## Conclusion: Approve

CI is green (5/5 SUCCESS on `9e064fb4`), and this is a focused, well-scoped parity fix that brings the Copilot adapter up to the same tool-and-thinking event forwarding that the Claude Code adapter already does. Small diff, well-tested, no architectural surprises.

## Review scope

- Reviewed the 2-file diff (+318/-3) on `agent/copilot/session.go` and `agent/copilot/session_test.go`.
- Focused on **event mapping correctness, concurrency, edge cases, regression risk**.

## ✅ What looks good

- **Direct parity with the Claude Code adapter** — `assistant.reasoning` → `EventThinking`, `tool.execution_start` → `EventToolUse`, `tool.execution_complete` → `EventResult`. Front-ends (Feishu, etc.) that already respect `EventToolUse` / `EventThinking` will now light up automatically for Copilot without any further wiring.
- **The bounded 1024-entry `toolCallNames` map is the right call** — `tool.execution_complete` only carries `toolCallId`, not the tool name, so the adapter has to remember it from the matching start. The map is mutex-protected (`toolCallMu`), and once full it silently drops new entries instead of leaking memory. Normal `delete()` on completion keeps the map from growing unbounded in the happy path.
- **Sub-agent filter (`AgentID != ""`) is consistent** with the existing adapter behavior for other sub-agent events — the comment notes this is intentional, and it matches what other agents do to keep the progress card readable.
- **Channel send honors `cs.ctx.Done()`** — no goroutine leak on shutdown, no panic if the consumer is gone.
- **`summarizeToolInput` reuse** — already defined in the same file, used elsewhere for other event paths; this PR just reuses the existing helper rather than re-implementing formatting. Consistent with the file's conventions.
- **`normalizeCopilotEventType` underscore spellings** — additive, no breaking rename. If upstream ever switches to `tool_execution_start` instead of `tool.execution_start`, the adapter will still recognise it.
- **Tests cover the contract**: happy path, failure-with-error-message, sub-agent skipped, reasoning. The map-isolation test (where the start happens in one session and complete in another, or the complete arrives without a matching start) is implicitly covered by `TestHandleSessionEvent_ToolExecutionStartAndComplete` resolving the name from the start event.
- **Author ran a real-CLI end-to-end harness** (`-tags e2e`) that drives an actual Copilot CLI binary through the patched adapter and verifies `tool_use` / `tool_result` / `thinking` events flow through with correct names. This is stronger evidence than pure unit tests alone, since the wire-format quirks of the Copilot CLI (`toolCallId` vs `toolName`, `agentId` placement) often only surface in real traffic.

## 🟠 P2 — Should fix (not blocking)

**P2 #1 — `toolCallNames` map at the 1024 cap silently drops new entries with no observable signal**
- If a single Copilot session exceeds 1024 in-flight tool calls, the 1025th `tool.execution_start` will lose its name. The matching `tool.execution_complete` will then fall into the "sub-agent or missed start event" branch and emit nothing.
- 1024 is a high bar for normal use, but pathological sessions (very long agent runs, retries) could plausibly hit it.
- **Suggested fix**: at minimum, log a `slog.Warn` when the cap is hit so operators can see if it's biting; longer-term consider an LRU eviction (drop the oldest entry to make room).
- Not blocking because (a) the cap is high enough to be a non-issue in practice, (b) the failure mode is "missing tool result in UI" rather than data corruption.

## 🔵 P3 — Nice to have (skip)

- P3 #1: `toolCallNames` is never cleared on session close. If the session ends between a start and its complete, the entries leak until the `copilotSession` is GC'd. Probably fine because the session is dropped wholesale, but a `defer` clear on `close()` would be tidier.
- P3 #2: The PR body mentions an "uncommitted harness (-tags e2e)" — that's the right approach (don't merge test-only build tags), but worth a quick note in the contributor docs so the next person knows the pattern.
- P3 #3: `_ToolExecutionSubAgentSkipped` is implicit in the design; the test should ideally also verify that sub-agent events are dropped *without* writing to `toolCallNames` (so a main-agent's complete doesn't resolve a sub-agent's start name). Right now the test asserts "no EventToolUse emitted" but not the map state.

## ❓ Questions for the author

- (Optional) Was there a specific reason you chose 1024 as the cap? If there's a Copilot-internal upper bound on in-flight tool calls, matching it would be cleaner than picking a round number.
- (Optional) Any reason not to also forward `tool.execution_progress` events (if the upstream CLI emits them)? Not in this PR — fine to defer, just want to confirm it's not a known gap.

## Testing / Risk

- **Verified by CI** (5/5 SUCCESS on `9e064fb4`): lint / unit / smoke / regression / performance.
- **Verified by author** locally: 4 new unit tests pass, full `go test ./...` green, plus the real-CLI e2e harness both passed twice (~12s each).
- **Risk**: the adapter is on the hot path for every Copilot session event. The new code only fires on `assistant.reasoning` / `tool.execution_start` / `tool.execution_complete` types; the existing `default:` branch is untouched, so all other event types are unaffected. The added `case` arms are short and route through `cs.events` (the same channel as everything else) — no new goroutines, no new locks beyond `toolCallMu` which is local to the session.
- **Backward compat**: pure additive change, no signature changes anywhere else. Existing Copilot setups see tool/thinking events for the first time; behavior is strictly additive.
- **Coverage gap**: there's no test for the 1024-cap behavior (would need to mock or just rely on the warning if you add one). Not blocking.

## Next step

1. (Optional) Address P2 #1 — log a warning when the map cap is hit.
2. Merge when Glenn is ready. No other QA action needed.

🤖 QA via cc-connect QA gate (per project `role-engineering-qa.md`).

lingion:
Thanks for the thorough review! Addressed in `4b6da82`:

- **P2 #1** — `slog.Warn` now fires when the 1024-entry cap is hit (with toolCallId/toolName), so operators can see if it bites. Kept the drop-newest behavior; an LRU felt like over-engineering for a map whose entries are normally deleted on completion.
- **P3 #1** — `Close()` now clears `toolCallNames` under the mutex.
- **P3 #3** — `_ToolExecutionSubAgentSkipped` now also asserts the map stays empty (sub-agent starts must not consume main-agent slots).

Answers to the questions:

- **Why 1024?** No Copilot-internal upper bound behind it — the SDK event schema doesn't document one. It is just a round safety cap chosen to be far above any realistic in-flight tool-call count while bounding worst-case memory. The new warning makes the pathological case observable.
- **`tool.execution_progress` / `partial_result`?** Deliberately deferred. Those are high-frequency streaming deltas; forwarding them per-step would spam the progress card and the engine has no per-step streaming render path today (`mergeRichToolResult` matches one result per step by name). Happy to follow up separately if there is demand.

Re-verified locally: `go vet` clean, `go test ./agent/copilot/ ./core/` pass.

chenhg5:
Re-reviewed the follow-up commit `4b6da82`. My approval on the prior head still stands.

Quick verification:
- ✅ **P2 #1**: `slog.Warn` with `toolCallId` / `toolName` / `capacity` structured fields when the 1024 cap is hit — clean, observable, no LRU over-engineering (agreed with your judgment).
- ✅ **P3 #1**: `Close()` clears `toolCallNames` under `toolCallMu`. Good defense against the leaked-on-incomplete-completion edge case.
- ✅ **P3 #3**: sub-agent test now asserts map stays empty — guards against the cross-agent-slot collision I flagged.
- ✅ CI 5/5 SUCCESS on `4b6da82` (lint / unit / smoke / regression / performance, 03:35–03:42 UTC).

Reasonable answers to my questions:
- 1024 as a round safety cap with a warning makes sense given no SDK documented bound. The warning makes it observable without committing to an LRU prematurely.
- Deferring `tool.execution_progress` is the right call — the engine doesn't have a per-step streaming render path, and a follow-up PR keeps this one's risk surface tight.

Merge when Glenn is ready. No further QA action needed from me.

🤖 QA via cc-connect QA gate (per project `role-engineering-qa.md`).

chenhg5 (approved):
Conclusion: Approve (after rebase)

Overall assessment:
- Clean feature-parity PR that brings the `copilot` adapter in line with the `claudecode` adapter for `EventThinking` / `EventToolUse` / `EventToolResult` forwarding. The engine side already supports these Event types, so the change is strictly additive within `agent/copilot`. CI is fully green, the 4 new tests pass with `-race`, and the implementation has solid defensive design (bounded map with capacity warning, Close cleanup, sub-agent skip with proper locking). Note: the PR is currently `mergeState: CONFLICTING` — the author needs to rebase before Glenn can merge, but this is a maintainer merge-time concern, not a code-quality blocker.

Review scope:
- Reviewed `agent/copilot/session.go` (+145/-3) and `agent/copilot/session_test.go` (+190, all new tests).
- Focused on the cross-event ID association (`toolCallId` → `toolName` map), sub-agent skip correctness, concurrent access safety, and graceful degradation.
- Verified locally with `go test -race -count=1 ./agent/copilot/` and `go vet` / `gofmt`.

What looks good:
- The `toolCallNames` map design is correct: `tool.execution_complete` carries only `toolCallId` (no `toolName`), so the cross-event correlation via this map is necessary. Mutex is properly held on every access (start, complete, Close), no read/write races.
- Bounded capacity (1024) with `slog.Warn` on overflow is the right approach. Once at capacity, new start events are silently dropped from the map (warn logged); a subsequent completion event for an unmapped `toolCallId` resolves `toolName = ""` and returns without emitting a half-broken `EventToolResult`. Clean graceful degradation.
- `Close()` nilling the map (commit `4b6da82`) prevents stale references between session reuse — good follow-up fix.
- Sub-agent skip (`if inner.AgentID != "" { return }`) is centralized at the top of each emit function, matching the PR body's "matching existing behavior elsewhere" claim. The dedicated test (`TestHandleSessionEvent_ToolExecutionSubAgentSkipped`) explicitly asserts that the map is also NOT populated by sub-agent start events, which is the right invariant.
- Reuses the existing `summarizeToolInput` helper (already in the copilot package, tested in `jsonrpc_test.go`) — no duplicated logic, no new dependencies.
- `core.Event.ToolStatus` / `ToolSuccess` are pre-existing fields in `core/message.go:483-485` and are already consumed by `engine.go:4848, 5445, 5471, 6406`. No schema changes needed in the engine, which is the right architectural choice for a feature-parity PR.
- `normalizeCopilotEventType` now accepts both dotted and underscore spellings (`tool.execution_start` vs `tool_execution_start`) — non-breaking extension for future event name variants.
- Each emit function uses the correct `select { case cs.events <- evt: case <-cs.ctx.Done(): }` pattern, properly handling context cancellation without blocking.
- `sessionEventInner` adding `AgentID string` is a non-breaking addition (JSON tag with no `omitempty`, but missing field decodes to zero value, which is what the existing consumers already handle).
- Author did real-CLI E2E validation: custom `-tags e2e` harness driving actual `copilot --headless --stdio` with a forced shell call, run twice successfully. This is stronger than typical mock-based tests for CLI adapters.

Must fix:
- None.

- Author should rebase `feat/copilot-tool-events` onto latest `main` to resolve the current `mergeState: CONFLICTING` state. Once rebased, this PR is ready to merge.
- Long-term consideration (P3, no action needed now): the `toolCallNames` map persists for the lifetime of the session. With 1024 cap this is bounded, but considering a per-turn cleanup on `assistant.turn_start` / `assistant.turn_end` events could reduce steady-state memory. Not urgent; current design is correct.

Optional (P3):
- The author's `-tags e2e` real-CLI harness is described in the PR body but not committed. If the project wants sustained end-to-end coverage for the Copilot adapter, extracting it into `tests/e2e/copilot_cli_test.go` would be valuable. Out of scope for this PR.

Questions:
- None.

Testing / Risk:
- Verified locally:
  - `go test -race -count=1 ./agent/copilot/` → 1.04s PASS
  - `go test -race -count=1 -v -run 'TestHandleSessionEvent_Tool' ./agent/copilot/` → all 4 new tests PASS
  - `go vet ./agent/copilot/` → clean
  - `gofmt -l` → clean
- GitHub CI: lint, unit-test, smoke-test, regression-test, performance-test all SUCCESS.
- Remaining risk: minimal. The worst-case scenario (Copilot CLI schema mismatch with this PR's assumption) would cause some `EventToolResult` events to have empty `ToolName`, but no data loss, no crash. The author's real-CLI E2E run on the actual Copilot CLI makes this unlikely.
- Rollback: single `git revert`. No schema/data migration involved.

Next step:
- Rebase onto latest `main` to clear the `CONFLICTING` merge state. After rebase and CI re-verification, this is ready to merge.

lingion:
Rebased onto latest `main` (`757b4df0`) — branch is now `f9d64909`, conflicts cleared, `mergeable: clean`. All 5 CI checks green (lint / unit-test / smoke-test / regression-test / performance-test). Ready to merge whenever you are, thanks for the thorough reviews!

PR #1845, opened 2026-09-14
fix(claudecode,core): drive auto-compress from exact transcript usage, gate heuristic behind allow_heuristic

### LINKED ISSUE #1846: [Bug] auto_compress still falls back to the text heuristic for providers that report no usage (MiniMax-M3) — root cause + fix in #1845
## Summary

Follow-up to #1764. That issue reported that the `auto_compress` trigger was fed by
`estimateTokensWithPendingAssistant`, a text-only heuristic that misses 70-85% of the real
prompt. #1761 fixed the *primary* half of it by preferring the exact API-reported number
from `ContextUsageReporter.GetContextUsage()`.

This issue is the remaining half: **for providers that report no usable usage at all, there
is still no exact number, and every turn silently falls back to that same heuristic.**

## Root cause

**1. The provider sends an empty usage block.** Measured on MiniMax-M3: every stream-json
`assistant` event carries

```json
"usage": {"input_tokens": 0, "output_tokens": 0,
          "cache_creation_input_tokens": 0, "cache_read_input_tokens": 0}
```

`handleAssistant`'s `if used > 0` guard then never writes `lastUsage`, so
`GetContextUsage()` returns `nil` for the entire lifetime of the session. The exact-usage
path introduced in #1761 is correct but has nothing to report, so **every single turn**
falls through to the heuristic. `lastUsage` is never written — not "written small".

**2. The `result` event's aggregate cannot stand in for it.** Its
`cache_read_input_tokens` sums every sub-call in the turn (measured: `in+cc+cr` walking
38377 → 38538 → 38671 *within one turn*), so it overstates the final prompt by a factor
that grows with the sub-call count — roughly 2x on a short turn, worse on long agentic
turns. Using it would trade an underestimate for an overestimate.

**3. The heuristic itself is badly wrong.** It counts only cc-connect's own history text,
ignoring `tool_use`/`tool_result` blocks and the fixed system-prompt + tools + skills
overhead. On our test instance it read **12,887** for a turn whose real prompt was
**37,184** — 2.9x low, and the ratio moves with how much the agent read.

## Where the exact number actually lives

The Claude Code JSONL transcript records one self-contained record per API call, and that
record's `in + cache_creation + cache_read` is exactly that call's prompt size. Unlike the
result aggregate it does **not** accumulate (verified live: `cr 38802` → `in+cr 38948` →
next `cr 38948`). It is the same number behind Claude Code's own `ctx N%` indicator.

Reading it is cheap as a **bounded tail read** rather than a scan. Measured across 11,061
real transcripts (2.5 GB total, largest 137 MB):

| percentile | last-line bytes |
|---|---|
| min | 82 |
| p50 | 478 |
| p90 | 668 |
| p99 | 1452 |
| max | 32510 |

So a 4 KiB initial window settles 99.7% of sessions, and doubling only when the record is
out of reach covers the rest. A full rescan would be O(n²) — one of our transcripts is
136 MiB, and auto-compress consults this on every turn.

## Reproduction

1. Any project whose agent is `claudecode` behind an Anthropic-compatible gateway that
   reports zero usage — MiniMax-M3 is the one we measured.
2. `auto_compress.enabled = true`, any `max_tokens`.
3. Watch the logs: `GetContextUsage()` stays `nil`, and the trigger value comes from the
   text-length heuristic no matter how large the real prompt gets. (Before the logging
   added in the PR, this was invisible — "the agent reported a small prompt" and "the
   agent reported nothing and we guessed from text" looked identical.)

## Proposed fix — PR #1845

- Recover the exact per-call prompt size from the transcript tail: in `handleResult` when
  the in-process snapshot is absent, and once at attach time so a `--resume`d process does
  not spend its first turn heuristic-blind. The result aggregate is used only for the
  `EventResult` billing fields.
- Coerce non-`float64` numbers in `parseClaudeUsage`, so a CLI build emitting token counts
  as JSON strings cannot silently zero them out either.
- Add `UsageSourceReporter` + `estimate_source` / `usage_source` on every decision log, so
  this ambiguity is visible next time.
- Add `auto_compress.allow_heuristic`, **default false**: a turn with no exact usage makes
  no decision and leaves `lastAutoCompressAt` untouched, so the next turn (which does have
  exact usage) decides on real data. The legacy fallback is preserved behind the opt-in.

Verified on an isolated instance against MiniMax-M3 over a multi-turn session: 9/9
decisions report `estimate_source=exact` with `usage_source=transcript`, zero heuristic
fallbacks, zero skips; `lastUsage` goes from never-written to written once per turn and
tracks real context growth (41656 → 41707 → … → jumps 42,597 → 58,489 when the turn loads
a large file). The one turn that crossed the threshold genuinely fired `/compact` on real
data, after which the next decision read 37,102.

Refs #1115

YongmaoLuo:
## What this fixes

`auto_compress` silently fell back to the text-length heuristic for providers that
report no usable usage. Two independent defects fed that fallback:

**1. Providers that send an empty usage block.** Measured on MiniMax-M3: every
stream-json `assistant` event carries

```json
"usage": {"input_tokens": 0, "output_tokens": 0, "cache_creation_input_tokens": 0, "cache_read_input_tokens": 0}
```

`handleAssistant`'s `if used > 0` guard therefore never wrote `lastUsage`, so
`ContextUsageReporter.GetContextUsage()` stayed `nil` for the entire session and
**every single turn** fell through to the heuristic. The merged fix in #1761 made
the exact number authoritative, which is correct — but it left this class of
provider with no exact number to be authoritative *about*.

**2. The `result` event's aggregate cannot stand in for it.** `cache_read_input_tokens`
on the result event sums every sub-call in the turn (measured: `in+cc+cr` walking
38377 → 38538 → 38671 within one single turn), so it overstates the final prompt by a
factor that grows with the sub-call count — roughly 2x on a short turn. We
deliberately do *not* use it as the fallback.

**3. The heuristic itself is badly wrong.** It counts only cc-connect's own history
text, ignoring `tool_use`/`tool_result` blocks and the fixed system-prompt + tools +
skills overhead. On our test instance it read **12,887** for a turn whose real prompt
was **37,184** — 2.9x low, and that ratio moves with how much the agent read.

## Why the transcript

The Claude Code JSONL transcript records one self-contained record per API call, and
that record's `in + cache_creation + cache_read` is exactly that call's prompt size —
it does **not** accumulate the way the result aggregate does (verified live:
`cr 38802` → `in+cr 38948` → next `cr 38948`). That is the same number Claude Code
itself uses for its `ctx N%` indicator.

This is deliberately a **bounded tail read**, not a full scan: a long-lived transcript
runs to hundreds of MiB (one measured at 136 MiB) and auto-compress consults this every
turn, so a full rescan would be O(n²). The window starts at **4 KiB** and doubles only
when the newest record is out of reach. Measured across **11,061 real transcripts**
(2.5 GB total, largest 137 MB):

| percentile | last-line bytes |
|---|---|
| min | 82 |
| p50 | 478 |
| p90 | 668 |
| p99 | 1452 |
| max | 32510 |

4 KiB covers 99.7% of sessions; the doubling window covers the rest (verified to reach a
record 240 KB back as `65536 → 131072 → 262144 → 524288`). A multi-MiB tool result in
the tail is skipped for free — a truncated line either fails to parse or fails the
`"type":"assistant"` check, so the backward walk just moves on.

## Changes

- **`agent/claudecode`**: recover the exact per-call prompt size from the transcript —
  in `handleResult` when the in-process snapshot is absent, and once at attach time so a
  `--resume`d process does not start each session heuristic-blind. The result aggregate
  is now used **only** for the `EventResult` billing fields.
- **`agent/claudecode`**: `parseClaudeUsage` coerces non-`float64` numbers, so a CLI build
  emitting token counts as JSON strings can no longer silently zero them out.
- **`core`**: `UsageSourceReporter` + `estimate_source` / `usage_source` on every decision
  log, so *"the agent reported a small prompt"* and *"the agent reported nothing and we
  guessed from text"* stop looking identical in the logs. That ambiguity is what kept this
  bug invisible.
- **`config`**: new `auto_compress.allow_heuristic`, **default false**.

## `allow_heuristic` (default false)

With it off, a turn with no exact usage makes **no decision at all** and leaves
`lastAutoCompressAt` untouched, so the next turn — which does carry exact usage — decides
on real data. With it on, the legacy heuristic fallback is preserved verbatim.

The heuristic was never a good signal: it is incomplete by construction (it cannot see
tool results or the fixed overhead) and so it errs in both directions — compacting early
and discarding context still in use, or never compacting at all. We think the safe default
is to not act on it, and we've kept the old behavior behind an opt-in rather than deleting
it outright.

## Verification

Run on an isolated cc-connect instance against MiniMax-M3 (the provider that triggers
this), over a multi-turn session:

```
INFO auto-compress: decision estimate=61552 estimate_source=exact usage_source=transcript heuristic_estimate=12879 max_tokens=60000 triggered=true
INFO auto-compress: decision estimate=70906 estimate_source=exact usage_source=transcript heuristic_estimate=12881 max_tokens=60000 triggered=false
INFO auto-compress: decision estimate=37102 estimate_source=exact usage_source=transcript heuristic_estimate=12883 max_tokens=60000 triggered=false
```

- **9/9 decisions** report `estimate_source=exact` with `usage_source=transcript`; **zero**
  heuristic fallbacks and zero skips.
- `lastUsage` goes from never-written to written once per turn; the value tracks the real
  context as it grows (41656 → 41707 → 41758 → 41809 → …) and jumps correctly when the turn
  loads a large file (42,597 → 58,489).
- The 61,552 turn genuinely **fired** `/compact` on real data, after which the next decision
  read 37,102 — i.e. compaction triggered on the right signal and had the expected effect.
- `--resume` attach recovery verified separately:
  `INFO recovered context usage from transcript used=37225 input=41 cache_read=37184`.

`go build ./...`, `go vet`, and the `core` / `agent/claudecode` / `config` / `cmd` suites
pass. (`TestCUJ_A3_ImageReachesAgent` / `TestCUJ_A5_FileReachesAgent` are flaky in this
repo under load — `TempDir RemoveAll cleanup: directory not empty` — and fail on a clean
tree without this branch; confirmed by stashing the diff.)

## Main-path safety (scope of the change)

The exact-usage recovery itself is **entirely inside `agent/claudecode`** — the transcript
reader, the attach-time recovery, and the `handleResult` fallback are all package-local.
Nothing in the shared path changed shape for other agents:

- `core/interfaces.go` only **adds** an optional `UsageSourceReporter`. Nothing is required
  to implement it; the engine type-asserts and degrades to `""`.
- `replyFooterSessionContextUsage` keeps its signature and its return value (it now
  delegates to `replyFooterContextUsage`).
- `SetAutoCompressConfig` keeps its signature and behaviour (it now delegates to
  `SetAutoCompressConfigWithSource(..., false)`), so existing callers and tests are
  unaffected.

The one shared-path behaviour change is the decision gate in `processInteractiveEvents`,
and it is deliberately narrow. Only **4** of cc-connect's agent packages implement
`ContextUsageReporter` (claudecode, codex, copilot, pi); gemini, cursor, kimi, qoder, iflow,
opencode, devin, antigravity, acp, tmux and reasonix never can. So the rule is:

| situation | decision |
|---|---|
| exact usage reported | decide on it |
| **no `ContextUsageReporter` at all** | **decide on the heuristic, exactly as before** |
| reporter present, nothing reported this turn (the MiniMax case) | wait for the exact number |

Only the third row is new. An agent that cannot report usage has no other mechanism, and
skipping the decision there would silently switch auto-compress off for most agents — a
main-path regression, not a fix. `TestAutoCompress_AgentWithoutReporter_StillUsesHeuristic`
pins that boundary, and it fails if the gate is widened to `estimateSource == "exact"`
alone (verified by mutation).

## On the name

`auto_compress` describes the trigger, not the timing. It fires **after** a turn completes,
never inside the agent's loop — so a turn that grows past the threshold mid-flight is only
compressed once it finishes. That is now spelled out in `config.example.toml` and on the
`AutoCompressConfig` struct. We did **not** rename the TOML key, since that would break
every existing config; happy to do it if you'd prefer a clearer name and consider it worth
the migration.

For Claude Code specifically this is why we'd suggest reaching for the native knob first:
cc-connect already passes `[projects.agent.options.env]` straight through to the spawned
`claude`, so `CLAUDE_CODE_AUTO_COMPACT_WINDOW = "200000"` gives mid-loop compaction at any
window from 100K to 1M — i.e. any value in the 200K–1M range, which is exactly the band
`auto_compress` was being used to approximate, but acted on inside the iteration rather
than after it.

There is one thing the native knob cannot do, though: `auto_compress` fires only *between*
turns, so it can pre-empt native compaction and keep the agent from being summarized mid-task
at all.

That distinction matters more than it first looks. A turn ends when the agent itself decided it
was done answering — it stopped to ask for a decision, or it finished the work and reported
back. That is the "sub-task has resolved" condition, which is exactly what
[Self-Compacting Language Model Agents](https://arxiv.org/html/2606.23525v2) spends a
metacognition rubric computing. So native compaction is the *fixed-schedule* arm of that paper's
experiment, and `auto_compress` is the *rubric-gated* arm approximated from a free behavioural
signal. The paper's timing-attributable gain is **+11.5 points** (oracle analysis, Table 3), and
Claude Code's own `/compact` prompt loses **~53% of stated constraints after one round** run
([The Compaction Cliff](https://arxiv.org/abs/2608.22752)).

We could not find anyone who has run the direct A/B — turn-boundary vs mid-turn compaction,
measured on task success, on Claude Code — so the full evidence trail, the exceptions where a
turn boundary is *not* a safe point, and a test design that would settle it are in
[#1846](https://github.com/chenhg5/cc-connect/issues/1846#issuecomment-5660333173).

## Related

- **#1846** — the follow-up issue this PR closes: `auto_compress` still falls back to the
  text heuristic for providers that report no usage. Full root cause and measurements there.
- #1764 — the original report on the text-only heuristic (closed by #1761). This PR
  addresses the half #1761 could not: providers that report nothing exact.
- #1115 — the open report that `estimateTokens` ignores system prompt / tools overhead.

Closes #1846
Refs #1115


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
Mergeable. This is a well-scoped BUGFIX for a real and well-characterised defect: under MiniMax-M3 (and any provider whose stream-json `assistant` events carry empty usage blocks), every turn silently fell back to a 2.9×-off text-length heuristic because three independent defects compounded (the `used > 0` guard zeroed `lastUsage`, the result aggregate `cache_read_input_tokens` overcounts by ~2×, and the heuristic itself ignores tool results + system overhead). The PR fixes all three by switching the ground truth to per-call `in+cache_creation+cache_read` recovered from the Claude Code JSONL transcript — the same number Claude Code itself uses for its `ctx N%` indicator. CI 5/5 green, branch-local tests pass, behaviour preserved for the 11 reporter-less agents. Three P2 follow-ups; no blockers.

✅ What looks good:

- **Defect decomposition is honest.** PR body and decision log explicitly distinguish the three independent defects that masked each other. The empty-usage provider issue (#1846) could not have been fixed without the source-tag work; both landed together as they had to.
- **`decidable` gate is correct and the boundary is pinned.** The `!hasReporter || estimateSource == "exact" || autoCompressAllowHeuristic` rule preserves behaviour for gemini/cursor/kimi/qoder/iflow/opencode/devin/antigravity/acp/tmux/reasonix (11 of cc-connect's agent packages), and `TestAutoCompress_AgentWithoutReporter_StillUsesHeuristic` pins that boundary with an `AgentSession` interface-embedded fixture whose test body explicitly checks the fixture does NOT implement the reporter. Mutation-tested by the author.
- **`UsageSourceReporter` is additive.** Optional refinement, type-asserted at the engine, falls back to `""`. codex/copilot/pi don't implement it but their existing `ContextUsageReporter` keeps working.
- **Bounded tail read is well-engineered.** 4 KiB → 8 → 16 → 32 → 64 KiB window covers 99.7% of real transcripts (p99 last-line 1452 bytes, max 32510), with `transcriptLineCap = 64 MiB` as a hard ceiling. The multi-MiB-tool-result case "falls out for free" because truncated lines fail to parse or fail the `type: assistant` check — no carry buffer needed, which is what makes the whole reader O(1)-ish rather than O(n).
- **`(usage, found)` two-value return is right.** `found = false` means "file unreadable", `found = true, usage = nil` means "read it, nothing usable". The caller can decide whether to log differently; tests pin both branches.
- **Test coverage is thorough.** `TestTailUsageFromTranscript_MatchesReference` cross-checks against an oracle read-the-whole-file implementation across many truncation boundaries including mid-multi-MiB-line cuts. `TestTailUsageFromTranscript_GrowsWindowOnlyAsNeeded` verifies the doubling actually reaches 240 KiB back. `TestHandleResultUsesTranscriptTailWhenAssistantEmpty` is the end-to-end repro of the reported bug; `TestHandleResultKeepsLiveAssistantUsage` is its dual — verifying the live path is NOT replaced by the transcript on a provider that does report per-call usage. `TestAutoCompress_NoUsage_MakesNoDecision` and `TestAutoCompress_HeuristicOptIn_StillTriggers` together pin the new gate.
- **Config example and doc-comments are bilingual (zh + en) and explain the "after turn" timing honestly.** The TOML comment explicitly tells users to prefer `CLAUDE_CODE_AUTO_COMPACT_WINDOW` (Claude Code's native mid-loop compaction) for the 200K–1M band — that is the right answer.

🚨 Must fix:
- None.

🟠 Should improve:

- **P2 #1 — release notes must flag the silent behaviour change.** Two things change by default: (1) `auto_compress.allow_heuristic = false` means a turn with no exact usage makes no decision and waits for the next turn; (2) `auto-compress: decision` is now an `slog.Info` log line (previously silent), which any log-scraping monitoring needs to know about. Both are intentional and the right call, but operators should see this in the release notes. (This is for the release, not the PR — but flagging here so it isn't lost.)
- **P2 #2 — `auto-compress: decision` log lacks `session_key`, but the skip log has it.** `core/engine.go` decision path emits `slog.Info("auto-compress: decision", "estimate", ..., "triggered", ...)`; the skip path emits `slog.Debug("auto-compress: skipped, ...", "session_key", sessionKey, ...)`. Add `session_key` to the decision log too so per-session correlation works in both branches.
- **P2 #3 — `TestTailUsageFromTranscript_GrowsWindowOnlyAsNeeded` comment over-promises.** The comment claims `Leaving a log at every successful read makes the sequence the test asserts on if this ever regresses to an always-large read`, but the test body never captures or asserts on any log output. Either capture slog and assert the doubling sequence, or weaken the comment.

🔵 Optional:

- **P3 #1 — `_ = tailUsageFromTranscript(...)` in `handleResult`.** The two-value return is correct for tests, but a one-line caller comment explaining `found` is intentionally discarded ("missing file" and "readable but no record" behave identically here) would prevent a future reader from "fixing" the call site to log differently.
- **P3 #2 — `asInt` string handling order.** `ParseInt` before `ParseFloat` is correct (exact first, lossy fallback), but the doc-comment doesn't say so. One sentence would help.

❓ Questions:

- None blocking. The PR description's measurements (2.5×–5× heuristic error, 11,061-transcript percentile table, in+cc+cr 38377 → 38538 → 38671 within a single turn, 136 MiB / 137 MB outliers) make the design decisions auditable without back-and-forth.

Testing / Risk:
- **Verified by QA locally on `29cb7b0f`:** `go build ./...` passes (only `web/embed.go:13 all:dist` complains, pre-existing, unrelated to this PR); `go vet ./agent/claudecode/... ./core/... ./config/... ./cmd/...` clean; `go test -count=1 -timeout 120s -run "TestAutoCompress|TestParseClaudeUsage|TestHandleAssistant|TestHandleResult|TestClaudeUsageFromTranscriptLine|TestTailUsageFromTranscript|TestHandleResultUsesTranscriptTailWhenAssistantEmpty|TestHandleResultKeepsLiveAssistantUsage" ./agent/claudecode/... ./core/...` — **PASS** (0.062s + 0.374s).
- **CI 5/5 green:** lint 2m25s, performance-test 51s, regression-test 28s, smoke-test 28s, unit-test 4m46s (Actions run 34814279784).
- **Remaining risk:** the `allow_heuristic = false` default is a behaviour change for any user whose agent CAN report exact usage but for some reason takes > 1 turn to do so — they will observe auto-compress triggering one turn later than before. In practice every reporter agent reports exact usage on the first result event after attach (verified 9/9 by the author), so this is a low-frequency corner case. `TestAutoCompress_NoUsage_MakesNoDecision` pins that `lastAutoCompressAt` is left untouched so the next turn's exact data still drives the decision.
- **Behaviour preserved for:** the 11 reporter-less agent packages (gemini, cursor, kimi, qoder, iflow, opencode, devin, antigravity, acp, tmux, reasonix) — heuristic remains the only trigger, identical to pre-PR behaviour.
- **Rollback:** revert is safe (9 files, ~1100 lines, additive optional interface). Users who want the old heuristic behaviour can set `auto_compress.allow_heuristic = true` in config without code changes.

Next step:
Merge. The three P2 follow-ups are minor and can land as a follow-up PR or directly in the release notes — they do not block this PR. If maintainer prefers to land P2 #2 (session_key in decision log) before merge, I'd happily re-approve on the follow-up head.

chenhg5 (approved):
## Conclusion: Approve

Strong fix for the MiniMax-M3 cluster (Issue #1846): three independent defects — empty assistant usage block, result event aggregate misuse, and a 2.9x-off text heuristic — are addressed with a bounded transcript tail read, an attach-time recovery for `--resume`, and a `UsageSourceReporter` for diagnostics. Default `allow_heuristic=false` is the right call; reporter-less agents (gemini/cursor/kimi/qoder/iflow/opencode/devin/...) keep the legacy path via `decidable = !hasReporter`.

## Verification

- CI: 5/5 green on `29cb7b0f` (lint 2m25s, performance 51s, regression 28s, smoke 28s, unit 4m46s).
- Local: `go test -count=1` green on all new + renamed tests in `agent/claudecode` and `core`. The `TestAutoCompress_AgentWithoutReporter_StillUsesHeuristic` boundary is exactly right and would fail if the gate is ever widened.
- Reporter evidence in PR body (9/9 exact, 0 heuristic, attach-time recovery verified, bounded window reaches a 240 KB-back record) is internally consistent and matches the implementation.

## Must fix
None.

- **`reloadConfig` keep-alive for `AllowHeuristic`**: `cmd/cc-connect/main.go:1782` only calls `SetAutoCompressConfigWithSource` when the new config has `auto_compress` populated; the `else` branch (line ~1785) falls back to `SetAutoCompressConfig(false, 0, 0)` and the false→true flip path goes through a different code path. Please confirm `proj.AutoCompress.AllowHeuristic` is preserved across a reload that disables then re-enables `auto_compress`, or note explicitly that a full config reload is required to pick up the new flag.

- **Release notes for `allow_heuristic` default flip**: existing configs that rely on the old "fall back to text heuristic" behavior will see turn-level decision timing change. The PR body's "Main-path safety" section covers it; please mirror that note into `CHANGELOG.md` / `docs/auto-compress.md` if either exists.

## Nits (optional)

- `reference_test.go:36` runs an exhaustive cut sweep per fixture. With current sizes this is fast, but consider adding `t.Parallel()` and a sampling step if the fixture grows.
- `reference_test.go:90` test comment claims it asserts a log sequence; the test does not. Either remove the promise or wire a `bytes.Buffer` log hook.
- `claudeContextWindow("MiniMax-M3[1m]")` (if such a variant appears) currently falls through to 200K. Add a unit test once the variant exists.

## Acknowledgments

The measurement-driven write-up — empty usage block, sub-call aggregate summing, 12,887 vs 37,184 heuristic gap, 4 KiB → doubling window reaching 240 KB — is the kind of evidence that makes this PR easy to approve. The reporter's `TestAutoCompress_AgentWithoutReporter_StillUsesHeuristic` boundary check is exactly the right shape for keeping the main path safe.

Full QA decision doc: `doc-20260914-ko768b`

PR #1847, opened 2026-09-14
fix(core): lock replacement session before releasing the old one in idle auto-reset

### LINKED ISSUE #1832: [Bug] maybeAutoResetSessionOnIdle: failed TryLock on the new session leaves the incoming turn running without the busy lock

Enominera:
> **Disclosure**: I'm an AI coding agent (Claude Code) working on behalf of @Enominera, who reviewed and approved submitting this. 仅供参考.

Fixes #1832.

## Summary

`maybeAutoResetSessionOnIdle` released the old session's busy lock (and tore down its interactive state) **before** attempting to lock the freshly created replacement. When that `TryLock` failed, the function returned nil — but the caller then ran its turn on the old session, which was now **unlocked** (and whose interactive state had been cleaned). A second concurrent turn could then acquire the busy lock for the same session, and two turns would interleave on one session.

## Fix

Pure reorder, no API change: lock the replacement session first; only on success release the old one and clean up its interactive state. On `TryLock` failure we now return with the old session still locked and its state intact — the caller's turn proceeds exactly as if no reset had been attempted.

## Testing

- `go test ./core/ ./config/ ./cmd/cc-connect` green.
- The same reorder has been running on our deployment since 2026-09-12 as an out-of-band patch (adapted there to the generation-counted lock API from #1838).

Deliberately sent separately from #1838 per the review discussion there — this fix is independent of the generation-counter work, and keeping them apart keeps both PRs' risk surfaces tight.


chenhg5 (approved):
## Conclusion: Approve

Clean, surgical fix for the lock-ordering bug in `maybeAutoResetSessionOnIdle`. The reorder is mechanical, the comment captures the contract precisely ("the caller assumes the session it passed in stays locked when we return nil"), and the slog message now tells ops why the new lock failed. CI 5/5 green on head `caa50470`.

## Verification

- Local: `go test -count=1 -run "TestMaybeAutoResetSessionOnIdle|TestAutoCompress|TestHandleMessage|TestResetOnIdle" ./core/...` → PASS (0.858s).
- CI: 5/5 green (lint 2m12s, unit 4m40s, smoke 29s, regression 41s, performance 52s), `mergeStateStatus=CLEAN`.
- Production validation: same fix has been running as an out-of-band patch on the author's deployment since 2026-09-12.
- Decoupling from #1838 is the right call — the generation-counter work and this lock-ordering fix are orthogonal, and bundling them would inflate #1838's blast radius.

## Must fix
None.

- **Add a regression test for the exact TryLock-failure path.** The two existing `TestMaybeAutoResetSessionOnIdle_*` tests cover the "fires reset" and "doesn't fire reset" branches, but neither exercises the `newSession.TryLock()` failure branch — which is the path that was actually broken. A small test that pre-locks the new session externally (forcing TryLock to fail), calls `maybeAutoResetSessionOnIdle`, then asserts `rotated == nil` AND that the original session is still locked would have caught the original bug and would catch any future regression of this exact form. Suggested outline:

  ```go
  func TestMaybeAutoResetSessionOnIdle_KeepsOldSessionLockedWhenTryLockFails(t *testing.T) {
      e := newTestEngine()
      e.SetResetOnIdle(30 * time.Minute)
      sm := NewSessionManager(t.TempDir())
      session := sm.GetOrCreateActive("user:sk")
      session.AddHistory("user", "hello")
      session.SetAgentSessionID("agent-id", "claudecode")
      session.TryLock()
      session.mu.Lock()
      session.LastUserActivity = time.Now().Add(-35 * time.Minute)
      session.mu.Unlock()

      // Pre-lock the new session so the function's TryLock must fail.
      preempt := sm.NewSession("user:sk", "")
      if _, ok := preempt.TryLock(); !ok {
          t.Fatal("preempt failed to lock")
      }

      p := &stubPlatformEngine{n: "test"}
      msg := &Message{SessionKey: "sk", ReplyCtx: "ctx"}
      rotated := e.maybeAutoResetSessionOnIdle(p, msg, sm, "ws:sk", session)
      if rotated != nil {
          t.Fatalf("expected nil, got %v", rotated)
      }
      if _, ok := session.TryLock(); ok {
          t.Fatal("original session lock was released; caller invariant violated")
      }
  }
  ```

- **Document the caller contract on the function's godoc.** The "if I return nil, your session is still locked" invariant is the load-bearing piece of this fix and lives only in an inline comment today. A one-line function doc would make it harder to regress. Suggested addition above the func signature:

  > Contract: when this function returns nil, the session the caller passed in remains busy-locked and its interactive state is intact. The caller may safely proceed to run a turn on that session.

## Nits (optional)

- The out-of-band patch note in the PR body could include a one-line follow-up reminder (e.g. "remove the production patch after merge") for whoever owns the deployment.

## Acknowledgments

The fact that this fix went through production as an out-of-band patch for two days before being PR'd is exactly the right validation discipline. Keeping it separate from #1838 keeps both risk surfaces tight, and the inline comment that names issue #1832 inline is a good touch for anyone back-reading.

Full QA decision doc: `doc-20260914-k5qyok`

PR #1848, opened 2026-09-14
fix(platform/max): reply to private-dialog callbacks and honor perm:* payloads

Voluka:
## Problem

MAX delivers callbacks from private dialogs **without `recipient.chat_id`**, and `platform/max` used that zero value as the send target. Reproduced on the live API with a real bot (long-poll mode, cc-connect v1.5.0).

Three symptoms, all from the same root cause:

**1. Every reply to a button press was lost.**

```
max: send message failed status=403 chat=0 body={"code":"chat.denied","message":"Invalid chatId: 0"}
platform reply failed platform=max error="max: send message: HTTP 403: ..."
```

The decision was applied internally, but nothing reached the user — the button looked dead.

**2. The press was routed to a parallel session.**

Session keys are built as `max:{chatID}:{userID}`, so a callback produced `max:0:<user>` — a different session from the conversation the button belonged to. The agent answered into a context nobody was reading:

```
11:38 [user]       perm:allow_all
11:38 [assistant]  I'm ready. What do you need?
```

**3. `Allow all` silently did nothing.**

Permission buttons forwarded the raw payload to the engine, which matches words (`allow`, `allow all`, `deny`). `perm:allow` worked *by accident* — tokenisation happens to find the `allow` token — while `perm:allow_all` matched nothing at all. From the user's side: `Allow` works, `Allow all` does nothing, with no error anywhere.

## Fix

Follows what `platform/telegram` already does.

- `replyContext` carries `userID` next to `chatID`, with `hasChat()` treating `"0"` as missing. `postMessage` targets `?user_id=` when there is no usable chat — the documented way to reach a user in a private dialog — and returns a clear error when neither target exists, instead of a silent 403.
- `handleMessage` records each user's dialog `chat_id` (`sync.Map`); `handleCallback` reuses it, so a press lands in the conversation's own session.
- `ReconstructReplyCtx` keeps the user id, so an existing `max:0:<user>` session key stays usable after a restart.
- Permission payloads are translated to `allow` / `allow all` / `deny`, and the message is flagged `IsPermissionResponse` so a stale click is dropped rather than forwarded to the agent as user input (same rationale as #826). The card is edited to show the chosen option. Non-permission payloads pass through unchanged.
- `StartTyping` skips the indicator when there is no chat — `POST /chats/{id}/actions` has no `user_id` form.

## Tests

`platform/max/max_chatfallback_test.go`, four cases:

- `TestPostMessageFallsBackToUserID` — stub API replies `403 chat.denied` to `chat_id=0` exactly like MAX does; asserts the `user_id` fallback, that a real `chat_id` still wins, and that an empty context errors out.
- `TestCallbackReusesLearnedChatID` — message then callback without `recipient.chat_id`; the press must land in the dialog's session, not `max:0:<user>`.
- `TestReconstructReplyCtxKeepsUserID` — a restored `max:0:<user>` key stays sendable.
- `TestCallbackTranslatesPermissionPayloads` — all three `perm:*` payloads plus an ordinary button that must pass through untouched.

`go test ./platform/max/` — 38/38 pass, `go vet` clean, `gofmt` clean.

## Verification

Built and run against the live MAX API with a real bot. Before the patch: `Allow` worked by accident, `Allow all` did nothing, replies to callbacks returned 403. After: both buttons work, the card updates with the chosen option, and no `chat.denied` appears in the logs.

Unrelated note for anyone hitting it: native **voice messages** in MAX never reach bots at all — the platform emits an empty `message_created` with no `message`, `chat_id` or attachment, and the message is absent from `GET /messages`. That is a MAX-side bug (max-messenger/max-bot-api-client-ts#250), not something cc-connect can work around.


chenhg5 (approved):
## Conclusion: Approve

Solid, well-scoped fix for three coupled bugs in the MAX platform that all trace back to one root cause: MAX omits `recipient.chat_id` in callbacks from private dialogs. The implementation mirrors the existing `platform/telegram` pattern, the `chatOfUser sync.Map` learning cache is well-scoped, and the test design reproduces the actual 403 response body byte-for-byte. CI 5/5 green on head `2a23f166`.

## Verification

- Local: `go test -count=1 -run "TestPostMessageFallsBackToUserID|TestCallbackReusesLearnedChatID|TestReconstructReplyCtxKeepsUserID|TestCallbackTranslatesPermissionPayloads" ./platform/max/...` → PASS (0.011s).
- CI: 5/5 green (lint 2m19s, unit 5m5s, smoke 33s, regression 33s, performance 1m0s), `mergeStateStatus=CLEAN`.
- Live reproduction evidence in PR body (cc-connect v1.5.0, long-poll mode) is concrete and matches the implementation.
- The four new tests cover the right boundary cases: the 403 fallback path (with the exact response body stubbed), the cross-session routing regression, the `max:0:<user>` reconstructibility, and the perm-payload translation (including a non-permission payload pass-through control case).

## Must fix
None.

- **Document the single-instance assumption for `chatOfUser`.** The cache is `sync.Map`, process-local. A multi-instance cc-connect deployment behind a load balancer would have one instance learning a user's chat_id and another routing a callback to a different instance — leading to the same bug shape this PR fixes. `ReconstructReplyCtx` covers session-key rebuild across restarts, but it doesn't cover routing of a callback to an instance that never saw the originating message. Either document "single-instance only" in the PR body / release notes, or open a follow-up issue for a shared cache (Redis/etcd, similar shape to existing `dedup`).
- **Tier the `UpdateMessage` error log.** Right now the best-effort card edit logs at `slog.Debug` on any error. A `403 chat.denied` or `404 chat.not_found` (race: user left the chat between the message and the callback) is operationally interesting; a transient network blip isn't. Suggested split:
  ```go
  if err := p.UpdateMessage(ctx, rctx, origText+"\n\n"+choice); err != nil {
      msg := err.Error()
      if strings.Contains(msg, "403") || strings.Contains(msg, "404") {
          slog.Warn("max: permission callback edit failed (chat no longer accessible)",
              "error", err, "user", cb.User.Name)
      } else {
          slog.Debug("max: permission callback edit failed", "error", err)
      }
  }
  ```

## Nits (optional)

- Extract the perm-payload → engine-token mapping into named constants. Right now `"allow"`, `"allow all"`, `"deny"` are duplicated in both the `switch` cases and the test cases; named constants (`payloadAllow`, `payloadAllowAll`, `payloadDeny`) plus a `permPayloads` map would also give the test a single source of truth.
- Add a one-line godoc on `replyContext` noting that `chatID == "0"` represents a callback-without-recipient-chat and that callers should use `hasChat()` before treating it as a send target.

## Acknowledgments

The decision to ship the perm-payload translation as part of the same PR as the chat-id fix is the right one — they share the callback handler and the bug shape ("button does nothing") is identical from the user's perspective. The exact reproduction of `{"code":"chat.denied","message":"Invalid chatId: 0"}` in the test fixture is the kind of detail that makes the fix auditable.

Full QA decision doc: `doc-20260914-hj3605`

Voluka:
Thanks for the thorough review. All four points addressed in 3a6bd90e — no behaviour change, CI should stay green.

**Single-instance assumption for `chatOfUser`** — documented at the field. Worth noting the constraint is not only about the cache: MAX long-polling already requires one poller per bot token (two pollers split updates between them), so a load-balanced deployment is broken for a different reason first. I wrote down the failure shape anyway, including the partial mitigation — replies still go out via the `user_id` fallback, only the session routing is wrong — plus the pointer to a shared cache in the shape of `dedup`. Happy to open a follow-up issue if you want that tracked separately.

**Tiered `UpdateMessage` log** — implemented as suggested: `403`/`404` (user left the chat between the message and the press) now logs at `Warn` with the user name; everything else stays at `Debug`.

**Named constants** — `payloadAllow` / `payloadAllowAll` / `payloadDeny` plus a `permPayloads` map holding both the engine token and the card label, which also collapsed the `switch` into a single map lookup. The test now takes payloads from the constants, but the expected engine wording (`"allow"`, `"allow all"`, `"deny"`) stays spelled out on purpose — reading it back from the map would make the test agree with itself instead of pinning the strings the engine actually matches on.

**`replyContext` godoc** — added, stating that `chatID == "0"` marks a callback without a recipient chat and that callers must check `hasChat()` before treating it as a send target.

One field note since the fix has been running here since the report: the patched build has been live on a real bot for several hours, and both buttons behave correctly — `Allow` and `Allow all` resolve the pending permission, the card updates with the chosen option, and no `chat.denied` has appeared in the logs since.

chenhg5:
## Re-review: head now `3a6bd90e`

Follow-up commit addresses all four suggestions from my cycle-148 review. CI 5/5 green on `3a6bd90e`. PR remains approved.

**P2 #1 — `chatOfUser` single-instance assumption**: explicit godoc on the field now reads:

> Process-local by design, like dedup: it assumes a single cc-connect instance per bot token, which is also what MAX long-polling requires (two pollers on one token split updates between them). Running several instances behind a load balancer would let one learn the chat_id while another handles the callback, reviving the bug this cache exists to fix; that deployment shape needs a shared cache and is not supported today. Replies still work in that case — postMessage falls back to user_id — but the press lands in a separate session.

This is the right level of honesty — it states the constraint, the rationale (long-poll already requires single instance), the failure mode if violated, and what still works vs what breaks. ✓

**P2 #2 — tier `UpdateMessage` error log**: implemented exactly as suggested, with a useful comment explaining the 403/404 case ("the chat is gone — the user left between the message and the press"). ✓

**P3 #1 — godoc on `replyContext`**: added the doc-comment block I suggested, referencing `hasChat()` and `postMessage`. ✓

**P3 #2 — named constants for perm payloads**: `payloadAllow` / `payloadAllowAll` / `payloadDeny` + `permDecision` struct + `permPayloads` map. The switch is gone; the lookup is a single source of truth. Clean. ✓

`go test -count=1 ./platform/max/...` → PASS (5.120s, all tests in package, not just the new ones).

PR remains approved. Awaiting Glenn's manual merge.

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Re-review of #1848. The new commit `3a6bd90eb7` ("refactor(platform/max): address review feedback on callback handling") is a clean +49/-17 follow-up with no behavior change — every change addresses documentation, log-level granularity, or naming/structure. CI 5/5 green on the new head. Prior approval on `2a23f166b8` stands.

Review scope:
- Reviewed the full +49/-17 in `platform/max/max.go` and the +7/-3 in `platform/max/max_chatfallback_test.go`.
- Cross-checked that the permPayloads map keys match the engine-side words the test pins (the test deliberately spells out expected wording rather than reading it back from the map, which is the right call — it pins the engine contract).
- Verified the 403/404 → Warn tier is conservative: only triggers on those exact substrings, anything else stays at Debug.

✅ What looks good:
- **No behavior change**: the switch statement was rewritten as a map lookup with the same payload-to-(token, label) mapping; `isPerm` semantics are preserved (true for perm:* payloads, false for ordinary buttons).
- **`chatOfUser` process-local documentation** is now explicit and includes the load-balancer caveat: "Running several instances behind a load balancer would let one learn the chat_id while another handles the callback, reviving the bug this cache exists to fix; that deployment shape needs a shared cache and is not supported today." This is exactly the kind of future-reader hazard operators need to know.
- **Log-level tiering is well-reasoned**: 403/404 means the chat is gone (user left between message and button press) → Warn; anything else is treated as a transient blip → Debug. The substring check on `err.Error()` is precise (no false positives for "Unexpected status 403" in unrelated error wrappers — they would also indicate a chat-gone scenario and rightfully escalate).
- **Named constants (`payloadAllow`, `payloadAllowAll`, `payloadDeny`) + `permPayloads` map** give a single source of truth shared between the sender and the test; the original magic strings only appeared in one place anyway, so this is mainly a maintainability improvement.
- **Comment on `replyContext`** explicitly states the `chatID == "0"` contract and tells callers to check `hasChat()` first. This prevents future bugs where someone skips the helper.
- **Test discipline**: the test deliberately spells out expected wording ("allow", "allow all", "deny") instead of reading it back from `permPayloads`. The new test comment explains why: "taking them from permPayloads would make the test agree with itself instead of pinning the exact wording the engine matches on." This is exactly right.

🟠 Should improve:
- None.

🔵 Optional:
- P3: the `permDecision` struct has `token` and `label`; could be flattened to two parallel maps (`permTokens`, `permLabels`) for symmetry with the constant declarations, but a struct is clearer here and the compile-time cost is negligible.

❓ Questions:
- None.

Testing / Risk:
- Verified: GitHub CI 5/5 green on head `3a6bd90eb7` (lint, unit-test, smoke-test, regression-test, performance-test all success).
- Coverage: existing `TestCallbackTranslatesPermissionPayloads` updated to use the new constants (table-driven cases unchanged); no test logic changes, just constant references.
- Diff is +56/-20 across 2 files; one commit; behaviorally equivalent to the previously approved `2a23f166b8`.
- Remaining risk: zero. The map lookup is identical to the switch statement for the three known payloads; the only behavior difference is the explicit `isPerm := false` initialization before the map lookup, which is what the switch already did in the `default` branch.

Next step:
- Merge when convenient. The refactor is exactly what was asked for in the prior review; no new blockers.

PR #1853, opened 2026-09-15
fix: enforce allow_from in Feishu onCardAction

### LINKED ISSUE #1852: Feishu card-action handler bypasses the allow_from user allowlist
## Summary

cc-connect bridges local AI coding agents to messaging platforms and gates who may command the agent with a per-platform `allow_from` setting. Every platform's plain-text message path enforces it. The Feishu platform additionally supports interactive message "cards" (buttons, forms, permission-approval prompts), handled by `onCardAction()`. That function checks only the chat-level `allow_chat` filter and never checks `allow_from`, so any user who can see and click a card the bot has posted, in any chat the operator's `allow_chat` config admits (which may include a shared group), can trigger a full agent command dispatch or approve the agent's own pending dangerous-action confirmation, entirely bypassing the per-user allowlist that protects the equivalent text-message path.

Affected versions: confirmed on commit `757b4df` (current `main`).

## Details

`platform/feishu/feishu.go`, the plain-text message handler correctly enforces both filters, and documents the intended model:

```go
// With history sharing enabled, observe ordinary text/post messages only
// after the chat-level allow list has admitted the chat. Do this before the
// existing mention filter and before allow_from: the chat controls whether
// cc-connect may observe the conversation, while allow_from controls who may
// trigger an agent turn.
...
if !core.AllowList(p.allowFrom, userID) {
	slog.Debug(p.tag()+": message from unauthorized user", "user", userID)
	p.replyUnauthorizedAccess(ctx, replyContext{...})
	return nil
}

if chatType == "group" && !core.AllowList(p.allowChat, chatID) {
	slog.Debug(p.tag()+": message from unauthorized chat", "chat_id", chatID)
	return nil
}
```

`onCardAction()`, as shipped, checks only `allow_chat`:

```go
func (p *Platform) onCardAction(event *callback.CardActionTriggerEvent) (*callback.CardActionTriggerResponse, error) {
	if event.Event == nil || event.Event.Action == nil {
		return nil, nil
	}

	// Check allow_chat filter: skip card actions from chats this platform doesn't own.
	if event.Event.Context != nil && event.Event.Context.OpenChatID != "" {
		if !core.AllowList(p.allowChat, event.Event.Context.OpenChatID) {
			return nil, nil
		}
	}

	actionVal, _ := event.Event.Action.Value["action"].(string)
	...
```

No `allow_from` check follows, for any of the five action prefixes the function handles:

- `cmd:` dispatches the button's payload as an arbitrary agent command, identical in effect to typing that text:
  ```go
  if strings.HasPrefix(actionVal, "cmd:") {
      cmdText := strings.TrimPrefix(actionVal, "cmd:")
      ...
      go p.dispatchCoreMessage(&core.Message{..., Content: cmdText, ...})
  }
  ```
- `perm:allow` / `perm:deny` / `perm:allow_all` approve or deny the agent's own pending permission/confirmation prompt:
  ```go
  if strings.HasPrefix(actionVal, "perm:") {
      ...
      go p.dispatchCoreMessage(&core.Message{..., Content: responseText, IsPermissionResponse: true, ...})
  }
  ```
- `nav:`, `act:`, and `askq:` similarly dispatch to the agent or update agent-facing state without any `allow_from` check.

`dispatchCoreMessage()` hands the message straight to the registered handler with no further authorization check downstream, so nothing else in the call chain re-validates the sender.

## POC

(available upon request)

## Impact

Any user who can see a card the bot has posted, in a chat admitted by `allow_chat` (which is frequently a shared group, not a 1:1 DM, and defaults to allow-all when unset), can:

- Click a `cmd:` button to dispatch an arbitrary command to the agent as if they were the `allow_from`-authorized operator.
- Click `perm:allow` or `perm:allow_all` to approve the agent's own pending confirmation for a dangerous action (the human-in-the-loop safety check the permission-prompt UI exists to enforce), on behalf of the operator, without being that operator.
- Click `nav:`/`act:`/`askq:` buttons to navigate cards, run in-place actions, or answer the agent's own questions, steering its behavior.

This defeats the `allow_from` allowlist that is the project's own documented mechanism for restricting who may command the agent, whenever the agent runs on Feishu and posts any interactive card (which is the default behavior; interactive cards are enabled unless explicitly disabled via `enable_feishu_card: false`).

carfeii:
Fixes #1852.

`onCardAction` handled interactive card actions (`nav:`/`act:`/`cmd:`/`perm:`/`askq:`) checking only `allow_chat`, never `allow_from`, unlike the sibling plain-text message handler which enforces both. `cmd:` dispatches arbitrary text as a full agent command; `perm:allow` and `perm:allow_all` approve the agent's own pending dangerous-action confirmation. Any user who could click a card the bot posted in a chat `allow_chat` admits could act as the `allow_from`-authorized operator. This adds the same `allow_from` check the text handler already uses, plus a regression test.


chenhg5 (commented):
## QA Review: PR #1853

**Verdict: Comment — recommend closing in favor of #1854.**

This PR fixes the same P0 security bug as #1854 (`onCardAction` bypasses the per-user `allow_from` allowlist, issue #1852), but with significantly smaller test coverage and a subtly weaker fail-closed guarantee.

### ✅ What looks good
- **Correct root cause.** The fix correctly identifies that `onCardAction` only checked `allow_chat` and ignored `allow_from` — `cmd:`/`perm:`/`askq:` actions can dispatch arbitrary text to the agent, so they must be gated by the same per-user allowlist as the text-message path. This is a real P0.
- **Single chokepoint.** The new check sits at the top of `onCardAction`, before the per-prefix dispatch block, so nothing can sneak past. Mirrors the placement of the existing plain-text `allow_from` check.
- **Reuses `core.AllowList`.** No duplicated parsing logic, so the card path and the text path stay in agreement on what "in allowlist" means.
- **CI green.** 5/5 pass on commit `b12f838da4c76bf3552970ba8853dfd6dd45f585`.

### 🟠 Should improve (substantive, but the right path is to defer to #1854)
- **🟠 P2 — Test coverage is much smaller than the security surface warrants.** This PR adds one negative test (`cmd:/whoami` from an unauthorized user). Compare to #1854 which adds **12 tests** covering all five action prefixes (`cmd:`, `perm:allow`, `perm:deny`, `perm:allow_all`, `nav:`, `act:`, `askq:`), explicit edge cases (empty operator), positive-path coverage for authorized users, the wildcard `*` regression case, and multi-list whitespace handling. For a P0 fix, the deeper test set is the safer choice.
- **🟠 P2 — Fail-closed on empty `Operator.OpenID` is not guaranteed.** Your check is:
  ```go
  if event.Event.Operator != nil && event.Event.Operator.OpenID != "" {
      if !core.AllowList(p.allowFrom, event.Event.Operator.OpenID) {
          slog.Debug(...)
          return nil, nil
      }
  }
  ```
  When `Operator` is nil or `OpenID == ""`, the entire block is skipped and execution falls through to the dispatch logic — i.e. an action with no identifiable user is allowed through. This is a fail-open path, exactly the opposite of what a security check should do. **#1854 closes this hole** with `if userID == "" || !core.AllowList(...) { ... return nil, nil }` — i.e. an unknown operator always gets blocked, even with `allow_from = "*"`. The plain-text handler at `feishu.go:1825` also uses the same fail-closed pattern. Please adopt this.

### 🔵 Optional / Nit
- **🔵 P3 — Test only covers `cmd:`.** For a P0 fix touching all five action prefixes, a single test gives future maintainers very little coverage confidence. Even if you decide to defer to #1854, please consider adding at least a `perm:allow_all` test (the highest-blast-radius variant) if you keep this PR open.

### ❓ Questions
- **❓ Are you aware of #1854 (chenhg5, opened the same day)?** Both PRs fix the same bug against the same base (`397d358528`). Maintainer typically picks one and closes the other to avoid double-fix conflicts. If you didn't see #1854, please review it before continuing.

### Recommendation
**Close this PR in favor of #1854.** Reasons:
1. #1854 has 12× the test coverage.
2. #1854 closes the empty-operator fail-open hole; #1853 leaves it open.
3. #1854 was reviewed by chenhg5 (who has merge authority) and is queued for merge.
4. Merging both would create merge conflicts in `feishu.go` and require one to be reverted.

If you want to keep this PR (e.g. you opened it first and prefer your style), the minimum change to make it merge-worthy is:
- Adopt the `userID == "" ||` fail-closed guard from #1854.
- Add the 11 missing test cases (or a curated subset covering each prefix + the empty-operator case).

Otherwise, please close this PR and let #1854 land.

### Testing / Risk
- **CI**: 5/5 green.
- **Diff size**: 12 lines of production code + 56 lines of test. Both small and easy to review.
- **Risk if merged as-is**: the empty-operator fail-open path is a real residual vulnerability — a card action with no `Operator` field would bypass the allowlist entirely. This is the same shape of bug the PR claims to fix.

**Impact**: P0 security bug, but better addressed via #1854.  
**Module**: `platform/feishu/feishu.go`.  
**Backward compat**: N/A (closes a security gap; deployments relying on the bypass would have been misconfigured).  
**Breaking change**: no.  
**Bug fix**: yes (security).

chenhg5 (commented):
## QA Follow-up: PR #1853 — still recommend closing in favor of #1854

Re-engaging one day after my initial comment. The branch is unchanged: still 1 commit `b12f838da4`, 1 negative test, fail-open path on nil `Operator` / empty `OpenID` intact. PR #1854 also unchanged: still 12 tests, fail-closed, ready for merge.

**Recommendation stands**: please close this PR. Either:

1. **Close as duplicate of #1854** (preferred — #1854 is maintainer-owned and has the fail-closed pattern + 12 tests), or
3. **Push a fix-up commit here** that adopts #1854's `userID == "" ||` fail-closed guard and at minimum adds tests for `perm:allow_all` (highest-blast-radius variant) + `perm:deny`. Even then, the maintainer would likely still prefer #1854 since the test set is broader.

If neither happens in the next cycle, I'll flag this to the maintainer for triage.

(Cycle 17 follow-up; no new code review needed since the branch is unchanged.)

PR #1854, opened 2026-09-15
fix(feishu): enforce allow_from in onCardAction() (#1852)

chenhg5:
## Summary

Fixes P0 security bug in `platform/feishu::onCardAction()`: card button clicks (cmd:/perm:/nav:/act:/askq:) previously checked only the chat-level `allow_chat` filter, bypassing the per-user `allow_from` allowlist that protects the equivalent plain-text message path.

Any user who could see a card the bot posted in an `allow_chat`-admitted chat (often a shared group) could:
- Click `cmd:` to dispatch arbitrary agent commands
- Click `perm:allow` / `perm:allow_all` to approve the agent's pending dangerous-action confirmations
- Click `nav:` / `act:` / `askq:` to steer card flow

## Fix

Mirrors the plain-text handler's check (see line ~1814 of `feishu.go`): after the operator user_id is extracted and before any of the five action prefix dispatches, fail closed if userID is empty or not in `core.AllowList(p.allowFrom, userID)`. Matches the documented two-layer model where `allow_chat` gates whether cc-connect observes the conversation and `allow_from` gates who may trigger an agent turn.

## Testing

12 new unit tests in `platform/feishu/card_allow_from_test.go`:

- **Negative**: cmd:/perm:allow/perm:deny/perm:allow_all/askq:/nav:/act: from unauthorized users all return `(nil, nil)` without invoking the handler or cardNavHandler
- **Empty Operator**: fails closed (userID == "" → reject) even with wildcard `allow_from`
- **Positive**: cmd: from authorized user dispatches with correct userID; perm:allow dispatches as `IsPermissionResponse: true` with content "allow"
- **Wildcard regression**: `allow_from = "*"` still permits anyone (no regression to deployments relying on the existing default)
- **Multi-list exact match**: `allow_from = "ou_a, ou_b , ou_c"` only allows listed members (whitespace + comma-separated)

All 12 pass with `go test -race -count=1`. Full `./platform/feishu/...` package test suite (37s) passes — no regressions.

## Scope

- **Files changed**: 2 (`platform/feishu/feishu.go`, `platform/feishu/card_allow_from_test.go`)
- **Lines**: +272 (test heavy by design — 12 distinct negative/positive cases)
- **Behavior change**: silent drop of unauthorized card actions (matching plain-text path's debug-logged silent return). No new user-facing strings, no i18n additions needed.

## Verification

```
go test -race -count=1 -run TestOnCardAction_AllowFrom ./platform/feishu/  # 12/12 PASS
go test -race -count=1 ./platform/feishu/...                              # full suite PASS
go vet ./platform/feishu/...                                              # clean
gofmt -l ...                                                               # clean
```

## Related

- Issue: https://github.com/chenhg5/cc-connect/issues/1852 (carfeii)
- Reporter POC available upon request
- Follow-up tracked separately: PM will request team-wide audit of other platform adapters (Telegram/Discord/Slack/钉钉/微信/企业微信/QQ/LINE) for similar callback handlers missing `allow_from`


chenhg5 (commented):
Conclusion: Approve

Overall assessment:
- This PR correctly fixes a real P0 security bug in `onCardAction()`. The implementation mirrors the existing plain-text handler's allow_from check, and the 12 new unit tests give full negative + positive coverage across all five action prefixes (cmd:, perm:allow/deny/allow_all, nav:, act:, askq:) plus edge cases. CI is green, local tests pass cleanly with `-race`.

Review scope:
- Reviewed `platform/feishu/feishu.go` (+11) and the new `platform/feishu/card_allow_from_test.go` (+261).
- Focused on correctness, security (fail-closed semantics), test coverage, and parity with the plain-text message handler.
- Verified `core.AllowList` semantics at `core/message.go:82` and confirmed the plain-text reference check at `feishu.go:1825`.

What looks good:
- Check placement is exact: after userID extraction (line ~803) and before all five prefix dispatches (line ~826+). Nothing can sneak past.
- `core.AllowList(p.allowFrom, userID)` is reused as-is, no duplicated parsing logic. This guarantees the card path and the plain-text path agree on what "in allowlist" means (whitespace-trim, comma-separated, case-insensitive, wildcard `*`).
- Intentional tightening vs. the plain-text path: card handler also fails closed on empty `userID` (`userID == "" || !core.AllowList(...)`), even with wildcard `*`. That's the correct security choice — when the operator is unknown, dropping the action is safer than accepting it. The plain-text path's `replyUnauthorizedAccess` feedback pattern would leak whitelist existence to a probing attacker, so the silent drop here is appropriate.
- `slog.Debug` with `p.tag()` prefix and `user` field matches the style of the surrounding handlers; useful for ops without being noisy.
- 12-test split into focused functions is more readable than a single table-driven test, and each negative case asserts a 200 ms "no message" timeout — strong guarantee the handler truly was bypassed, not just delayed.
- Test `BlocksPermAllowAll` is the most important one — it specifically blocks the blanket-approve action, which is the highest-impact exploit.

Must fix:
- None.

- Consider whether the docs (e.g. `docs/` page on `allow_from`) should explicitly mention that the filter covers both plain-text messages and card button clicks. Most existing deployments probably assumed this, but the existence of this bug suggests at least some users did not. This is a docs follow-up, not a blocker for this PR.
- Other platform adapters (Telegram/Discord/Slack/钉钉/微信/企业微信/QQ/LINE) likely have the same gap in their button/callback/interaction handlers. The PR description correctly scopes this PR to Feishu only, and the PM-tracked audit is the right follow-up channel. Just flagging that the audit should not slip.

Optional (P3):
- Future work could collapse the 12 test functions into a table-driven test to lower maintenance cost when new action prefixes are added. Current form is more readable, so this is purely a style preference.

Questions:
- None.

Testing / Risk:
- Verified locally:
  - `go test -race -count=1 -run TestOnCardAction_AllowFrom ./platform/feishu/` → 12/12 PASS
  - `go test -race -count=1 ./platform/feishu/...` → full package 37.08 s PASS, no regressions
  - `go vet ./platform/feishu/...` → clean
  - `gofmt -l` → clean
- GitHub CI run 34961204318: lint, unit-test, smoke-test, regression-test, performance-test all SUCCESS.
- Remaining risk: minimal. The only behavior change is that card actions from users not in `allow_from` are now silently dropped — this is the intended fix. Deployments that were relying on the buggy behavior (any group member can click cards) will need to add those members to `allow_from`, which is correct. No schema change, no config change, no data migration.
- Rollback: a single one-line revert restores original behavior.

Next step:
- Merge. The remaining platform-adapter audit should proceed in parallel via the PM-tracked follow-up; it does not block this PR.

chenhg5 (commented):
Conclusion: Approve (posted as COMMENT due to self-author rule)

Overall assessment:
- Comprehensive 12-test P0 security fix for the same `allow_from` bypass as #1853, with the fail-closed guard that #1853 lacks. 1 commit, +272/-0 across 2 files (261 lines of new test + 11 lines of production code). CI 5/5 green on `7256450c36`. Already-self-reviewed by chenhg5 (the repo maintainer) on Sep 22 with `Verdict: Approve — ready to merge` (per my prior research). This is the fix that the cycle 17 follow-up on #1853 was waiting for.

Review scope:
- `platform/feishu/feishu.go` (+11/-0): the new `userID == "" || !core.AllowList(...)` guard in `onCardAction`
- `platform/feishu/card_allow_from_test.go` (+261/-0, new file): 12 test cases
- Cross-checked `core.AllowList` semantics (it returns `true` for empty/wildcard `allowFrom`, so the `userID == ""` short-circuit is **load-bearing** for fail-closed behavior).
- Cross-checked the placement: guard sits after `userID` extraction but before `chatID/messageID` block, mirroring the plain-text message handler's structural position.

✅ What looks good:
- **Correct, complete fix**: the new guard sits at the top of `onCardAction`, after `userID` extraction but before any per-prefix dispatch (cmd/perm/nav/act/askq). Card actions from unauthorized users are dropped silently before they can reach any downstream dispatch — no race window, no shadow path.
- **Fail-closed semantics with the empty-userID check** is the right call given `core.AllowList`'s behavior of returning `true` for empty/wildcard `allowFrom`. Without `userID == "" ||`, an action with `Operator == nil` and `allowFrom == "*"` would be admitted by `AllowList("*", "")` and dispatch. The PR test `TestOnCardAction_AllowFromBlocksEmptyOperator` explicitly pins this — it asserts the action is dropped even when `allow_from = "*"`. This is exactly the failure mode #1853 leaves open.
- **12 test cases are well-partitioned**:
  - 5 negative tests, one per action prefix (`cmd:`, `perm:allow`, `perm:deny`, `perm:allow_all`, `nav:`, `act:`, `askq:`) — every dispatch surface is covered.
  - 1 negative test for the empty-Operator case (fail-closed guarantee).
  - 3 positive tests (`TestOnCardAction_AllowFromAllowsAuthorizedCmd`, `TestOnCardAction_AllowFromAllowsAuthorizedPermAllow`, `TestOnCardAction_AllowFromWildcardAllows`) to guard against over-blocking.
  - 1 positive test for multi-list whitespace handling (`TestOnCardAction_AllowFromMultiListRespectsExactMember` — covers the `ou_a, ou_b , ou_c` parse edge).
- **`newAllowFromTestPlatform` helper** is the right test scaffolding: it constructs a real `*interactivePlatform` via the public `New()` constructor and overrides only the `handler` channel, so the test exercises the production check path end-to-end (no mock-only assertions).
- **`waitNoMessage`/`waitAnyMessage` helpers** use `time.After` (not polling), so tests are deterministic and fast (~200 ms for negatives, 2 s for positives — appropriate for "must dispatch" assertions).
- **Test naming convention** (`TestOnCardAction_AllowFromXxxYyy`) is consistent and self-documenting.

- **P2 — Inconsistency with existing message handlers at `feishu.go:1825` and `feishu.go:5772`**: both the plain-text message handler and the bot-menu-event handler still use `if !core.AllowList(p.allowFrom, userID)` without the `userID == ""` short-circuit. Given that `core.AllowList("", "")` returns `true`, a missing `OpenID` on a plain-text event would currently be admitted when `allowFrom` is wildcard/empty. The new onCardAction is strictly more defensive than the message path it mirrors. Recommend a follow-up PR that hardens the message + menu paths the same way. **Not a blocker for this PR** (the P0 in scope is fixed), but worth opening an issue so it doesn't get lost.
- **P2 — `slog.Debug` for the blocked path is silent by design**, but operators investigating "why didn't my button click work?" will see no user-facing signal. Consider a single `Info` log line at lower verbosity for the first hit per session, or a counter metric, so debuggability doesn't suffer from the fail-closed default.
- **P2 — `cardActionEvent` helper hard-codes `oc_test_chat` / `om_test_message`**, which is fine for current coverage but will get noisy if you want to test the interaction with `allow_chat` filtering (currently the chat-level filter runs first and isn't exercised by these tests). Not required for this PR; noted for future hardening.

❓ Questions:
- None.

Testing / Risk:
- Verified: GitHub CI 5/5 green on head `7256450c36` (lint, unit-test, smoke-test, regression-test, performance-test all success).
- New tests: 12 in `card_allow_from_test.go`, all of them deterministic and well-scoped.
- Coverage comparison vs #1853: 12× more cases, including the 6 critical ones #1853 doesn't test (`perm:deny`, `perm:allow_all`, `nav:`, `act:`, `askq:`, empty-Operator, wildcard-regression, multi-list whitespace).
- **Risk if this lands**: the latent inconsistency at `feishu.go:1825` and `feishu.go:5772` is **not** fixed by this PR. Operators relying on wildcard `allow_from` and sending malformed events without `Operator.OpenID` would still be admitted on the message + menu paths. **Not a blocker** because the failure mode requires (a) wildcard `allow_from`, (b) malformed events with no OpenID, (c) an actual attacker control of event source — narrow intersection — but it should be hardened.
- **Risk if merged alongside #1853**: would create a merge conflict in `feishu.go`. **#1853 must be closed first**.

Next step:
- **Merge** — this is the maintainer's own fix with self-approval, CI green, and the only blocker is that #1853 needs to be closed first to prevent `feishu.go` conflict. Recommend the maintainer close #1853 and merge #1854 in the same maintenance window.
- **Follow-up issue**: open a follow-up to harden `feishu.go:1825` and `feishu.go:5772` with the same `userID == "" ||` guard. This is a latent P2 that should not block #1854 but should not be forgotten.

PR #1855, opened 2026-09-15
fix(claudecode): drop Stop-hook-rejected draft from aggregated reply

openclaw202602-lang:
## Problem

When a Claude Code **Stop hook** blocks a turn (`{"decision":"block","reason":...}`), Claude writes a rewritten answer within the same turn. With `tool_messages = false` (and stream preview disabled), `processInteractiveEvents` joins **all** text parts of the turn:

```go
if len(textParts) > 0 && segmentStart == 0 && !e.display.ToolMessages {
    fullResponse = strings.Join(textParts, "")
}
```

So the rejected draft and the rewrite are delivered together to the chat. If the hook blocks N times, the user receives N+1 concatenated versions.

## Reproduction

Claude Code 2.1.272, a Stop hook that blocks once, prompt "reply AAA". The `stream-json` output is:

```
assistant text: "AAA"
user:           [{"type":"text","text":"Stop hook feedback:\n<reason>"}]
system:         notification (key=stop-hook-error)
assistant text: "BBB"
result:         result="BBB"
```

cc-connect (v1.4.1 and v1.5.0, Feishu) delivers `AAABBB`.

There is a clean boundary signal (the `user` text item starting with `Stop hook feedback:`), but `handleUser` only processes `tool_result` items, so the engine never learns the draft was discarded.

## Fix

- `agent/claudecode`: when a user `text` item starts with `Stop hook feedback:`, emit a new `EventHookRejected`. All other user text is still ignored; the `tool_result` path is unchanged.
- `core`: in `processInteractiveEvents`, on `EventHookRejected` reset the accumulated answer state (`textParts`, `segmentStart`, `silentHold`, `partialText`, rich-card answer buffer). Later text accumulates normally.

The other aggregation paths (`runUnsolicitedReader`, `processCompressEvents`, `HandleRelay`) already prefer `event.Content` from the result event and only fall back to joining when it is empty, so they are unaffected.

Note: if a streaming card is enabled, the draft may already have been previewed before the reset; this PR only fixes the final aggregated reply.

## Tests

- `TestProcessInteractiveEvents_DropsStopHookRejectedDraft`: `AAA → hook_rejected → BBB → result` delivers only `BBB` (fails on current main with `AAABBB`).
- `TestProcessInteractiveEvents_NoHookRejectionKeepsTextAcrossToolBoundary`: without a rejection, `AAA → tool_use → tool_result → BBB` still delivers `AAABBB`.
- `TestHandleUserEmitsHookRejectedFromStopFeedback`.
- `go test ./core/ ./agent/claudecode/` passes on current main.


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Clean, minimal, well-tested bugfix for a real Claude Code 2.1.272 issue. The Stop hook rejection mechanism in Claude Code emits a `Stop hook feedback:` user text item between the rejected draft and the rewrite; without this fix, cc-connect concatenates both into the final reply, producing duplicated or corrupted output for users relying on Stop hooks. CI is fully green; 3 new tests pass with `-race`; full `./agent/claudecode` and `./core` packages pass with no regressions; `go vet` and `gofmt` clean.

Review scope:
- Reviewed `core/message.go` (+1, EventType constant), `agent/claudecode/session.go` (+12, `handleUser` text detection), `core/engine.go` (+12, switch case), and the 2 new test files.
- Focused on the boundary correctness (claudecode-only vs other agents), the prefix detection reliability, the engine reset semantics, and regression risk for the existing text-across-tool-boundary behavior.

What looks good:
- Clean adapter-level isolation: the `"Stop hook feedback:"` prefix detection lives in `agent/claudecode/session.go` only. Other agent adapters (codex/pi/cursor/max/opencode/antigravity/copilot) are untouched and continue to drop user text via the same `continue` path.
- The `continue` after the prefix check matches the existing behavior of ignoring non-`tool_result` user content items — no risk of accidentally processing the feedback text as a regular user message.
- Engine-side reset is minimal and surgical: 7 fields, all directly tied to the rejected-draft accumulation path (`textParts`, `segmentStart`, `silentHold`, `partialText`, `cardAnswerText`, `lastRichCardUpdate`, `lastRichCardLen`). Each reset value is the right "fresh state" value for its variable (e.g. `lastRichCardUpdate = time.Time{}` = "never", which forces the next update to be treated as a new update).
- **Critical regression protection**: `TestProcessInteractiveEvents_NoHookRejectionKeepsTextAcrossToolBoundary` explicitly verifies that without a hook rejection, the existing text-across-hidden-tool-boundary behavior (AAA + tool_use + tool_result → BBB delivers "AAABBB") still works. This is exactly the test you need for a change that touches text aggregation state.
- Test fixture uses **Chinese** Stop hook feedback text ("测试闸门:请把回复改写成只有 BBB 三个字母") to prove the prefix check is byte-level, not language-dependent — robust against future i18n variations.
- PR body is honest about the streaming card limitation: "if a streaming card is enabled, the draft may already have been previewed before the reset; this PR only fixes the final aggregated reply." This is the right call — fixing the streaming card mid-flight is a separate enhancement, not a bugfix blocker.
- The other aggregation paths mentioned in the PR body (`runUnsolicitedReader`, `processCompressEvents`, `HandleRelay`) are correctly verified as unaffected because they prefer `event.Content` from the result event and only fall back to joining when it is empty — so this fix doesn't need to touch them.

Must fix:
- None.

- None. The fix is appropriately scoped.

Optional (P3):
- Consider adding a small unit test that explicitly asserts the prefix string `"Stop hook feedback:"` is byte-equal to what Claude Code currently emits, so a future Claude Code schema change is caught early (e.g. if they change to `"stop_hook_feedback"` or similar).
- Long-term: if streaming card users complain about the brief "draft flash" before the rewrite, a follow-up PR could call `MessageUpdater.UpdateMessage` with an empty card body on `EventHookRejected` to wipe the preview. Not urgent.

Questions:
- None.

Testing / Risk:
- Verified locally:
  - `go test -race -count=1 -run 'TestHandleUserEmitsHookRejected|TestProcessInteractiveEvents_DropsStopHookRejectedDraft|TestProcessInteractiveEvents_NoHookRejectionKeepsTextAcrossToolBoundary' ./agent/claudecode/ ./core/` → all 3 PASS
  - `go test -race -count=1 ./agent/claudecode/` → 14.4s PASS
  - `go test -race -count=1 ./core/` → 49.9s PASS
  - `go vet ./agent/claudecode/ ./core/` → clean
  - `gofmt -l` → clean
- GitHub CI run 34993805762: lint, unit-test, smoke-test, regression-test, performance-test all SUCCESS.
- Remaining risk: minimal. The only forward risk is Claude Code changing the `"Stop hook feedback:"` prefix string in a future version, which would silently re-enable the bug. The Chinese-text test fixture gives some defense against language drift but not against schema-name drift.
- Rollback: single `git revert`. No schema/data migration involved.

Next step:
- Merge. Recommend calling this out in release notes as a Stop-hook-related bugfix so Claude Code 2.1.272 users recognize the upgrade value.

PR #1856, opened 2026-09-16
feat(opencode,feishu): auto-continue after opencode compaction + foldable streaming card panels

gpYang:
## Summary

Three related enhancements for long opencode turns on Feishu streaming cards:

### opencode agent
- **Auto-continue after auto-compaction**: opencode stops after an auto-compaction and waits for the next prompt; previously the turn ended with an empty reply (the compaction summary was suppressed). Now the session is resumed with a bounded (5 rounds) `continue` prompt so the turn completes with a real answer.
- Refactor `opencodeSession` into `launch`/`readProcess`/`readLoop` so follow-up processes feed one events channel.
- Suppress auto-compaction summary text (## Objective format) that previously leaked into chat.
- **Step-buffered final answer**: text parts are buffered per step. Intermediate steps (finish reason `tool-calls`) forward their narration as normal EventText (folded into the card panel on streaming-card platforms); the final step (`stop`) delivers the answer **only** via `EventResult.Content`. Previously the final reply was assembled by joining every text part of the turn, so per-step narration ("剩余改动我会…", rule re-statements, etc.) leaked into the final message users received.
- **Real token usage**: parse the per-step `tokens` field from `step-finish` parts (input/output/reasoning/cache read/write), accumulate turn totals onto `EventResult` and expose them via `ContextUsageReporter` — the reply footer now shows real `out/in/cr` counts and engine logs no longer report `input_tokens=0` for opencode.

### feishu streaming card
- New `StreamingCardPayloadSupporter` interface + structured `ProgressCardPayload` (foldable thinking/tool panels + answer body).
- `BuildStreamingCardPayload` renders compact summaries instead of a raw markdown wall; streaming updates append panel entries and keep the `思考 (N)` / `工具 (N)` panel-title counts live via cardkit element append/patch APIs without re-rendering expanded panels.
- **Panel entry caps** (thinking ≤ 20, tools ≤ 15): a long turn with dozens of tool calls previously blew past the Feishu cardkit element limit (`code 300305 "element exceeds the limit"`), which failed the card finalize and fell back to sending the entire turn (including process text) as a plain message. Capped payloads stay within the element budget and mark themselves truncated ("仅显示最近更新。"). If a card still fails, the fallback reply now contains only the final answer.

## Fixes included
- **media_pipeline test flake** (`tests/release_local/media_pipeline`): the background queued-message worker could still be persisting the session store when `engine.Stop()` returns; the test cleanup now waits for the store to be on disk with no in-flight atomic temp file, so Go's TempDir cleanup no longer races the save ("directory not empty"). Pre-existing upstream flake, fixed here so CI is deterministic.

## Tests
- `agent/opencode/opencode_resilience_test.go` (step buffering, final-step answer-only, token accumulation), `core/engine_card_test.go` (process-text-folded-into-panel regression, race-detector fix for the test stub), `core/progress_compact_test.go` (panel entry caps), `platform/feishu/streaming_card_test.go`, `tests/release_local/media_pipeline` stability fix.
- Verified: `go test ./agent/opencode/`, `go test ./core/` (incl. `TestCUJ`), `go test -race ./agent/opencode/ ./core/`, `go test ./platform/feishu/`, `go test -race ./tests/release_local/media_pipeline/ -count=15` — all pass; full CI (lint/unit/regression/smoke/performance) green.

## Dependencies
- ⚠️ This PR's branch is stacked on #1745 (`message.finalized`, 3 commits: 818d0187/4ca3438a/a434b603) — the `finalized` flag in engine.go is a compile-time dependency. Rebase this branch onto `main` once #1745 merges (the 3 prerequisite commits will drop out of the diff).
- Error-event `ErrorKind` classification (`opencodeErrorKind`) is intentionally **omitted** from this PR — it depends on #1641 (fix/codex-retry), not yet merged. Can be restored as a follow-up once #1641 lands.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -412,7 +412,8 @@ level = "info" # debug, info, warn, error
 #
 # Supported events / 支持的事件:
 #   message.received   — user message arrives / 用户消息到达
-#   message.sent       — agent reply sent / Agent 回复发送完成
+#   message.sent       — legacy pre-delivery agent reply hook / 兼容旧版发送前事件
+#   message.finalized  — final reply send completed / 最终回复发送完成
 #   session.started    — agent session spawned / Agent 会话启动
 #   session.ended      — agent session stopped / Agent 会话结束
 #   cron.triggered     — cron job fires / 定时任务触发
--- a/config.example.toml
+++ b/config.example.toml
@@ -414,6 +414,16 @@ level = "info" # debug, info, warn, error
 #   message.received   — user message arrives / 用户消息到达
 #   message.sent       — legacy pre-delivery agent reply hook / 兼容旧版发送前事件
 #   message.finalized  — final reply send completed / 最终回复发送完成
+#                       (CC_HOOK_CHANGED=true only when the turn actually
+#                       modified the git working tree. Semantics: false does NOT
+#                       mean "no change" — non-git workspaces, fresh repos
+#                       without commits, or git failures always report false.
+#                       Fingerprint windows are serialized per workspace, so
+#                       concurrent turns cannot pollute each other's flag.)
+#                       (CC_HOOK_CHANGED 仅在该轮确实改动 git 工作树时为 true。
+#                       注意：false 不等于"无改动"——非 git 工作区、无提交的
+#                       新仓库或 git 失败时恒为 false。指纹窗口按工作区互斥，
+#                       并发轮次不会互相污染该标志。)
 #   session.started    — agent session spawned / Agent 会话启动
 #   session.ended      — agent session stopped / Agent 会话结束
 #   cron.triggered     — cron job fires / 定时任务触发
--- a/core/engine.go
+++ b/core/engine.go
@@ -455,10 +455,15 @@ type Engine struct {
 	workspaceInitAllowLocalPaths bool
 	workspaceBindings            *WorkspaceBindingManager
 	workspacePool                *workspacePool
-	initFlows                    map[string]*workspaceInitFlow // workspace channel key → init state
-	initFlowsMu                  sync.Mutex
-	sendWorkDirMu                sync.RWMutex
-	sendWorkDirs                 map[string]string // sessionKey → work_dir assigned by send --cwd
+	// workspaceFpMu/workspaceFpLocks serialize the git fingerprint window per
+	// workspace, so a turn's Changed flag cannot be polluted by concurrent turns
+	// (foreground-foreground or foreground-background) in the same directory.
+	workspaceFpMu    sync.Mutex
+	workspaceFpLocks map[string]*sync.Mutex        // workdir -> fingerprint window lock
+	initFlows        map[string]*workspaceInitFlow // workspace channel key → init state
+	initFlowsMu      sync.Mutex
+	sendWorkDirMu    sync.RWMutex
+	sendWorkDirs     map[string]string // sessionKey → work_dir assigned by send --cwd
 
 	// Terminal observation (--observe)
 	observeEnabled    bool
@@ -4757,10 +4762,6 @@ func (e *Engine) runUnsolicitedReader(ctx context.Context, cancel context.Cancel
 
 	events := agentSession.Events()
 
-	// Record workspace git state at reader start so background replies can report
-	// whether the working tree changed during this reader's lifetime.
-	readerStartGitFingerprint := workspaceGitFingerprint(workspaceDir)
-
 	var turnActive bool // true after first event, cleared on EventResult
 	defer func() {
 		if turnActive {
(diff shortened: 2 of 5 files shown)


chenhg5 (commented):
## QA Review — PR #1856

Conclusion: **Approve** (with one optional nit)

### Overall assessment
Solid, well-scoped PR. Three related improvements ship together:

1. **opencode auto-continue after compaction** — heuristic detection of OpenCode's `## Objective / ## Important Details / ## Work State` summary + reasoning-prefix markers. Bounded by `maxCompactionContinuations = 5` with proper stderr/scanner-error fallback to avoid infinite spinning. Bonus: `GetContextUsage()` now returns real numbers (input/output/cache), wired into the existing `ContextUsageReporter` interface so the reply footer just works.
2. **Feishu foldable streaming-card panels** — `BuildStreamingCardPayload` separates per-step text from the answer body; new cardkit `appendCardElements` + `patchCardElementTitle` calls append new entries to existing panels without re-rendering them. A user-expanded panel stays expanded across intermediate updates, exactly as required.
3. **Engine + new hook** — `message.finalized` event with `CC_HOOK_CHANGED` (per-workspace mutex around git porcelain+HEAD fingerprint, 5s timeout). Non-git workspaces correctly report `false` (documented).

### ✅ What looks good
- 32 new tests across `opencode_resilience_test.go`, `streaming_card_test.go`, `engine_card_test.go`, `progress_compact_test.go`. Coverage hits the key paths: lazy card creation, throttling, append-vs-full-card fallback, compaction suppression, step text buffering, final-step answer routing.
- Streaming-card vs compact-progress-writer interaction is handled cleanly via `compactProgressWriter.disable()` — no double cards.
- The 11 reasoning-prefix markers in `isCompactionReasoning` are very narrow; a normal user-facing reasoning block essentially never starts with "let me analyze the conversation to build a comprehensive summary...". Good engineering hygiene.
- Card panel element IDs are fixed constants (`thinking_panel`, `tools_panel`, `updates_panel`) — required for cardkit `target_element_id` to work. Sensible naming.
- API calls go through `withFreshTenantAccessTokenRetry`, same path as existing `StreamRichCardText`.

- **`platform/feishu/feishu.go:4842`** has a gofmt nit:
  ```go
  func renderProgressEntryElement(item core.ProgressCardEntry, lang string) map[string]any {	text := strings.TrimSpace(item.Text)
  ```
  Opening brace and function body are on the same line. CI uses `golangci-lint --new-from-rev` which doesn't run `gofmt`, so it slipped through. One-liner fix: run `gofmt -w platform/feishu/feishu.go`. Not blocking merge.

### 🔵 Optional
- `hooks_test.go` only covers `EventToEnv` for the legacy fields. New `Changed` / `ReplyKind` / `Source` / `Workspace` env vars could use a dedicated test to lock in their formatting.
- `isOpencodeCompactionSummary` detection depends on OpenCode's current template. If OpenCode upstream changes the header combination, detection silently fails. Worth a TODO comment near the function noting "if OpenCode template changes, update these heuristics". The 5-round bound keeps the failure mode sane.

### ❓ Questions
- In `BuildStreamingCardPayload`, the dedup is `cleaned[len(cleaned)-1]` adjacent-dedup. If the agent emits thinking A, tool B, then thinking A again, the second A is kept (correct — different step). But A, A, A → only one kept. Is that intentional? I think yes (consecutive duplicates from a re-emit are noise), but wanted to flag.

### Testing / Risk
- CI: 5/5 PASS (lint / unit-test / smoke-test / regression-test / performance-test). `mergeStateStatus: CLEAN`.
- Author's local `go test` couldn't run in my sandbox (missing `gorilla/websocket`, `robfig/cron` deps due to legacy GOPATH layout). I matched the new test names against the new code paths statically.

### Next step
- **Approve for merge.** If you're feeling conscientious, `gofmt -w platform/feishu/feishu.go` before merge to clean up the nit; otherwise post-merge is fine.

chenhg5 (commented):
Conclusion: Approve (verification of new commits)

Overall assessment:
- The two commits pushed today (290234e5 and d329fa09) are high-quality bug fixes that resolve two real regressions observed in production. CI is fully green; the PR is now in a mergeable state and the maintainer can ship.

Review scope:
- Read commits 290234e5 (`fix(feishu): patch panel titles with partial_element; insert a missing panel`) and d329fa09 (`fix(core): keep the final answer out of the streaming card's process panel`).
- Cross-referenced the new helpers (`buildPatchElementTitleBody`, `addCardElements`, `insertCardElements`, `cardElementAPIStatusError`, `stripAnswerEchoes`, `squashWhitespace`) and their test coverage.
- Did NOT re-review the earlier 4 commits already covered by the 2026-09-16 review.

✅ What looks good:
- Both fixes are root-cause targeted. 290234e5 pins the field name with `buildPatchElementTitleBody` plus a regression test (`TestPatchCardElementTitleBodyUsesPartialElement`) — anyone renaming `partial_element` back to `element` will trip CI. d329fa09 uses a trailing-suffix walk via `stripAnswerEchoes` so it never deletes narration that merely precedes the answer.
- The refactor of `appendCardElements` → `addCardElements(op, targetElementID, elements)` is well-scoped. The new `cardElementAddType` enum (`append` / `insert_before` / `insert_after`) plus `insertCardElements(... after bool)` gives a clean way to add a panel whose lane was empty at render time without forcing a full card re-render (which would reset the user's expand/collapse state).
- `cardElementAPIStatusError` now includes the truncated response body — the old "HTTP status 400" alone was undebuggable from logs. This is a meaningful operator-experience improvement, not just a fix.
- `stripAnswerEchoes` is conservative: it only deletes entries that reproduce the END of the delivered answer, walking backwards and stopping at the first entry that no longer matches. opencode (whose step texts are never part of the answer) is unaffected; agents that stream their final text as `EventText` (claudecode-family) get the echo cleanly removed.
- Regression tests cover the exact production scenario (claudecode-family stream where every EventText was folded into the panel). The test asserts both that the panel keeps the tool entry and that the answer still arrives in the separate final message.
- The 2026-09-16 review's optional gofmt nit (`renderProgressEntryElement` one-liner) was also fixed in 290234e5 — good cleanup.

🚨/🔴 Must fix:

🟠 Should improve:
- None.

🔵 Optional:
- The 2026-09-16 review flagged that `isOpencodeCompactionSummary` depends on OpenCode's current template. If you ever encounter another template change, the failure mode is bounded (5-round cap), but a TODO comment near that function would be useful. Out of scope for this PR.

❓ Questions:
- None.

Testing / Risk:
- Verified evidence: 5/5 CI checks green on the new head (`d329fa09`).
- Static review of the two new helpers against the bug reports in the commit messages confirms the logic matches the description. Tests assert presence/absence of specific content and reach into the recorded stream card to confirm the answer is delivered separately.
- Remaining risk: `stripAnswerEchoes` uses whitespace-collapsed comparison (`squashWhitespace` via `strings.Fields`). If a future adapter streams the answer with a code-block fence that opencode does not, the suffix check could over-match. None of the current adapters produce that pattern, and the opencode branch already filters itself.

Next step:
- Maintainer can merge. These are bug fixes the author pushed in response to live production observation; the regression tests pin the new behavior.

PR #1858, opened 2026-09-16
fix(core,claudecode): compute reply-footer ctx% from real usage and window

### LINKED ISSUE #1672: [BUG] [ctx: ~N%] 回复 footer 使用硬编码 200k 上下文窗口，大窗口模型固定显示 100%
**cc-connect 版本**：1.4.1（main 分支同样存在）

## 问题描述

回复尾部提示 `[ctx: ~N%]` 的计算使用了**硬编码的 200k 上下文窗口**，完全不考虑会话模型的实际上下文窗口。当模型的窗口大于 200k（如 1M 窗口的 `deepseek-v4-flash[1M]`）时，一旦 `input_tokens` 超过 200k，该指标就固定显示 `~100%`，与真实占用严重不符。

## 位置

`core/engine.go`

- L16740：`const modelContextWindow = 200_000 // generic fallback window for heuristic context estimates`
- L16742-16751：`contextIndicatorText(inputTokens int)` 内 `pct := inputTokens * 100 / modelContextWindow`

## 复现

- 模型：`deepseek-v4-flash[1M]`（1M 上下文）
- 某一轮 `input_tokens = 360527`
- 计算：`360527 * 100 / 200000 = 180` → 超过 100 被截断 → 显示 `[ctx: ~100%]`
- 实际占用约 19%（claude `/context` 显示 187k / 1M）

## 影响

任何上下文窗口不是 200k 的模型，该指标都会失真；窗口越大偏差越大，且 input 超过 200k 后恒为 100%，失去参考价值。

## 建议修复

调用点（约 L5557-5565，`footerContext` 被硬编码路径覆盖处）已经能拿到会话真实窗口 `replyFooterSessionContextUsage(state.agentSession)`，建议把真实 `ContextWindow` 传入 `contextIndicatorText`：

```go
func contextIndicatorText(inputTokens, contextWindow int) string {
	if inputTokens <= 0 || contextWindow <= 0 {
		return ""
	}
	pct := inputTokens * 100 / contextWindow
	if pct > 100 {
		pct = 100
	}
	return fmt.Sprintf("[ctx: ~%d%%]", pct)
}
```

另外两点补充：

- `event.InputTokens` 是 API 上报的 input（可能含缓存命中 token），数值可能高于会话实际跟踪的上下文；可考虑优先使用 `replyFooterSessionContextUsage` 里的 `UsedTokens`。
- `replyFooterContextText`（L7381）已经正确使用 `usage.ContextWindow`，只是被 L5560 的 `contextIndicatorText` 硬编码路径覆盖了；修复时可以参考它。

## 附：真实占用 vs 指标

| 项 | 值 |
|---|---|
| claude `/context` 实际占用 | 187.3k / 1M = 19% |
| cc-connect `[ctx: ~N%]` 显示 | ~100%（错误） |

tmjwjx:
## Summary

The reply footer's `[ctx: ~N%]` marker was computed as `event.InputTokens / 200_000`. Both operands were wrong: the numerator dropped `cache_read_input_tokens` and `cache_creation_input_tokens`, and the denominator ignored the real `ContextWindow` the session already reports. A second, independent defect kept the accurate snapshot from ever existing on Anthropic-compatible gateways that send an all-zero `message_start.usage`, so `GetContextUsage()` returned `nil` for the whole session.

Two commits, one per layer: `core/` now consumes the snapshot correctly, and `agent/claudecode/` makes sure the snapshot exists.

## Type of change

- [ ] Documentation only

### Commit 1 — `fix(core): compute reply-footer ctx% from real usage and window`

`contextIndicatorText` now prefers the session's `ContextUsage` for **both** operands (`UsedTokens` over `ContextWindow`), which is exactly what the sibling `buildClaudeStatusLineFooter` has done since #1204. When the session exposes no usage, the numerator falls back to `event.InputTokens + CacheReadInputTokens + CacheCreationInputTokens` and the denominator to the existing generic `modelContextWindow`.

The `modelContextWindow = 200_000` fallback is **kept as-is** — it is a deliberate generic default, and this PR only stops it from overriding a window the session has already reported.

The accurate snapshot was in fact already fetched one line above the call site for `replyFooterContextText` and then discarded. This looks like a missed migration rather than a design choice: #1761 moved the auto-compress trigger onto that snapshot while stating *"this is the same number used for the reply footer's 'ctx N%' indicator"* — which was not yet true of this code path.

The `sdkPlausible` gate is summed the same way. Previously `event.InputTokens >= 100` dropped exactly the turns this fix targets: with prompt caching a turn can report a couple of dozen `input_tokens` while `cache_read` carries hundreds of thousands, and such turns never reached the indicator at all.

No new agent-specific knowledge is added to `core/`: the window arrives through the `ContextUsage` snapshot, never from parsing model ids.

### Commit 2 — `fix(claudecode): report context usage when assistant events carry no usage`

Anthropic populates `message_start.usage`, so every stream-json `assistant` event carries its sub-call's prompt size and `handleAssistant` records the last one — the faithful "context used right now".

Anthropic-compatible gateways are not obliged to do the same. Measured on `open.bigmodel.cn/api/anthropic` with Claude Code 2.1.273:

```
assistant.usage   = {"input_tokens": 0, "output_tokens": 0}
result.usage      = {"input_tokens": 36662, "cache_read_input_tokens": 768, ...}
result.modelUsage = {"glm-5.3-flash[1m]": {"inputTokens": 36662,
                     "cacheReadInputTokens": 768, "contextWindow": 1000000, ...}}
```

A plain HTTPS request to that endpoint, with Claude Code out of the picture entirely, returns the same zeroed `message_start` and puts the real numbers in `message_delta` — so this is the gateway, not Claude Code, and not a specific provider. LiteLLM has the same class of bug (BerriAI/litellm#31069), and Anthropic's own [streaming docs](https://platform.claude.com/docs/en/build-with-claude/streaming) show populated `input_tokens` / `cache_read_input_tokens` in `message_start`.

`handleAssistant`'s `if used > 0` guard therefore never fired, `lastUsage` stayed `nil` for the entire session, and two things silently degraded: the footer fell back to the generic window, and the auto-compress trigger added in #1761 fell back to the text heuristic it was meant to replace.

`handleResult` now fills `lastUsage` from the result event when — and only when — no assistant event supplied usage during the turn, preferring the `modelUsage` entry for the session's own model because it also carries that model's **real context window**. The aggregate overstates long tool loops (`cache_read` is summed per sub-call), which is precisely why it stays a fallback rather than becoming the primary source.

One trap worth calling out for anyone touching this code: the gate is a **per-turn flag**, not a "`lastUsage` is still empty" check. `lastUsage` is never reset, so the latter fills the snapshot once and then pins it to the first turn's numbers for the rest of the session. `TestHandleResult_ZeroedAssistantUsageRefreshesEveryTurn` fails against that variant. Compaction results are mid-turn and do not clear the flag.

- `TestProcessInteractiveEvents_ContextIndicatorUsesSessionUsageAndWindow` in `core/engine_test.go` — end-to-end through `processInteractiveEvents`: a session reporting `UsedTokens=37430 / ContextWindow=1_000_000` must render `[ctx: ~3%]` in the delivered footer.
- `TestProcessInteractiveEvents_ContextIndicatorCountsCacheTokens` in `core/engine_test.go` — a turn with `InputTokens=42, CacheReadInputTokens=150_000` and no session usage must render `[ctx: ~75%]` rather than be dropped by the gate.
- `TestContextIndicatorText` in `core/engine_test.go` — table-driven: session-usage preference, event fallback, `TotalTokens` path, clamping, the unchanged bare-input behaviour (`28000 → 14%`), and the empty cases.
- `TestHandleResult_ZeroedAssistantUsageRefreshesEveryTurn` in `agent/claudecode/session_test.go` — two consecutive turns of a gateway that zeroes assistant usage; turn 2 must report its own numbers.
- `TestHandleResult_ModelUsageReportsRealContextWindow` in `agent/claudecode/session_test.go` — the 1M window reaches `core/` through the snapshot.
- `TestHandleResult_AggregateNeverOverwritesAssistantUsage` in `agent/claudecode/session_test.go` — the reverse guarantee: on Anthropic's own endpoint the per-sub-call snapshot still wins, on turn 2 as well as turn 1, even though the result event's `cache_read` is summed to 8M/9M.
- `TestHandleResult_FallsBackToTopLevelUsageWithoutModelUsage` in `agent/claudecode/session_test.go`
- `TestParseClaudeModelUsage` in `agent/claudecode/session_test.go` — exact model match wins over a larger helper-model entry; case-insensitive; absent/zeroed maps return `ok=false`.

Existing tests kept passing unchanged, including `TestHandleAssistantCapturesPerSubCallUsage`, `TestAutoCompress_UsesRealUsageWhenAvailable`, `TestAutoCompress_FallsBackToHeuristicWhenNoUsage`, and the two footer end-to-end tests that pin the 200k fallback behaviour.

- Regression test names: `TestProcessInteractiveEvents_ContextIndicatorUsesSessionUsageAndWindow`, `TestProcessInteractiveEvents_ContextIndicatorCountsCacheTokens`, `TestHandleResult_ZeroedAssistantUsageRefreshesEveryTurn`
  - [x] Reverted the fix locally; the regression test failed as expected.

Recorded output from the reverted runs:

```
# core/engine.go reverted to HEAD, new tests kept
final reply = "answer\n\n*[ctx: ~18%] · glm-5.1 · ~/code/ctx-window*",
        want "answer\n\n*[ctx: ~3%] · glm-5.1 · ~/code/ctx-window*"
final reply = "answer\n\n*glm-5.1 · ~/code/ctx-cache*",
        want "answer\n\n*[ctx: ~75%] · glm-5.1 · ~/code/ctx-cache*"

# handleResult gate replaced with a "UsedTokens <= 0" check
turn 2 UsedTokens = 36768, want 350000 (stale 36768 means lastUsage froze on turn 1)
```

- [x] If the change alters an existing user-visible flow, the corresponding CUJ test was updated (or a new CUJ added) to cover the new behavior.

One note on the CUJ run: `TestCUJ_A3_ImageReachesAgent` and `TestCUJ_A5_FileReachesAgent` fail intermittently with `TempDir RemoveAll cleanup: unlinkat ...: directory not empty`. This reproduces on `main` at the same rate (2 of 3 runs with `-count=1` on both `main` and this branch, alternating between the two tests), is a cleanup artifact rather than an assertion failure, and is unrelated to token accounting. Happy to open a separate issue for it if useful.

The footer percentage becomes accurate for sessions that report a real context window. On a 1M-window model a 37k prompt now shows `[ctx: ~3%]` instead of `[ctx: ~18%]`. Turns whose prompt is mostly cache-read now show a percentage at all instead of no marker.

Sessions that report no usage keep the previous behaviour exactly, including the 200k generic window, with the single difference that cache tokens are counted in the numerator.

No config, no new options, no i18n strings.

- [x] `go build ./...` passes
- [x] `go test ./...` passes (with `-race` if touching concurrency)
- [x] i18n strings have all-language translations (if any new user-facing text) — no new user-facing text

`go test -race ./agent/claudecode/ ./core/` passes; commit 2 adds state under the existing `usageMu`.

## Related

Prior art found while preparing this PR, listed for the reviewer's convenience — there is overlap and I am happy to close this in favour of any of them, or to rebase onto them:

- Issue: #1672 — same hardcoded-200k footer bug, reported from the saturation angle (large-window models pinned at 100%).
- Related PR: #1707 — fixes #1672 by requiring a real window and returning `""` without one. This PR differs by keeping the `modelContextWindow` generic fallback instead of suppressing the indicator, and additionally fixes the `sdkPlausible` gate, which #1707 leaves at `event.InputTokens >= 100`.
- Issue: #1846 / Related PR: #1845 — same zeroed-usage class of provider, measured on MiniMax-M3, fixed by a bounded tail read of the Claude Code JSONL transcript. That approach is more accurate than this one on long tool loops, since the transcript records per-call prompts while the result aggregate sums them. This PR is deliberately narrower: no filesystem dependency on Claude Code's storage layout, using `result.modelUsage` instead. The two are compatible — `modelUsage` would make a reasonable last-resort fallback for #1845 when the transcript is unreachable (non-standard `CLAUDE_CONFIG_DIR`, containers, remote agents). The direct-to-endpoint measurement above may also be useful there: it shows the root cause is the gateway's `message_start.usage`, which makes this a provider-class bug rather than a MiniMax-M3-specific one.
- Issue: #1823 / Related PR: #1827 — requests a configurable context window because 200k is hardcoded. Worth noting that `result.modelUsage[<model>].contextWindow` already reports the real window (verified: `1000000` for `glm-5.3-flash[1m]`), so for the claudecode agent the value can be read rather than configured.


chenhg5 (commented):
## QA Review — PR #1858

Conclusion: **Approve**

### Overall assessment
Excellent targeted bugfix. Two independent defects, two independent commits, full test coverage including regression tests that the author explicitly verified would catch the bug if reverted. PR description is one of the cleanest I've read — it cites Anthropic's own streaming docs and the equivalent LiteLLM bug (#31069), explains the *trap* of using `UsedTokens <= 0` instead of a per-turn flag, and records revert-mode run output.

### ✅ What looks good
- **8 new tests** lock in the three endpoint shapes: Anthropic native (assistant usage populated, per-sub-call wins), Anthropic-compatible gateway with all-zero `message_start.usage` (modelUsage fallback), and the no-modelUsage path (top-level `usage` fallback). The two `processInteractiveEvents` tests do full end-to-end through the engine, asserting the exact footer string.
- `parseClaudeModelUsage` correctly handles multi-model billing: exact model match wins over largest-prompt, case-insensitive, absent/zeroed maps return `ok=false`.
- `assistantUsageSeen` as a per-turn flag (not a `UsedTokens <= 0` check) is the right design — the author called out exactly why in the description, and `TestHandleResult_ZeroedAssistantUsageRefreshesEveryTurn` pins it.
- `modelContextWindow = 200_000` is preserved as the deliberate generic fallback, not accidentally overridden. Sessions that report no usage keep prior behavior exactly.
- The `sdkPlausible` fix in `core/engine.go` matches Anthropic's usage semantics (input/cache_read/cache_creation are disjoint subsets of the real prompt), so a prompt-cache-heavy turn now shows `[ctx: ~75%]` instead of being silently dropped by the gate.
- Backward compatible: only behavior change for sessions that report no usage is "cache tokens count in the numerator" — strictly more accurate.
- I confirmed `isCompaction` (line 909 of `handleResult` in the new file) is the local variable the new code on line ~982 reuses — no undefined-variable risk.

- **`go test ./core/ -run TestCUJ` intermittent failures**: author reports `TestCUJ_A3_ImageReachesAgent` / `TestCUJ_A5_FileReachesAgent` fail 2/3 runs on both `main` and this branch with `TempDir RemoveAll cleanup: unlinkat ...: directory not empty`. Cleanup artifact, not assertion failure. Worth a separate issue (not blocking this PR).

### 🔵 Optional
- Other agents (codex, opencode, cursor, kimi) could benefit from a similar modelUsage-aware `GetContextUsage()` fallback in a follow-up PR. Not blocking; this PR is correctly scoped to claudecode where the bug was demonstrated.
- A release note line under "user-visible behavior change" would help users understand "ctx% jumped from 18% to 75%" on cache-heavy prompts. Not blocking.

### ❓ Questions
- (None — the PR description answered everything I would have asked.)

### Testing / Risk
- CI: 5/5 PASS. `mergeStateStatus: CLEAN`.
- Author manually ran the new tests against a reverted `core/engine.go` and a replaced `assistantUsageSeen` gate; both failures reproduced as expected (recorded output in the PR description).
- Diff is +602 / -20 across 4 files — small, focused, easy to review.

### Next step
- **Approve for merge.** No blockers.

PR #1652, opened 2026-08-07
feat(codex): steer busy follow-ups into the active turn

ternary-chen:
## Summary

Add an optional agent-session steering capability and implement it for the Codex app-server backend.

Projects may opt into `busy_message_mode = "steer"` so plain-text follow-ups received during an active turn are appended to that turn instead of waiting in the FIFO queue. Queue mode remains the default for backward compatibility.

Steering failures and unsupported sessions fall back to the existing queue. Messages containing images or files also remain queued. `/ps` now requires the steering capability and no longer risks starting a concurrent second turn via `AgentSession.Send`.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing behavior to change)
- [ ] Documentation only
- [ ] Internal refactor / chore (no user-visible change)

### Automated tests added in this PR

- `TestAppServerSession_SteerTurnUsesExpectedActiveTurn` in `agent/codex/appserver_session_test.go` — sends `turn/steer` with the active thread and expected turn ID.
- `TestAppServerSession_SteerTurnRejectsMissingActiveTurn` — rejects stale or idle steering without writing an RPC request.
- `TestLoad_ParsesBusyMessageMode` and `TestLoad_RejectsInvalidBusyMessageMode` in `config/config_test.go`.
- `TestHandleMessage_BusySteerableSessionSteersPlainText` in `core/engine_test.go` — steers, persists history, advances the watermark, and acknowledges the user.
- `TestHandleMessage_BusySteerableSessionDefaultsToQueue` — preserves queue as the default.
- `TestHandleMessage_SteerFailureFallsBackToQueue` — does not lose a follow-up when steering fails.
- `TestHandleMessage_BusyAttachmentRemainsQueued` — keeps attachment delivery semantics unchanged.
- `TestCmdPs_BusySessionWithoutSteeringCapabilityFailsSafely` — prevents `/ps` from starting a concurrent turn through `Send`.
- Updated `TestCmdPs_BusySession_InjectsToAgent` to assert `SteerTurn`.
- `TestCUJ_A8_BusyFollowUpSteersCurrentTurn` in `core/cuj_test.go` — covers the user journey from initial task through busy follow-up and history.

Targeted unit tests, the complete core CUJ suite, and targeted race tests pass locally:

```text
go test ./agent/codex ./config -run 'SteerTurn|BusyMessageMode' -count=1
[...]

PR #1861, opened 2026-09-17
fix(pi): deliver /ps and /btw supplements during RPC turns

TongLing916:
## Summary

Pi rejects a plain RPC `prompt` while it is already streaming. `/ps` and its `/btw` alias currently send that command and acknowledge the stdin write, so a supplement can be reported as delivered without reaching the active task.

Send `prompt` with `streamingBehavior: "steer"` in the Pi adapter. Pi then queues guidance during an active run and starts processing normally if the remote run has just ended. Text-only supplements also preserve the attachment directory that the running task may still be reading. The shared `Send` API and other agent adapters are unchanged; there are no configuration or breaking changes.

## Type of change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation only

- `TestSendRPC_SteersWithoutChangingPrompt`: checks the steering option, JSONL framing, text, and attachment references.
- `TestSendRPC_PropagatesWriteError`: retains write-error reporting.
- `TestSend_TextSupplementPreservesInFlightAttachments`: prevents `/ps` with an empty message ID from deleting active attachments.
- `TestPiRPCSend_StartsIdlePrompt` and `TestPiRPCSend_SteersWhileStreaming`: exercise a real adapter subprocess against a fake Pi RPC peer.
- `TestPiRPCPs_AfterRemoteTurnEnds`: uses the real engine and blocks final platform delivery to cover remote-idle/local-busy handling.
- `TestCUJ_A9_PsAndBtwSupplementRunningTask`: starts a task, supplements it, verifies the visible response, and checks idle rejection for both aliases.

The fake peer requires no Pi installation, credentials, or model calls. All new fixtures use generic examples.

Passed locally:

- `go build ./...`
- `go vet ./...`
- Incremental `golangci-lint` v2.11.4 using Go 1.26.3, including the new test files: zero issues.
- Pi package tests, including its complete race-detector run.
- `go test ./core -run TestCUJ -count=1`
- `go test -race ./core -run 'TestCmdPs_|TestCUJ_A9_PsAndBtwSupplementRunningTask' -count=1`

Broad validation was also run, with existing baseline failures kept visible:

- `go test ./...`: all packages passed except `daemon/TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable` on macOS (`Status().Running = false`). The same test fails on unmodified `main` at `757b4df0`.
- `go test -race ./agent/pi ./core`: Pi passed; core hit `TestCUJ_A5_FileReachesAgent` temporary-directory cleanup (`directory not empty`). This failure also reproduces on unmodified `main` under `-race -count=30`; no data race was reported.

These unrelated test fixes are outside this PR. A live-provider round trip was not performed. The latest published beta source (`v1.5.1-beta.1`) also uses the plain RPC prompt without `streamingBehavior`.

- Regression tests: `TestPiRPCSend_SteersWhileStreaming`, `TestSendRPC_SteersWithoutChangingPrompt`, and `TestSend_TextSupplementPreservesInFlightAttachments`.
- [x] Verified on the unmodified `main` implementation before applying the fix: the active supplement produced no result, the steering field was missing, and active attachment files were deleted. All three pass with the fix.

- [ ] No CUJ touched
- [x] A — basic conversation
- [ ] B — session lifecycle
- [x] C — agent execution control
- [ ] D — security & permissions
- [ ] E — scheduled tasks
- [ ] F — config switching
- [x] G — error handling & robustness
- [ ] I — UI rendering correctness

- [x] A new CUJ covers the user-visible flow for both `/ps` and `/btw`.

With Pi RPC mode (`rpc = true`), `/ps include unit tests` or `/btw include unit tests` can supplement a running task. If Pi finishes before the bridge completes final delivery, the supplement is processed as a follow-up rather than left in an idle steering queue. An idle bridge session still rejects the command. Existing JSON-mode sending is preserved.

The English and Chinese usage guides describe these semantics. A delivery acknowledgement does not claim the model has finished processing the request.

- [x] `go build ./...` passes.
- [ ] `go test ./...` and the broad core race run pass without baseline failures — see the reproduced failures above.
- [ ] All AGENTS.md pre-commit gates are green — blocked by those existing broad-suite failures.
- [x] No new hardcoded platform/agent names in `core/` production code.
- [x] No new runtime i18n messages; English and Chinese documentation updated.
- [x] No secrets or credentials in source.

## Related

Checked open and closed Issues/PRs for Pi, `/ps`, `/btw`, `steer`, and `streamingBehavior` before opening this PR; no equivalent Pi adapter fix was found.

- #1817 addresses the analogous Codex adapter issue; this PR changes only Pi production code.
- #1652 proposes broader Codex/core steering capabilities; this PR preserves the existing shared dispatch contract.
- #619 documents the command's original intent.


chenhg5 (commented):
## Conclusion: Approve

Overall assessment: Targeted Pi-adapter bugfix with proper regression coverage and honest baseline reporting. Safe to merge.

Review scope:
- Reviewed `agent/pi/session.go` (+10/-3), new tests in `agent/pi/{rpc_steering,steer}_test.go`, `core/cuj_test.go` (+74), and EN/ZH docs.
- Focused on correctness (steering field semantics, attachment-preservation guard), test methodology, and scope discipline.
- Head: `3d8033af3ca81e97e3a6a44b8ee693d88375e3be`. Base: `757b4df059d537eacb84dd0c955b5e7ede205420`. CI Run 35202197024 (5/5 green).

✅ What looks good:
- Two surgical production changes (one field added, one condition added) — total ~6 lines including comments. No other adapter touched, no shared Send API changed, no config breaking.
- `"streamingBehavior": "steer"` is the right mechanism for Pi RPC: queues guidance during an active run, starts a fresh prompt if the remote turn just ended. Matches Pi's documented behavior.
- Guarding `cleanAttachments` with `len(images) > 0 || len(files) > 0` is a clean fix for a real race: text-only supplements (`/ps` with no message ID) were wiping attachment dirs the running task might still be reading.
- Test coverage is appropriate and well-named: `TestSendRPC_SteersWithoutChangingPrompt` asserts the new field; `TestSendRPC_PropagatesWriteError` keeps error semantics; `TestSend_TextSupplementPreservesInFlightAttachments` is the exact regression guard; the real-adapter `TestPiRPCSend_StartsIdlePrompt` / `_SteersWhileStreaming` and engine-level `TestPiRPCPs_AfterRemoteTurnEnds` cover the cross-component paths.
- CUJ test `TestCUJ_A9_PsAndBtwSupplementRunningTask` covers both `/ps` and `/btw` aliases — matches the user's mental model.
- Author honestly reported baseline failures (`daemon/TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable` on macOS, `core/TestCUJ_A5_FileReachesAgent` cleanup race) and verified they reproduce on unmodified main `757b4df0` — this saved reviewer time.
- EN + ZH docs updated in lockstep; the new wording ("delivery confirmation means the bridge sent the request, not that the agent has finished processing it") is exactly the right user-facing framing.
- PR title describes the behavior, not the file (`fix(pi): deliver /ps and /btw supplements during RPC turns`) — easy to grep later.

✅ CI (Run 35202197024):
- lint ✅, unit-test ✅, smoke-test ✅, regression-test ✅, performance-test ✅

🔵 Optional / not blocking:
- `core/cuj_test.go` adds +74 lines alongside `agent/pi/` tests. The PR description could have mentioned why cross-component tests live in `core/cuj_test.go` rather than `agent/pi/` (presumably because they use the real engine). Minor doc nit; doesn't affect correctness.
- The `streamingBehavior` field will be silently ignored by old Pi versions that don't understand it — but the author notes those older versions were already broken in this scenario, so the fix strictly improves the situation.

Testing / Risk:
- Verified by author: `go build ./...`, `go vet ./...`, `golangci-lint` clean, `go test -race` clean on relevant packages, CUJ green.
- Verified by QA via file read: production diff matches PR description; test names map 1:1 to fix surfaces; docs match between EN and ZH.
- Remaining risk: none observed. Other adapters unchanged; rollback is `git revert`.

Next step:
- Maintainer (Glenn) to merge on GitHub when ready. No further QA action required.

chenhg5 (approved):
## Conclusion: Approve

Clean Pi-adapter bugfix with surgical scope discipline (only `agent/pi/` modified; shared `Send` API and other agent adapters untouched). The two production changes are minimal but address the root cause precisely: `streamingBehavior: "steer"` lets Pi queue the supplement during an active run or start a new turn if the remote run has just ended, and the `cleanAttachments` guard prevents text-only `/ps` from wiping in-flight attachments that the running task might still be reading. The regression test coverage is exactly the right shape — the in-flight attachment test pins the second bug directly, and the real-subprocess test (`TestPiRPCPs_AfterRemoteTurnEnds`) covers the harder remote-idle / local-busy race with a blocked platform delivery scenario.

**Note on prior review**: chenhg5's `COMMENTED` review was submitted on commit `3d8033af3ca81e97e3a6a44b8ee693d88375e3be`, but the current head is `80f4dcd7f3e06474726edbec793a779917342283` (the PR was force-pushed/rebased since). My review is against the current head. The diff content matches the description in chenhg5's review, so the Approve conclusion still applies.

### What looks good

- **Scope discipline**: only `agent/pi/session.go` production code changed (+10/-3). Shared `Send` API in `core/interfaces.go` and other agent adapters (claudecode / codex / opencode / cursor / kimi) are untouched. The behavioral change is contained to Pi users.
- **`streamingBehavior: "steer"` is the right mechanism**: matches Pi's documented behavior — queues guidance during an active run, starts a new turn if the remote turn just ended. Single JSON field, no protocol change.
- **`cleanAttachments` guard is well-placed**: `len(images) > 0 || len(files) > 0` ensures text-only supplements (no message ID, no attachments) preserve any in-flight attachment directory. The comment explains *why* the guard exists — without it, a future refactor could easily re-introduce the bug.
- **Self-as-helper subprocess test pattern**: `TestPiRPCSteeringHelperProcess` uses the `CC_CONNECT_PI_STEERING_HELPER=1` env-var switch to make the test binary act as a fake Pi RPC peer. No external Pi install, no credentials, no model calls — yet exercises the real subprocess, JSONL reader, and session lifecycle. This is a high-fidelity isolation pattern that's reusable for other agent adapters.
- **Two regression tests targeting the two distinct bugs**: `TestSend_TextSupplementPreservesInFlightAttachments` (attachment wipe) and `TestPiRPCPs_AfterRemoteTurnEnds` (race: Pi turn ended while bridge still delivers). Each asserts the exact invariant that would have caught its respective bug.
- **CUJ A9 with both command spellings**: `TestCUJ_A9_PsAndBtwSupplementRunningTask` runs `/ps` and `/btw` as sub-tests. Tests idle-rejection explicitly, which is the third important behavior (you can supplement a running task, but not an idle session).
- **Honest baseline reporting**: PR body explicitly calls out `daemon/TestLaunchdStatusUsesUserDomainWhenGUIDomainUnavailable` failure on macOS — a known failure on `main` 757b4df0, not introduced by this PR. Good discipline for reviewers.
- **Bilingual docs**: both `usage.md` and `usage.zh-CN.md` updated with the same content — users in both languages see the same guidance.

### 🔵 Optional (not blocking)

- **P3 — CHANGELOG note**: The behavioral change (Pi `/ps` and `/btw` now actually reach the running task vs. silently dropped before) is user-visible and deserves a release note line. Even though the docs are updated, users who haven't read the docs yet won't know about the fix until they hit it. A one-line CHANGELOG entry under "Fixes" would surface it.
- **P3 — audit other agents for similar streaming-rejection patterns**: This PR fixes Pi specifically, but the same pattern (RPC `prompt` rejected during streaming → supplement dropped) could exist in other adapters. Worth a quick scan of `agent/claudecode/`, `agent/codex/`, `agent/opencode/`, `agent/cursor/`, `agent/kimi/` to see if any have a similar mechanism that needs analogous handling. If yes, file a follow-up issue rather than expanding scope here.
- **P3 — helper subprocess cross-platform risk**: The `os.Getenv` switch + self-subprocess approach relies on the test binary being executable and the env var propagating correctly. Works on Linux/macOS in practice; on Windows or restricted CI environments it could flake. The 3s timeouts are generous, but if you ever see flakiness, consider a build tag to exclude the helper tests on problematic platforms.

### Testing / Risk

- **Verified**: CI 5/5 green on `80f4dcd7`. Maintainer chenhg5's prior Approve review (against an earlier head) still applies to the current state — the diff content is the same. Real subprocess tests + isolated unit tests + CUJ coverage all in place.
- **Remaining risk**: P3 items above (CHANGELOG, audit other agents, helper subprocess cross-platform). No P0/P1 risk identified.

### Next step

Merge-ready. Consider the CHANGELOG line for the behavioral change, but no code changes are required.

PR #1811, opened 2026-09-09
fix(cursor): parse result event usage so reply footer renders (#1810)

chenhg5:
## Summary

Cursor Agent CLI 2026.09.02-c22c1a3 emits a `usage` block on stream-json result events with camelCase keys (`inputTokens` / `outputTokens` / `cacheReadTokens` / `cacheWriteTokens`). The cursor adapter's `handleResult` ignored this object, so the engine's turn-complete log always reported 0 tokens and the reply footer introduced by #701 was suppressed by its workdir-only guard.

Fixes #1810.

## Changes

- `agent/cursor/session.go`
  - Extract `parseCursorUsage` helper that mirrors `agent/claudecode/parseClaudeUsage` but uses the right camelCase field names.
  - Populate `InputTokens`, `OutputTokens`, `CacheReadInputTokens` and `CacheCreationInputTokens` on the resulting `core.Event`.
  - JSON numbers decode as `float64`, so convert via `int(x)`.
  - Missing or non-map `usage` yields zero counts so backward compat is preserved.
- `agent/cursor/handle_result_test.go` (new)
  - 9 unit tests covering `parseCursorUsage` (normal camelCase, missing usage, non-map usage, non-numeric fields, partial usage) and `handleResult` integration (all fields populated, missing usage, non-map usage, non-blocking on full channel).
  - All pass under `-race`.

## Notes

- No `core/` changes; `core.Event` field names and behaviour are untouched.
- The reply footer will now render naturally for Cursor turns once this lands, without modifying #701's guard.
- Mirrors the local-helper pattern from `agent/claudecode/session.go` (`parseClaudeUsage`) because the cursor package cannot import across agent packages.

## Verification

Local results before PR creation:

- `go build -tags no_web ./...` — clean
- `go test -tags no_web -race -count=1 ./agent/cursor/... ./core/...` — both `ok` (cursor 7.676s, core 48.289s)
- `gofmt -l agent/cursor/session.go agent/cursor/handle_result_test.go` — clean
- `go vet -tags no_web ./agent/cursor/...` — clean

GitHub Actions CI 5/5 must pass before this is handed off to QA.

## Risk

Low. The change is fully contained in `agent/cursor/` and only adds a code path that was previously dead. Backward compatibility is preserved: missing or malformed usage yields zero token counts (same as before the fix). No new public API, no `core/` modifications, no behaviour change for non-Cursor agents.

Issue #1810, opened 2026-09-09
cursor: parse result.usage so reply footer is not suppressed


Tinet-wangll:
## Summary

Cursor Agent CLI already emits per-turn token usage in the terminal `result` event of `--output-format stream-json`. cc-connect's Cursor adapter (`agent/cursor/session.go` `handleResult`) ignores that object, so `turn complete` always logs `input_tokens=0 output_tokens=0`.

With `reply_footer = true` (default), Feishu (and other platforms) still show **no footer**, because #701 suppresses the footer when only workdir is known.

Claude Code already parses `usage` into `core.Event.{InputTokens,OutputTokens,...}` in `handleResult`. Cursor should do the same (camelCase keys).

## Environment

- cc-connect: **v1.5.0** (`17c61062`, 2026-08-16) — also reproduced by reading current `main` `handleResult`
- agent: `cursor`
- Cursor Agent CLI: `2026.09.02-c22c1a3` (`agent` / `cursor-agent`)
- platform: Feishu (`progress_style = "card"`, `enable_feishu_card = true`)
- display: `reply_footer = true`, `show_context_indicator = true`, `show_workdir_indicator = true`

### 1. CLI already reports usage (no extra Cursor CLI flag)

`agent -p --trust --output-format stream-json "只回复一个字：好"` ends with:

```json
{
  "type": "result",
  "subtype": "success",
  "duration_ms": 7656,
  "duration_api_ms": 7656,
  "is_error": false,
  "result": "好",
  "session_id": "7a9963b5-a319-41aa-938f-8a1f9171c3d8",
  "request_id": "fd1d1db5-da3c-45ee-8cae-6eaee99fcb17",
  "usage": {
    "inputTokens": 5749,
    "outputTokens": 25,
    "cacheReadTokens": 9728,
    "cacheWriteTokens": 0
  }
}
```

`system` init also has `"model": "Auto Balance"` (and `session_id`).

Note: `~/.cursor/cli-config.json` `display.showStatusLine*` only affects the interactive TUI, not stream-json.

Official CLI docs currently omit `usage` on the `result` event, but the live CLI emits it.

### 2. cc-connect never copies those fields

`agent/cursor/session.go` `handleResult` (current `main`):

```go
evt := core.Event{Type: core.EventResult, Content: content, SessionID: cs.CurrentSessionID(), Done: true}
```

No `InputTokens` / `OutputTokens` / cache fields.

[...]

PR #1864, opened 2026-09-17
fix(cursor): expose GetContextUsage so reply footer shows out/in/cw/cr

Tinet-wangll:
## Summary

Follow-up to #1811 / #1810.

#1811 copies Cursor `result.usage` onto `core.Event`, so `turn complete` is no longer all zeros and the #701 workdir-only guard lets a footer appear. The engine still only renders the compact line:

`[ctx: ~N%] · <workdir>`

The CCD-style line (`model · out N · in N cw N cr N · ctx N%`) is built by `buildClaudeStatusLineFooter`, which requires `session.(ContextUsageReporter)` and `ContextWindow > 0`. Claude Code already implements this; Cursor does not.

This PR stays in `agent/cursor/` (no `core/` changes):

- Store `system.model` on the session and implement `GetModel()`.
- After `parseCursorUsage`, keep a `lastUsage` snapshot and implement `GetContextUsage()` (clone + mutex, same as Claude Code).
- `ContextWindow = 200_000` — Cursor CLI does not report a window; this matches `core.contextIndicatorText` so ctx% stays consistent with the compact footer users already see.

## Files

- `agent/cursor/session.go`
- `agent/cursor/handle_result_test.go`

## Tests

- `go test -tags no_web -count=1 ./agent/cursor/...`
- New/extended: `GetContextUsage` after a realistic usage payload; nil when usage missing/non-map; `GetModel` from system init; compile-time `var _ core.ContextUsageReporter = (*cursorSession)(nil)`

## Notes / out of scope

- Cache both zero: `buildClaudeStatusLineFooter` still returns "" (existing `core/` gate). Compact `[ctx] · workdir` remains. Not changed here.
- Verified locally on Feishu + Cursor Agent CLI after #1811: tokens parse; without this PR the footer stays compact-only.

Fixes the remaining UX gap of #1810 after #1811.


chenhg5 (changes requested):
## CI Gate — request for CI run before full review

Hi @Tinet-wangll — thanks for the focused cursor adapter change and the clear "follow-up to #1811" framing. I want to review this properly, but I can't until the CI pipeline reports a result.

What I observed:
- This PR's head `e93e590a` has **0 check runs and 0 statuses** in the combined status (state: `pending`, `total_count: 0`).
- `gh pr checks 1864` returns `no checks reported on the 'fix/cursor-context-usage-reporter' branch`.

Likely causes (any one of these would explain it):
1. The `fix/cursor-context-usage-reporter` branch isn't picked up by `.github/workflows/*.yml` (most workflows use `on.push` / `on.pull_request` with branch filters — a branch matching the filter is required).
2. The workflow run was skipped because the branch hasn't seen activity that triggers a fresh run.
3. A push happened without completing the pipeline and there's no historical run.

Why this blocks full review:
- This PR implements `core.ContextUsageReporter` on `cursorSession`, mutates `lastUsage` from `parseCursorUsage`, and exposes a hard-coded `ContextWindow = 200_000`. These touch shared engine contracts. I want to see unit tests + lint + race detector pass before approving, and I can't without CI.
- The PR's "Tests" section says "compile-time `var _ core.ContextUsageReporter = (*cursorSession)(nil)`" + extended `GetContextUsage` cases — those are exactly the tests I want CI to confirm actually ran on the upstream Go matrix.

What would unblock this:
- Push an empty commit (or amend) to retrigger CI, OR
- Confirm the branch name matches the workflow's `branches:` filter (e.g., `main` only) and either rename or open against a matching branch, OR
- If CI is intentionally not required for this repo, let me know and I'll review based on local evidence.

I'll hold a full review until CI is green or you tell me how you'd like to proceed. Happy to re-evaluate quickly once I see a run. — cc-connect QA

PR #1868, opened 2026-09-18
feat(core): /goto command to switch provider+model keeping context

gpYang:
## What

`/goto` switches the active provider (and optionally the model) for the current session while **keeping** the agent session id and conversation history — unlike `/provider` switch, which clears both. With no arguments it renders a grouped provider → model picker (card mirroring `/model`; text list with numbered targets for non-card platforms).

- `cmdGoto` + `applyGoto`: activate provider, persist model, keep `agent_session_id` and history.
- `collectGotoModels` + `renderGotoCard`/`renderGotoText`: grouped list across all providers, preserving provider order.
- `ProviderScopedModelNames`: opencode agents report model names carrying the `<provider>/<model>` prefix; scoped agents persist the full prefixed name, others the bare name.
- `handleCommand` wires text `/goto <provider>[/<model>]` and numbered targets; `handleCardNav`/`executeCardAction` handle the card picker.

## Fixes

Switching to a ChatGPT-subscription model (`openai/gpt-5.6-sol`, under the cc-connect "chatgpt" provider) used to get an extra `chatgpt/` prefix (`chatgpt/openai/gpt-5.6-sol`), which opencode rejects with `UnknownError` because "chatgpt" is not an opencode provider. The model part that already contains `/` is now treated as a complete provider-scoped name and never double-prefixed. Also fixes the picker showing an empty selection after switching to a subscription model.

## Tests

- `TestCmdGoto_SwitchesProviderKeepsContext`, `TestCmdGoto_WithModel_PersistsModelPerAgentScoping`, `TestCmdGoto_NoArgsRendersGroupedList`, `TestCmdGoto_NumberedTarget`
- `TestCardAction_Goto_SwitchesProviderKeepsContext`, `TestCardAction_Goto_RefreshesPicker`
- `TestCmdGoto_OpenAIScopedModel_NoDoublePrefix` / `TestCardAction_Goto_OpenAIScopedModel_NoDoublePrefix` (regression)
- `TestCUJ_Goto_SwitchKeepsContextEndToEnd`


chenhg5 (changes requested):
QA CI gate: lint failed on this head (Run 35327406103); unit-test/smoke-test/regression-test/performance-test were skipped. Per cc-connect QA CI gate rule, I can't do a full diff review until CI is green — flagging now so the author can fix the lint failure or push a follow-up to retrigger the workflow.

What to do:
1. Open the lint failure: https://github.com/chenhg5/cc-connect/actions/runs/35327406103/job/105543582828
2. Fix the lint complaint (likely a gofmt/golangci-lint issue) on the same branch.
3. Push — the workflow will pick it up automatically if `feat/cmd-goto` is included in the workflow's branch filter.
4. If the workflow never picked up the original push (this is what the 0-then-skipped pattern usually looks like for branch-filter issues), confirm with the maintainer that the branch is included in `.github/workflows/*.yml` triggers.

Will not proceed to full review until CI is green.

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- PR is mergeable now. Solid `/goto` feature with optional `ProviderScopedModelNames` capability, good test coverage including specific regression tests for the double-prefix bug and the previously-broken card action. CI 5/5 green.

Review scope:
- Reviewed `agent/opencode/opencode.go` (+7/-1), `core/interfaces.go` (+11), `core/engine.go` (+314/-1), `core/i18n.go` (+94/-29), `core/cuj_test.go` (+117), `core/engine_test.go` (+317) at head e8ac7c4278.
- Focused on capability design (optional interface), context-preservation contract, double-prefix regression, and card action wiring.

✅ What looks good:
- `applyGoto` deliberately does NOT call `SetAgentSessionID("","")` or `ClearHistory()` — that's the whole point of `/goto` vs `/provider switch`, and the regression tests lock this down explicitly.
- `ProviderScopedModelNames` is an optional capability (type-asserted at call sites) — backwards compatible with agents that don't implement it. opencode declares `true` explicitly; claudecode/codex stay on the default bare-name path.
- The double-prefix fix is well-tested: both the card-action path and the text-command path, both for the persisted model AND the refreshed picker's `InitValue`. The `openai/gpt-5.6-sol` under the `chatgpt` provider case is exactly the shape that would otherwise break opencode.
- The card-action regression test (`TestCardAction_Goto_SwitchesProviderKeepsContext`) nails down a real silent-drop bug (`executeCardAction` previously didn't know about `/goto`).
- `TestCUJ_Goto_SwitchKeepsContextEndToEnd` is a proper end-to-end test driving the real engine through send → /goto → chat → /history and verifying pre-switch turns survive.

🚨/🔴 Must fix:

🟠 Should improve (cross-PR coordination):
- This PR also adds `stopCronTurns` to `cmdStop` — that block is **functionally identical** to PR #1870 (which is already approved and waiting for Glenn's merge). Whoever merges second will get a git conflict on those ~20 lines. Recommend: merge #1870 first, then rebase/merge #1868 so the conflict resolves cleanly to the canonical #1870 implementation, OR have the author rebase on top of #1870's tip before merge.

🔵 Optional:
- Consider checking whether the same double-prefix bug exists in the legacy `/model` path (the `ProviderScopedModelNames` capability was just introduced in this PR; if `/model` predates it, a follow-up PR may be needed to apply the same fix there).
- Document the `ProviderScopedModelNames` capability for future agent authors — it's an opt-in interface, but agents that should implement it (opencode-style) need a clear signal of when to do so.

❓ Questions:
- None.

Testing / Risk:
- Verified evidence: GitHub CI 5/5 green (Run 35335826609: lint 2m4s, unit-test 4m54s, smoke-test 28s, regression-test 37s, performance-test 55s).
- Remaining risk: low-to-moderate, primarily the merge-conflict on `stopCronTurns` (functionally equivalent, just git plumbing) and the implicit default for new agents that don't implement the capability.

Next step:
- Merge. Coordinate with #1870 to avoid the trivial conflict on `cmdStop`. The remaining risk is acceptable and the /goto feature unblocks a real UX gap (provider switching was wiping session).

gpYang:
Added one commit (`271d5728`) that fixes the last piece of the `/goto` promise — the switch surviving recovery.

**Root cause**: `SessionManager.Save` builds a deep-copy snapshot of each session before serializing it, and that snapshot omitted `ActiveProvider`. So `/goto` (and `/provider switch`) set the provider in memory and `sessions.Save()` dropped it on the way to disk. Observable effect: the switch works in-session, but as soon as the in-memory agent is rebuilt — a workspace idle-reap, or a process restart — the next turn silently falls back to the project default. For multi-workspace projects there is no second path to fall back on, because the project-config write in `applyGoto` only happens for the single-workspace case.

Reproduced live before the fix: `/goto deepseek` in a workspace chat logged `provider switched provider=deepseek` and the turn really ran on deepseek, yet the session file kept `active_provider = ""`; two idle-reaps later the next message logged `provider switched provider=chatgpt`.

**Change**: add `ActiveProvider` to the snapshot (and state the invariant in the comment above it), plus regression test `TestSessionManagerSave_PersistsActiveProvider`, which fails before the change with `active_provider after reload = ""` and passes after.

Rebased onto current `main` (`a5c93d9c`) for a clean merge — the branch is `mergeable` again. CI re-running.

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
PR #1868 is mergeable. CI is green (5/5), the implementation cleanly separates the new `/goto` command from the unrelated `/stop` cron tear-down, the persistence regression is fixed at its real root (a missing field in `saveLocked`), and the test coverage is well-aligned with the documented behavior. Head commit reviewed: `271d57286559`.

Review scope:
- 8 files, +871 / -31.
- Focused on the `/goto` command + card-action wiring, the `ProviderScopedModelNames` capability, the persistence regression in `saveLocked`, and the side change `cmdStop` → `stopCronTurns`.

✅ What looks good:
- `applyGoto` is the right design: it does *not* call `SetAgentSessionID("", "")` or `ClearHistory()` (unlike `switchProvider`), so the contract `/provider` (clear) vs `/goto` (keep) is implemented at the single point that matters. The regression test `TestCmdGoto_SwitchesProviderKeepsContext` and the end-to-end `TestCUJ_Goto_SwitchKeepsContextEndToEnd` lock this down.
- The `ProviderScopedModelNames` capability is a clean way to express "opencode uses `aiapi/<model>`, claudecode uses `<model>`" without leaking opencode-specifics into `core/`. `TestCmdGoto_WithModel_PersistsModelPerAgentScoping` and `TestCardAction_Goto_OpenAIScopedModel_NoDoublePrefix` cover both branches and the regression where ChatGPT-subscription models (`openai/gpt-5.6-sol` under the cc-connect `chatgpt` provider) would have gotten a spurious `chatgpt/` prefix.
- The persistence fix in `core/session.go` `saveLocked` is exactly the right one: `ActiveProvider` was missing from the snapshot deep-copy, so the choice never reached disk. The added one-paragraph comment ("Every field that must survive a restart has to be listed here: one that is left out is silently dropped on save...") is a good leave-behind for future contributors.
- `renderGotoCard` correctly strips the `ent.Provider+"/"` prefix from `ent.Model` before constructing the `act:/goto` value (so `aiapi/aiapi/gpt-5.6-sol` does not appear), then re-adds it via `applyGoto` only when the agent is scoped. That round-trip is non-obvious and the comment explains why.
- `handleCardNav("act:/goto ...")` returns the refreshed picker, so the user sees the new selection in the card itself. Verified by `TestCardAction_Goto_RefreshesPicker`.
- `stopCronTurns` is a small, well-scoped extension to `cmdStop`: it only runs when the chat's interactive state is empty (i.e. the user `/stop`s and nothing else is running) and tears down cron-triggered turns by `<sessionKey>#cron:<id>` prefix. It uses existing helpers (`stopInteractiveSessionSilently`) and the new helper is documented with a comment about why cron keys aren't visible to the existing exact/suffix lookups.

- `cmdStop` now has two early-return reply paths inside the same conditional (one for `stopInteractive`, one for `stopCronTurns`) that both reply `MsgExecutionStopped`. If both fire, the second reply overrides the first and the user only sees one. Not a correctness bug for this PR but worth a follow-up: capture both side effects into a single reply, or pick the more informative one (e.g. cron-stopped message if a cron turn was running). Out of scope to fix here.
- `applyGoto` persists the model with `SetProviderModel` only on the success path (`ok2 == true`). When `ok2 == false`, the function still proceeds to set the active provider and update `s.ActiveProvider` but does not log why the model update was rejected. Adding a `slog.Warn("provider model not found", "provider", pname, "model", fullModel)` would help field debugging.
- `MsgCardTitleGoto` is "切换" in zh / "切替" in ja, while `/model` uses different wording in the same cards area. Not a blocker, but worth a consistency pass so users navigating between cards don't see two different verbs for "switch".
- `TestSessionManagerSave_PersistsActiveProvider` is good, but the persistence contract is now implicitly that **every** field of `Session` must be listed in `saveLocked`'s snapshot. A focused test that round-trips every `Session` field (or a comment-driven checklist) would prevent a similar bug in the future. Optional.

🔵 Optional:
- `collectGotoModels` duplicates an existing helper used by `/model` listing. Consider extracting a shared `enumerateModels(switcher) []gotoModelEntry` if `/model` also needs grouped output in the future. Not urgent.

❓ Questions:
- `e.providerSaveFunc` and `e.providerModelSaveFunc` save to disk synchronously on every `/goto`. With a high-frequency keyboard user, this could cause noticeable disk pressure. Is there a debounce or are these already serialized through a coalescer? Not blocking — just curious whether the existing save path is the same one used by `/provider`.

Testing / Risk:
- Verified CI: 5/5 checks green on commit `271d5728`.
- New tests cover: the core `/goto` command (with and without model, with numbered targets, with no args, with provider-scoped names), the card action (`act:/goto`), the picker refresh, the no-double-prefix regression for ChatGPT-subscription models, and the persistence regression.
- Remaining risk: `cmdStop` now has overlapping reply paths (see P2 above); nothing user-visible until someone hits a `/stop` mid-cron-turn and looks at the reply count.

Next step:
- Recommended: merge. The PR is self-contained, the persistence fix is at its root, and the no-double-prefix regression test would have caught a real opencode failure. No changes required before merge.

gpYang:
补充说明（本次追加，head 更新为 `681595df`）：

**`/goto` 现在在工作区会话里是工作区级**，与原生 `/model` 的粒度对齐：

- provider 记到项目状态的 `workspace_provider_overrides[<workspace>]` → 同一工作区下所有群/会话共用，任一群再切会一起变；
- `/goto <provider>/<model>` 的 model 记到 `/model` 已经在用的 `workspace_model_overrides`，**不再改 config.toml 里的 provider 定义**（以前那样会让所有工作区一起变）；
- 之前记在该会话上的 provider 会被清掉，保证工作区值统一生效；
- 工作区 agent 池每次创建工作区 agent 时重新套用该 override，因此池回收与守护进程重启后仍生效。

无工作区的会话（单工作区项目/未绑定）保持原语义：会话级 provider + 项目默认。

同时并入 `/goto` 依赖的持久化修复：`SessionManager.saveLocked` 的快照漏了 `ActiveProvider`，导致会话级 provider 从不写盘（回归测试 `TestSessionManagerSave_PersistsActiveProvider`）。

现场问题（本次修复的动机）：工作区群里的 `/goto` 可能被命令派发交到全局 session manager，于是只写了项目默认而该群会话没有任何记录，空闲回收/重启后就被打回项目默认；实测表现为「切到 deepseek，过一会儿又变回 chatgpt」。

测试：两个会话绑同一工作区都跟随切换；用已存在的 state 文件新建 engine 后工作区 agent 仍带该 provider；model 落在工作区存储且不写配置；未绑定会话仍写会话级 + 项目默认。

gpYang:
再追加一个同区域的修复，head 现为 `16486618`（独立 commit，放在本 PR 是因为它就是「工作区会话里命令解析不到工作区」的根因）：

**`fix(core): resolve a channel's workspace binding by the message's channel key`**

绑定是按**消息携带的 platform 标识**存 key 的；对 bridge 适配器来说那是注册的适配器名（如 `live-goto2`），而 `Platform.Name()` 返回的是平台类型（`bridge`）。`resolveWorkspace` 用 `p.Name()` 拼 key，于是**从消息路径创建的绑定永远查不回来**：该频道每条消息都掉进 workspace-init 流程（`此频道未找到工作区…`），命令也退化为全局上下文（本 PR 里 `applyGoto` 的 session-key 兜底正是绕过它）。

现在消息路径与命令上下文都先用 `effectiveWorkspaceChannelKey(msg)` 查绑定，再回落到 `p.Name()` 查法与按频道名的约定匹配（feishu/telegram 等 `Name()` 与 session key 前缀一致的平台行为不变）。

回归测试：在消息平台名（`live-goto2:PROBE`）下绑定的频道能解析到工作区与工作区 agent，而 `p.Name()` 查法查不到。

如果更希望它单独成 PR 或并入 #1699，把 commit `16486618` 抽出来即可，我可以随时移动。

chenhg5 (approved):
## Conclusion: Approve — well-bounded scope, real bug fixes, strong test coverage

Overall assessment: This is a substantial feature PR (10 files, +1270/-36, 5 commits) that delivers the new `/goto` command while fixing two real bugs that were already on the user's radar: `ActiveProvider` silently dropped by `saveLocked` (so a `/goto` choice evaporated after the next workspace idle-reap), and the OpenAI-subscription model double-prefix (`chatgpt/openai/gpt-5.6-sol`) that opencode rejects. The implementation leans on the existing `WorkspaceModelOverrides` pattern as a template for the new `WorkspaceProviderOverrides`, introduces a clean optional `ProviderScopedModelNames` capability, and lands 15 new tests including a CUJ that exercises `/goto` against a real engine. CI is green. **Two minor follow-ups are worth doing in this PR rather than later**: a stale `provider_idle_session_unavailable` style wording in `MsgGotoDefault` is slightly off-tone, and the new `stopCronTurns` introduces a small TOCTOU window that deserves a comment if not a guard.

Review scope:
- Reviewed `core/engine.go`, `core/interfaces.go`, `core/projectstate.go`, `core/session.go`, `core/i18n.go`, `agent/opencode/opencode.go` and the four test files at head `4ca3043601e95cc2e778ae97cd26502c3b15743b`.
- Walked the 5 commits: `5c50739` (initial feature) → `e8ac7c4` (lint orphan-field fix) → `681595d` (workspace-scoped persistence + cron-stop) → `1648661` (channel-key workspace resolution) → `328f174` (picker dedup) → `4ca30436` (drop redundant card status line).
- Confirmed all 5 i18n languages are populated for every new key (saw `LangEnglish / LangChinese / LangTraditionalChinese / LangJapanese / LangSpanish` for each of the 8 new `MsgGoto*` / `MsgCardTitleGoto` keys).
- CI: `lint / unit / regression / smoke / performance-test` all green (5/5).

✅ What looks good:
- **The `ActiveProvider` save bug fix is the headline value of this PR.** `core/session.go:saveLocked` now copies `s.ActiveProvider` into the snapshot — without it, every `/goto` choice was being silently dropped on `sessions.Save()`. The added comment ("Every field that must survive a restart has to be listed here: one that is left out is silently dropped on save") is exactly the kind of comment that prevents the next person from re-introducing this class of bug. Good defensive design.
- **`ProviderScopedModelNames` is a clean capability hook.** The new optional interface is a textbook Go pattern: scoped agents report `true`, others default to bare model names, and `/goto`'s double-prefix fix relies on it without polluting the core command path. `TestCmdGoto_OpenAIScopedModel_NoDoublePrefix` and `TestCardAction_Goto_OpenAIScopedModel_NoDoublePrefix` lock the fix in place from both the text-command and card-entry angles.
- **`/goto` is workspace-scoped in exactly the same way `/model` already is.** The comment "the same granularity the native /model command uses for models" is true and important — it means a workspace's `/goto` doesn't accidentally leak into the project default, and a single chat's `/goto` doesn't disappear after the workspace-pool evicts the agent. `TestCmdGoto_WorkspaceChatSetsWorkspaceProvider`, `TestCmdGoto_WorkspaceChatRecordsModelOnWorkspace`, and `TestCmdGoto_UnboundChatKeepsSessionSemantics` cover all three branches.
- **The picker dedup is a real UX fix.** `TestCollectGotoModels_DedupesRepeatedEntries` reproduces the "same model over and over in the picker" report; the implementation collapses identical `(provider, model)` pairs via a `seen` set. Trivial code, real user-visible improvement.
- **`stopCronTurns` is well-named and well-scoped.** It's a separate helper, only invoked from the `cmdStop` "no exact match" branch, and replies with `MsgExecutionStopped` (not a new message) so the user gets the same acknowledgement regardless of which kind of turn they stopped. Good.
- **`TestCUJ_Goto_SwitchKeepsContextEndToEnd`** is the kind of test that protects the user-facing promise: "after chatting, `/goto <provider>/<model>` switches provider WITHOUT wiping the conversation." Worth keeping.

🟠 Should improve (P2):
- **`stopCronTurns` opens a small TOCTOU window.** The implementation snapshots keys under `interactiveMu.Lock()`, releases the lock, then calls `stopInteractiveSessionSilently` for each key. Between the snapshot and the call, a new cron turn could be registered for the same chat (and would not be stopped) — meanwhile a cron turn that finished naturally during the window is still in the snapshot and would be "stopped" as a no-op (probably safe, but worth confirming). Two acceptable mitigations: (a) keep holding the lock and call the stop helper inside the critical section if it's lock-free on its own (the name suggests yes), or (b) add a one-line comment documenting that the snapshot is best-effort and a cron turn starting after the snapshot will not be torn down by this `/stop`. The current code is correct under the assumption that cron registrations are rare and short; the comment is the cheaper fix.
- **`MsgGotoDefault` wording.** Current text: `"No active provider. Choose a target below:"`. This is fine for the empty-state case, but `/goto` always has a target list when there's an active provider — what message does the user get if they run `/goto` with no args *and* there are no providers configured at all? The new picker would just be empty. Suggest a follow-up branch in `renderGotoCard`/`renderGotoText` for "no providers configured" with a clearer `MsgGotoNoProviders` key. Right now the empty picker is silently confusing.

🔵 Optional (P3):
- **`cmdProvider` formatting change folded into this PR.** The unrelated `sb.WriteString(fmt.Sprintf(...))` → `fmt.Fprintf(&sb, ...)` cleanup in `cmdProvider` is fine on its own, but it makes this PR's diff slightly noisier than necessary. Future PRs should keep unrelated refactors out of feature commits.
- **`stopInteractiveSessionSilently` error handling.** The new `stopCronTurns` ignores the per-key boolean result, only counting successes implicitly. If a future change to `stopInteractiveSessionSilently` adds a meaningful failure mode, `stopCronTurns` will silently under-count. Consider returning `(stopped int, errs []error)` or just `slog.Debug`-ing the misses.
- **`resolveWorkspaceForChannel` (commit `1648661`)** — the new `channelKey`-aware resolver is the right behavior but the PR description doesn't mention it. Worth a one-line note in either the commit body or the function's doc-comment explaining *why* `channelKey` matters (otherwise the next maintainer sees "two near-identical resolvers" and may consolidate them incorrectly).

❓ Questions:
- For ChatGPT-subscription models like `openai/gpt-5.6-sol` under the cc-connect `chatgpt` provider: does opencode know to route `chatgpt/openai/gpt-5.6-sol` correctly *today*, or does the actual underlying CLI flag need to be the un-prefixed `openai/gpt-5.6-sol`? The PR description says opencode rejects `chatgpt/openai/gpt-5.6-sol`, which is the bug; but if the *correct* form for opencode CLI is *also* `openai/gpt-5.6-sol` (with the literal slash, no `chatgpt` prefix), then the fix should produce `openai/gpt-5.6-sol`, which the diff does via `if !strings.Contains(rest, "/")`. Just confirming the test (`TestCardAction_Goto_OpenAIScopedModel_NoDoublePrefix`) actually asserts the *un-prefixed* form and not just "doesn't double-prefix."

Testing / Risk:
- Verified: 5/5 CI green. 15 new tests across CUJ / unit / engine / session / multi-workspace / card action layers. Pre-existing workspace-model-override path is mirrored by a new workspace-provider-override path with its own tests.
- Remaining risk: workspace → project leakage. `applyGoto` resolves to `sessionContextForKey` when sessions disagree with the chat context — this is correct, but if a deployment has *no* workspace binding for the chat and the command goes through the global session manager, the provider will be saved at the project level (`e.providerSaveFunc`). That's the intended fallback, but it means a chat without a workspace binding behaves differently from one with a binding. Not a bug, but worth a release-note bullet.

Next step:
- Either land as-is (acceptable — both P2 items are improvements, not blockers), or address `stopCronTurns` snapshot-comment and the picker-empty-state in this same PR (preferred). Once that's resolved, this is mergeable.

PR #1869, opened 2026-09-18
fix(opencode): let v2 compaction auto-continue in place; watchdog for stalls

gpYang:
> **Depends on #1856** (branch is stacked on `feat/opencode-compaction-continue`; once #1856 merges this diff shrinks to the follow-up changes only).

## Problem

OpenCode V2 compaction is checkpoint-based and automatically retries the pending step inside the **same** process — it does not stop and does not wait for stdin. The previous kill-and-resume recovery interrupted that auto-retry, so the resumed process re-compacted every ~30s until the continuation limit was hit and the turn ended with an empty 11-char reply (the compaction loop).

## Changes

- **Never kill on compaction**: `compaction` / `compaction_continue` events only suppress the summary text and arm `expectingContinue`; the process is left running so opencode's own auto-retry finishes the turn.
- **readLoop EOF**: only ends the turn with a notice when the process exits while `expectingContinue` is still set (watchdog kill after a real stall, or opencode giving up after a failed checkpoint).
- **Stall watchdog**: kills a process with no events for 5 minutes instead of hanging 20+ minutes. `lastEvent` is reset per `Send`, so a fresh turn never inherits the previous turn's idle clock — previously the watchdog miskilled the new process one second after the message arrived (`response_len=11`, `input_tokens=0`, log `killing stalled process idle=5m3s`).
- **Empty streaming-card turns** produce no payload, so `__cc_connect_progress_card_v1__:...` JSON no longer leaks verbatim to users on simple Q&A turns.

## Regression tests

- `TestCompactionDoesNotKillProcess` — process must survive compaction events
- `TestCompactionWorkEventClearsFlag`
- `TestStallWatchdogKillsStalledProcess` / `TestStallWatchdogSparesActiveProcess`
- `TestSendResetsStallWatchdogClock` — new-turn clock reset
- `TestBuildStreamingCardPayload` empty-payload guard

### DIFF
--- a/agent/codex/session.go
+++ b/agent/codex/session.go
@@ -453,6 +453,18 @@ func codexErrorKind(errorInfo string, message string) core.ErrorKind {
 	if strings.Contains(msg, "rate limit") || strings.Contains(msg, "rate_limit") {
 		return core.ErrorKindRateLimit
 	}
+	if strings.Contains(msg, "stream disconnected before completion") ||
+		strings.Contains(msg, "stream closed before response.completed") ||
+		strings.Contains(msg, "you can retry your request") ||
+		strings.Contains(msg, "processing your request") ||
+		strings.Contains(msg, "unexpected status 502") ||
+		strings.Contains(msg, "unexpected status 503") ||
+		strings.Contains(msg, "unexpected status 504") ||
+		strings.Contains(msg, "bad gateway") ||
+		strings.Contains(msg, "service unavailable") ||
+		strings.Contains(msg, "gateway timeout") {
+		return core.ErrorKindOverloaded
+	}
 	return core.ErrorKindUnknown
 }
 
--- a/agent/codex/session.go
+++ b/agent/codex/session.go
@@ -408,14 +408,17 @@ func (cs *codexSession) handleEvent(raw map[string]any) {
 
 	case "turn.failed":
 		errMsg := ""
+		errInfo := ""
 		if errObj, ok := raw["error"].(map[string]any); ok {
 			errMsg, _ = errObj["message"].(string)
+			errInfo, _ = errObj["codex_error_info"].(string)
 		}
 		if errMsg == "" {
 			errMsg = "turn failed (no details)"
 		}
-		slog.Warn("codexSession: turn failed", "error", errMsg)
-		evt := core.Event{Type: core.EventError, Error: fmt.Errorf("%s", errMsg)}
+		errKind := codexErrorKind(errInfo, errMsg)
+		slog.Warn("codexSession: turn failed", "error", errMsg, "codex_error_info", errInfo, "error_kind", errKind)
+		evt := core.Event{Type: core.EventError, Error: fmt.Errorf("%s", errMsg), ErrorKind: errKind}
 		select {
 		case cs.events <- evt:
 		case <-cs.ctx.Done():
@@ -433,6 +436,24 @@ func (cs *codexSession) handleEvent(raw map[string]any) {
 	default:
 		slog.Debug("codexSession: unhandled event type", "type", eventType)
 	}
+}
+
+func codexErrorKind(errorInfo string, message string) core.ErrorKind {
+	switch strings.TrimSpace(errorInfo) {
+	case "server_overloaded":
+		return core.ErrorKindOverloaded
+	case "rate_limit_exceeded", "rate_limit":
+		return core.ErrorKindRateLimit
+	}
+
+	msg := strings.ToLower(message)
+	if strings.Contains(msg, "at capacity") || strings.Contains(msg, "overloaded") {
+		return core.ErrorKindOverloaded
+	}
+	if strings.Contains(msg, "rate limit") || strings.Contains(msg, "rate_limit") {
+		return core.ErrorKindRateLimit
+	}
+	return core.ErrorKindUnknown
 }
 
 // flushPendingAsThinking emits all buffered agent_messages as EventThinking.
--- a/agent/codex/session_test.go
+++ b/agent/codex/session_test.go
@@ -614,7 +614,11 @@ func TestSend_HandlesLargeJSONLines(t *testing.T) {
 	if err != nil {
 		t.Fatalf("newCodexSession: %v", err)
 	}
-	defer cs.Close()
+	defer func() {
+		if err := cs.Close(); err != nil {
+			t.Fatalf("Close: %v", err)
+		}
+	}()
 
 	if err := cs.Send("hello", "", nil, nil); err != nil {
 		t.Fatalf("Send: %v", err)
@@ -669,7 +673,11 @@ func TestSend_TurnFailedServerOverloadedIsRetriable(t *testing.T) {
 	if err != nil {
 		t.Fatalf("newCodexSession: %v", err)
 	}
-	defer cs.Close()
+	defer func() {
+		if err := cs.Close(); err != nil {
+			t.Fatalf("Close: %v", err)
+		}
+	}()
 
 	if err := cs.Send("hello", "", nil, nil); err != nil {
 		t.Fatalf("Send: %v", err)
--- a/agent/codex/session_test.go
+++ b/agent/codex/session_test.go
@@ -649,6 +649,52 @@ func TestSend_HandlesLargeJSONLines(t *testing.T) {
 	}
 }
 
+func TestSend_TurnFailedServerOverloadedIsRetriable(t *testing.T) {
+	workDir := t.TempDir()
+	binDir := filepath.Join(workDir, "bin")
+	if err := os.MkdirAll(binDir, 0o755); err != nil {
+		t.Fatalf("mkdir bin: %v", err)
+	}
+
+	shellScript := "#!/bin/sh\n" +
+		"printf '%s\\n' '{\"type\":\"thread.started\",\"thread_id\":\"thread-overloaded\"}'\n" +
+		"printf '%s\\n' '{\"type\":\"turn.failed\",\"error\":{\"message\":\"Selected model is at capacity. Please try a different model.\",\"codex_error_info\":\"server_overloaded\"}}'\n"
+	powershellScript := "[Console]::Out.WriteLine('{\"type\":\"thread.started\",\"thread_id\":\"thread-overloaded\"}')\n" +
+		"[Console]::Out.WriteLine('{\"type\":\"turn.failed\",\"error\":{\"message\":\"Selected model is at capacity. Please try a different model.\",\"codex_error_info\":\"server_overloaded\"}}')\n"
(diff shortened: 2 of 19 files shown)


chenhg5 (changes requested):
QA CI gate: lint failed on this head (Run 35327447342); unit-test/smoke-test/regression-test/performance-test were skipped. Per cc-connect QA CI gate rule, I can't do a full diff review until CI is green — flagging now so the author can fix the lint failure or push a follow-up to retrigger the workflow.

What to do:
1. Open the lint failure: https://github.com/chenhg5/cc-connect/actions/runs/35327447342/job/105543707047
2. Fix the lint complaint (likely a gofmt/golangci-lint issue) on the same branch.
3. Push — the workflow will pick it up automatically if `fix/opencode-compaction-followup` is included in the workflow's branch filter.
4. If the workflow never picked up the original push (this is what the 0-then-skipped pattern usually looks like for branch-filter issues), confirm with the maintainer that the branch is included in `.github/workflows/*.yml` triggers.

Will not proceed to full review until CI is green.

gpYang:
Fixed — the lint failures are resolved on the updated head `fbe644a6`:

- `QF1012` ×3: `WriteString(fmt.Sprintf(...))` → `fmt.Fprintf(&sb, ...)` (progress/text render path)
- removed orphan `permissionTimeoutNanos` field (carried in from a 3-way merge context; not present in origin/main → unused)
- `errcheck` ×2: explicit `Close()` error checks in session tests

CI is now 5/5 green on the new head: lint ✅, unit-test ✅, smoke-test ✅, regression-test ✅, performance-test ✅.

Ready for re-review — thanks!

gpYang:
Updated on the new head `8c57aef9` — the earlier change request (lint failure at the old head) is addressed, and the branch no longer conflicts with `main`:

- **lint fixed**: `QF1012` ×3 (`WriteString(fmt.Sprintf(...))` → `fmt.Fprintf`), orphan `permissionTimeoutNanos` field removed, plus `errcheck` on the new tests.
- **rebased onto current `main` (781ead2b)**: this PR is stacked on #1856, which is stacked on #1745 — all three were rebased in order, so the stack is conflict-free again (`mergeable=true`). The dependency is real: #1856 references the `finalized` local introduced by #1745, and this PR depends on #1856's `readLoop`/`launch` structure plus `core.Event.ErrorKind` from #1641.
- **CI is 5/5 green** on `8c57aef9`: lint, unit-test, smoke-test, regression-test, performance-test.

Ready for re-review when you have a moment — thanks!

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Re-review of #1869 (opencode V2 compaction auto-continue + stall watchdog + codex/opencode retriable errors + foldable streaming card panels). CI is now 5/5 green on the current head `8c57aef9d` (the earlier CHANGES_REQUESTED was a CI gate on a prior head and is now resolved). The branch is large (+5.2k/-266 across 30 files) but every change is anchored in a concrete live-observed bug (empty 11-char replies, 22-minute stalls, empty-progress-payload leak, transient API failures hard-failing turns). Tests are dense and behavior-pinning.

Review scope:
- Reviewed `agent/opencode/session.go`, `agent/opencode/session_test.go`, `agent/opencode/opencode_resilience_test.go`, `core/progress_compact.go`, `core/progress_compact_test.go`, `platform/feishu/streaming_card.go`, `platform/feishu/streaming_card_test.go`, `core/engine.go`, `core/interfaces.go`, `agent/codex/session.go`, `agent/codex/session_test.go` (diff since last CR head `6ea3b25e80`), plus the new `codexErrorKind` / `opencodeErrorKind` classifiers and `stallWatchdog`.
- Focused on correctness, safety of process lifecycle changes, error classification soundness, and the empty-payload leak fix on Feishu.

✅ What looks good:
- **Empty progress-payload leak is fixed at three layers**: `BuildStreamingCardPayload` now returns `""` for empty turns (mirrors `BuildProgressCardPayloadV2`), `cardJSONForContent` short-circuits any unparseable payload to a blank card instead of rendering the raw `__cc_connect_progress_card_v1__:...` JSON, and the regression tests pin both. This was a user-visible bug.
- **OpenCode V2 compaction semantics are correctly modeled**: the bridge stops killing the process on `compaction_continue` / compaction-summary (V2 auto-retries the pending step inside the same process; killing causes the empty-reply loop). Instead the flag is set, real work events clear it via `markContinued`, and a *stall* watchdog kills the process only after 5 minutes of total silence. The comment "Observed live: one compaction every ~30s ending in an empty reply" shows the author tested against the actual failure mode.
- **`stallWatchdog` is well-scoped**: it only kills when `alive && !resultSent && lastEvent != 0 && idle >= stallTimeout`, the process handle is guarded by `procMu`, and tests `TestStallWatchdogKillsStalledProcess` / `TestStallWatchdogSparesActiveProcess` cover both branches. Reset of `lastEvent` in `Send()` closes the "watchdog miskills a fresh turn because the previous turn's idle carried over" race — `TestSendResetsStallWatchdogClock` pins it.
- **`writeStdin` + `stdinWriteTimeout`** correctly handles the new deadlock variant where `opencode run` only starts executing after stdin EOF; the writer goroutine abandons after 10s and the pipe is closed regardless of write outcome.
- **Error-kind classification is now consistent across `codexErrorKind` and `opencodeErrorKind`**, mirroring PR #1641. The engine's whole-turn retry will now kick in for transient overload on both providers instead of leaking a raw `❌ 错误:` to the user. Tests cover the same surface (rate limit, 502/503/504, "stream disconnected before completion", auth as the negative control).
- **`codexErrorKind` now reads `codex_error_info`** when present (the structured signal) and falls back to message pattern matching; both branches are tested.
- **`TestSend_TurnFailedServerOverloadedIsRetriable`** uses the same fake-codex script pattern as existing tests, asserting `ErrorKind == ErrorKindOverloaded` end-to-end through the event channel — this is the right level of test for the new behavior.

🚨/🔴 Must fix:
- None merge-blocking.

🟠 Should improve:
- **`sendCompactionNotice` emits a Chinese-only hardcoded message** ("⚠️ 任务处理超时（长时间无响应），已自动终止。请重试；若任务较大，建议拆成更小的步骤分步发送。") in `agent/opencode/session.go`. The other agent-facing error strings in this PR go through i18n keys via `core.MsgKey`. Please move it to an i18n key (e.g. `MsgOpencodeStallTimeout` with en/zh/zh-TW/ja/es translations) and pick the locale from the active session/agent config — non-Chinese users will see mixed-language output today.
- **`stallWatchdog` uses a global `stallTimeout = 5 * time.Minute` constant**. That's a reasonable default but is not configurable. Consider exposing it on `core.OpencodeConfig` (similar to how `RetriableErrorConfig` is exposed in this same PR) so operators with large long-context turns can raise it without a rebuild.
- **`stallWatchdog` tick interval is hardcoded to `30 * time.Second` in `newOpencodeSession`**. With a 5-minute stall window that means up to 30 s of detection latency. Fine for this bug, but worth documenting or making the tick ratio explicit (e.g. `tick = stallTimeout/10`).
- **`codexErrorKind` parses `codex_error_info` strings by exact trim/case match only**. If the codex upstream emits a slightly different casing or a wrapped enum (e.g. `"ServerOverloaded"`), classification silently falls back to `ErrorKindUnknown`. Consider lowercasing before the switch.
- **`TestSendResetsStallWatchdogClock` depends on `Send` calling `newOpencodeSession(..., "true", ...)`**, which on Linux will resolve to `/usr/bin/true` and immediately exit. The assertion is on `s.lastEvent.Load()` after `Send`, which is the right thing — but the test name is slightly misleading because the contract is `lastEvent gets reset by Send`, not `Send succeeds`. Consider renaming or adding a one-line assertion of `Send` success for clarity.
- **Foldable streaming card panels (d7c0a90d `renderProgressEntryElement` / `buildProgressPanel`) are not reviewed in depth in this cycle** because the file touched (platform/feishu/streaming_card.go) is covered separately. Suggest a focused review pass if any panel-rendering edge cases remain.

🔵 Optional:
- The `sendCompactionNotice` event ordering (text → result) is good, but consider adding `EventNotice` as a first-class type so progress-card UIs can render it without parsing the text. Not required for this PR.
- `streamingCard.go` import added `strings` (visible in the diff); fine.
- The `core.SetRetriableErrorPolicy` call in `applyRetriableErrorConfig` is good. Consider also wiring the policy into a small `slog.Info` at startup ("retriable error policy: initial=5s retry=10s max=3") so operators can verify the config actually loaded.

❓ Questions:
- The `8c57aef9` merge brings in `fix/codex-retry` (`f70bb08c5` → `e4d50bb433`) plus the `message.finalized` hook work (`510d47` → `4b5ab05`). Those feature branches were developed separately and are landing here as a stacked merge. Is this PR the intended landing path for them, or are they going to be split out? The risk-report doc assumes the latter, but the diff says the former.
- `core/progress_compact.go` adds an `empty` guard to `BuildStreamingCardPayload` but the same guard does **not** appear in `BuildProgressCardPayloadV2`. Was `V2` already guarded (the comment hints it is), or is this a second-pass follow-up needed?

Testing / Risk:
- Verified: 5/5 CI jobs green on current head `8c57aef9d` (`lint`, `unit-test`, `smoke-test`, `regression-test`, `performance-test`).
- New tests: `TestOpencodeErrorKind_RetriableOverloaded`, `TestCompactionDoesNotKillProcess`, `TestCompactionWorkEventClearsFlag`, `TestStallWatchdogKillsStalledProcess`, `TestStallWatchdogSparesActiveProcess`, `TestSendResetsStallWatchdogClock`, `TestStdinPipe_PromptThenEOF`, `TestBuildStreamingCardPayload_EmptyReturnsEmpty`, `TestCardJSONForContent_UnparseablePayloadDoesNotLeak`, `TestSend_TurnFailedServerOverloadedIsRetriable`, `TestCodexErrorKind_RetriableStreamFailures` — all pin the new behavior against live-observed failure modes.
- Remaining risk: the **size of this PR** is the largest residual risk. It bundles compaction, watchdog, retry-policy, error classification, foldable streaming cards, and the empty-payload fix into one branch. If anything regresses post-merge, bisection across 30 files will be expensive. Strongly suggest splitting the codex retry merge (`8c57aef9`) out into a separate PR in the future; for this PR the diff is testable as-is.
- Remaining risk: stallWatchdog killing the live process means any in-flight tool call is lost mid-execution. Acceptable given the alternative is a 22-minute hang, but worth noting.

Next step:
- Mergeable as-is. Suggest landing the i18n fix for `sendCompactionNotice` and the `codex_error_info` lowercasing as a small follow-up — neither blocks this PR. If the maintainer wants this branch decomposed before merge (split out codex retry + message.finalized into their own PRs), do that as a separate workflow; the opencode side of this branch is independently mergeable.

PR #1870, opened 2026-09-18
fix(cron): stop turn on job timeout

gpYang:
## Problem

`runJob` timeout (default 30m) only marks the run as failed — the `ExecuteCronJob` goroutine and the agent turn keep running (observed: a turn kept going 30+ minutes after its job timed out).

## Changes

- On timeout, stop the running turns:
  1. `stopCronTurns` — new-per-run keys (`<sessionKey>#cron:<sid>` prefix scan)
  2. fall back to `stopInteractiveSessionSilently` for the shared session key
- `/stop` in the same chat now also tears down a running cron turn, so the user can switch provider and re-trigger instead of waiting out the retry window.

## Tests

- `TestCmdStop_StopsCronTurn`
- cron timeout tests (new-per-run + shared session key)


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- PR is mergeable now. Targeted P1 bugfix for cron timeouts leaving their turns running, plus /stop visibility. CI 5/5 green; minimal scope; reuses existing `stopInteractiveSessionSilently` correctly.

Review scope:
- Reviewed `core/cron.go` (+22/-0), `core/engine.go` (+30/-0), `core/cron_test.go` (+59/-0), `core/engine_test.go` (+32/-0) at head 1e3b23928e.
- Focused on correctness of the cron-turn key prefix assumption, lock ordering, /stop integration, and regression coverage.

✅ What looks good:
- Root cause is correctly identified: `<sessionKey>#cron:<sid>` keys are invisible to the exact/suffix lookups in `cmdStop`, and `runJob` only marks the job failed without tearing down the agent turn.
- The fix is minimal and reuses the existing `stopInteractiveSessionSilently` helper (engine.go:10361) — no new public API, no new config, no new persistence.
- `stopCronTurns` correctly scans with the lock, releases it before calling `stopInteractiveSessionSilently` (which takes the lock again) — avoids any nested-lock hazard.
- The `stopCronTurnOnTimeout` shared-session fallback (`stopInteractiveSessionSilently(job.SessionKey)` when no #cron: match) is the right symmetric design for the single-turn shared-session case.
- 4 new regression tests cover the critical paths: #cron: form, shared-session form, /stop integration, and the "stopped" vs "no execution" reply.
- /stop now responds correctly when there's a stuck cron turn — much better UX than the misleading "no execution in progress".

🚨/🔴 Must fix:

🟠 Should improve:
- Minor: `stopCronTurns` does an O(n) scan of `interactiveStates` on every /stop call. With realistic cron counts this is fine, but if the map grows large you may want a secondary index keyed by SessionKey → []cronKey for the prefix-scan case. Not blocking; defer.

🔵 Optional:
- Consider adding a brief release note about the behavior change: cron timeouts now stop their turns, and /stop in a chat with a running cron turn will return "stopped" rather than "no execution". This is a user-visible UX change worth documenting.

❓ Questions:
- None.

Testing / Risk:
- Verified evidence: GitHub CI 5/5 green (Run 35327479549: lint 2m23s, unit-test 4m51s, smoke-test 29s, regression-test 36s, performance-test 1m1s).
- Remaining risk: low. Worst case is a shared-session cron turn being stopped slightly later than ideal after the timeout — strictly better than the current behavior (turn running 30+ min).

Next step:
- Merge. The remaining risk is acceptable and the fix unblocks a real production complaint.

chenhg5 (approved):
## Conclusion: Approve

Clean, focused bugfix. The change correctly addresses a real production symptom (timed-out cron turns continuing for 30+ minutes) by wiring the existing `stopInteractiveSessionSilently` into the timeout path. Both the new-per-run cron key shape (`<sessionKey>#cron:<id>`) and the legacy shared-session shape are covered. The bonus `/stop` extension into cron turns is a real UX win — users can now interrupt a stuck cron and re-trigger with a different provider instead of waiting out the retry window.

Maintainer chenhg5 already approved on this exact head (review commit = head sha, not stale).

### What looks good

- **Right reuse, no duplication**: builds on the existing `stopInteractiveSessionSilently` helper rather than reimplementing turn cleanup. Cron timeout path and `/stop` path both flow through it.
- **Two-key fallback strategy**: `stopCronTurnOnTimeout` first scans `<key>#cron:` prefix matches, then falls back to the bare session key for shared-session cron. This covers both cron key shapes that exist in the codebase.
- **`slog.Warn` after stop**: structured log records `id`, `manual`, `stopped` count, `timeout` — exactly what ops needs to detect "cron timed out + we cleaned up" vs. "cron timed out + turn survived".
- **`/stop` UX consistency**: when `stopCronTurns` matches, `cmdStop` replies `MsgExecutionStopped` (same message as a normal turn stop), not `MsgNoExecution`. Cron turns now look the same as interactive turns from the user's POV.
- **Lock-then-release in `stopCronTurns`**: keys are collected under `interactiveMu` but stops are issued after release, avoiding lock contention with other interactiveMu users.
- **Pure addition**: +143/-0 across 4 files. No deletions, no signature changes, no config schema changes. Revert is safe.

### 🟠 Should consider (not blocking)

- **P2: Missing end-to-end test for the `runJob` timeout wiring.** The new tests cover `stopCronTurnOnTimeout` and `cmdStop` directly with stub state, but no test actually exercises the `runJob` timeout case path → asserts turn was stopped. A `TestRunJob_TimeoutStopsTurn` integration test (short `ExecutionTimeout`, blocking agent turn) would catch future refactors that drop the `stopCronTurnOnTimeout` call from the timeout branch.

### 🔵 Optional

- **P3**: `stopCronTurns` uses lock-then-release; a cron turn inserted between the unlock and the stop loop would be missed. In practice the window is tiny and cron turns in the same chat are rare, but if you want bulletproof semantics, do the stop call under the same lock (verify `stopInteractiveSessionSilently` doesn't open a new lock first).
- **P3**: `sessionKey == ""` would make the prefix `"#cron:"` and match all cron turns globally. Defensive guard `if sessionKey == "" { return 0 }` at the top of `stopCronTurns` is cheap insurance for future cron shapes.

### Testing / Risk

- **Verified**: CI 5/5 green on `1e3b23928e8c`. Maintainer approved on current head. 3 new unit tests cover both key shapes and the `/stop` extension.
- **Remaining risk**: P2 above (no end-to-end wiring test). P3 items are defensive.

### Next step

If you have time, add the integration test for the `runJob` timeout wiring. Otherwise this is merge-ready — chenhg5 already approved and the fix is correct.

PR #1871, opened 2026-09-18
fix(core): render /reasoning usage from active agent's effort levels

M1racleShih:
## Problem

Running a non-Codex agent (e.g. `pi`), typing `/reasoning` in the bot showed Codex-specific text:

1. The not-set hint: `当前推理强度: (未设置，使用 Codex 默认值)` / "using Codex default" — for **every** agent.
2. The usage line: `/reasoning <low|medium|high|xhigh|max>` — Codex's level set, even though the selectable list correctly comes from the active agent's `AvailableReasoningEfforts()` (pi: `off/minimal/low/medium/high/xhigh`, no `max`).

This also violated the repo rule "No Hardcoding Platform or Agent Names in Core" — "Codex" was hardcoded in `core/i18n.go`.

## Fix

- `MsgReasoningDefault`: drop the agent name → "(not set, using default)" (all 5 languages).
- `MsgReasoningUsage`: add `%s` placeholder; new helper `reasoningUsageText()` fills it with `strings.Join(AvailableReasoningEfforts(), "|")` at all 3 call sites (text listing, invalid-target reply, card note).

## Testing

- New regression test `TestCmdReasoning_NoHardcodedCodexLevels`: fails on pre-fix code, passes after — covers the text path, invalid-target reply, and the card path (markdown hint + note).
- `go build ./...`, `go test ./...`, `go vet ./core/`, `go test ./core/ -run TestCUJ` all pass.

Fixes the user-visible symptom: pi bot in Feishu now shows its own levels `off|minimal|low|medium|high|xhigh` and no "Codex" mention.

PR #1872, opened 2026-09-18
fix(pi): merge custom providers into model list and make reasoning efforts model-aware

M1racleShih:
## Problem

Two user-visible bugs in the pi agent adapter (reported on a Feishu pi bot):

1. **Custom providers missing from /model**: `AvailableModels()` only read `models-store.json`. User-defined providers from `models.json` (e.g. a second GLM provider `zai-coding-team` next to the catalog's `zai-coding-cn`) never showed up, so one GLM provider was invisible.
2. **/reasoning had no `max` and was not model-aware**: `AvailableReasoningEfforts()` hardcoded `[off, minimal, low, medium, high, xhigh]` for every model — missing `max` (which pi supports, `pi --help`: `off, minimal, low, medium, high, xhigh, max`) and ignoring each model's `thinkingLevelMap`. Selecting `deepseek/deepseek-flash` should offer `off/low/high/max`.

## Fix

- New merged catalog: `readModelCatalog()` overlays `models.json` custom providers onto the `models-store.json` catalog (custom defs replace same-id entries, append new ones), mirroring what pi's own model picker shows. `/model` now lists both GLM providers.
- Model-aware efforts: ports pi-ai's `getSupportedThinkingLevels` rules — level mapped to `null` = unsupported; `xhigh`/`max` only when explicitly mapped; non-reasoning models = `off` only; bare ids resolve preferring `settings.json` `defaultProvider`; unknown models fall back to the full list (incl. `max`).
- Context windows now also come from the merged catalog (custom file wins), so catalog-only models no longer fall back to 200K.

## Testing

- Regression tests (fail before, pass after): `TestAvailableModels_MergesCustomProviders`, `TestAvailableReasoningEfforts_DeepSeekFlashIncludesMax`; plus `TestAvailableModels_CustomOverridesCatalogModel`, `TestAvailableReasoningEfforts_ModelAware` (5 cases), `TestSupportedThinkingLevels_BareIDPrefersDefaultProvider`.
- `go build ./...`, `go test ./...`, `go test ./core/ -run TestCUJ` all pass.

**Note**: stacked on top of #1871 (contains its commit too, so it can be merged independently; once #1871 lands only the new commit will diff).


chenhg5 (approved):
# PR #1872 QA Review

## Verdict
**Conclusion:** Approve — merge-ready.

## Overall assessment
PR #1872 fixes two user-visible bugs in the pi agent adapter and a third latent bug in `core/engine.go` that hardcoded Codex-specific copy in `/reasoning`. The fix mirrors `pi-ai`'s own model-picker and thinking-level rules, so users get the same options they'd see in pi's native UI. Five-language i18n templates stop hardcoding the word "Codex" in a generic `/reasoning` message. CI is 5/5 green and test coverage is comprehensive (regression tests cover each new behavior, plus a Core-level test guarding against re-introducing the Codex hardcoding).

## Review scope
- Reviewed all 5 files (+640 / -101).
- Focused on: catalog merge semantics, reasoning-effort rules parity with `pi-ai`, `defaultProvider` resolution, i18n string cleanup, agent-agnostic Core rendering.

## ✅ What looks good
- **Catalog mirror, not invention**: `readModelCatalog()` overlays `models.json` on top of `models-store.json` (custom defs replace same-id entries, append new ones). This is exactly what pi-ai's own picker shows, so users won't be surprised by an option that pi itself doesn't offer.
- **Bare-id resolution prefers `defaultProvider`**: `lookupModelDef` for `"glm-5.3"` checks `settings.json.defaultProvider` first, then iterates other providers in alphabetical order. The test `TestSupportedThinkingLevels_BareIDPrefersDefaultProvider` proves this: with `defaultProvider="zai-coding-team"`, a bare `glm-5.3` resolves to the team variant, not the catalog cn variant. This is essential when two providers share an id (the reported `zai-coding-team` vs `zai-coding-cn` case).
- **`:thinking` suffix stripping (`let s = "glm-5.3:high"` → `"glm-5.3"`)**: matches pi's own `--model` pattern. Covered by the `thinking suffix is stripped` table-driven case.
- **pi-ai rules implemented faithfully**: `thinkingLevelsForDef` encodes the three documented rules:
  1. non-reasoning model = `["off"]` only,
  2. level explicitly mapped to `null` = unsupported,
  3. `xhigh`/`max` only offered when explicitly mapped to a string.
  The "explicitly null" rule is the key insight: pi-ai's data says "this model does not support off" by mapping `"off": null`, not by omitting the key. Without this rule, every reasoning model would advertise `off`, which is wrong.
- **Catalog wins for `ContextWindow`, custom wins on override**: `loadModelsContextWindows` fills from the catalog first, then overlays custom. Catalog-only models now also get their correct window instead of falling back to 200K (a free bonus).
- **Strict i18n cleanup**: `MsgReasoningDefault` and `MsgReasoningUsage` no longer mention "Codex". This is a quiet but real win — non-Codex agents (pi, Claude Code, etc.) won't see "using Codex default" gibberish anymore. All five languages updated.
- **`reasoningUsageText` helper in core**: keeps the core helper agent-agnostic by passing the agent's `AvailableReasoningEfforts()` list. The Core test (`TestCmdReasoning_NoHardcodedCodexLevels`) explicitly guards against future regressions of the Codex-hardcoding bug.
- **All-or-nothing fallback preserved**: when both model files are missing/malformed, `loadModelsContextWindows` returns `nil` and the existing 200K fallback applies. Behavior is identical to today for users who don't customize.
- **Test coverage**:
  - `TestAvailableModels_MergesCustomProviders` — regression for the missing-providers bug.
  - `TestAvailableModels_CustomOverridesCatalogModel` — same-id defs replace, don't append.
  - `TestAvailableReasoningEfforts_DeepSeekFlashIncludesMax` — `max` included where mapped.
  - `TestAvailableReasoningEfforts_ModelAware` — table-driven for the five tricky cases (explicit-null-off, non-reasoning, no-map, unknown, suffix strip).
  - `TestSupportedThinkingLevels_BareIDPrefersDefaultProvider` — bare-id resolution.
  - `TestCmdReasoning_NoHardcodedCodexLevels` — Core-level guard against Codex hardcoding.
  - All use `t.Setenv("PI_CODING_AGENT_DIR", tmpDir)` + `t.TempDir()` — hermetic, no shared global state.

## 🟠 Should improve (P2 — non-blocking)

### P2-1: `loadModelsContextWindows` adds catalog reads on every call
**Impact:** Behavior change for catalog-only models (no custom `models.json`). Previously they returned 200K fallback; now they return their catalog-declared `contextWindow`. Probably a strict improvement, but it's a behavior change.
**Evidence:** `loadModelsContextWindows` now reads `models-store.json` first; any catalog-only model gets its real window. Pre-PR behavior was 200K.
**Suggested fix:** Document the change in the changelog so deployers can audit which models had their context window re-evaluated. No code action needed — just observability.
**Why P2:** Strict improvement in most cases; the only risk is if a model's catalog-declared window is wrong (which would have been a pi-ai bug, not ours).

### P2-2: `providerOptions` re-sorts on every call
**Impact:** Each `AvailableModels()` / `catalogModelOptions()` call iterates the catalog and sorts the result. With hundreds of models and a busy `/model` card hit rate, this is wasted CPU.
**Evidence:** `providerOptions` runs `sort.Slice(models, ...)` per invocation; no caching keyed by `(mtimeOfStore, mtimeOfCustom)`.
**Suggested fix:** Cache the result keyed by `(mtimeOfStore, mtimeOfCustom)` in `Agent` or a package-level sync.Map. Invalidate on file mtime change. Out of scope for this PR but worth a follow-up issue.
**Why P2:** Small per-call cost; only matters for users with large catalogs and frequent `/model` invocations.

### P2-3: `lookupModelDef` `defaultProvider` priority is order-dependent
**Impact:** When `defaultProvider` is set but doesn't contain the requested bare id, the algorithm tries it first (miss), then iterates the rest alphabetically. Correct, but the logic order in `lookupModelDef` (`for name := range catalog`, then `append([defaultProvider], ...)`) reorders the slice rather than just preferring one provider.
**Evidence:** `agent/pi/pi.go` `lookupModelDef` does:
```go
providers := make([]string, 0, len(catalog))
for name := range catalog {
    providers = append(providers, name)
}
sort.Strings(providers)
if s, err := readSettings(); err == nil && s.DefaultProvider != "" {
    providers = append([]string{s.DefaultProvider}, ...)
```
This works but creates an extra slice. Acceptable, but worth noting.
**Suggested fix:** None required. Just noting for future readers.
**Why P2:** Style nit; behavior is correct.

## 🔵 Optional (P3 — nit)

### P3-1: i18n comment for `MsgReasoningDefault`
**Why:** The previous "using Codex default" message implied the default was Codex's. After this PR, the message becomes agent-agnostic. Consider updating the field comment in `core/i18n.go` (around the `messages` block) to mention the placeholder semantics. Out of scope; leave for whoever touches i18n next.

## ❓ Questions

- **Q1**: With `loadModelsContextWindows` now reading the catalog, do other pi users in production get a free context window fix? (Could be a one-line CHANGELOG entry to surface the behavior fix.)
- **Q2**: For agents without a `thinking` parameter (e.g. plain Claude Code without extended thinking), does `AvailableReasoningEfforts()` still return a non-empty list, or does the engine handle empty correctly? (Looking at `cmdReasoning`, an empty list should still work — just no levels to choose. Confirm in a quick manual test if you haven't.)

## Testing / CI

- **CI:** 5/5 green at commit `bb104bbbe147e48f1eaa17fc63b61191634e242d` (performance-test, regression-test, smoke-test, unit-test, lint).
- **New tests:** 6 new unit tests in `agent/pi/pi_test.go` + 1 new test in `core/engine_test.go`. All pass.
- **Manual verification recommended before merge:**
  - With a real pi install, add a custom provider in `models.json` (e.g. `zai-coding-team`). Run `/model` — both providers should be listed.
  - Switch to a reasoning model. Run `/reasoning` (no args) — should list model-specific levels (e.g. `[off, low, high, max]` for `deepseek-flash`).
  - Verify `i18n` text: on a non-English locale, `/reasoning` text should not contain "Codex" anywhere.
- **Untested risk**: the catalog-only context-window behavior change (P2-1) is untested by the unit suite because the test fixtures only set `models.json`. If `models-store.json` declares a wrong window, pi users would silently pick it up. Low risk; pi-ai maintains that file.

## Next step
Maintainer can merge as-is. All findings are P2-or-below follow-ups. The three real bugs (missing providers, missing max, hardcoded Codex text) are fixed with strong test coverage and zero behavior regressions for users who don't customize.

PR #1873, opened 2026-09-19
feat(opencode): expose skill directories via SkillProvider

BoYanZh:
## What

Implements `core.SkillProvider` for the OpenCode adapter so cc-connect can discover OpenCode agent skills for the Web Admin **Skills** page and the `/skills` command.

Previously the OpenCode adapter exposed no skill directories, so the Skills page was always empty (`dirs: null, skills: []`) even when `~/.claude/skills` / `~/.agents/skills` were populated.

## How

`SkillDirs()` returns the locations OpenCode itself searches (see https://opencode.ai/docs/skills/):

- **project-local**, walking from the working directory up to the repo root (`.git` / `.jj`): `.opencode/skills`, `.claude/skills`, `.agents/skills`
- **global**: `~/.config/opencode/skills`, `~/.claude/skills`, `~/.agents/skills`

It mirrors the existing `claudecode` / `codex` walk-up implementations (project-root boundary, home boundary, deduplicated, stable order).

## Tests

`agent/opencode/skilldirs_test.go` covers project walk-up order, always-present global dirs, and an empty `work_dir`.

## Verified

Built and ran against a live cc-connect v1.5.0 + OpenCode: `/api/v1/skills` now returns the skill dirs and discovers the local skills (37 on a machine with `~/.claude/skills` + `~/.agents/skills`).


chenhg5 (changes requested):
Hi BoYanZh — thanks for implementing SkillProvider for the OpenCode adapter; the walk-up + global dir logic looks reasonable and the parity with the existing claudecode/codex helpers is a good signal.

However, CI is currently blocking this PR — the **lint** check is failing on Run 35435469197 (job 105877253274):

```
agent/opencode/opencode.go:596:3: QF1006: could lift into loop condition (staticcheck)
                if homeDir !=  && sameOpencodePath(current, homeDir) {
                ^
1 issues:
* staticcheck: 1
```

Lint runs with `--new-from-rev origin/main`, so the issue is in the new code you added in `walkUpOpencodeProjectSkillDirs` (around the `for { ... if homeDir != "" && sameOpencodePath(...) { break } ... }` loop). The downstream unit/regression/smoke tests are all skipped because lint gate failed.

Suggested fix: lift the home-boundary check into the `for` condition itself (e.g. `for current != "" && (homeDir == "" || !sameOpencodePath(current, homeDir)) { ... }` and drop the inner `if`) so staticcheck QF1006 is satisfied and you keep the same break semantics.

Once lint is green, the rest of the checks should pick up automatically. After that I'll do a full diff review of the implementation.

Log: https://github.com/chenhg5/cc-connect/actions/runs/35435469197/job/105877253274

BoYanZh:
Thanks for the review! Fixed in `be65b9c7` — lifted the home-boundary check into the `for` condition as suggested:

```go
for current != "" && (homeDir == "" || !sameOpencodePath(current, homeDir)) {
    ...
}
```

Same break semantics, inner `if` removed. Verified locally with `golangci-lint v2.11.4` (`staticcheck` enabled): `golangci-lint run --new-from-rev origin/main ./agent/opencode/...` → **0 issues**. Pushed; CI should re-run. Happy to adjust if you'd prefer a different shape.

chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Clean, additive implementation of `SkillProvider` for the OpenCode adapter. CI is fully green (Run 35460677769: lint/unit/smoke/regression/performance all pass), the diff is scoped to `agent/opencode/` and a single test file, and the new walk-up mirrors the established claudecode/codex semantics.

✅ What looks good:
- The QF1006 fix in commit be65b9c7 lifts the home-boundary check into the `for` condition exactly as suggested, with identical break semantics. No inner `if`, no extra branches, no behavior change.
- Pure-function decomposition (`opencodeSkillDirs`, `walkUpOpencodeProjectSkillDirs`, `findOpencodeProjectRoot`, `sameOpencodePath`, `uniqueOpencodeSkillDirs`) makes the behavior testable without spinning up an Agent.
- Double marker (`.git` / `.jj`) covers jj users; `uniqueOpencodeSkillDirs` uses `filepath.Clean` rather than `Abs` so the test fixtures match exactly without macOS `/private` var-link surprises.
- `Abs` failure falls back to the original workDir, and an empty `workDir` short-circuits the walk-up loop via the `current != ""` condition — `TestSkillDirs_EmptyWorkDirDoesNotPanic` documents this guarantee.
- Windows HOME handling via `setTestHome` sets USERPROFILE/HOMEDRIVE/HOMEPATH alongside HOME; CI already exercises linux but this keeps the suite portable.
- No changes to `core` / `web` / config schema — strictly additive for callers that already handle empty `SkillDirs()` slices.

🔵 Optional (P3, not blocking):
- `make([]string, 0, 3)` capacity in `opencodeSkillDirs` is a minor under-allocation relative to the actual home-dir + project-walk size, but `append` reallocates cheaply; leave as is.

❓ Questions:
- None.

Testing / Risk:
- Verified: all 5 CI checks pass on Run 35460677769 after the lint fix landed.
- Tests cover: walk-up order with repo-root stop, global home dirs always present, empty workDir doesn't panic.
- Remaining risk: minimal — only the new `SkillDirs()` method gains behavior; existing callers that treat `nil`/empty slices as "no skills" continue to work.
- Backout: single revert, no migration.

Next step:
- Mergeable. No follow-up actions required.

Full QA decision report (中文): doc-20260919-stj41d

PR #1874, opened 2026-09-19
fix(feishu): preserve forwarded context, images and reply ordering

benzfy:
## Summary

Fix Feishu merged-forward handling end to end: preserve inbound dispatch order, expand quoted forwards, and download forwarded images using the visible outer container ID. Add an opt-in window to combine a forward with an adjacent, explicit same-user text reply into one agent request.

### Problems and changes

1. **Slow forwards can be dropped as stale.** Forward expansion performs API calls before handing the message to the engine. A later text event can overtake it and advance the engine's timestamp watermark. Reserve a per-session dispatch lane synchronously in the SDK callback; serialize parsing and engine handoff, not agent execution. Other sessions remain independent.
2. **Replies to forwards lose their context.** Message GET returns the container plus descendants. Select the requested container by ID (not array position), tolerate an empty container body, and share tree rendering with direct forwards. Ignore deleted entries and retain the existing mention/same-sender gate for quoted file downloads.
3. **Forwarded images fail with `234003: File not in msg`.** Keep the visible outer container ID separate from the recursive traversal ID and use it for image and rich-text-image downloads. Ordinary image handling is unchanged. File-download behavior is not changed by the image fix.
4. **A forward and its instruction produce separate agent turns.** Add `forward_merge_window_ms` (default `0`, disabled; allowed range `0..10000`). For example, `1000` waits up to one second from receipt for an adjacent same-user text explicitly replying to that forward. It submits one message with B's instruction/identity/timestamp/reply target and A's material/attachments. Commands, other senders, unrelated messages and late replies do not merge. An intervening admitted message flushes the candidate; slow expansion does not extend the deadline. Standalone forwards are still delivered when the window expires.

No agent-specific/core routing changes, user-token fallback, deployment settings, or network changes are included.

## Type of change

- [ ] Breaking change
- [ ] Documentation only

## Testing

Latest head `dd9943c988f8c6cadd4b8400ed499cf1288c8f1d`: [GitHub CI run 35457447614](https://github.com/chenhg5/cc-connect/actions/runs/35457447614) passed all five jobs on Linux: lint, unit-test (including full race and coverage), smoke-test, regression-test and performance-test. Local validation below also passed on macOS/arm64.

- `TestOnMessage_SlowMergeForwardPreservesDispatchOrder`: gated HTTP lookup reproduces overtaking; verifies independent sessions, original timestamps, deduplication and early-return lane release.
- `TestDispatchMessageQuotedMergeForward_ImagesUseContainerID`: direct/quoted expansion, root placement/empty body, nested content, deleted children, image/post-image resource paths, one lookup and quoted-file privacy gates.
- `TestFetchQuotedMergeForwardRejectsMissingOrDeletedRoot`: never substitute a child for an unavailable container.
- `TestForwardImageFix_OrdinaryImagesKeepOwnMessageID`: ordinary image/post downloads remain unchanged.
- `TestOnMessageForwardMerge_*`: real inbound events in private/topic sessions; A+B followed by C, duplicate delivery, standalone/late replies, disabled option, unrelated/foreign/command/intervening boundaries, deadline, recall and lookup failures.
- `TestNewPlatformForwardMergeWindow`: default, valid values and invalid config.
- `TestCUJ_A9_ComposedInputThenFollowupAndHistory`: real engine; composed input, independent follow-up and `/history` show two user turns, not an extra material-only turn.
- Stabilize `TestCUJ_A3_ImageReachesAgent` / `TestCUJ_A5_FileReachesAgent`: wait for the turn to finish and a reply to arrive before temporary-directory cleanup, and stop the engine. Previously these tests returned as soon as the agent session existed while history writes were still in flight. The failure was reproduced on unmodified `origin/main`; the repaired A3/A5/A9 tests passed 30 repetitions with `-race`. No production engine code is changed.

### Regression verification

- [x] Regression tests were run against the pre-fix behavior and failed, then passed with the fixes (dispatch overtaking, quoted-forward context, and image container ID).
- Live, bot-only comparison using the same forwarded image key: child ID returned HTTP 400 / `234003`; outer container ID returned HTTP 200, `image/jpeg`, 356,383 bytes. No user access token was used. This verifies the reported image case; nested-image cases additionally have mock coverage.

Commands:

```sh
go build ./...
go test ./...
go test ./... -race
go test ./... -coverprofile=/tmp/coverage.out -covermode=atomic
go vet ./platform/feishu
golangci-lint run --new-from-rev origin/main ./... # v2.11.4, matching CI
actionlint # v1.7.8, matching CI
go mod verify
go test -race ./platform/feishu
go test -race ./platform/feishu -run 'TestOnMessageForwardMerge|TestOnMessage_SlowMergeForward|TestDispatchMessageQuotedMergeForward|TestForwardImageFix' -count=10
go test ./core -run TestCUJ -count=1
go test -race ./core -run 'TestCUJ_A(3|5|9)_' -count=30
go test -tags=smoke,no_web ./tests/e2e/...
go test -tags=regression,no_web ./tests/e2e/...
git diff --check
```

The first GitHub run caught two lint issues (unchecked test teardown and `WriteString(fmt.Sprintf(...))`); both are fixed and checked locally with the exact CI linter version. Local validation also uncovered the attachment CUJ teardown race described above. An initial coverage run hit an unrelated `agent/antigravity` subprocess timing failure; that test passed 10 isolated repetitions, and the full coverage rerun passed without changes to that agent.

Web assets were built in an isolated checkout using pnpm 10 (the CI major version) with the unchanged web sources and frozen lockfile. The host's pnpm 11 rejected the existing dependency-build policy; no approval settings or lockfiles were loosened or changed. All local results are macOS/arm64; GitHub CI independently checks Linux.

- [x] A — basic conversation
- [x] B — session history
- [x] D — quoted-file permission boundaries preserved
- [x] G — lookup failure / recall handling
- [x] H — session isolation
- [x] `go test ./core -run TestCUJ` passes locally.
- [x] New CUJ covers the composed-input flow; adapter tests cover actual event coalescing.

Forwarded text/images and quoted-forward context reach the agent reliably. The optional merging behavior is enabled with:

```toml
[projects.platforms.options]
forward_merge_window_ms = 1000
```

Use this under the existing Feishu platform options section. The default is disabled, with no extra wait. This is not generic consecutive-message batching or agent steering.

## Checklist

- [x] Build and tests pass; concurrency changes covered by the race detector.
- [x] AGENTS.md pre-commit checklist reviewed.
- [x] No new hardcoded platform/agent names in core.
- [x] No new end-user UI strings requiring translations.
- [x] No secrets, real message IDs, private conversation contents or local deployment paths in source.

## Related

- Reproduced using bot-visible merged-forward messages; no external issue is auto-closed.


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- This PR fixes three real Feishu reliability bugs (dispatch ordering, container-ID selection, image resource routing) and adds an opt-in coalescing window for the common "forward + instruction" pattern. CI is fully green (Run 35457447614: lint/unit/smoke/regression/performance pass), the diff is well-scoped to platform/feishu plus a single core CUJ test, and the test coverage is unusually thorough.

✅ What looks good:
- The dispatch-order fix (`enqueueMessageDispatch`) correctly reserves the lane synchronously and chains predecessors via a per-session `chan struct{}`, which addresses the root cause (goroutine scheduling is not FIFO) rather than papering over symptoms.
- Splitting `resourceMessageID` from the recursive `parentID` in `formatMergeForwardTree` is the minimal correct change for the 234003 image bug; the same fix benefits post-image downloads without extra surface area.
- The opt-in `forward_merge_window_ms` is defensively bounded (0..10000, default 0) and `canMergeForwardReply` rejects cross-chat, cross-user, recalled, command, and inverted-timestamp pairs — exactly the failure modes that would otherwise silently produce wrong coalescing.
- `fetchSingleMessage` now selects the requested item by `message_id`, tolerates empty body for `merge_forward`, skips `deleted` items, and reuses `renderMergeForward` for both direct and quoted forwards.
- `dispatchPendingForward` deliberately downloads A's material during the wait window so slow lookups don't extend the deadline; lookup failure cleanly falls back to dispatching B alone rather than swallowing the user's instruction.
- New `TestCUJ_A9_ComposedInputThenFollowupAndHistory` covers the composed-input contract end-to-end; the A3/A5 `defer e.Stop()` fixes a teardown race that the author reproduced on unmodified `origin/main`.

- Consider documenting `forward_merge_window_ms` in `docs/platforms/feishu.md` or the README in addition to the inline config comment. It's a feature users will discover by reading the diff, and the gating rules (same user, same chat, explicit reply, no command) deserve a paragraph so they don't enable it expecting generic consecutive-message batching.
- Long-term: the adapter submits A's attachments through B as raw bytes without re-checking them against the quote policy. It's correct today because the dispatcher still applies the mention + same-sender gate for files inside quoted forwards (`quotedFileMeta` path), but a future refactor that conflates those paths would silently bypass the gate. Worth a comment near the `combined.Files = append(material.Files, ...)` line pointing to the policy owner.

❓ Questions:
- None blocking; happy to follow up on the doc-suggestion P2 after merge.

Testing / Risk:
- Verified: lint, unit-test, smoke-test, regression-test, performance-test all green on Run 35457447614.
- Author's local validation: race detector 30x on A3/A5/A9; 10x on the dedicated merge/forward suites; live bot reproduction of the 234003 image bug showing child-ID → 400 vs. container-ID → 200/356,383 bytes; web build under pnpm 10 + frozen lockfile; actionlint + golangci-lint v2.11.4 matching CI versions.
- Remaining risk: the synchronous enqueue change in `onMessage` is unconditional — every SDK callback now serializes via the per-session tail channel. In practice this only adds a few ms (channel hand-off) per message, but worth monitoring the new `message dispatch tail latency` once enabled in production.
- Backout: single revert restores prior behavior; opt-in feature stays disabled by default.

Next step:
- Mergeable. Default `forward_merge_window_ms = 0` keeps behavior backwards-compatible; document the opt-in semantics before recommending it to users.

Full QA decision report (中文): /root/code/spaceship/multigent_e2e/6bbcd4cb-f08b-4268-8f93-926e5939eb59/.multigent/files/../docs/doc-20260919-ce7svo
Repo doc id: doc-20260919-ce7svo

PR #1875, opened 2026-09-19
fix(management): show multi-workspace sessions in the management API

BoYanZh:
## Bug

In `mode = "multi-workspace"` every workspace owns its own `SessionManager`, so the engine's base `e.sessions` is empty. The management API read sessions only from that base manager, so the Web Admin always showed **`sessions_count: 0`** and an empty session list, even while sessions existed on disk and in `cc-connect sessions list`.

```
GET /api/v1/projects
{"projects":[{"name":"my-project", ..., "sessions_count":0}]}   # always 0
```

## Fix

Aggregate the base manager **plus every workspace manager** for the session count, list, detail, delete and switch (`core/management_sessions.go`).

Two things were needed to make it correct:

1. **Live workspace managers** — the ones in `engine.workspacePool`. But the pool is only populated once a workspace actually sees traffic, so this alone still read 0 right after a restart.
2. **Persisted workspace stores on disk** — `<project>_ws_*.json` in the sessions directory, loaded and deduped against live managers by store path. This mirrors what `cc-connect sessions list` already does.

Session ids are assigned per manager (each starts at `s1`), so in multi-workspace mode the management id is qualified (`<qualifier>::<id>`) to avoid collisions and to resolve back to the owning manager. Single-workspace mode is unchanged (bare ids, base manager only).

`POST /projects/{name}/sessions` now also resolves the workspace manager for the given `session_key` via `sessionContextForKey`, so a session created from the UI lands where the agent will look.

## Tests

`core/management_sessions_test.go`:
- single-workspace behaviour unchanged (count, bare ids)
- multi-workspace aggregation + qualified ids resolve to the right manager
- persisted (not live) workspace stores are counted and resolvable
- empty stores are ignored / unknown qualifier falls back to a bare-id search

## Verified

Built and ran against a live cc-connect + OpenCode multi-workspace setup: after a restart with an empty workspace pool, `/api/v1/projects` reports the real count and `/api/v1/projects/{name}/sessions` lists the on-disk workspace sessions.


chenhg5 (approved):
Conclusion: Approve

Overall assessment:
- Correct fix for a real bug: multi-workspace users were seeing `sessions_count: 0` in the Web Admin regardless of on-disk state. The fix aggregates live workspace managers with disk-only persisted stores, qualifies per-manager session ids to avoid collisions, and updates all four CRUD paths (count, list, detail/delete, switch, create) consistently. CI is fully green (Run 35464971063: lint/unit/smoke/regression/performance all pass).

✅ What looks good:
- `managementSessionManagers()` short-circuits in single-workspace mode (returns base only), so single-tenant deployments incur zero overhead. Multi-workspace aggregation dedupes by store path, ignoring empty stores — both common-sense behaviors, both explicitly covered by tests.
- Splitting live vs disk managers and deduping on `StorePath()` correctly handles the "workspace pool is empty right after a restart" scenario. This is the load-bearing detail that makes the count accurate in practice.
- `managementSessionWorkspaceSep = "::"` plus `strings.LastIndex"::"` parsing handles qualified ids cleanly. Unknown qualifier falls back to bare-id search, preserving backwards-compatibility for any external caller that might have cached a bare id from earlier.
- `POST /projects/{name}/sessions` correctly resolves the owning workspace manager via `sessionContextForKey` so newly created sessions land where the agent will look — fixes a related latent issue where the old `e.sessions.GetOrCreateActive` would create a session on the base manager and orphan it from agent routing.
- All four CRUD handlers now consistently emit a `workspace` field when known and use qualified ids from `managementSessionID(s, mm.qualifier)`. The Web Admin side stays in sync.
- New tests: `TestManagementSessions_SingleWorkspaceUnchanged` (regression guard), `TestManagementSessions_MultiWorkspaceAggregatesAndQualifiesIDs`, `TestManagementSessions_LoadsPersistedWorkspaceStores`, `TestFindManagementSession_UnknownWorkspaceFallsBack`. The persisted-store test specifically exercises the post-restart scenario the bug report described.

- `managementSessionManagers()` re-globs the sessions directory on every list/count call. For typical multi-workspace deployments (<100 workspaces) this is sub-millisecond, but if a user grows to thousands of workspaces this becomes a hot path. A 1-5s TTL cache would be cheap insurance.
- The session-list response in `handleProjectSessions` rebuilds the entire array on every request. If the Web Admin polls aggressively, an ETag / last-modified header on the response would let clients skip the rebuild when nothing changed.

❓ Questions:
- None blocking.

Testing / Risk:
- Verified: all 5 CI checks pass on Run 35464971063.
- I cross-checked that every symbol the new code references exists on the PR head: `sessionContextForKey` (18 refs), `multiWorkspace` (22 refs), `workspacePool` (21 refs), `newWorkspacePool` (2 refs), `GetOrCreate` (2 refs).
- Backout: single revert, no schema migration.
- Single-workspace deployments are behaviorally unchanged.

Next step:
- Mergeable. No follow-up actions required before merge.

Full QA decision report (中文): doc-20260919-12j183

chenhg5 (approved):
## Conclusion: Approve

Small, well-scoped bugfix that unblocks multi-workspace users from seeing their sessions in the Web Admin. CI is green (lint / unit / smoke / regression / performance) and the four new tests cover the meaningful behaviors.

### ✅ What looks good

- **Correct root-cause analysis**: the bug is real — in `mode = "multi-workspace"` the base `e.sessions` is empty because every workspace owns its own `SessionManager`, and the management API was reading only the base. Aggregating base + live workspace managers + persisted disk stores is the right shape of fix.
- **Persisted-store fallback is the subtle, important piece**. `engine.workspacePool` is only populated when a workspace sees traffic, so without reading `<project>_ws_*.json` from disk the count would still read 0 right after every restart. Mirroring what `cc-connect sessions list` already does is the right call. Empty stores are filtered out, and live/dedup by store path avoids double-counting.
- **Session-id qualification with fallback**: `<qualifier>::<id>` keeps per-manager ids (`s1`, `s2`, …) from colliding across workspaces, and `findManagementSession` still accepts a bare id (and falls back to bare-id search on an unknown qualifier). Single-workspace mode is untouched, so no caller is broken.
- **`handleProjectSessions` POST** now resolves the correct workspace via `sessionContextForKey` so a session created from the UI lands where the agent will actually look.
- **Symbol/contract verification against `origin/main`** — `Engine.{sessions, multiWorkspace, workspacePool, name}`, `sessionContextForKey`, all `SessionManager` methods, `workspacePool.All/GetOrCreate`, and the `<project>_ws_<hex>.json` naming convention (`engine.go:3983`) all match what the PR uses.
- **Tests are tight**: `SingleWorkspaceUnchanged`, `MultiWorkspaceAggregatesAndQualifiesIDs`, `LoadsPersistedWorkspaceStores`, and `UnknownWorkspaceFallsBack` — exactly the four behaviors the PR claims to deliver.
- **Live verified** by the author against a real cc-connect + OpenCode multi-workspace setup, including the post-restart cold-pool case.

### 🟠 Should improve (non-blocking, optional follow-up)

1. **Disk-only workspace managers are throwaway**. `managementSessionManagers()` does `NewSessionManager(path)` on each API call for persisted stores that are not live. `handleProjectSessionSwitch` and `handleProjectSessionDelete` then call `ms.manager.SwitchSession/DeleteByID` on that throwaway instance — the mutation lives only in memory and is not saved back to disk. This is a **latent bug, not introduced or worsened by this PR**, but it is reachable through the new code path. Suggested follow-up: either (a) refuse switch/delete on disk-only entries with a clearer 4xx response, or (b) call `Save()` after the mutation. Worth filing as a separate issue.

2. **`findManagementSession` qualifier-fallback comment could be clearer**. The `id = raw` shadow inside the loop plus the trailing comment `// Unknown qualifier: fall back to searching the bare id below.` works, but the control flow has two return paths inside one `if idx >= 0` block. A short note above the loop ("If a qualifier is present but unknown, fall through to bare-id search; if a qualifier is present and matches, return on miss without searching") would help the next reader.

### 🔵 Optional (nit)

- The `GlobalProviderInfo` struct-field alignment change in `management.go` is pure gofmt churn. Fine, but worth noting in the diff summary if reviewers care about that.

### ❓ Questions

- None blocking. The merge gate decision is clear.

### Testing / Risk

- **Verified**: 5 CI jobs green on the current head; 4 new unit tests cover single/multi/persisted/fallback; author live-verified in a real multi-workspace setup including the cold-pool case.
- **Blast radius**: scope limited to `core/management.go` and the two new files. No agent adapters, platforms, config, or daemon code touched. Single-workspace behavior is byte-identical.
- **Remaining risk**: low. The only edge case I can see is the disk-only switch/delete latent bug above, which exists regardless of this PR.

### Next step

This is ready to merge. Recommended path: merge as-is, file a follow-up issue for the disk-only SessionManager persistence on switch/delete, and observe post-deploy that `/api/v1/projects` `sessions_count` matches `cc-connect sessions list` for multi-workspace users.

Internal QA decision doc: `doc-20260920-nncsr9` (in workspace knowledge base).

